PR #145, opened 2017-02-26
Easytest fix to make exceptions from code under test result in test e…

tmciver:
…rrors.


tmciver:
Not sure why checks failed.  Local tests passed (when running `stack test`).

Issue #139, opened 2016-12-27
Get Travis build working again, seems to be broken due to glibc mismatch?


pchiusano:
See: https://travis-ci.org/unisonweb/unison/builds/187051809

PR #147, opened 2017-04-06
Fix/travis

peterbecich:
I believe this fixes https://github.com/unisonweb/unison/issues/139

The experimental Trusty build environment fixed the broken libc6 dependency: https://docs.travis-ci.com/user/trusty-ci-environment/

I put a few new things from this (https://docs.haskellstack.org/en/stable/travis_ci/) in `.travis.yml`, but these things weren't necessary to solve the problem.  I think the only necessary change was `dist: trusty`

Successful build: https://travis-ci.org/peterbecich/unison/builds/219097173

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ addons:
       - hvr-ghc
     packages:
       - ghc-8.0.1
+      - libc6
+      - libc6-dev
 install:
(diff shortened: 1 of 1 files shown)


pchiusano:
Wow, nice! Thanks for figuring this out.

peterbecich:
@pchiusano thanks, sure thing!

PR #152, opened 2017-06-25
autogenerate FunctionApplication.scala

refried:
This patch autogenerates `FunctionApplication.scala`.

`FunctionApplication.scala` is obviously simpler than `Runtime.scala`, but I was thinking to get started on the latter, I could copy the contents over as a static template, and could begin factoring pieces out from there.  

Let me know if you'd like me to do this.


pchiusano:
I think what I'd like to do is move any repetitive logic in `Runtime` to separate modules (like was done with `FunctionApplication.scala`, but would like to apply this more consistently).

`org.unisonweb.compile.{LetRec, LetRec1, Let1, , ...}`. Existing `FunctionApplication.scala` would go here. Idea would be that all these `org.unisonweb.compile.*` modules are generated.

I kind of wish we could have function args in "normal" left-to-right order. It would simplify the code - `fn(fn, arg1(rec, r), r.unboxed, arg2(rec, r), r.unboxed, ...)` rather than needing to declare all those intermediate variables.

PR #157, opened 2018-03-09
Decompilation of terms (including terms with cyclic references)

pchiusano:
This is really important. We can now compile terms, execute them using the JIT'd runtime representation, but retain enough metadata to be able to decompile results after (or during) execution, with human-readable results.

This is useful just for having a really nice interactive editor (the result of any computation can be displayed back to the user as an "ordinary" Unison program, rather than showing them some ad hoc `toString` representation that the user often has to write and which can't print out functions or cyclic structures), for debugging (imagine being able to pause a computation at any point, obtaining the continuation as a regular Unison function whose source you can inspect), and all sorts of other stuff (perhaps serialization of continuations for inter-node protocol). 

An example, after evaluating this:

```Haskell
let 
  id x = x
  let rec
      pong x = ping x
      ping x = pong (id x)
      ping
```

We get:

```Haskell
let rec
    pong x = ping x
    ping x = pong ((x -> x) x)
    x -> pong ((x -> x) x)
```

The `id` definition has been substituted away during execution, but we've retained enough metadata about the resulting value to know to decompile it to `x -> x` (since `x` was the parameter name chosen for `id`).

Some other minor cleanup:

* Removed `RunUtilTests`, added `AllTests` and `RunAllTests`, which is the sole `main` in the test directory.
* Handy: `RunAllTests` takes a command line arg for the prefix, ex: `> test:run "util"` or `> test:run "util.Critbyte"`. I got tired of editing the code just to select a subsuite of tests.
* Added `org.unisonweb.util.Monoid`, using it in a couple places instead of passing a pair of args.

@refried The simplified algorithm I proposed seems like it works, but we should try to beef up the tests at some point and see if can break it.

### DIFF
--- a/runtime-jvm/main/src/main/scala/ABT.scala
+++ b/runtime-jvm/main/src/main/scala/ABT.scala
@@ -30,12 +30,14 @@ object ABT {
       AnnotatedTerm(f(annotation), get.map(_.map(f)))
     def reannotate(f: A => A): AnnotatedTerm[F,A] = AnnotatedTerm(f(annotation), get)
 
+    /** Discard all the annotations and just annotate with the set of free variables. */
     def annotateFree(implicit F: Traverse[F]): AnnotatedTerm[F, Set[Name]] = get match {
       case Var_(name) => Var(name)
       case Abs_(name, body) => Abs(name, body.annotateFree)
       case Tm_(tm) => Tm(F.map(tm)(_ annotateFree))
     }
 
+    /** Push an annotation down from the root. */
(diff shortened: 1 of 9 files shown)


refried:
Did you mean to pick up the critbyte commits, ie will this cause a merge conflict when we try to add from topic/critbyte?

Otherwise lgtm

pchiusano:
I'll switch this over to be a PR against topic/critbyte on Monday. Then we can merge this PR, then merge topic/critbyte to master, since it's in a stable state (can always fill in more functions or speed up impl later).

aryairani:
👍

pchiusano:
Sync'd this with topic/critbyte and changed it to be a merge into that branch.

Note, just do `> test:run "util.Critbyte"` to run just those tests, or leave off the argument to `test:run` if you want to run the whole suite.

PR #161, opened 2018-03-14
unionWith and faster union

runarorama:


runtime-jvm/main/src/main/scala/util/Bytes.scala
```diff
@@ -66,21 +66,21 @@ object Bytes {
...
     /** Lexicographical minimum, assuming unsigned bytes. */
-    def min(b2: Seq): Seq = {
+    def min(b2: Seq): Seq = try {
       val sdi = this.smallestDifferingIndex(b2)
       if (sdi >= this.size) this
       else if (sdi >= b2.size) b2
       else if (unsigned(this(sdi)) < unsigned(b2(sdi))) this
       else b2
-    }
+    } catch { case Bytes.Seq.NotFound() => this }
```
runarorama: If the sequences are equal, I'm returning the first one rather than rethrowing `NotFound`.

runtime-jvm/main/src/main/scala/util/Bytes.scala
```diff
@@ -115,8 +115,8 @@ object Bytes {
 
     def emptyBase: Base = One(Array())
 
-    case object OutOfBounds extends Throwable
-    case object NotFound extends Throwable
+    case class OutOfBounds() extends Throwable
+    case class NotFound() extends Throwable
```
runarorama: I've made these classes rather than objects to make stack traces useful. A `Throwable` will forever retain the stack trace from the first place it was thrown. Exception objects don't play nice with stack traces in general since stack tracing is a side effect.

runtime-jvm/main/src/main/scala/util/Critbyte.scala
```diff
@@ -179,9 +230,11 @@ object Critbyte {
       // `sdi` is either the index of the first byte that differs between
       //        `key` and `smallestKey`, or it is the length of `key`,
       //        (because `key` is shorter than `smallestKey`).
-      val sdi = key smallestDifferingIndex smallestKey
+      val sdi =
+        try key smallestDifferingIndex smallestKey
+        catch { case Bytes.Seq.NotFound() => -1 }
 
-      if (sdi >= critbyte) {
+      if (sdi >= critbyte || sdi == -1) {
```
runarorama: *cringe*

runtime-jvm/main/src/main/scala/util/Bytes.scala
aryairani: That is interesting.

runtime-jvm/main/src/main/scala/util/Critbyte.scala
```diff
@@ -138,6 +148,55 @@ object Critbyte {
...
+                   newSmallest,
+                   newRunt,
+                   newChildren).unionWith(otherRunt)(f)
+          }
+          else
+            // The sdi of smallest keys is after both critbytes, telling us nothing.
+            worstCase
+        } catch { case Bytes.Seq.NotFound() =>
+          // The smallest key of both trees is the same
+          worstCase
```
runarorama: We can probably expand `worstCase` to compute the prefixes and then descend into one of the other cases if there's a common prefix.

runtime-jvm/main/src/main/scala/util/Bytes.scala
pchiusano: That is okay for debugging but we should switch back to `object` once we've gotten all the kinks out - otherwise we are paying a performance cost that exceptions were meant to avoid (vs using `Option`).

runtime-jvm/main/src/main/scala/util/Bytes.scala
pchiusano: Seems reasonable.

runtime-jvm/main/src/test/scala/util/CritbyteTests.scala
```diff
@@ -57,6 +57,23 @@ object CritbyteTests {
...
+        val y = genCritbytes(intIn(0, i), int)
+        val u = x union y
+        note("x keys: " + x.keys)
+        note("y keys: " + y.keys)
+        note("u keys: " + u.keys)
+        equal1(u.keys.sorted, (x.keys union y.keys).sorted)
+        y.keys.foreach(k => (for {
+          a <- u lookup k
+          b <- y lookup k
+        } yield equal1(a,b)) getOrElse ok)
```
pchiusano: Idea for a stronger property - 

```
u.keys.foreach { k => equal1(u.lookup(k), y.lookup(k) orElse x.lookup(k)) }
```

PR #168, opened 2018-03-29
move .sbtopts to build.sbt

aryairani:
This moves the `javaOptions` into `build.sbt`, where you can reload with a `reload` instead of having to kill sbt.  

Also sets `fork := true` to fork a separate JVM for runs (which will then use those flags), in order to be able to track performance of different benchmark runs independently of the the main SBT process used for compilation etc.


aryairani:
sure

PR #173, opened 2018-04-05
add sbt-scoverage plugin

aryairani:
easy enough

### DIFF
--- a/runtime-jvm/project/plugins.sbt
+++ b/runtime-jvm/project/plugins.sbt
@@ -2,3 +2,5 @@ scalacOptions ++= Seq(
   "-feature",
   "-deprecation"
 )
+
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")


aryairani:
screen shot 2018-04-05 at 5 01 04 pm

pchiusano:
That’s it?? 😀 How does it know about EasyTest?

aryairani:
I guess it doesn't care.  You issue a command to enable coverage instrumentation, then run whatever executable (maybe more than one), and then generate a report.

If we want to tie it automatically into the build or CI process it'll be a little more involved, but this is enough to manually log coverage from the sbt prompt.

`; clean; coverage; main/test:run; coverageReport`

pchiusano:
Cool. Can you add a note about how to run to development.markdown (and while you are at it include instructions for running the EasyTest jvm tests with or without prefix). Then ok to merge.

pchiusano:
Actually I just noticed development.markdown doesn’t even mention how to build and run the runtime-jvm subproject. Let’s just revamp that in separate commit.

PR #174, opened 2018-04-11
Mutators gonna mutate

runarorama:


aryairani:
Also, awesome!

pchiusano:
Ok think am starting to get this... rather than patterns taking in their RHS (and doing some pushing before calling their RHS), patterns just mutate the Result object with variables they’ve captured. Compilation of each case then runs the pattern, extracts the bound variables from the Result, and then calls the RHS.

aryairani:
Oh, I see now where it returns U.

pchiusano:
I think might have a nice simplification to this basic idea, will explain!

PR #177, opened 2018-04-18
Fix for Term.selfToLetRec

pchiusano:
This was called as a preprocessing step before pretty-printing and was causing us to see goofy results.


pchiusano:
Testing

PR #178, opened 2018-04-19
Topic/streamwork 2 wip

aryairani:


runtime-jvm/main/src/main/scala/Builtins.scala
```diff
@@ -1,18 +1,45 @@
...
 /* Sketch of convenience functions for constructing builtin functions. */
 object Builtins {
 
+  // Stream.empty : Stream a
+  val Stream_empty =
+    c0l("Stream.empty", Stream.empty[Value])
+
+
+  // Stream.fromInt : Integer -> Stream Integer
+  val Stream_fromInt = // Stream.iterate(unison 0)(Integer_inc)
```
aryairani: `Stream.fromInteger`

runtime-jvm/main/src/main/scala/Builtins.scala
```diff
@@ -57,51 +87,54 @@ object Builtins {
     Sequence_size
   )
 
+  val Integer_inc =
+    fl_l("Integer.inc", "x", _ + 1)
```
aryairani: `increment`

runtime-jvm/main/src/main/scala/Builtins.scala
```diff
@@ -181,66 +209,127 @@ object Builtins {
...
+    val body: Computation.C2P = (r,x1,x0,_,x0b) =>
       B.encode(r, f(x1, A.decode(x0, x0b)))
     val decompiled = Term.Id(name)
-    val lambda = new Lambda.ClosureForming2(decompiled, n1, n2, body)
+    val lambda = new Lambda.ClosureForming2(arg1, arg2, body, None, decompiled)
     name -> Return(lambda)
   }
 
-  abstract class FPP_U[A,B] { def apply(a: A, b: B): U }
+  def flp_z[A:Decode,B:LazyEncode](name: Name, arg1: Name, arg2: Name, f: FLP_P[A,B]) =
```
aryairani: This could use a comment on naming convention for `z`

PR #179, opened 2018-05-17
stream work

aryairani:


runtime-jvm/main/src/main/scala/Builtins.scala
```diff
@@ -11,33 +11,78 @@ import org.unisonweb.util.{Sequence, Stream, Text}
 
 /* Sketch of convenience functions for constructing builtin functions. */
 object Builtins {
+  def env =
+    (new Array[U](20), new Array[B](20), StackPtr.empty, Result())
 
   // Stream.empty : Stream a
   val Stream_empty =
-    c0l("Stream.empty", Stream.empty[Value])
+    c0z("Stream.empty", Stream.empty[Value])
```
aryairani: "L" for Lazy became "Z" for laZy. "L" is now for Long.

runtime-jvm/main/src/main/scala/Builtins.scala
```diff
@@ -11,33 +11,78 @@ import org.unisonweb.util.{Sequence, Stream, Text}
...
 
 
-  // Stream.fromInt : Int64 -> Stream Int64
-  val Stream_fromInt = // Stream.iterate(unison 0)(Int64_inc)
-    fp_l("Stream.fromInt", "n", Stream.fromUnison)
+  // Stream.fromInt : Integer -> Stream Integer
+  val Stream_fromInt = // Stream.iterate(unison 0)(Integer_inc)
+    fp_z("Stream.fromInt", "n", Stream.fromUnison)
```
aryairani: just `Stream.from`, or even define this function in Unison as `Stream.iterate 0 (+)`

runtime-jvm/main/src/main/scala/Builtins.scala
```diff
@@ -11,33 +11,78 @@ import org.unisonweb.util.{Sequence, Stream, Text}
...
+          (f: Value, s: Stream[Value]) =>
+            s.map(UnisonToScala.toUnboxed1(f.asInstanceOf[Lambda])(env)))
+
+  // Stream.foldLeft : b -> (b -> a -> b) -> Stream a -> b
+  val Stream_foldLeft =
+    fppp_p("Stream.foldLeft", "acc", "f", "stream",
+           (acc: Value, f: Value, s: Stream[Value]) =>
+             s.foldLeft(acc)(
+               UnisonToScala.toUnboxed2(f.asInstanceOf[Lambda])(env))
+           // todo: env needs to be bigger here
```
aryairani: I don't remember exactly what I meant by "bigger" here; bigger than 20, maybe?  But I do find it a little sketchy that I'm allocating new stacks here and there.

runtime-jvm/main/src/main/scala/Builtins.scala
```diff
@@ -11,33 +11,78 @@ import org.unisonweb.util.{Sequence, Stream, Text}
...
+  }
 
   val streamBuiltins = Map(
     Stream_empty,
     Stream_fromInt,
     Stream_cons,
     Stream_drop,
+    Stream_take,
+    Stream_map,
+    Stream_foldLeft,
```
aryairani: Add some more builtins?

PR #181, opened 2018-05-17
implement/test `flatMap`/`filter` and relax type constraints on `filter`/`choose`

aryairani:


aryairani:
Let's merge this.  The CI builds are only failing because the CI script isn't up-to-date relative to the Haskell deleting and cleanup of 42ea0604d26847406d73ae9120a534319b381ee6.  Also, let's fix that :)

PR #183, opened 2018-05-25
fix & speed up travis script

aryairani:
* Switch away from using `apt` and hard-coded paths in the travis script; use `stack` to install and cache per-user `ghc` distro.
* Run `runtime-jvm` tests in addition to Haskell tests.
* Reintroduce `sbt-coursier` plugin to hopefully speed up sbt dependency resolution.
* Brings CI time from ~15min to ~3min

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,25 +14,37 @@ cache:
 
 before_cache:
 # Cleanup the cached directories to avoid unnecessary cache updates
-- find $HOME/.ivy2/cache -name "ivydata-*.properties" -print -delete
-- find $HOME/.sbt        -name "*.lock"               -print -delete
+- find $HOME/.ivy2/cache -name "ivydata-*.properties" -delete
+- find $HOME/.sbt        -name "*.lock"               -delete
+# diagnostic stuff, temporary
+- find $HOME/.stack
+# - find $HOME/.ivy2/cache
+# - find $HOME/.sbt
+# - ls -ld /opt/ghc/*
 
 before_install:
 # Download and unpack the stack executable
 - mkdir -p ~/.local/bin
 - export PATH=$HOME/.local/bin:$PATH
 - travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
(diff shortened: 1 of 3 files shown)


pchiusano:
Maybe still pass that skip-ghc-check flag to stack build? Might prevent an unnecessary download of ghc

aryairani:
Looks like the download of ghc might be necessary now?  Could you take a look at the latest log and see what you think.  Complains about`Semigroup`.

pchiusano:
Ok I guess let’s go back to the version that worked, even though it looked like it was downloading and installing ghc on every build. Builds will just take longer.

Seems like there ought to be a way to only download if needed.

pchiusano:
Also I like having both pr and push builds - if push succeeds and pr fails, that tells you useful info when reviewing the PR.

aryairani:
I think it is possible to cache the installed ghc after it's done once, but it looked like a headache to figure out which paths to cache to achieve that, especially since I don't know an easy way to ssh into the CI build environment to tinker around.

WRT pr vs push build checks, when are they different?  I guess it would be that push represents the branch itself, and the pr is whatever the result of the merge would be?

aryairani:
It looks like 42ea0604d26847406d73ae9120a534319b381ee6 switched us from ghc-8.0.1 to ghc-8.4.2.  The ubuntu package is 8.0.1 (gets used when `--no-build-ghc`), but 8.4.2 isn't backwards compatible.  This wasn't caught by CI because I don't think there was PR for it. :-)

So we either have to 
- figure out why stack can't / doesn't cache its build of ghc-8.4.2
- set up a custom PPA for 8.4.2 (requires sudo? dunno the implications of enabling sudo in `.travis.yaml`)
- build 8.4.2 with stack each time
- revert to 8.0.1 (our macs are on 8.4, so that could mean stack having to build ghc locally)

aryairani:
I guess `sudo: false` allows containerized nodes instead of VM nodes.

aryairani:
@pchiusano This is ready to merge now.  Let me know if you want me to squash.

aryairani:
Accidentally / purposely left `$HOME/.coursier` cache entry, although it's not being used now that we've re-removed the plugin.

PR #184, opened 2018-05-26
try using ppa to get ghc-8.4.3

aryairani:


aryairani:
just going to use `stack` and not `apt`

PR #191, opened 2018-06-14
Patterns now checked against scrutinee type

pchiusano:
Without this PR, you could write things like:

```Haskell
case 42 of
  Optional.None -> true
  Optional.Some _ -> false
```

and it would still typecheck. Now, you'll get a type error as expected. Yay! (Error messages are terrible though - we will need to do a pass to improve that)

The main change is [here](https://github.com/unisonweb/unison/compare/topic/typechecker...wip/pattern-typechecking?expand=1#diff-ba491a6b28216fbef57af9bf084d0d5cR625), just adding an `Term.ann patTerm scrutineeType`, where before we just had the `patTerm` without any annotation.

In order to make this possible, there were a bunch of rather mechanical changes:

* We were using the `Term v` type alias everywhere, which has a single type parameter `v`, which is the type of both _term variables_ (like you see in the body of a lambda) and _type variables_ (like you see in a `forall` that is attached to a term as a type annotation). This requirement that both kinds of variables be of the same type needed to be relaxed.
  * [`Term v`](https://github.com/unisonweb/unison/blob/88d62e43c4e936dae2a8eec8e3ec019129d38132/parser-typechecker/src/Unison/Term.hs#L94) is an alias for `ABT.Term (Term.F v) v ()` - an annotated ABT where the annotation is `()` and the term and type variables are the same - `v`.
  * There is the alias `Term' vt v`, which allows the `vt` (variable type used for types appearing in term annotations) and `v` (used for term variables) to be different.
  * And there is the alias `AnnotatedTerm' vt v a` which is like `Term' vt v` but has `a` as the annotation type at each node in the ABT.
* So I moved to using `Term' vt v` or `AnnotatedTerm' vt v a` everywhere. There were a bunch of random helper functions and smart constructors that needed updating / more general type signatures.
* Locally within the typechecker, I define the alias [`Term v = Term' (TypeVar v) v`](https://github.com/unisonweb/unison/compare/topic/typechecker...wip/pattern-typechecking?expand=1#diff-ba491a6b28216fbef57af9bf084d0d5cR49). That is, the typechecker is expecting terms whose type annotations can have existential or universal type vars. Using this type alias means that none of the type signatures in the typechecker needed to change and you can see the diff in that file is pretty small.
* Then I just fixed all the type errors.

### DIFF
--- a/parser-typechecker/src/Unison/ABT.hs
+++ b/parser-typechecker/src/Unison/ABT.hs
@@ -224,19 +224,39 @@ freshNamed' used n = fresh' used (v' n)
 -- | `subst v e body` substitutes `e` for `v` in `body`, avoiding capture by
 -- renaming abstractions in `body`
 subst :: (Foldable f, Functor f, Var v) => v -> Term f v a -> Term f v a -> Term f v a
-subst v r t2@(Term fvs ann body)
+subst v r t2 = subst' (const r) v (freeVars r) t2
+
+-- Slightly generalized version of `subst`, the replacement action is handled
+-- by the function `replace`, which is given the annotation `a` at the point
+-- of replacement. `r` should be the set of free variables contained in the
+-- term returned by `replace`. See `substInheritAnnotation` for an example usage.
+subst' :: (Foldable f, Functor f, Var v) => (a -> Term f v a) -> v -> Set v -> Term f v a -> Term f v a
+subst' replace v r t2@(Term fvs ann body)
   | Set.notMember v fvs = t2 -- subtrees not containing the var can be skipped
   | otherwise = case body of
-    Var v' | v == v' -> r    -- var match; perform replacement
+    Var v' | v == v' -> replace ann -- var match; perform replacement
            | otherwise -> t2 -- var did not match one being substituted; ignore
-    Cycle body -> cycle' ann (subst v r body)
+    Cycle body -> cycle' ann (subst' replace v r body)
     Abs x _ | x == v -> t2 -- x shadows v; ignore subtree
     Abs x e -> abs' ann x' e'
-      where x' = freshInBoth r t2 x
+      where x' = fresh t2 (fresh' r x)
             -- rename x to something that cannot be captured by `r`
-            e' = if x /= x' then subst v r (rename x x' e)
-                 else subst v r e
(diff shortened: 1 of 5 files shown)


aryairani:
:+1: Thanks for comprehensive PR description.

parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -34,7 +34,7 @@ import qualified Unison.Note as Note
 import           Unison.Pattern (Pattern)
 import qualified Unison.Pattern as Pattern
 import           Unison.Reference (Reference)
-import           Unison.Term (Term)
+-- import           Unison.Term (Term)
```
aryairani: delete?

parser-typechecker/src/Unison/ABT.hs
```diff
@@ -224,19 +224,39 @@ freshNamed' used n = fresh' used (v' n)
...
 -- renaming abstractions in `body`
 subst :: (Foldable f, Functor f, Var v) => v -> Term f v a -> Term f v a -> Term f v a
-subst v r t2@(Term fvs ann body)
+subst v r t2 = subst' (const r) v (freeVars r) t2
+
+-- Slightly generalized version of `subst`, the replacement action is handled
+-- by the function `replace`, which is given the annotation `a` at the point
+-- of replacement. `r` should be the set of free variables contained in the
+-- term returned by `replace`. See `substInheritAnnotation` for an example usage.
+subst' :: (Foldable f, Functor f, Var v) => (a -> Term f v a) -> v -> Set v -> Term f v a -> Term f v a
```
runarorama: Is there a coherency problem here? The `Set v` should be computed from the `a`, but that's not enforced. When would the `Set v` be different from the `freeVars` of whatever gets passed to `replace`?

parser-typechecker/src/Unison/Typechecker/Context.hs
pchiusano: Sure

parser-typechecker/src/Unison/ABT.hs
aryairani: Should we choose a naming convention for pseudo-private functions?  A convention other than "has an ugly name".  I bet we have more instances of pseudo-private functions kicking around.

PR #192, opened 2018-06-14
use updated body of synthesized UnisonFile when serializing

aryairani:
`FileParsers.synthesizeFile` was picking apart a `UnisonFile `, updating its `term` by substing the free vars for builtins, and then returning the new term and synthesized type.  

Bootstrap.hs was using that `synthesizeFile` to typecheck the `UnisonFile` sub, but then serialized the original `UnisonFile` which had the free vars.

This PR introduces a `synthesizeUnisonFile`, which produces an updated `UnisonFile` instead of an updated  `Term`, and updates Bootstrap to serialize the resulting `UnisonFile`.


pchiusano:
Looks good - yea that is a much less error prone approach than what we were doing.

PR #195, opened 2018-06-17
`-Wall` warnings for parser-typechecker bootstrap target

aryairani:


aryairani:
disregard, I'll move this to the PR where the warnings also have been fixed

PR #199, opened 2018-06-19
when checking body of Handle, the body is checked against a type `{e}…

pchiusano:
We were forgetting to do this. 

Need to do something similar in `synthesize`, except for `synthesize`, before returning we have to strip out the effect that is removed by the handler.


pchiusano:
Also that one test `heff` now passes.

PR #200, opened 2018-06-22
Some progress..

pchiusano:
@runarorama we were missing a case in `check`:

``` haskell
      go (Term.Lam' body) (Type.Arrow' i o) = do -- =>I
        x <- ABT.freshen body freshenVar
        modifyContext' (extend (Ann x i))
        let Type.Effect'' es _ = o
        scope ("pushing effects: " ++ show es) . withEffects0 es $ check (ABT.bind body (Term.var x)) o
        modifyContext (retract (Ann x i))
```

So the body of the lambda gets access to the abilities introduced by the arrow type (and no other abilities - like it cannot access abilities from outer scopes). We'd talked about this IIRC.

We also had [this snippet, which I think was just wrong](https://github.com/unisonweb/unison/compare/wip/typechecker...pchiusano:wip/typechecker2?expand=1#diff-ba491a6b28216fbef57af9bf084d0d5cL582).

I got the tests passing after that and decided to write some more tests, which uncovered something very wonky I don't understand yet - this (incorrect) program somehow typechecks (you can run via `stack exec tests "typechecker.--Type.apply"):

```Haskell
  , bombs [r|--Type.apply
            |type List a = Nil | Cons a (List a)
            |
            |map : ∀ a b . (a -> b) -> List a -> List b
            |map f as = case as of
            |  List.Nil -> List.Nil
            |  List.Cons h t -> List.Cons h (map f t) -- should not typecheck, missing (f h)
            |
            |-- definitely should not typecheck!
            |map2 : ∀ a . a
            |map2 = map
            |
            |c = List.Cons
            |z = List.Nil
            |
            |ex = c 1 (c 2 (c 3 z))
            |
            |pure-map : List Int64 -- should fail, output is a `List Text`
            |pure-map = map (a -> "hi") ex
            |
            |()
            |]
```

### DIFF
--- a/parser-typechecker/src/Unison/ABT.hs
+++ b/parser-typechecker/src/Unison/ABT.hs
@@ -394,6 +394,6 @@ instance (Show1 f, Var v) => Show (Term f v a) where
   -- annotations not shown
   showsPrec p (Term _ _ out) = case out of
     Var v -> (show v ++)
-    Cycle body -> showsPrec p body
+    Cycle body -> ("Cycle " ++) . showsPrec p body
     Abs v body -> showParen True $ (Text.unpack (Var.shortName v) ++) . showString ". " . showsPrec p body
     Tm f -> showsPrec1 p f
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -243,8 +243,10 @@ modifyContext' f = modifyContext (pure . f)
 appendContext :: Var v => Context v -> M v ()
 appendContext tl = modifyContext' (\ctx -> ctx `append` tl)
 
-scope :: String -> M v a -> M v a
-scope msg (M m) = M (\menv -> Note.scope msg $ m menv)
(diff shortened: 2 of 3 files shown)


pchiusano:
At least one bug, it looks like variables introduced by patterns don't have their types refined at all, for instance this program typechecks -

```Haskell
type List a = Nil | Cons a (List a)

head' : ∀ a b . List a -> b
head' as = case as of
  -- this typechecks!!
  List.Cons h t -> t
  -- these cases don't typecheck as expected:
  -- List.Cons h t -> "hi"
  -- List.Cons 10 t -> t
  -- Woot.Woot w -> _
```

I feel like this should be straightforward to track down, has to be something with how we are typechecking patterns.

pchiusano:
ITS WORKING!!!!!1!

The issue was that `_` was synthesized to `forall a . a`. Blanks are used for pattern variables so these pattern variables were too unconstrained - correct thing to do is to instantiate pattern variables as existentials whose types will get fixed to some concrete type.

Tests all pass now, I think this is okay to merge.

PR #201, opened 2018-06-23
Fix remaining bugs re: effect checking and add nullary effects

pchiusano:
This PR is a bit of a mess, but wanted to get all the latest stuff merged. Tests all pass and I haven't been able to find any more typechecking bugs. A change is that existentials can now be instantiated to effect types - we didn't have logic for this before which was resulting in counterintuitive ability check failures. There was also an issue with checking of `EffectBind r cid args k` - the `k` continuation in an effect bind should be of the form `exists a . a -> {e} a` for the same effect type as is produced by the constructor.

Also this works now:

```Haskell
effect IO where 
  launch-missiles : {IO} ()

foo : () -> {IO} ()
foo unused = IO.launch-missiles -- no parens needed!
```

If you don't want to evaluate `IO.launch-missiles`,  you have to guard it with a lambda.

### DIFF
--- a/parser-typechecker/src/Unison/Type.hs
+++ b/parser-typechecker/src/Unison/Type.hs
@@ -73,10 +73,6 @@ arity (ForallNamed' _ body) = arity body
 arity (Arrow' _ o) = 1 + arity o
 arity _ = 0
 
-unEffect0 :: Type v -> ([Type v], Type v)
-unEffect0 (Effect' es t) = (es, t)
-unEffect0 t = ([], t)
-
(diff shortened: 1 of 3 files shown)


aryairani:
Which parens aren't needed?  The signature looks like a lambda.

pchiusano:
@aryairani - the IO.launch-missiles effect. It used to be that all effect constructors needed to take at least one arg.

The foo definition is a lambda because top level definitions don’t have access to any effects.

Just noticed a typo in foo, fixing.

PR #205, opened 2018-06-25
merge master into wip/typechecker

aryairani:
specifically to merge #202 into wip/typechecker to restore CI


aryairani:
CI still fails, but now due to `typechecker.--handle inference` test, rather than broken stackage.

PR #216, opened 2018-07-07
Streams allocate a single extra stack when consumed, + stuff

aryairani:
Added/tested:
```haskell
Stream.single :: a -> Stream a
Stream.constant :: a -> Stream a
Stream.take-while :: (a -> Boolean) -> Stream a -> Stream a
Stream.drop-while :: (a -> Boolean) -> Stream a -> Stream a
Stream.flat-map :: (a -> Stream b) -> Stream a -> Stream b
Stream.iterate :: forall a . a -> (a -> a) -> Stream a
Stream.reduce :: forall a . a -> (a -> a -> a) -> Stream a -> a
Stream.to-sequence :: forall a . Stream a -> Sequence a
Stream.filter :: forall a . (a -> Boolean) -> Stream a -> Stream a
Stream.scan-left :: forall a b . b -> (b -> a -> b) -> Stream a -> Stream b
Stream.sum-int64 :: Stream Int64 -> Int64
Stream.sum-uint64 :: Stream UInt64 -> UInt64
Stream.sum-float :: Stream Float -> Float
Stream.append :: forall a . Stream a -> Stream a -> Stream a
Stream.zip-with :: forall a b c . (a -> b -> c) -> Stream a -> Stream b -> Stream c
Stream.unfold :: forall a b . (a -> Optional (b, a)) -> b -> Stream a
```
Using `Stream.unfold` requires a known `Optional` type and known `Optional.Some` and `Optional.None` constructors, so I updated`Builtin.builtinTerms` to derive these constructors from `Builtin.builtinDataDecls :: Map.Map R.Reference (DataDeclaration v) `.  

I also updated `builtinDataDecls` to use a parser instead of hard-coded ABTs.  **Note**: Our parsers currently can't parse `type () = ()`, so this remains as a hard-coded ABT.

On the Scala side, I updated the runtime representation of Streams to `Env => Stream[Value]`, and added the `StackDecode` typeclass for parameters that need a unison runtime environment to decode to a Scala value.  I fixed a bug in `Stream.flatMap`, added pretty-printing for Unison tuples and sequences, and added implicit syntax for `Seq[A <% Term] => Term` for comparing Unison terms to Scala values.


aryairani:
Whoops, I forgot to update the benchmarks project.

aryairani:
@pchiusano Oh yeah, I forgot to update the title of the PR from "allocate 1 extra stack" to "allocate 0 extra stacks", which I got working in a later commit.

PR #217, opened 2018-07-20
Refactor typechecker to track location information

pchiusano:
Previously, typechecker worked on `Term v` values, with no location information, and with errors just reported as strings. Now it works on `AnnotatedTerm v loc` values and reports errors as a data type, with location information, and can recover from and report multiple errors (though we haven't implemented that yet, but doing so should be some surgical edits to `Context.hs`).

We made what we thought were reasonable choices re: propagation of location info during typechecking but weren't able to test this until we can feed it a location-annotated tree like what new parser will produce.

`Result.hs` is used for possibly failing computations that may log multiple errors. We are using it in a number of places, like the typechecker.

``` Haskell
-- Comes equipped with a monad
data Result note a = Result { notes :: Seq note, result :: Maybe a }

data Note v loc
  = Parsing String
  | InvalidPath Path (Term v loc)
  | UnknownSymbol v loc
  | UnknownReference Reference
  | Typechecking (Context.Note v loc) deriving Show
```

We work with a `Result (Note v loc) a` in various places.

@runarorama I'm not sure what sort of errors will come out of the parser, but we will replace the `Parsing String` there with `Parsing ParseError` for whatever richer parse error type.


aryairani:
👍

pchiusano:
@aryairani have a quick look at last few commits, I generalized `Result` to not care what kind of notes were being logged:

``` Haskell
data Result note a = Result { notes :: Seq note, result :: Maybe a }
-- previously Result v loc = Result { notes :: Seq (Note v loc), result :: Maybe a }
```

And then also revamped the `Context.Note` type to be in a more convenient form for processing. The previous representation was super non-uniform which I realized would be annoying to work with. New representation has the root cause always available in the same spot, with path to that cause as a regular `Seq`:

```Haskell
data Note v loc = Note { cause :: Cause v loc, path :: Seq (PathElement v loc) } deriving Show

data PathElement v loc
  = InSynthesize (Term v loc)
  | InSubtype (Type v loc) (Type v loc)
  ...

data Cause v loc
  = TypeMismatch (Context v loc)
  | IllFormedType (Context v loc)
  ...

-- Add `p` onto the end of the `path` of this `Note`
scope' :: PathElement v loc -> Note v loc -> Note v loc
scope' p (Note cause path) = Note cause (path `mappend` pure p)

-- Add `p` onto the end of the `path` of any `Note`s emitted by the action
scope :: PathElement v loc -> M v loc a -> M v loc a
scope p (M m) = M go where
  go menv =
    let (notes, r) = m menv
    in (scope' p <$> notes, r)
```

I got rid of the various `withinBlah` functions. `withinCheck e t a` became `scope (InCheck e t) a`, etc.

aryairani:
@pchiusano Cool yeah I meant for us to generalize Result like that but got distracted and forgot.

pchiusano:
Okay merging!

PR #219, opened 2018-07-22
Make 'sbt test' run the tests.

paulp:
Maybe it's just me, but I hate when 'sbt test' succeeds vacuously despite there being actual tests.

### DIFF
--- a/runtime-jvm/build.sbt
+++ b/runtime-jvm/build.sbt
@@ -37,6 +37,7 @@ lazy val testSettings = Seq(
 
 lazy val root = project.in(file(".")).
   settings(name := "unison-runtime-root").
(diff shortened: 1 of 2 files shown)


runtime-jvm/build.sbt
```diff
@@ -37,6 +37,7 @@ lazy val testSettings = Seq(
 
 lazy val root = project.in(file(".")).
   settings(name := "unison-runtime-root").
+  settings(test := (run in Test in main toTask "").value).
```
dwijnand: Even better, wire this in `main` (and drop the `main` reference in the RHS), and the rest will be done by the aggregation.

aryairani:
Thanks, didn’t realize it could be done in one line.

PR #222, opened 2018-08-02
A whole bunch of language tweaks and syntax updates

pchiusano:
Here's a quick summary. We can try out new syntax changes and make tweaks, I would not consider any of this set in stone. Open to bikeshedding on the syntax if anyone has other ideas.

* Syntax sugar for delayed computations (`() -> a`): `'a`, `!a`, and `delay <layout block>`
* `Sequence a` type can be written as `[a]`.
* Added some missing `Sequence` functions (`drop`, `++`).
* Type signatures are generalized over any lowercase free variables. Example: `a -> a` becomes `forall a . a -> a`.
* `namespace` blocks, ex: `namespace Foo where blah = 24` is a circuitous way of saying `Foo.blah = 24`. They can be nested.
* imports, via the `use` keyword, syntax `use Math.sqrt` to allow `sqrt` to be referenced unqualified, other examples: `use Sequence.++` (import a single operator), `use Sequence.++ map filter` (an operator and some other names). One or more `use` statements can appear at start of any block. No wildcard imports yet. Later imports shadow earlier ones.

## Some more notes on the syntax for delayed computations

* `'43` is syntax for `() -> 43`. It has highest possible precedence, so `get-or-else '43 x` parses as `get-or-else ('43) x`. You might use the idiom `'(foo x)` if you want to pass around a delayed call to `foo x`. 
* The _type_ `'x` is sugar for `() -> x`. It has higher precedence than `->` but lower precedence than application, so the type `fork : '{Remote} a -> {Remote} (Future a)` parses as `('{Remote} a) -> {Remote} (Future a)`.
* `!x` is syntax for `x()`. It has same precedence as `'`, so `at !spawn '(factorial 22)` parses as `at (spawn ()) (() -> factorial 22)`.
* `delay` is like `let`, it introduces a layout block, but wraps the whole block in a delay. Example:

```
go : '{IO} ()
go = delay
  x = 1
  launch (x + 1)
```

I feel like we might want to give `delay` higher precedence than function application, so you can say things like `fork delay <layout block on next line>`.

Implementation notes - 

* All the syntax changes were just parser changes, no AST or typechecker changes.
* It is awkward having to know the `Reference` and constructor id of any pattern - it means you need to know about all data types before you start parsing terms. This is responsible for the restriction that `use` statements have to appear first in a block.
* We might want delays to have a separate AST node rather than just being syntax sugar. This would be an easy refactoring for later.


pchiusano:
@aryairani build is failing due to 

```
    /Users/pchiusano/Dropbox/projects/unison/parser-typechecker/src/Unison/Type.hs:107:12: error: [-Wincomplete-patterns, -Werror=incomplete-patterns]
        Pattern match(es) are non-exhaustive
        In a case alternative: Patterns not matched: []
        |
    107 | unApps t = case go t [] of [_] -> Nothing; f:args -> Just (f,args)
        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

aryairani:
@pchiusano Sorry not sure how that slipped by.  It is fixed now but something else is broken.

aryairani:
@pchiusano It's now failing because `cannot serialize program with blanks`

aryairani:
in https://github.com/unisonweb/unison/blob/master/unison-src/tests/stream/unfold.u

Any ideas?

aryairani:
@pchiusano Could you add test programs that exercise these features to `unison-src/tests/`?  They will be picked up by the `runtime-jvm` test runner.  You can make use of nested directories under `unison-src/tests/` if desired.  

(We should also have some negative tests, and there is already a `unison-src/errors/` directory, but it isn't wired in for testing yet; I've just been using them for manual inspection of parse/type errors.)

aryairani:
@pchiusano By default, the files under `unison-src/tests/` are just checked to make sure that they evaluate. (To anything.)  To check that they evaluate to some expected result (recommended), add them to [FileCompilationTests.checkResultTests](https://github.com/unisonweb/unison/blob/a91b6c81a81098423adacaeeb6b54a0f16f86030/runtime-jvm/main/src/test/scala/FileCompilationTests.scala#L14).

parser-typechecker/src/Unison/Parser.hs
```diff
@@ -33,8 +33,31 @@ import qualified Unison.Var as Var
 debug :: Bool
 debug = False
 
-type PEnv = UnisonFile.CtorLookup
-type P v = P.ParsecT (Error v) Input ((->) PEnv)
+data PEnv v = PEnv { constructorLookup :: UnisonFile.CtorLookup }
```
aryairani: Phantom type v?  How is it used, or is its use coming later?

aryairani: Maybe it's just for the Monoid / Semigroup?  I did the same thing for `Rendered`.

atacratic:
Looking at the `fork` example above, I wonder if the ' apostrophe syntax is definitely better than the syntax used by Frank?  Frank's use of 'suspenders' { } for delay makes it easy to see what is being delayed without parsing for precedence.

I think Frank's rendering of fork would be

`fork : { [Remote] a } -> [Remote] (Future a)`

(Excuse my gatecrashing this PR!  Love all the progress you're making - really seems to be steaming along at the moment 😀)

aryairani:
@atacratic Gatecrashing is welcome 😁  We are pondering the suggestion now.

pchiusano:
Yeah, we are currently using `{}` for term level effect values, but maybe can change that syntax or overload it somehow...

pchiusano:
hey going to merge this. @atacratic thanks for comments, that file was a major work in progress and not surprised parts of it were nonsense. :) 

Re: syntax, I got rid of `delay` keyword, just have `'` bind looser than keyword-based layout constructs like `let`, `case`, `handle`. So you can say `spawn 'let <blah>` or `spawn 'case x of`, no special rule needed.

PR #237, opened 2018-08-24
Parsing improvement - type and effect decls can be intermingled with term definitions, `use` can appear anywhere in a block

pchiusano:
Currently, the file parser had a rigid structure where it requires the type and effect declarations to appear first in the file, followed by term definitions. Also the basic term block parser requires the `use` statements all at the start of the block. 

_This PR relaxes all that._ You can order the declarations (of terms or types) however you like in the file, and you can add add ` use` statement in the middle of a block.

The file parser is unchanged, it's implemented via a normalization pass in the lexer, which brings the type and effect decls to the top of the file, and also reorders the `use` statements within a block to appear at the start. (Am undecided about the `use` reordering, but easy to turn that off if we don't like it.)

The function `Unison.Lexer.lexer` does this normalization, and `lexer0` is the non-normalizing implementation if you ever want that. All code is now going through this new implementation and all tests pass. 

If you're curious about implementation, [start reading here](https://github.com/unisonweb/unison/blob/fb44a09d533d8b9c976149c06d972c2175a2869d/parser-typechecker/src/Unison/Lexer.hs#L116). Lexer output is parsed into a tree structure based on matching `Open` and `Close` tokens, and we can then [easily reorder `Semi` delimited sections within this tree based on their starting token](https://github.com/unisonweb/unison/blob/afb98b5354291953ff8cbddc54a66c94a273f58f/parser-typechecker/src/Unison/Lexer.hs#L165).

Other changes -

* Shuffled some of the declarations around in the tests to get some testing of this feature.
* The `debugLex` functions can be a bit simpler and just use the `Show` instance for `Lexer.T`, the tree structure used to implement the above.


atacratic:
Oops, finger slipped!

PR #238, opened 2018-08-25
A few failing typechecker tests

atacratic:
- tests/delay_parse.uu - () -> {T} '() does not parse - interaction between the ' and the {T}
- tests/match_unit.uu - can't match on () in a pattern
- tests/hole.uu - holes don't seem to be supported yet
- errors/effect_unknown_type.uu - no error when using an unknown type in an effect declaration

These are the things I found while getting the following to compile, which it now does! :-)  It's a sketch of how the Remote API might break down in to constituent effects.

https://gist.github.com/atacratic/99fb707a7cb96bcd714cc1c024a7e08e


atacratic:
OK, cool.  Have removed the dud tests.  Thanks for the info about foralls and patterns.  

I think it would be convenient if bindings with signature but no definition were allowed.  (Even if those bindings couldn't then be used in other bindings.)  It's helpful to be able to compile unfinished work.  Won't lose sleep over it though!

atacratic:
I'm going to cross my fingers and hope that I have permission to merge :)

pchiusano:
Thanks @atacratic. For uncontroversial PRs like this that have already gotten review I think okay to go ahead and merge, but I like practice of saying “I’m merging this tomorrow if no objections” so there’s no surprise. If there’s any controversy or outstanding discussion then I’d hold off until there’s an explicit “ok to merge”. 😀

Realizing we should put this in a contributing doc of some sort.

Appreciate you diving in and experimenting with what we have so far!! 😀 Keep it coming. Would you be interested in taking on a bigger project that deals with compiler internals? I’ve got some ideas...

atacratic:
OK, got it on the merge process.

Interested but slightly daunted!  I'll drop you an email...

PR #239, opened 2018-08-27
Add `expectEqual` to EasyTest.

tmciver:
I started using EasyTest in a personal project and found I needed a function like this.  Is this reasonable or is there another way to achieve this?


pchiusano:
Hey @tmciver! Seems reasonable, merging.

PR #246, opened 2018-09-05
Topic/broken continuations

pchiusano:
Unpending'd some tests. Fixed a couple bugs:

* Fixed a bug with `compose`, which was used for building up continuations inside the `Requested` exceptions. `compose` wasn't handling `Requested` exceptions internally, which resulted in portions of the continuation getting discarded. Hat tip to @stevekrouse for helping to notice this issue.
* Moved `ANF` call to `compileTop`, as programs that use effects need to be ANF'd to work correctly.
* Fixed a bug with let rec minimization that could scramble the order of actions (unnamed bindings) if the dependencies did not force some other order.
* Tuple pretty-printer on Scala side doesn't crash on ANF'd tuples.


pchiusano:
Merging this...

PR #250, opened 2018-09-21
Type pretty-printer

atacratic:
Can I merge the type pretty-printer?  This is also the merge of Arya's port of the pretty-printing layout code.

* It's had all the testing I'm planning on doing.  
* It has some pending tests, which I want to leave til I've got at least a basic Term printer done.
   * Including some which boil down to 'write functions which reverse generalizeEffects and generalizeLowercase', which looks like it will be tricky...
   * And some which are due to lexer issues which are on my list to fix.
   * And a couple which look like they could be issues in the pretty-printer layout code, which I will look into.
* It passes a round-trip test on all the types in the test suite (apart from a few which fail on one of the pending issues mentioned above.)
* It has some test code in Builtin.hs which you may want me to remove.  (If so actually I'd rather commit it and then commit its removal separately so I can reuse the commit later.)

In the mean time I'll make a start on the Term pretty-printer.


atacratic:
Thanks for the review @pchiusano - I will merge this tomorrow unless anyone shouts.

pchiusano:
Thanks Chris!! Merging now...

PR #252, opened 2018-09-24
Spreadsheet-like file watcher

pchiusano:
screen shot 2018-09-24 at 2 39 55 pm

This is hooked up to the `stack exec watcher` executable. General idea: you can put `>` before any line in the program to 'watch' that line, and on each save, the result of that line's expression will be evaluated and printed out as shown. It's incredibly snappy (so snappy that you sometimes don't even notice it's updated) and gives me a feeling like I'm editing a spreadsheet... only it's a spreadsheet with access to a higher order language! Moreover, I get to keep using my favorite text editor to get this interactivity.

The idea is that this sort of thing totally replaces any need for a REPL. I think REPLs have a poor DX, and they aren't how you'd design an interactive programming environment today, even if you are sticking to just plain text.

Also added recently, you can use `(expr)?` and save the file to learn the type of `expr` and any constraints on replacement expressions:

screen shot 2018-09-24 at 11 57 33 pm

For later:

* We could do more to avoid redundant evaluations.
* It would be great if we can optionally obtain a full execution trace for watched expressions. This is potentially a huge amount of data, so it would need to return some sort of value that user could then slice and dice in various ways to explore it.
* A bit easier - it would be great if you showed not just the line the expression comes from, but the expanded expression with free variables all substituted away.


pchiusano:
Going to merge this tomorrow if no objections.

unison-src/tests/language-reference.u
```diff
@@ -32,22 +32,22 @@ ex =
...
+    part1 ++ part2
 
   number = 42.24 + 92.283
 
   (msg, number)
 
--- The type `(Text, Float)` is a pair of a text value and a floating point number. The type `(Text, Int64, Text)` is, yup, a triple of a text value, an `Int64` value, and another `Text`. And so on: the syntax for tuples of any size is just a comma separated list in parentheses.
+-- The type `(Text, Float)` is a pair of a text value and a floating point number. The type `(Text, Int, Text)` is, yup, a triple of a text value, an `Int` value, and another `Text`. And so on: the syntax for tuples of any size is just a comma separated list in parentheses.
 
 -- The `concatenate part1 part2` is the Unison syntax for function application. The arguments are separated by spaces. This syntax binds tighter than any infix operator, so for instance, in this program:
```
TomasMikula: `concatenate` is no longer present after this PR.

PR #259, opened 2018-10-11
TDNR resolves definitions in the same source file

runarorama:
This change allows TDNR to resolve situations like:

```
Local.foo x = x + 1

> foo 99 
```


pchiusano:
Nice!

PR #269, opened 2018-12-06
Pretty-printer (kind of) finished

atacratic:
Please can I merge this stuff?  

The final commit is me running the key files through an auto-formatter - to see the diff without that noise look [here](https://github.com/unisonweb/unison/compare/topic/codebase-editor...c3f3933).

Sorry, I did something screwy with git at some point so there are a bunch of duplicate commits on the branch.  

It's kind of finished in that
- I don't know of anything that it can't render in some kind of semantically correct way
- the known issues are pretty much all blocked on niggles in the parser (TODOs in the test files).

I'm sure there are things that could be improved in terms of the syntax and layout of the output it emits.  ([This page](https://github.com/lspitzner/brittany/blob/master/doc/showcases/Layout_Interactions.md) from the brittany haskell auto-formatter was thought-provoking.)

It doesn't attempt to emit `use` directives, which it could do in order to avoid lots of repetitive fully-qualified names.  I think that's what I ought to tackle next, unless someone has better ideas.  Haven't got any thoughts yet on how to approach it so any advice welcome.


pchiusano:
👍 Yes, merging. Am slightly annoyed that we broke the build on topic/codebase-editor, I'll probably fix that up after just so we can at least run the old codebase editor until wip/codebase-editor is ready to merge back in.

PR #270, opened 2018-12-08
Change comment to match function behavour.

noahhaasis:
`bound'` returns the list of all bound variables whereas `bound` returns the set of all bound variables. Changed the comment accordingly.


aryairani:
I was just thinking we could use more comprehensive docs...

PR #271, opened 2018-12-11
New pretty-printer

pchiusano:
With the work we're doing to produce nicely formatted output in the codebase editor, I decided to do some yak shaving and make the pretty-printing library flexible enough to be used for both user-facing messages and also for pretty-printing types and terms.

The main things missing from the previous library were:

* Support for wrapped text (so you can just type a paragraph and have it automatically flow into whatever width you choose)
* Better support for dealing with explicit line breaks (for things like bulleted lists, two-column lists / tables, etc).
* Colors. You should be able to produce color output (if showing in console) or not (if writing to a file).

The new type is called just `Pretty`, it's in `Unison.Util.Pretty` for now. Really, anywhere you'd be tempted to use string munging to assemble a message or some output, consider using `Pretty` instead. :) I'm pretty happy with how it came out. Here's a [demo file](https://github.com/unisonweb/unison/blob/7a68e3fbe8435aa7efd252028cd34d1d7e50d1e9/parser-typechecker/prettyprintdemo/Main.hs) and see attached output.

screen shot 2018-12-11 at 12 41 54 am

@atacratic I updated the type and term pretty printers to use this new library. It was really helpful having such an extensive test suite! LMK if you have questions about any of the changes I made. I found it was pretty mechanical as a lot of the API is the same or similar as before. One thing I'll point out - there were a lot of places where we were using the idiom `blah <> b " " <> PP.Nest "  " somestuff`, which I replaced everywhere with just `PP.hang blah somestuff`.

If you're curious, peep [the implementation here](https://github.com/unisonweb/unison/blob/71a35c5df2860c5e65574c9608f76fee7004d5d0/parser-typechecker/src/Unison/Util/Pretty.hs#L114)

Idea for a fun task - the pretty printers could easily produce a `Pretty ColorText` rather than `Pretty String`.


aryairani:
Looks great!  But I’m wondering if it wouldn’t be better if wrapped words was the default, and non-breaking was optional.  Also, is fromString redundant here:
```
    PP.indent (PP.red (fromString ">> ")) (PP.wrapWords loremIpsum),
```

aryairani:
Like, I’m thinking all of our output should always be pretty-wrapped at the terminal size, but you should be able to designate substrings that don’t get wrapped in the middle.  Maybe indented or bulleted text would be the same except with width reduced by the bullet/leader width.

pchiusano:
@aryairani No, that `fromString` isn't needed. 

I kinda think we should keep default as is, though I could see the arguments either way. For structured pretty printing like for printing terms and types, you need to be very careful about where you are allowing breaks, so IMO that default isn't really appropriate there IMO (also if you accidentally introduce some breakability it can be kinda annoying to track down, moreso than if you just have a message you forgot to call `wrapWords` on). For printing messages and such, you do sometimes want unbroken segments like you mention (with current API you can create via `wrapWords someStuff <> "An unbroken thing" <> moreStuff`).

PR #274, opened 2018-12-15
Add the language overview to the README

noahhaasis:
Added the reference to the README because I think an overview of the syntax is interesting and it's quite hard to find it otherwise.


aryairani:
Probably many of these docs need a bigger overhaul, but every little bit helps.

PR #275, opened 2018-12-16
Add a nicer error message for empty match lists

noahhaasis:
If a match the match list is empty you know get the error `expecting at least one case in case/of`


runarorama:
Thanks!

PR #281, opened 2018-12-18
Calculate width in watchPrinter

noahhaasis:
The width of the printed expression in watchPrinter now depends on the terminal width.
For these two are equal because I'm unsure what the best width would be.


noahhaasis:
Maybe terminal width minus length of the lead ("          ") would be a good width?

pchiusano:
@noahhaasis we are going to be deleting this executable, functionality has been folded into the codebase editor, so going to close this.

PR #283, opened 2018-12-18
Change filesInPathMatchingSuffix to use System.FilePath.TakeExtension

noahhaasis:


aryairani:
@noahhaasis Would you mind changing the function name from `*MatchingSuffix` to `*MatchingExtension` to match the implementation? :)

aryairani:
@pchiusano @runarorama I ran `stack exec unison` and confirmed that it would load a .u file before merging, but it would be good if we could catch that sort of thing with automated tests.

aryairani:
P.S. Thanks @noahhaasis!

Issue #289, opened 2018-12-21
Parser revamp to improve error reporting


pchiusano:
The goal: do a pass through the parser and generally improve error reporting. At the moment, there's been little thought put into this and the errors are just whatever the default that Megaparsec generates, in whatever state it happens to be in when it fails.

So, there's a couple steps to doing this better:

* I'd probably start by brainstorming a big list of programs with parse errors (at least a dozen to start, and add to it as you think of new cases), and add these to the repo in the unison-src/parser-tests directory. For each file, `foo.u`, create `foo.u.message.txt` (or whatever name) with the current parse error (and you can add a note here about what you think could be improved about it). Suggest writing a little main function to generate all this info for a directory full of .u files.
* Armed with that, go through the grammar and do some work to encode error states in a more deliberate way. This will require a good understanding of how Megaparsec generates errors (the most important thing to know is that Megaparsec will commit to a branch after a single token is consumed, unless there is a use of `try`, which is used in like 2 places in the grammar).
* Use a custom error rather than doing string munging to produce a message inside the parser. Here's an example of [how you can generate a custom error](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/TermParser.hs#L424). Just add to [this list of parse errors](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Parser.hs#L37) as needed.
* Very often, you'll just be adding a custom failure at the end of a chain of `<|>` possibilities. __I wouldn't bother with using the label / `<?>` function__ - just have the custom failure have enough info in it to craft a nice message.
* Each time you add a new error constructor, you might get a warning in `PrintErrors` about it not being handled. This is the place to do whatever string munging you want to actually construct the error message that is shown to a human. :)
* As you improve the messages, regenerate those files containing the messages so we can all see how things are progressing. Once we're happy with the messages, we can celebrate and close this issue.  🎉 
If / when we later encounter a poor error message, we can add the example to that same directory and then create an issue to improve it.

@noahhaasis I think this is a good first issue, LMK if you have any questions or want help getting started. :)

PR #291, opened 2018-12-22
Parser revamp

noahhaasis:
Implemented the directory and script described in issue #289. Is this how you imagined it @pchiusano?
I will add more error message files later.

### DIFF
--- a/unison-src/parser-tests/GenerateErrors.hs
+++ b/unison-src/parser-tests/GenerateErrors.hs
@@ -1,0 +1,38 @@
+{- For every file foo.u in the current directory write the parse error to foo.message.txt -}
+module GenerateErrors where
+import qualified Data.Text                as Text
+import           Data.Text.IO             ( readFile )
+import           Prelude           hiding ( readFile )
+import           System.Directory         ( listDirectory, getCurrentDirectory )
+import           System.FilePath          ( takeExtension, dropExtension )
+import           System.IO                ( putStrLn )
+import qualified Unison.Builtin           as B
+import           Unison.Parser            ( Err )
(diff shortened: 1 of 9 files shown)


unison-src/parser-tests/GenerateErrors.hs
```diff
@@ -0,0 +1,38 @@
...
+errorFileName :: String -> String
+errorFileName n = dropExtension n ++ ".message.txt"
+
+processFile :: FilePath -> IO ()
+processFile f = do
+  content <- Text.unpack <$> readFile f
+  let res = P.parseFile f content B.names
+  case res of
+    (Left err) ->
+      let prettyErr = B.showParseError "" (err :: Err Symbol)
```
noahhaasis: I'm not sure if I'm using the first argument of showParseError correctly?

unison-src/parser-tests/if-without-condition.message.txt
```diff
@@ -0,0 +1,3 @@
+if-without-condition.u:1:10:
+unexpected then
+    1 | [0m
```
pchiusano: Since parse errors have ansi codes in them (and we want to see them in color), can you modify your script to print out all the messages to the terminal? Still generate the file too but for the file, let’s emit those as plain text.

@aryairani can you show how to display a parse error sans ANSI codes?

pchiusano:
Looking good so far, thanks for diving in and helping!

pchiusano:
@noahhaasis I want to merge this even though it's still in progress. Can you just add yourself to CONTRIBUTORS file?

pchiusano:
👍

PR #292, opened 2018-12-27
Remove unused Language Pragmas

noahhaasis:
Remove all unused Language Pragmas found by hlint.


aryairani:
I’m thinking a better fix would be putting most of the pragmas into the build file.  Thoughts?

noahhaasis:
Would you add it for each executable separately or just to the library section?

aryairani:
OverloadedStrings

runarorama:
Merging this since it's a good change and self-contained.

PR #298, opened 2019-01-18
The `todo` command

pchiusano:
ITS WORKING!! In the screenshotted example, I edited the `Optional` type, so a couple builtin functions that use that type are marked as needing updates.

image

You can now edit definitions that have transitive dependents, and the `todo` command will tell you what the "frontier" of your refactoring is and walk you though which definitions need updating.

Still missing:

* A `dependents`/`usages` command
* An `edit` command, for conveniently starting an edit on a definition (basically `view` into a scratch file)

Kind of interesting - nothing stops you from editing a builtin type or term, though that's probably not something you would commonly do.

### DIFF
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -5,29 +5,35 @@
 {-# LANGUAGE TypeApplications #-}
 module Unison.Builtin where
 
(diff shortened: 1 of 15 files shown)


runarorama:
It shows “1 transitive dependent” even though there are clearly at least 2.

PR #299, opened 2019-01-20
`edit` command

pchiusano:
`edit` behaves just like view, but dumps the contents of the definition to the top of the most recently edited file (otherwise to a file called `<currentBranchName>.u`), "above the fold". Am open to tweaking this behavior.

screen shot 2019-01-20 at 4 18 07 pm

In `1.u`, it generates:

screen shot 2019-01-20 at 4 19 37 pm

One thing I'd like to address later is that the newly edited file triggers a file change event, so the file gets immediately typechecked. Maybe that is okay, but I think it's a little weird. To fix, we'll need some way of telling the code that responds to those file change events that it should ignore the event we just triggered by writing to the file.

Haven't played with it much, but so far, I would say it makes the editing process more fluid. The only thing I dislike is you still have to switch to the appropriate text buffer after you do the edit... it would be cool if somehow that text buffer were just opened automatically, and if it's already open, switch to it.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/CommandLine2.hs
+++ b/parser-typechecker/src/Unison/Codebase/CommandLine2.hs
@@ -72,11 +72,11 @@ notifyUser :: forall v . Var v => FilePath -> Output v -> IO ()
 notifyUser dir o = do
   -- note - even if user's terminal is huge, we restrict available width since
   -- it's hard to read code or text that's super wide.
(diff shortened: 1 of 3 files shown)


aryairani:
Cool. Maybe we just need to change the phrasing to something more neutral, like “foo.u contains the following definitions”.

pchiusano:
@aryairani okay I added a bit of LoopState to prevent the file update from triggering a round of typechecking

Issue #315, opened 2019-02-04
Parser should support hash-qualified names


pchiusano:
This is so that you can write code which references hash-qualified definitions and have it still parse / typecheck.

Needs some thought: what about viewing a hash-qualified definition? What happens when you parse and typecheck it again? I’d say it should probably strip out the hashing from the name (and tell the user it’s done so).

aryairani:
No, instead of stripping them out, the output should warn of use of / list deprecated references.

pchiusano:
@aryairani - Agree on the warning, but I was talking about if the name of the definition itself is hash qualified, as in, the thing to the left of the = sign. You get this when viewing something conflicted..

pchiusano:
Clarifying with an example:

```
edit foo#2823
```

And that dumps a definition to your file...

```
foo#2823 = 42 + 1
```

which you then edit (`42 + 2`) and then `add`.

Maybe we just don't support this, or put the `#2823` as a comment?

pchiusano:
Arya: "Make sure to look up in oldnames too if needed."

aryairani:
Thought on implementation: have the lexer generate hash qualified tokens. Take these tokens, produce a `Set HashQualified`, use some magic function `Set HashQualified -> Branch m -> m Names0` which produces a `Names0` covering the input, by looking through the `Branch` history.

Side note: I'm thinking we will have multiple uses for a function that folds through causal "merge" nodes in parallel (instead of recursively following one path back to the root), with early termination after some condition, such as having found all the things you're looking for.

aryairani:
I'm currently implementing `findHistoricalNames0 :: Set HQ.HashQualified -> Branch m -> m Names0` for #535.

Issue #314, opened 2019-02-04
`list` and `view` commands should support search/display of hash-qualified types


pchiusano:
See this comment: https://github.com/unisonweb/unison/commit/c14cbbfcd35a833d6ba14b993fe06607479074d4#r32141412

aryairani:
Fixed in #354 IMO:
![image](https://user-images.githubusercontent.com/538571/53782016-6a138400-3ed9-11e9-9ea4-ff60e762ec67.png)

Issue #313, opened 2019-02-04
`rename` should support hash-qualified names for the source name (but not for the target name)


pchiusano:
That is, should be able to write:

    > rename foo#9sdjf oldFoo

aryairani:
Still relevant in #354

pchiusano:
Is this done? I thought so... if yes you can close. @aryairani

runarorama:
Is this still relevant?

pchiusano:
```
> rename.term foo.bar#asldfkj

  ⚠️
  
  `rename.term` takes two arguments, like `rename.term oldname newname`.

> rename.term foo.bar#34 baz

  extra characters after foo.bar: "#34"
```

aryairani:
@runarorama We do have a similar problem in branchless, though maybe for a different reason now.

I'd start by looking at how `Path.parseHQ'Split'` behaves on `foo.bar#34`.  Does it return `Right` or `Left`?

aryairani:
Fixed in bfbec5a
![image](https://user-images.githubusercontent.com/538571/60341394-530f0b00-997c-11e9-96f8-0655e5f7c5d9.png)

PR #316, opened 2019-02-04
Support for hash-qualified names in `list` and `view`

pchiusano:
See #314.

There's still a couple loose ends for this: #315 and #313 (I think @aryairani might have one or two more) but I think this is in good shape to merge to master.


pchiusano:
@aryairani any idea what this Travis failure is about?

PR #333, opened 2019-02-11
gives guard expressions access to variables

### LINKED ISSUE #327: Pattern match guard expressions don't have access to variables introduced by the pattern
As I understood unison should support guards on a pattern match but this does not seem to be working

```
f blah =
  case blah of
    x | x == "woot" -> false
    y | y == "foo" -> true
```

```
master> 
I'm not sure what x means at line 135, columns 9-10

  135 |     x | x == "woot" -> false

Whatever it is, it has a type that conforms to Text.
```

francisdb:
fixes #327 

@pchiusano I tried adding a test case to `test/Termparser.hs` but it did not trigger the issue. Are there any similar tests that also invoke the TypeChecker to trigger this issue?


pchiusano:
@francisdb Excellent!! You can just add a `.u` file to `unison-src/tests` - all .u files in that directory will get parsed and typechecked. Could you add that as part of this PR and also add yourself to CONTRIBUTORS file? :)

pchiusano:
👍  thanks @francisdb

PR #335, opened 2019-02-12
Haskell runtime integration

pchiusano:
Still more work to do but this might be close to being ready to merge (but hold off for now). The new runtime is behind a flag, `-haskell`, which you can pass to the `unison` executable.

```
stack exec unison -- -haskell
```

Also in this branch - watch expressions are now part of the file structure and are done in a non-janky way.


parser-typechecker/src/Unison/Codecs.hs
```diff
@@ -4,6 +4,7 @@ module Unison.Codecs where
 
 -- A format for encoding runtime values, with sharing for compiled nodes.
 
+-- import Debug.Trace
```
noahhaasis: I assume this was just for debugging and can be removed

parser-typechecker/src/Unison/Runtime/ANF.hs
```diff
@@ -0,0 +1,145 @@
...
+  go e@(ABT.out -> ABT.Cycle body) = ABT.cycle' (ann e) (go body)
+  go e@(ABT.out -> ABT.Abs v body) = ABT.abs' (ann e) v (go body)
+  go e = e
+
+  -- test for whether an expression `let x = y in body` can be
+  -- reduced by substituting `y` into `body`. We only substitute
+  -- when `y` is a variable or a primitive, otherwise this might
+  -- end up duplicating evaluation or changing the order that
+  -- effects are evaluated
+  canSubstLet (Var' _) _body = True
```
noahhaasis: Since leaves are always pure primitives can Thisbe replaced be a guard `isLeaf e`?

pchiusano: @noahhaasis yeah that sounds right - `canSubstLet` is just `isLeaf`. 

I was thinking we could also do the substitution if the expression is pure (requires no abilities) and only occurs once in the body of the let, but we don't have an easy way to check for purity here.

pchiusano:
Okay great, we got the Travis issues sorted. 

Still a bunch of work to do on this before it's usable but going to merge the WIP.

Issue #339, opened 2019-02-15
Allow running/compiling/testing unison from anywhere


francisdb:
Creating this ticket as a discussion starting point to get a view on what the vision is regarding this:

- What binaries will be available?
- Do we expect to need some kind of build tool, do we plan to plug in to an existing one?
- How would you develop/test/distribute a library?

---
_Current state of things below (this is not what this ticket is about)_

I create a git repo to store some unison files I am working on and want to compile/test it.

Symlinking the folder into unison-src does not work
https://github.com/haskell-fswatch/hfsnotify/issues/87

Running the `unison` binary from that folder does not work as the jvm runtime is not relative to the current path:
```
../unison.francisdb/parser-typechecker/.stack-work/dist/x86_64-osx/Cabal-2.2.0.1/build/Unison/unison
unison: sbt: createProcess: runInteractiveProcess: chdir: does not exist (No such file or directory)
```

The new haskell runtime can start from a unison file but is not yet usable

Will make the `scratchFilePath` customizable with an argument which should help me forward

aryairani:
There will be a single `unison` executable that can be run from any directory. This is the compiler and build tool and app launcher. There can also be a flag to specify the watch directory, or else watch the current directory.

The `.unison` folder can be versioned on Github along with any scratch .u files, although any definitions that typecheck can just be added to the `.unison` codebase.  We plan to have a way to browse the codebase in text form at github.com.

francisdb:
How would one then create / share a library for user by other projects?

aryairani:
@francisdb Initially, you could share a library with others just by publishing libraries as branches in your .unison directory in your github repo, someone could download from github and merge into their local branch.  This will naturally be easier once you're not limited to running `unison` within the Unison tool's build directory.  I set up an example repo [here](https://github.com/aryairani/unison-libs), in case you're interested in playing around with it.  In not too long, I think we'll have a syntax in the CLI (or even in a .u file) for referencing a library's branch on github.  This would automatically add the branch to your codebase, and you could merge that branch into yours to gain the library functions it published.  Eventually this process will be more polished, including probably some more fine-grained options.

Uploading to a central unison repo might become a thing.

The story for tests is a work-in-progress, but you would write them like any other unison function.

francisdb:
Somehow storing .unison in a git repo is like storing one version control system into an other one

How would you work with that example repo in case you want to add it to your own project?

aryairani:
[...]

PR #340, opened 2019-02-15
Allow providing scratch file path

francisdb:
Allows setting the scratch file path to something else than `.`
Instead of `-haskell` you will have to use `--haskell`

Question: Is that `_scratchFilePath` / `_initialFilePath` something that will be added in the future as it is currently not used? Otherwise I can remove that part of the code

see also #339


pchiusano:
@francisdb can you explain this one a bit more? Maybe post an example session of how the `unison` executable can be launched after this change? (Also, what do you see if you do something incorrectly when launching `unison` exe)

francisdb:
See the related ticket, wanted to have external sources and symlinks did not work so needed option to change the scratch path

pchiusano:
@francisdb cool, can you just show the usage of it, like what do I type at the command line to make use of this new functionality, and what is the output I see? :)

francisdb:
Check out https://github.com/francisdb/unison-json and then run this, it only changes the scratch path, nothing more

```
stack exec unison -- -s ../unison-json/src
```

Now what remains is running tests, something like this?

```
stack exec unison -- test -s ../unison-json/src

[test output here]
```

Full output below

```
stack exec unison -- -s ../unison-json/src
...

  ┌
  │  ✅
  │
  │  I found and typechecked these definitions in /Users/francisdb/workspace/unison-json/src/json.u:
  │
  │    type Either a b
  │    type Json.Value
  │    type Token
  │    Json.format : Json.Value -> Text
  │    Json.format : Json.Value -> Text
  │    Json.lex : Text -> [Token]
  │    Json.parse : Text -> Optional Json.Value
  │    Sequence.flatMap : (a -> [a]) -> [a] -> [a]
  │    Sequence.fold : a -> (a -> a -> a) -> [a] -> a
  │    Sequence.intersperse : a -> [a] -> [a]
  │    Sequence.map : (a -> b) -> [a] -> [b]
  │    Sequence.prependToAll : a -> [a] -> [a]
  │    Text.concat : [Text] -> Text
  │    Text.concatSeq : Text -> Text -> Text -> [Text] -> Text
  │    formatProp : (Text, Json.Value) -> Text
  │    formatProp : (Text, Json.Value) -> Text
  │    lexString : Text -> Either Text (Text, Text)
  │
  │  Now evaluating any watch expressions (lines starting with `>`) ...
  └

  ┌
  │  131 | > Json.format (Json.Value.Object [("foo", Json.Value.Number "123"), ("bar", Json.Value.Array [Json.Value.Null, Json.Value.String "test"])])
  │          ⧩
  │          "{\"foo\":123,\"bar\":[null,\"test\"]}"
  └
  ┌
  │  133 | > Json.lex "\"test\""
  │          ⧩
  │          Con #5cbWz9X1yckJnAbgh5HMxCAWwB9otVmzy6km1BE5gQ1KscRk9ZZZrD55tSXZzm2qWreQV6FsX5DkpmyCiAFei3La 1
  │            ("test", "")
  └
  ┌
  │  135 | > Json.lex "\"test"
  │          ⧩
  │          Con #5cbWz9X1yckJnAbgh5HMxCAWwB9otVmzy6km1BE5gQ1KscRk9ZZZrD55tSXZzm2qWreQV6FsX5DkpmyCiAFei3La 0
  │            "Missing closing\""
  └
  ┌
  │  137 | > Json.lex "[]"
  │          ⧩
  │          [ Con #2pPZi4FuaByP6hms9vHr2DYG37ZUgoDFixMDXryCC5Y3ZFKBjRi5x5PUiqkmJqau2ZxnpHNBAaHTpHWLZC1FDb1k 0
  │            "[",
  │           Con #2pPZi4FuaByP6hms9vHr2DYG37ZUgoDFixMDXryCC5Y3ZFKBjRi5x5PUiqkmJqau2ZxnpHNBAaHTpHWLZC1FDb1k 0
  │            "]"]
  └
  ┌
  │  139 | > Json.lex "{\"foo\":123,\"bar\":[null,\"test\"]}"
  │          ⧩
  │          Con #5cbWz9X1yckJnAbgh5HMxCAWwB9otVmzy6km1BE5gQ1KscRk9ZZZrD55tSXZzm2qWreQV6FsX5DkpmyCiAFei3La 1
  │            ("foo", ":123,\"bar\":[null,\"test\"]}")
  └
  ┌
  │  141 | > case (Json.parse "{'foo':123,'bar':[null,'test']}") of Some a -> format a
  │          ⧩
  │          "\"TODO\""
  └

```

francisdb:
unison can be run in any path with the new hasell runtime so no need for this any more

PR #382, opened 2019-03-13
Parse underscores correctly

### LINKED ISSUE #381: Small parser bugs with underscores
None of these parse:

```
_foo_ = x
```

Doesn't like underscores in the beginning of identifiers.

```
foo_ _ = x
```

Doesn't like arguments named `_`.

mrdziuban:
This fixes #381 by updating the lexer and parser to support underscores in both prefix and infix function names. The changes necessary to get this working were:

- Support `Blank`s as prefix vars, as `_foo` is parsed as a `Blank`
- Prepend `_` to the string value of a `Blank` when using it as a prefix var
- Recognize `_` as a valid starting character of a wordy id to allow the parsing of backticked terms with leading underscores


pchiusano:
@mrdziuban looks good, thanks, could you add yourself to CONTRIBUTORS file? then will merge

mrdziuban:
👍 Thanks, all set

pchiusano:
Thanks! Merging..

PR #385, opened 2019-03-14
Fix printing braces in effects

### LINKED ISSUE #375: Missing brackets around multiple- or higher-order effects in Effect value printer
e.g.
```
multiHandler : s -> [w] -> Nat -> Effect State s, Writer w, Random a -> (s, [w], Nat, a)
```
vs
```
multiHandler : s -> [w] -> Nat -> Effect {State s, Writer w, Random} a -> (s, [w], Nat, a)
```

mrdziuban:
@aryairani this should fix #375. Given the code in `multiple-effects.u`, the output is:

```
-- Before
ability Console
ability State s
Console.state : s -> Effect (State s) a -> a
multiHandler : s -> [w] -> Nat -> Effect State s, Console a -> ()

-- After
ability Console
ability State s
Console.state : s -> Effect (State s) a -> a
multiHandler : s -> [w] -> Nat -> Effect {State s, Console} a -> ()
```

I just had one question &mdash; I haven't fully grokked when it's necessary to pass 0 or 3 to `go`, but I tried changing it to 3 in this case and got this for `multiHandler`:

```
multiHandler : s -> [w] -> Nat -> Effect {(State s), Console} a -> ()
```

Is having the parentheses around `State s` desirable since it's printed with parentheses in the `Console.state` type?

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -727,8 +727,10 @@ renderType
   -> AnnotatedText a
 renderType env f t = renderType0 env f (0 :: Int) (Type.ungeneralizeEffects t)
  where
-  paren :: (IsString a, Semigroup a) => Bool -> a -> a
-  paren test s = if test then "(" <> s <> ")" else s
+  wrap :: (IsString a, Semigroup a) => a -> a -> Bool -> a -> a
+  wrap start end test s = if test then start <> s <> end else s
+  paren = wrap "(" ")"
+  curly = wrap "{" "}"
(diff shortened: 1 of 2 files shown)


pchiusano:
That number is the “ambient precedence” - generally you only want to put parens if the immediately surrounding expression has higher precedence. The general idea is this ambient precedence follows the precedence of the parser, so the prettyprinter prints things with a minimum of parens needed to produce the same AST.

In this case, the parens inside the ability list are redundant so I think it’s right to pass 0 as the ambient precedence.

Hope that is helpful!

mrdziuban:
Thanks @pchiusano, that does help!

PR #392, opened 2019-03-15
fix #378 close empty = blocks

### LINKED ISSUE #378: Incomplete definition block not closed by 0 indentation
I would expect the incomplete `test =` block to end at `x = ...` but it seems not to be the case. In a large file this is annoying because the error is reported far away from the source and sometimes you get totally unrelated errors.

```haskell
type Either a b = Left a | Right b

test : Text -> Boolean
test =

x = Right 1
y = Right 2

test2 : Text -> Boolean
test2 t = t == "ok"
```

currently the error is reported at the end of the file somewhere

```
master>
I found a block that ends with a binding instead of an expression at line 9, columns 1-6:

    9 | test2 : Text -> Boolean

Based on the context, I'm expecting an expression of type Text -> Boolean after this binding.
```

haskell will report on the `x = ...` line with `parse error (possibly incorrect indentation or mismatched brackets)`

francisdb:
The result of this fix is:

```haskell
test =
  test3 =
    test4 =

> test
```

image

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -236,7 +236,15 @@ lexer0 scope rem =
       (spaces, '-':'-':rem) -> spanThru' (/= '\n') rem $ \(ignored, rem) ->
         pushLayout b l (incBy ('-':'-':ignored) . incBy spaces $ pos) rem
       (spaces, rem) ->
-        let pos' = incBy spaces pos in go ((b, column pos') : l) pos' rem
+        let topcol = top l
+            pos'   = incBy spaces pos
+            col'   = column pos'
+        in
+          if b == "=" && col' <= topcol then
+            -- force closing by introducing a fake col +1 layout
+            popLayout0 ((b, col' + 1) : l) pos' rem
+          else
+            go ((b, col') : l) pos' rem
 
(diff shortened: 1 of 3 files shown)


parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -236,7 +236,14 @@ lexer0 scope rem =
       (spaces, '-':'-':rem) -> spanThru' (/= '\n') rem $ \(ignored, rem) ->
         pushLayout b l (incBy ('-':'-':ignored) . incBy spaces $ pos) rem
       (spaces, rem) ->
-        let pos' = incBy spaces pos in go ((b, column pos') : l) pos' rem
+        let topcol = top l
+            pos' = incBy spaces pos
+            col' = column pos'
+        in
+          if b == "=" && col' <= topcol then
+            popLayout0 ((b, col' + 1) : l) pos' rem
```
francisdb: the +1 here is a bit dirty but it forces a close on popLayout0

pchiusano: Ah interesting, thanks for pointing this out, but could you include this as a comment in the actual source?

francisdb: @pchiusano ok like this?

pchiusano:
Looks good, thanks!

PR #394, opened 2019-03-18
Universal.{<,>,<=,>=,compare} functions

pchiusano:
Same idea as #393. Ordering is lexicographical for products like data constructors. Functions are ordered just based on their hash, we don't compare their actual source lexicographically. The ordering of constructors is just based on their constructor number, which is arbitrary for structural types, but when we add other kinds of types that number will be less arbitrary and the ordering thus more meaningful.

Screen Shot 2019-03-18 at 1 45 56 PM


aryairani:
Could you document `Unison.Util.Hashtable` to explain why it exists (vs just using `Data.HashTable.IO` directly); and maybe even explain cyclicEq and cyclicOrd in general?

pchiusano:
@aryairani sure, can add some docs before merging. Short answers: HashTable.IO doesn't store its size, which we need fast access to, and cyclicEq/Ord are needed to be able to compare potentially cyclic structures like Unison values.

PR #395, opened 2019-03-18
add ANSI-patched haskeline as a submodule

aryairani:
This PR adds haskeline with [an alignment fix](https://github.com/judah/haskeline/pull/115) and [some flags tweaking](https://github.com/unisonweb/unison/commit/64e26f386facd6574a9d26d0082e393060b50aa2) as a git submodule and stack subproject.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "yaks/haskeline"]
+	path = yaks/haskeline
+	url = https://github.com/aryairani/haskeline.git
+	branch = topic/align-ansi-completions
--- a/stack.yaml
+++ b/stack.yaml
@@ -6,6 +6,7 @@ allow-different-user: true
 
 packages:
 - yaks/easytest
+- yaks/haskeline
 - parser-typechecker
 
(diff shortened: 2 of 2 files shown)


aryairani:
https://github.com/BlockScope/unicode-homoglyph
https://gitlab.haskell.org/ghc/ghc/issues/15775#ticket
https://github.com/agda/agda/commit/5ff3fc5155af087c3881915ed82e45c9d06523df
bizzare

aryairani:
`ghcid` was propagating haskeline's .cabal's global use of `CPP` extension into our project too, which blew up a couple of our files (it doesn't like certain unicode characters we used for formula in comments, and doesn't like multiline strings we used in some tests).  

The most recent commit disables the haskeline fork submodule's global use of `CPP`.

We can get rid of this fork dependency if the fix (https://github.com/judah/haskeline/pull/115 or equivalent) goes into haskeline stackage or if we stop trying to use ANSI in auto-completion display (i.e. highlighting fuzzy search match regions).

PR #402, opened 2019-03-20
Switched implementation of `[a]` type to use Data.Sequence and cleanup naming

pchiusano:
Not a very exciting change, but the `[a]` type is now backed by `Data.Sequence`. I've been meaning to clean this up for a while and was reminded of it because this [pure Unison `Map` type](https://github.com/unisonweb/unison/blob/b37b1231a6de9f78e9248154baf4498b933b3a42/unison-src/Map.u) relies on having a sequence type that supports efficient, slicing, random access, and concatenation.

I also cleaned up the naming - like 50% of places were still using "Vector", so now everything uses "Sequence" consistently.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Serialization/V0.hs
+++ b/parser-typechecker/src/Unison/Codebase/Serialization/V0.hs
@@ -3,7 +3,6 @@
 module Unison.Codebase.Serialization.V0 where
 
 -- import qualified Data.Text as Text
-import qualified Data.Vector                   as Vector
 import qualified Unison.PatternP               as Pattern
 import           Unison.PatternP                ( Pattern )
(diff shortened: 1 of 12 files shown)


aryairani:
:+1:

francisdb:
There is no slice / split exposed?

pchiusano:
@francisdb Any non higher-order sequence methods would be a welcome PR for new builtins. 😀 To add a builtin, you have to tell Unison what it’s type is in Builtin.hs (can follow other examples there, like search for Sequence.drop). And then you have to tell Unison how to evaluate it, check Rt1.hs, can also follow examples there.

There is take/drop so you can use that to implement slice and split in pure Unison.

Higher order sequence functions should probably be pure Unison because of request handling stuff.

Issue #407, opened 2019-03-21
type error showing a bare builtin reference for a function name defined in .u


aryairani:
```
The 1st argument to the function ##makeDyn is Nat ->{𝛆, inferred-effect} Nat, but I was expecting (a ->{State (Map a b)} b) -> a ->{State (Map a b)} b:

   15 | fib n = handle State.run Map.empty in fibDyn n
    .
   44 | fib' = makeDyn fib
   45 | 
   46 | makeDyn : ((a ->{State (Map a b)} b) -> a ->{State (Map a b)} b) -> a -> b

More specifically, I found Nat where I was expecting a ->{State (Map a b)} b:

    7 | fibDyn : Nat ->{State (Map Nat Nat)} Nat
    .
   46 | makeDyn : ((a ->{State (Map a b)} b) -> a ->{State (Map a b)} b) -> a -> b
```
full source: https://gist.github.com/aryairani/91adb7ccd4aa42578bc5e3dc54de86eb

Probably need to use a different prettyprint environment or something.

aryairani:
"Or change how the typechecker does that."

PR #411, opened 2019-03-22
Fix for #407

pchiusano:
I'd call this a not ideal fix, but still an improvement over what is happening now, which is #407.

The issue, #407 is caused by the fact that the typechecker treats "previously typechecked definitions" exactly the same as new builtins (as in, after obtaining a type for some binding, `foo`, it replaces all references to the var `foo` with a builtin, also called `foo`, which has the given type). The PrettyPrintEnv has no idea the typechecker does this so it just tries to print that builtin like any other.

All I did for the fix was give these "virtual builtins" a unique prefix, which the PrettyPrintEnv can detect and print differently. I'd consider this kind of ugly.

I think a more comprehensive fix would be to make the `Ref` constructor inside a Term be a more structured type than just `Reference`, or make it a type parameter of the `Term` rather than hardcoding `Reference` there. Then the typechecker could use a richer type there. All these options seemed like huge refactorings though.

### DIFF
--- a/parser-typechecker/src/Unison/PrettyPrintEnv.hs
+++ b/parser-typechecker/src/Unison/PrettyPrintEnv.hs
@@ -4,14 +4,16 @@ module Unison.PrettyPrintEnv where
 
 import Control.Applicative ((<|>))
 import Data.Maybe (fromMaybe)
-import Unison.Reference (Reference)
+import Unison.Reference (Reference(Builtin))
 import qualified Data.Map as Map
 import Unison.HashQualified (HashQualified)
 import qualified Unison.HashQualified as HQ
+import qualified Unison.Name as Name
 import Unison.Names (Names)
 import qualified Unison.Names as Names
 import Unison.Referent (Referent)
 import qualified Unison.Referent as Referent
+import qualified Data.Text as Text
 
 data PrettyPrintEnv = PrettyPrintEnv {
   -- names for terms, constructors, and requests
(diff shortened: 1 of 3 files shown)


aryairani:
Is it not possible to give it a `DerivedId` ref, and add the appropriate name to the `PrettyPrintEnvironment`?

pchiusano:
Maybe, though that seems weird also. 😶 What hash do you pick for that derived id?

The type checker doesn’t currently have access to a prettyprinting environment so you’d need to come up with some way of it outputting information to add to it. Maybe tweak the typechecking monad to output this and then propagate this. Doesn’t seem too promising.

aryairani:
Maybe I misunderstood, but I was thinking you’d pick the hash the same way as all the other hashes — based on their contents?

pchiusano:
Ah yes that makes more sense, not sure why I wasn’t thinking that! There’s still the issue of getting the names for those hashes, will think on that.

aryairani:
I haven't traced it all the way through, but it looks to me like typechecker currently receives a PPE that is populated by all the references known in the `UnisonFile`; however, only the decls have references known in the `UnisonFile`.  

Terms aren't currently put into the PPE because we thought we don't know their hashes.

We do know the referents for constructor terms, and adding those to the PPE will fix the first render error in #419.

For non-constructor terms, there isn't an obvious hash.  We could compute one from their contents, but must keep in mind that their contents can change over the course of typechecking, as Var references are resolved one way or another.

Could you back up and explain why the typechecker makes virtual builtins?  How does it distinguish  virtual builtins from real builtins in the end? (prior to this fix, which I assume doesn't change anything about that.)  Would it break currently up if someone named a local definition after a builtin?

aryairani:
It looks like the typechecker can only currently look up term types by `Reference`, and I'm guessing that's the reason these defs that aren't yet in the codebase are being coerced into References anyway (as builtins, eep! This seems super weird and hacky to me.)

If we don't want a separate mechanism for the typechecker to know about these auxiliary defs that aren't in the codebase, we could have a 3rd type of `Reference` to represent "previously typechecked definitions".

I think in general we should lean more towards putting domain knowledge into the Haskell types.  e.g. If there's a function that can handle 3 different cases, we pass a type with 3 cases; if another function can only handle 2 of those 3 cases, we pass a different type that encompasses only 2 of those cases.  If we have a type that represents 3 cases, but we need to distinguish 4, we make a new type with 4 cases, rather than overloading a subset of one of the existing 3 cases.

pchiusano:
Close in favor of #474

Issue #413, opened 2019-03-23
parse error in file with no definitions above `---`


aryairani:
A .u file with no definitions produces nice message, but .u file with `---` and no definitions produces a parser error:
```
/Users/arya/unison/drew.u:1:1:
unexpected end of input
expecting ability, effect, or use
```
😭

aryairani:
🎉

PR #416, opened 2019-03-24
keyword based expressions like `if` can be parenthesized

### LINKED ISSUE #373: trouble parsing parenthesized if-expression
Each of these
```
foo = (if true then 0 else 0)
> (if true then 0 else 0)
```
yields an error
image

Adding spaces inside the parens doesn't help.

pchiusano:
Fixes #373, so things like `(if true then 1 else 2) + 1` now parse fine. 😎

The fix treats `(` as an open token in the lexer, and `)` emits `Close` tokens and pops the lexer stack back to the nearest `Open "("`. This lets the `)` close the `else` block, even though the `else` would otherwise be expecting an outdent to finish the block.


pchiusano:
Added a fix for #413 to this PR also.

PR #422, opened 2019-03-28
Update development documentation

BenFradet:
There doesn't seem to be any scala left in the repo so I though I'd cut some things down.


aryairani:
Oof, thanks for bringing this up.  Even the Haskell parts were out of date.

PR #426, opened 2019-03-29
Fix demo examples

BenFradet:


unison-src/tests/stream-shouldnt-damage-stack.ur
```diff
@@ -1 +0,0 @@
-(4950, 9999)
```
BenFradet: This seems like an artifact of https://github.com/unisonweb/unison/commit/3af0cb54755f8ef5e7cf5109b7b4f121547f5472

aryairani: 👍

Issue #425, opened 2019-03-29
an ugly parse error on empty watch expression


aryairani:
```
>
```
yields
```
Token {payload = Close, start = Pos 51 1, end = Pos 51 1} :| []
/Users/arya/unison/scratch.u:51:1:
unexpected <outdent>
expecting and, binding, infixApp, lambda, or or
   51 | 
```
Probably drop the Token stuff, at least?  Can we do even better?

pchiusano:
Good catch. This is should be an easy fix and you could knock it out if you want. :) This would also be doable by another contributor too if anyone would like to pipe in and volunteer. Here's how:

I'd add a custom error constructor for this in `Unison.Parser (Error)`, say it's called `EmptyWatch` and then modify the watch expression parser: a `>` followed immediately by a `Close` token gets a `Text.Megaparsec.customFailure (EmptyWatch ..)`.

Compiler warnings should then guide you to add some custom logic for printing out this parse error, which you can make nice.

This is also generally the approach to adding better parse error messages - at the point in the parser where you are expecting a certain parser `p` to succeed, you add a `p <|> customFailure blah` and if it fails in an uncommitted state, you get `blah` as your error instead (if `p` fails in a committed state, then it's consumed some characters and `blah` isn't the right description of the error anyway as it's "too far back").

BenFradet:
I can have a look :+1:

PR #430, opened 2019-04-01
Print custom error message in case of an empty watch expression

BenFradet:
c.f. #425

What's the best way to go about unit testing? Should I add to https://github.com/unisonweb/unison/blob/master/parser-typechecker/tests/Unison/Test/FileParser.hs ?

### DIFF
--- a/parser-typechecker/src/Unison/FileParser.hs
+++ b/parser-typechecker/src/Unison/FileParser.hs
@@ -29,7 +29,6 @@ import qualified Unison.UnisonFile as UF
 import           Unison.Var (Var)
 import qualified Unison.Var as Var
 import qualified Unison.PrettyPrintEnv as PPE
--- import Debug.Trace
 
 file :: forall v . Var v => P v (PPE.PrettyPrintEnv, UnisonFile v Ann)
 file = do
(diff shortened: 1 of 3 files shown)


pchiusano:
@BenFradet looks good am going to merge!

At the moment, we don't have a good way of testing error messages. I'd like to avoid having tests that depend on the exact formatting or wording of messages since I think that'll be too fragile.

I think it would be doable to have a little test harness that lets you compare the failing parse value, so something like:

``` Haskell
import EasyTest
expectFileParseFailure :: String -> (Parser.Error -> Test ()) -> Test ()
```

Which you then use to construct a bunch of tests, just making sure that the right kind of error makes its way into the parse error.

BenFradet:
Thanks, I'll look into it.

aryairani:
Hey @BenFradet — we need, but don't yet have, a test harness for CLI workflow.

It is meant to be something like:  send some `Input` to the Actions.hs loop, and then interpret the resulting `Free (Command i v) a` in the test harness to make sure it produces your anticipated `Notify` command.

aryairani:
Oh, I'm late to the party.  That's another way.

pchiusano:
@aryairani 😀yeah I think what I suggested is probably better if the goal is just to unit test the errors that the parser generates. If it's more of an integration test of how parser interacts with CLI and codebase state, then what you said. 😀

PR #432, opened 2019-04-02
Fix #429 - Remove duplicate elements from branch search

### LINKED ISSUE #429: duplicate search results from ls
image

Probably needs a `nubOrd` somewhere.

noahhaasis:
By creating a set of all names first and then converting it to a list there are no duplicate names.


aryairani:
Hey @noahhaasis, thanks for tackling this, and digging up a 

I think that this proposed fix doesn't quite do the trick though, and I should have documented the following better: `searchTermNamespace` returns a set that is ordered by score, so it's important to retain that score-order by going directly `toList` (i.e. `fmap snd . toList`.  If we do `Set.map snd` first (i.e. `toList . Set.map snd`), then we are no longer using the score to determine the output order.

I think the right fix in this case would be `nubOrd . fmap snd . toList`.  What do you think? :)  `nubOrd` is in `Data.List.Extra`

noahhaasis:
Sounds good @aryairani . Totally forgot about the order

pchiusano:
@noahhaasis 👍🕺

PR #433, opened 2019-04-02
Reorder functions to be top down

noahhaasis:


aryairani:
Hey @noahhaasis, thanks for this PR; can you describe it a little bit?  Does it just rearrange the functions in the file, or are there any logic changes?

noahhaasis:
@aryairani It just rearranges the function. It doesn't change the behaviour but I recently noticed that I tend to order my programs button up and wanted to change that.

pchiusano:
@noahhaasis can you show screenshot before and after? I'm not understanding based on the description... 😶

aryairani:
@pchiusano I'm not positive, but I think it literally just rearranges the order of the Haskell definitions in `parser-tests/GenerateErrors.hs`

noahhaasis:
Bildschirmfoto 2019-04-04 um 06 24 08
The new version is on the left and the old on the right. @pchiusano is right I just rearranged the functions. I don't know if you prefer the program to be top down or bottom up

pchiusano:
@noahhaasis Ahh... okay. Gotcha. :)

I'd say let's leave it as is. I don't really have much preference one way or another but you can make a case for either (like do you use `let` or `where`?). If you're writing new code feel free to pick whatever order you prefer. Also, if you're doing major surgery on a module and aren't expecting the text diff to be very meaningful then any formatting changes or reordering is also fine. (If you're doing a more targeted change and want the text diff to be meaningful, I'd leave existing formatting in place where possible.) 

I was just thinking how "order of functions" won't be a thing in Unison... you could view the same set of of definitions in whatever order: alphabetically, in dependency order, reverse dependency order, etc.

Issue #427, opened 2019-03-30
add more Float functions


aryairani:
e.g. these: http://www.cplusplus.com/reference/cmath/

pchiusano:
Here’s how: in Builtin.hs, add the name and type, you can follow the other examples.

Then in Rt1.hs, add the implementation here: https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Runtime/Rt1.hs#L291

Then add some tests.

PR #434, opened 2019-04-03
Float Library Functions

billy1kaplan:
Adds float functions as described in #427


francisdb:
Any chance you could add a fromText: Text->Either[String, Float]?

billy1kaplan:
As far as I can tell, using the Unison either type in these functions might be a bit more challenging.

Either is defined here:
https://github.com/unisonweb/unison/blob/ed69a95128440f7976014d2826a0e0872662ba43/parser-typechecker/src/Unison/Runtime/IOSource.hs#L119


We could pretty easily add `fromText: Text -> Optional Float` if that's an acceptable alternative.

pchiusano:
@billy1kaplan this is great, thanks, can you add yourself to CONTRIBUTORS? :)

Also, `Float.fromText : Text -> Optional Float` sounds good to me.

francisdb:
Optional should indeed be enough. The user can easily come up with a proper error message for this simple case.

pchiusano:
Thanks @billy1kaplan 🎸😀 merging...

PR #436, opened 2019-04-05
Fix cloning instruction in the README

jhrcek:
```
git checkout --recursive ...
error: unknown option `recursive'
```


pchiusano:
Thanks!

PR #438, opened 2019-04-07
Escape annotation highlighting when enclosed with quotation marks

BenFradet:
Otherwise this leads to:

2019-04-06-18:50:25_1920x1080_scrot


pchiusano:
Thanks!

Issue #455, opened 2019-04-09
Typechecker should disallow unguarded mutually recursive definitions


pchiusano:
See #454 and #453 

Idea for implementation - after checking a cycle minimized let rec and all the binding types are known, make sure at least one of those types is an arrow type. If not, emit an error.

pchiusano:
Proposal after discussion with @aryairani: given the graph of the definitions in the cycle, it's an error if any two non-lambdas are directly connected. If all the non-lambdas only depend on lambdas, that could be allowed, since we can move all the lambdas to the top of the cycle.

Issue #410, opened 2019-03-22
Spurious ability check failures when calling an underapplied function


pchiusano:
See https://github.com/unisonweb/unison/blob/master/unison-src/tests/spurious-ability-fail-underapply.uu minimal pending test illustrating the issue.

It looks like the ability check that happens when typechecking function application isn't taking into account that the function may be getting underapplied and hence not all its abilities may be required for the underapply. Might be an easy fix, but I didn't see anything obviously busted in the typechecker that would cause this..

pchiusano:
A bit more detail on this, it only occurs when the underapplied function has no type annotation. So it seems to be some weird interaction between ability inference and the ability checking that happens as part of function application.

Issue #374, opened 2019-03-09
Weirdness when stacking handlers, sometimes?


aryairani:
(Full program is below, excerpts are shown here.)

```haskell
> handle stateHandler "hello" in
  handle writerHandler [] in
  replicate 5 main

main = '(tell get)
```
doesn't pick up the ambient ability provided by `stateHandler`:

![image](https://user-images.githubusercontent.com/538571/54072842-44c2a500-424e-11e9-952b-df352de07f3b.png)

But if I inline `main`,
```haskell
> handle stateHandler "hello" in
  handle writerHandler [] in
  replicate 5 '(tell get)
```
it works fine:

![image](https://user-images.githubusercontent.com/538571/54072853-615edd00-424e-11e9-9714-b90044dece9e.png)

It also works fine if I eliminate the call to `replicate`:
```haskell
> handle stateHandler "hello" in
  handle writerHandler [] in
  main

main = '(tell get)
```

Using the signature `replicate : Nat -> 'a -> ()`, or letting the type of replicate be inferred, doesn't change anything.

Adding a type annotation in the handle block produces the same error as calling passing main to replicate:
```haskell
> handle stateHandler "hello" in
  handle writerHandler [] in
  replicate 5 ('(tell get) : '{State Text, Writer Text} ())
```
![image](https://user-images.githubusercontent.com/538571/54072950-5d7f8a80-424f-11e9-98b6-9e82aa70812c.png)

---
Full program:
```haskell
use State get put
use Writer tell

> handle stateHandler "hello" in
  handle writerHandler [] in
  replicate 5 main

main = '(tell get)

replicate : Nat -> '{e} () -> {e} ()
replicate n x =
  if n == 0 then () else
    !x
    replicate (n `drop` 1) x

effect State a where
  get : {State a} a
  put : a -> {State a} ()

effect Writer w where
  tell : w -> {Writer w} ()

stateHandler : s -> Effect {State s} a -> (s, a)
stateHandler s x = case x of
  { State.get -> k } -> handle stateHandler s in k s
  { State.put s -> k } -> handle stateHandler s in k ()
  { a } -> (s, a)

writerHandler : [w] -> Effect {Writer w} a -> ([w], a)
writerHandler ww x = case x of
  { Writer.tell w -> k } -> handle writerHandler (ww `snoc` w) in k ()
  { a } -> (ww, a)
```

PR #460, opened 2019-04-15
Improved ability inference

### LINKED ISSUE #390: Revisit effect generalization and inference
Type _checking_ of effects is straightforward. We maintain an "ambient ability", which is pushed down into subexpressions following user-provided type annotations, and when typechecking a function call for a function `a ->{e} b`, we verify that the function's required abilities, `{e}` are a subset of the ambient abilities. This is the basic gist behind the Frank paper and I feel pretty good about it.

Effect inference (not in the Frank paper) is used in a few places as well - a lambda without an annotation will try to infer the effects used, and in `handle` blocks, the body of the block gets access to whatever effects the handler eliminates. I'd characterize our implementation of inference as pretty ad hoc and I feel less good about it.

Also on somewhat shaky ground: we do effect generalization. The idea behind effect generalization is that type signatures like `(a -> b) -> [a] -> [b]` can be generalized to `(a ->{e} b) -> [a] ->{e} b`, which makes them "effect polymorphic" (so `map` can be used as `traverse` without a separate definition). Effect generalization is fine by itself, except it interacts weirdly with effect inference, which has to introduce existential type variables for the ambient ability. I suspect #332 is an instance of this.

A few things I'd like to do:

* Start by documenting how we'd like this to work.
* Likely thing: don't do any generalization for non HOFs, as they don't benefit from it.
* Do effect inference at the level of a whole n-ary lambda. So the whole lambda gets one existential ambient ability, which is refined during typechecking of the body. This should generate predictable results for inferred types, which could be spec'd out in the above document.
* Think carefully about the inference provided by `handle` blocks. It may not be terrible to just eliminate this inference if it can't really be spec'd out and explained easily.

pchiusano:
This PR implements the [scheme discussed here](https://github.com/unisonweb/unison/blob/master/docs/ability-typechecking.markdown), making ability typechecking and inference much more predictable.

Closes #390, #374, #410, #455 

Am feeling pretty good about it, but hold off on merging, have a few things I'd like to clean up:

- [x] Refactor `Var` to make variable type ("user provided", "inference variable") explicit. This is needed to properly clean up the rendering of type signatures.
- [x] Clean up rendering of type signatures. No more eliding ids unless safe to do so, and can dynamically choose names for "synthetic" variables.

Motivation for the above is that ability variables show up much more regularly in types presented to the user, so want the types to not look super ugly. 😀

Maybe later:

- [x] Remove AskInfo hack, or at least convert it to use new var type
- [x] `Ord` instance for ABTs, uses alpha equivalence just like `Eq`
- [x] Add a `nubOrd` when finalizing a type ability list, avoid duplicates which I think can show up if two existentials get refined independently.


pchiusano:
Okie doke, this is ready to merge. 

@aryairani some "good" news, the crazy typechecking bugs we discovered in `stream.uu` are also present in master. I haven't tried to bisect, but I'm guessing they've been there for a while and aren't related to this refactoring.

I'll create separate ticket to track those bugs.

PR #461, opened 2019-04-15
Parser error tests

BenFradet:
This is what was discussed in https://github.com/unisonweb/unison/pull/430#issuecomment-479599804.

If this first test looks ok, I can extend it to other parser errors :+1:


pchiusano:
LGTM

pchiusano:
This seems like a good framework that we can extend as we add other kinds of parse errors, feel free to fill in more!

BenFradet:
that's the plan :+1:

Issue #439, opened 2019-04-07
Initial branch load doesn't verify builtins


aryairani:
It should do the same checks (specifically, for missing builtins) as the switch branch command.

This is likely contributing to [this](https://gitter.im/unisonweb/unison?at=5ca219d9759abc043c4aa950)
> There are implementations like Float.ceiling in Rt1.hs but no types in Builtins.hs. And adding Float.ceiling to Builtin.hs doesn't work for me

and [this](https://gitter.im/unisonweb/unison?at=5ca23cd3a84e0c501abbe714)
>@noahhaasis make sure you do `builtins.update` or create a fresh branch otherwise the branch namespace won’t have a mapping for those new functions

Currently, the CLI will warn you of this condition and tell you which command to run, but only in response to a `branch` command, not the initial branch load.

I'm thinking we should just inject [the `SwitchBranchI` action](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Codebase/Editor/Actions.hs#L263
) when initializing [the input event queue](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/CommandLine/Main.hs#L97).

aryairani:
Closing this until we decide it's a problem again.

PR #467, opened 2019-04-21
Add pattern matching on sequences

### LINKED ISSUE #266: Add pattern matching on vectors / sequence
Looks like we might be missing a way to pattern match on vectors?

The pattern parser is missing handling for them.  When given 
```haskell
case x of [a, b] -> a
```
it throws 
```
unexpected [
expecting (, blank, false, symbol, or true
```
... and there's nothing in the pattern AST for it to parse to anyway (unless maybe a special constructor, like with Pair and ().)

(Related: in the term language, does Vector really pay its way as a member of the AST definition?  How about if [a,b,c] was sugar that mapped to `App (Constructor Vector 0) (the tuple (a, b, c))`?  Would that prevent Vector being its own thing at runtime?)

mrdziuban:
Resolves #266. Adds pattern matching on sequence literals and cons, snoc, and concat operations. Note that at least one side of a concat must be of a fixed length, otherwise both typechecking and compilation will fail.

I tried adding `+:` and `:+` functions as aliases of `cons` and `snoc` in https://github.com/mrdziuban/unison/commit/b0f68e1b9cdd7b9347f086b053d000aaf4159bbf, but I'm getting this error when trying to use `+:`:

```
I'm not sure what User "Sequence.+:" means at line 4, columns 16-18

    4 | prepend1 l = 1 +: l

Whatever it is, it has a type that conforms to Nat ->{𝕖} [Nat] ->{𝕖} [Nat].
```

so I excluded them for now, but if you have any tips on how to add them, I'd appreciate it.


aryairani:
Ah yes, sorry.  Now that you mention it, we don't really have an appropriate way to add aliases for builtins until after #468.  

What happened for you is that you added a new builtin with the same implementation, which is not what we want.  And although your build of unison now has that new builtin, your pre-existing branch doesn't, and you didn't get any warning about it because of #439.  (You will get a warning if you change branches and change back. :|  You will also get the new builtin if you create a new branch, as they are populated with all the bulitins that unison knows about.)

So we'll need to hold off on the aliases until after #468.  I'd say: take them back out of this PR, and open a new ticket for adding them.

mrdziuban:
Thanks for the explanation @aryairani, this should be all set and I created #471 to track adding them.

parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -21,7 +21,8 @@ import           Prelude hiding (and, or, seq)
 import qualified Text.Megaparsec as P
 import qualified Unison.ABT as ABT
 import qualified Unison.Lexer as L
-import           Unison.Parser
+import           Unison.Parser hiding (seq)
+import qualified Unison.Parser as Parser (seq)
```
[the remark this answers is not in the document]
mrdziuban: Since there's another function named `seq` in this file, this makes it so this one has to be referenced as `Parser.seq` [here](https://github.com/unisonweb/unison/pull/467/files#diff-211d62c2ee7dccd868dc459a4e36136cR221).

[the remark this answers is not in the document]
aryairani: Ah roger.  I think you can leave the (seq) off the second one and still reference it as `Parser.seq`. nbd though.

pchiusano:
Minor merge conflict due to renaming of `UnknownEffectConstructor`, which I just resolved on web interface... assuming Travis passes will merge!

parser-typechecker/src/Unison/Codebase/Serialization/V0.hs
```diff
@@ -324,6 +338,10 @@ putPattern putA p = case p of
   Pattern.EffectBind a r cid args k
     -> putWord8 9 *> putA a *> putReference r *> putLength cid
                   *> putFoldable args (putPattern putA) *> putPattern putA k
+  Pattern.SequenceLiteral a ps
+    -> putWord8 10 *> putA a *> putFoldable ps (putPattern putA)
+  Pattern.SequenceOp a l op r
```
pchiusano: So, unless there's a super compelling reason (is there, here?), I'd just make the order of the fields match left to right order of the pattern. So `a`, then `l`, then `op`, then `r`.

Minor reason is serialization code should be as dumb as possible, using left-to-right order requires less thinking, less opportunity for errors, and is easier to inspect for correctness. Also, not for nothing, the deserialization case becomes simpler, can use applicative rather monadic syntax.

parser-typechecker/src/Unison/Codebase/Serialization/V0.hs
```diff
@@ -338,6 +356,13 @@ getPattern getA = getWord8 >>= \tag -> case tag of
   7 -> Pattern.As <$> getA <*> getPattern getA
   8 -> Pattern.EffectPure <$> getA <*> getPattern getA
   9 -> Pattern.EffectBind <$> getA <*> getReference <*> getLength <*> getList (getPattern getA) <*> getPattern getA
+  10 -> Pattern.SequenceLiteral <$> getA <*> getList (getPattern getA)
+  11 -> do
+    a <- getA
+    op <- getSeqOp
+    l <- getPattern getA
+    r <- getPattern getA
+    pure $ Pattern.SequenceOp a l op r
```
pchiusano: If you switch the order, this would just become 

```haskell
let pa = getPattern getA
in Pattern.SequenceOp <$> getA <*> pa <*> getSeqOp <*> pa
```

aryairani: @pchiusano Ah, I had recommended writing `op` first; alternatively we could change the order of the constructor?

pchiusano: Sure, that's fine also.

pchiusano:
Actually, I'm just going to merge and then fix up the serialization after... thanks @mrdziuban

PR #477, opened 2019-05-02
Watch expression caching, `test` command, incremental testing, and "branchless" WIP

pchiusano:
There's more to do here, but I think it's in good shape to merge and I think it makes more sense to polish after the branchless refactoring is done, since that will shake up a bunch of stuff in the CLI anyway.

__Quick summary:__ Watch expressions get cached, based on the hash of their source. Cached watches are shown without color, and watches not found in the cache get a purple triangle, shown here (open to other suggestions on this). 

Screen Shot 2019-05-01 at 10 42 28 PM

The `FileCodebase` stores watches in `watches/<hashOfWatchSrc>.ub`, where the file contains the evaluated term.

The syntax `test>` defines a test. A test is just another watch expression, it gets cached as well, it just happens to be of type `Test.Result`, where `type Test.Result = Fail Text | Ok Text`.

Screen Shot 2019-05-01 at 10 55 07 PM

The idea is that `Ok` and `Fail` are more like compilation targets for a higher-level testing library (anyone wanna port EasyTest?? 😀) - you don't actually program with that type directly.

Tests can be added to the codebase (they are given a GUID name if the user doesn't pick one). They get refactored like any other definition. All tests that are added to the codebase The `test` command "runs" (actually just looks up in the cache) all the tests which intersect with the set of named definitions in the branch, and prints out a little summary:

Screen Shot 2019-05-01 at 10 40 21 PM

### Remarks

* There's no `test.propagate` command yet. It should be easy to write though - for every definition, `x` in the branch, check if it is the target of a type preserving edit whose source hash is something in the test cache. If yes, evaluate `x` and put its result in the test cache.
* Alternately, you could do test rerunning as part of propagation, rather than a separate command.
* It would be nice to elide the autogenerated test name in more places, and I have some ideas for that.
* Fixed some stuff in the lexer, like `1+1`, `x-2` now lexes like you'd expect, even without spaces. `-` is not a valid token in a wordy identifier anymore.


pchiusano:
This PR closes #444 once merged (though comments on any of this still welcome anytime).

atacratic:
I'm excited about this, it seems really neat.
- I like the fact that tests are just definitions in the codebase like any other.
- I like the way that propagating updates out to test results is going to feel all like a single step.
- I like the fact that tests are just a specialization of watches.

I think you used the phrase 'perfect incremental testing' somewhere, and I just wanted to get clear on what that means.  When I have
```
test> f(g(x))
```
and I change the definition of `f`, is `g(x)` recomputed as part of recomputing the test result, despite being unchanged?  (I'm thinking yes?)  

When you talk about incremental testing, do you mean, after a code change it only reruns tests that actually hit the changed code?

I didn't quite understand this bit:

> The test command "runs" (actually just looks up in the cache) all the tests which intersect with the set of named definitions in the branch

Does it mean those tests which have a named definition in the branch pointing to them?  Or those tests which transitively invoke a named definition?  And either way I don't quite get it.  

You mentioned eliding the autogenerated test name in more places.  Why can't we just have a user-chosen name for a test (or a watch), in the same way that other terms can be named?  Is it because there isn't a space for it in the `test> f(x)` syntax?  Is it really right that we'll have to execute the test to get its name (from the `Ok Text` result)?  I think from reading "they are given a GUID name if the user doesn't pick one" that I might have misunderstood this.

Will any of this framework carry over to tests in `IO`?  Will the user list and run those tests in a similar way (albeit without caching)?

aryairani:
Hi @atacratic,

> When you talk about incremental testing, do you mean, after a code change it only reruns tests that actually hit the changed code?

Right :)

> Does it mean those tests which have a named definition in the branch pointing to them? Or those tests which transitively invoke a named definition? And either way I don't quite get it.

I'm thinking it's tests which transitively invoke a named definition.  e.g. we won't run try to tests of code that isn't in your namespace.

> Why can't we just have a user-chosen name for a test (or a watch), in the same way that other terms can be named?

We can allow user-chosen names, but need to have a nice experience for even when the user doesn't care to think about test names.

> Is it really right that we'll have to execute the test to get its name (from the Ok Text result)?

No, I'm not sure what this means either; I think the last two screenshots might be a little half-baked :)  Maybe that `Text` in the `Ok` field was meant to be what's printed to the screen in a test summary, but it isn't related to the test's name.

IO tests will have some overlap with this, but it's TBD.  For one, they shouldn't be executed without user intervention.

Thanks again for the questions and feedback!

@pchiusano Here's a feature idea: before I `add` my .u to the codebase, let me see which tests would newly start to pass or fail if I did.

francisdb:
Great to see this coming along, will test this on my json lib once it is merged on master

atacratic:
Thanks @aryairani, that all makes sense.

I realise I wrote it slightly wrong with 

> after a code change it only reruns tests that actually hit the changed code

Actually it's that it reruns tests that transitively reference changed code - so, a static property rather than a dynamic one.

That is actually an interesting difference in the case someone has written some kind of scheduler or work dispatcher in their unison code. Depending on how they do it, that could easily degenerate into 'any code change makes it want to rerun all the tests.'  But anyway, what we're doing here is still all good.

pchiusano:
Am going to merge this, a couple things just reading through the comments - you can name any watch expression, including tests, just by declaring a binding, for instance, this is valid syntax:

```
test> x = Fail "oh noes!"

> z = 1 + 1

> y = z * 2

> y + 1
```

The name of the test is just the name of the definition, it's currently how you refer to the test if you want to view or edit its source. This is different than the diagnostic information produced when the test completes with an `Ok` or `Fail`, which I'll call the "test message". The test message might be like:

`Sequence.map (checking that map id xs == xs): FAILED for input [1,2,3]`

so it includes a little description of the test and even information about the failure as well. Again, this `Test.Result` type is just meant as a compilation target, you don't create `Ok` and `Fail` values directly most likely. You could imagine a little EasyTest-like library that gives you an `ability Test` which gives you access to a `scope` function, a `fail` function, repeatable randomness, etc. And then you handle that ability in your watch expression, perhaps via `test> test '(1 + 1 === 2)`, where `test : '{Test} a -> Test.Result`.

Agree the display of the `test` command needs work. One thing I was thinking, don't show the test name  show the test name, show the source instead, and just present it as a numbered list, which `view` and `edit` can access via numbered lookup.

PR #492, opened 2019-05-08
unique types

pchiusano:
This PR allows you to create types whose identity is a combination of type structure + GUID. [Here's the main change to the implementation](https://github.com/unisonweb/unison/pull/492/files#r282575045). 

Useful for things like enums, whose meaning is determined "by convention". For instance:

```Haskell
unique type Day = Sun | Mon | Tue | Wed | Thu | Fri | Sat

-- alternately, if you want to supply your own guid
unique[anyValidWordyIdentifierHere] type Day = Sun | Mon | Tue | Wed | Thu | Fri | Sat
```

`unique` is a modifier you can place in front of `ability` or `type`, followed by an optional `wordyId` inside `[]`. If the user doesn't provide a GUID, one will be generated using system randomness.

Screen Shot 2019-05-08 at 3 44 14 PM

__A couple things still todo before mergeable:__

- [x] File watch typechecking display should render the modifier
- [x] `unique` should introduce a layout block
- [x] test `.u` files
- [x] `view` should show the `unique[..]` on its own line (and add test for this)
- [x] `find` should elide the `unique` guid

### DIFF
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -21,7 +21,6 @@ import           Unison.DataDeclaration         ( DataDeclaration'
                                                 , EffectDeclaration'
                                                 )
 import qualified Unison.DataDeclaration        as DD
-import qualified Unison.FileParser             as FileParser
 import qualified Unison.Lexer                  as L
 import           Unison.Parser                  ( Ann(..) )
 import qualified Unison.Parser                 as Parser
@@ -72,14 +71,6 @@ constructorType r =
   else if any f (builtinEffectDecls @Symbol) then CT.Effect
   else error "a builtin term referenced a constructor for a non-builtin type"
   where f = (==r) . fst . snd
-
--- todo: does this need to be refactored if we have mutually recursive decls
-parseDataDeclAsBuiltin :: Var v => String -> (v, (R.Reference, DataDeclaration v))
-parseDataDeclAsBuiltin s =
-  let (v, dd) = either (error . showParseError s) id $
-        Parser.run (Parser.root FileParser.dataDeclaration) s mempty
-      [(_, r, dd')] = DD.hashDecls $ Map.singleton v (DD.bindBuiltins names0 dd)
-  in (v, (r, const Intrinsic <$> dd'))
 
 -- Todo: These definitions and groupings of builtins are getting a little
 -- confusing.  Sort out these labrinthine definitions!
--- a/parser-typechecker/src/Unison/Codebase/Editor.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor.hs
@@ -698,7 +698,7 @@ commandLine awaitInput rt notifyUser codebase command = do
       fileToBranch handler codebase branch unisonFile
     Typecheck ambient branch sourceName source -> do
       -- todo: if guids are being shown to users, not ideal to generate new guid every time
-      namegen <- Parser.uniqueBase58Namegen 8
+      namegen <- Parser.uniqueBase58Namegen
       typecheck ambient codebase (namegen, Branch.toNames branch) sourceName source
     Evaluate ppe unisonFile           -> evalUnisonFile ppe unisonFile
     ListBranches                      -> Codebase.branches codebase
--- a/parser-typechecker/src/Unison/Codebase/Serialization/V0.hs
+++ b/parser-typechecker/src/Unison/Codebase/Serialization/V0.hs
@@ -539,15 +539,27 @@ putDataDeclaration :: (MonadPut m, Ord v)
                    -> DataDeclaration' v a
                    -> m ()
 putDataDeclaration putV putA decl = do
-  putA $ DataDeclaration.annotation decl
+  putModifier (DataDeclaration.modifier decl)
+  putA (DataDeclaration.annotation decl)
   putFoldable (DataDeclaration.bound decl) putV
   putFoldable (DataDeclaration.constructors' decl) (putTuple3' putA putV (putType putV putA))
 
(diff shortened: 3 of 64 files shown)


pchiusano:
Okay, this is ready for merge I'd say. We decided to not display the GUID that is generated, it's discouraged to make up your own GUID and this will probably be a hidden, undocumented feature, used internally for things like `IO` and `Either` which need a stable identifier and must always be added to a new codebase. The only reason to ever show it is if you want to be able to copy/paste blobs of text to share Unison code, but that's a pretty janky way of sharing code which isn't even guaranteed to work (since what if identifiers in the code refer to something different for me than for you?). We can come up with something a lot nicer for that or if needed a separate `export` command that's distinct from `view`.

If you leave a definition in your scratch file and attempt to re-add, although the GUID will be different, the name of the type will be the same, so Unison will complain that you should rename one or the other and will refuse to add it. Which seems like fine behavior. Maybe a more specific message in this case would be nice (when the names and structures of two unique types are the same, but only the GUID differs).

The "I typechecked" output now uses the same logic as the "I added" output, and I deleted that weird duplicate code in `PrintError` that printed the same information but a little differently. BONUS: definitions are nicely aligned in that message now. Also added a bit of color to type headers, I thought this made it easier to scan search results which include both types and terms.

Screen Shot 2019-05-09 at 12 05 22 PM

Type headers show their type parameters now (so `type Optional a` shows up in the search results, rather than just `type Optional`)

Screen Shot 2019-05-09 at 11 59 27 AM

Random things:

* Got rid of `effect` keyword. It's all `ability` now. It would have been extra work to maintain both keywords, based on how the lexer handles the `unique` keyword, so I figured now was a good time.
* No linebreak allowed after `unique`, I didn't think that was as important now that guid isn't ever being displayed. This would be not too annoying to add later if desired. `unique` starts a layout block, at a column one greater than the column where `unique` appears.

parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -73,14 +72,6 @@ constructorType r =
   else error "a builtin term referenced a constructor for a non-builtin type"
   where f = (==r) . fst . snd
 
--- todo: does this need to be refactored if we have mutually recursive decls
```
pchiusano: pretty funny, this definition was just not being used! DELETED

parser-typechecker/src/Unison/Codebase/Editor.hs
```diff
@@ -698,7 +698,7 @@ commandLine awaitInput rt notifyUser codebase command = do
       fileToBranch handler codebase branch unisonFile
     Typecheck ambient branch sourceName source -> do
       -- todo: if guids are being shown to users, not ideal to generate new guid every time
-      namegen <- Parser.uniqueBase58Namegen 8
+      namegen <- Parser.uniqueBase58Namegen
```
pchiusano: Namegen now takes a base58 length as a parameter, at each usage site.

parser-typechecker/src/Unison/Codebase/Serialization/V0.hs
```diff
@@ -539,16 +539,28 @@ putDataDeclaration :: (MonadPut m, Ord v)
                    -> DataDeclaration' v a
                    -> m ()
 putDataDeclaration putV putA decl = do
-  putA $ DataDeclaration.annotation decl
+  putModifier (DataDeclaration.modifier decl)
```
pchiusano: Nothing interesting here, just needed to add serialization and deserialization of the modifier (`unique` or `structural`)

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -215,18 +214,21 @@ notifyUser dir o = case o of
...
+  Typechecked sourceName ppe uf -> do
+    Console.setTitle "Unison ✅"
+    let terms = sortOn fst [ (HQ.fromVar v, typ) | (v, _, typ) <- join $ UF.topLevelComponents uf ]
+        typeDecls =
+          [ (HQ.fromVar v, Left e)  | (v, (_,e)) <- Map.toList (UF.effectDeclarations' uf) ] ++
+          [ (HQ.fromVar v, Right d) | (v, (_,d)) <- Map.toList (UF.dataDeclarations' uf) ]
+    if UF.nonEmpty uf then putPrettyLn' . ("\n" <>) . P.okCallout . P.sep "\n\n" $ [
+      P.wrap $ "I found and" <> P.bold "typechecked" <> "these definitions in "
+            <> P.group (P.text sourceName <> ":"),
+      P.indentN 2 . P.sepNonEmpty "\n\n" $ [
```
pchiusano: This is a good idiom if you have two or more lists of things that you want separated by blank lines.

pchiusano:
⚠️ IMPORTANT NOTE - because hashing of data types now includes an extra "level" (with the modifier), the hash values are now completely different. You'll have to destroy and recreate your codebase from text files.

These kinds of breaking format changes are something we should stop doing after getting to M1. After that we can have a more controlled codebase upgrade process, so the codebase format can be ground truth and you can delete your text files. I don't think it's worth trying to do this before M1 though.

parser-typechecker/src/Unison/DataDeclaration.hs
```diff
@@ -47,7 +47,11 @@ type ConstructorId = Int
 
 type DataDeclaration v = DataDeclaration' v ()
 
+data Modifier = Structural | Unique Text -- | Opaque (Set Reference)
```
pchiusano: This is really the key change. All type declarations have a modifier, and then I went and fixed all the compile errors.

pchiusano:
Will merge this tomorrow morning if no objections, I think it's in good shape.

PR #503, opened 2019-05-19
[topic/cbor] implemented V0 and V1 in CBOR.

joshcough:


pchiusano:
Context on this PR - Josh was looking for a project to start with and I thought it could be a good idea to move codebase format to use CBOR rather than custom binary format. It has the added bonus of having a bijection with JSON, so we can produce JSON from the same serialization code.

pchiusano:
Looks really good so far. I like that `Encoding` is just a monoid.

@joshcough I'm curious to know about the format sizes and how the JSON looks. Could you grab a couple definitions (use `ls -l` to get hashes which you can use to track down the `compiled.ub` file) and report the results here of:

* How many bytes is V0 format?
* How many bytes is the binary CBOR format?
* How many bytes is the JSON-ified version of that CBOR?
* Can you paste the JSON-ified version here (and tell us the original source) so we can take a look.

pchiusano:
@joshcough and I just chatted, this was excellent work to investigate all this but we aren't going to merge this. Rationale is:

* The JSON produced by the CBOR-JSON conversion is not really human consumable as is. You'd need to write a decoder for it to get it into some more computeable form. Once you're in that territory, it's the same or less work to just decode from the binary format directly, and we could provide (say) a PureScript library for that purpose.
* File sizes are bigger, even the binary form of the CBOR was 20-40% bigger.
* CBOR means we give up some control over the format. Which could be okay if we were gaining something else, but doesn't seem like we are.

Closing. It's still on the plan to get a sweet JS codebase viewer but work on this probably won't start until after M1.

Issue #500, opened 2019-05-16
Parser regression regarding int array literals


BenFradet:
A fresh copy of master gives:

```
unexpected +
expecting ], and, infixApp, lambda, or or
   34 | x = [+1, +2, +3]
```

whereas `[ +1, +2, +3 ]` works.

BenFradet:
I'll try having a look :+1:

PR #504, opened 2019-05-20
Fix parser regression regarding int array literals

BenFradet:
c.f. #500 

turns out it was introduced by 89e39f95f957340f6933448923d2fa036bf5f077

I went a bit ham with the fix, happy to change it to something more fine-grained :+1:


pchiusano:
Excellent, thanks!

PR #505, opened 2019-05-21
Remove Type.stream and Type.streamRef

TomasMikula:
as `Stream` is not a builtin type.


pchiusano:
Thanks for this! could you add yourself to CONTRIBUTORS file as part of this PR?

TomasMikula:
Sure, done.

PR #506, opened 2019-05-25
`view` reports missing names.

### LINKED ISSUE #490: `view <name>` should emit a message if <name> is not found
I type `view foo`, where `foo` is a name that does not exist in the codebase, and I just see

```
master> view foo

  

```
There should be an explanatory message.

TomasMikula:
Resolves #490.

Before:

```
master> view Int.+ foo

  -- Int.+ is built-in.

master> 
```

After:

```
master> view Int.+ foo

  -- Int.+ is built-in.


  ⚠️
  -- foo not found in the codebase.

master> 
```

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -903,21 +903,19 @@ asSearchResults b =
   tp(n,r) = SR.typeResult (hashQualifiedTypeName b n r) r (hashNamesForType r b)
 
 -- note: I expect these two functions will go away
--- Returns matching search results ordered by score
-searchBranch :: forall score. Ord score => Branch0 -> (Name -> Name -> Maybe score) -> [HashQualified] -> [SearchResult]
+-- Returns matching search results ordered by score.
+-- The order of results corresponds to the order of queries.
+searchBranch :: forall score. Ord score => Branch0 -> (Name -> Name -> Maybe score) -> [HashQualified] -> [[SearchResult]]
 searchBranch b score queries =
-  nubOrd . fmap snd . toList $
-    searchTermNamespace b score queries <> searchTypeNamespace b score queries
+  map (\q -> nubOrd . map snd . toList $ searchTermNamespace b score q <> searchTypeNamespace b score q) queries
   where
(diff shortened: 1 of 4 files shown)


aryairani:
Hey @TomasMikula, thanks for submitting this PR. I've held off on merging it though, because it's in a part of the code that we're trying to rewrite completely in another branch. I'm going to leave it open for now as a reminder that we want this functionality in the rewrite too; or maybe when we get there, we can call on you to add it again? :) If it does end up getting closed without merging though, it'd be for this same reason.

TomasMikula:
No problem, thanks for letting me know. I can have a look when the rewrite is ready (if it does not already contain this functionality).

I also wanted to investigate other `view` features:
 - show type signatures of built-in terms,
 - ability to show _full_ type signatures (i.e. with explicit `forall`s).

I suppose these are better held off as well.

aryairani:
Hi @TomasMikula — if you are still open to it, please go ahead and submit a PR for this functionality against the `wip/branchless` branch. :)

TomasMikula:
Will do.

TomasMikula:
@aryairani I'm now unable to determine the code path for the `view` command. I thought it would be handled in the `ShowDefinitionI` case in `HandleInput.hs`, but if I change that to

```haskell
     case input of
      -- [ ... ]
      ShowDefinitionI outputLoc (fmap HQ.fromString -> hqs) ->
       let !_ = error "SHOW DEFINITION"
       in -- the original body follows
```

then run `stack build && stack exec unison` and then in Unison type `view foo`, it does not crash. ¯\\\_(ツ)\_/¯

aryairani:
@TomasMikula Oh no, sorry!  Use `stack build && stack exec unison2` — counterintuitively, the one you're running is the one you had built previously from the `master` branch.  (I kept both around for now, to help compare behavior between the old and `wip/branchless` executables.

aryairani:
@TomasMikula Also I just reworked that function a bit for #535, so I wanted to warn you to pull the latest before putting in a lot of work!  I don't anticipate changing it for a while now going forward, so it should be safe.

TomasMikula:
Thanks for clarification!

TomasMikula:
Superseded by #554.

Issue #499, opened 2019-05-15
Support pluggable syntax for Unison


pchiusano:
Unison's syntax is just used for parsing and pretty-printing of definitions, but the actual definitions are stored in the codebase as their AST. This means we could support multiple syntaxes for the the language. Alice could write some Unison code in Python-like syntax, add it to the codebase, and then Bob could view Alice's definition in C-like syntax or Lisp-like syntax. Your `~/.unison` file might have `syntax = Pythonish` in it to control your personal default for reading and writing.

If you want to help make this happen, it would be pretty straightforward. The first step is abstracting the current parser and prettyprinter into an interface, perhaps something like:

```Haskell
data Syntax v = Syntax { 
    parseFile     :: Text -> Either (Parser.Error v) (UnisonFile v Ann)
  -- Pretty (ColorText, Ann) means we can add location-based highlighting after the fact
  , prettyTerm    :: PrettyPrintEnv -> AnnotatedTerm v Ann -> Pretty (ColorText, Ann)
  , prettyType    :: PrettyPrintEnv -> AnnotatedType v Ann -> Pretty (ColorText, Ann)
  , prettyData    :: PrettyPrintEnv -> DataDeclaration' v Ann -> Pretty (ColorText, Ann)
  , prettyAbility :: PrettyPrintEnv -> EffectDeclaration' v Ann -> Pretty (ColorText, Ann) }
```

So define this (perhaps in `Unison.Syntax`), and then provide an implementation of it for the current syntax (can put in `Unison.Syntax.Default`). Then it's just an exercise in plumbing a `Syntax` value everywhere that the current implementation calls the existing parser and prettyprinter directly (maybe rename or move the existing implementations to make sure that the compiler will alert you to all the places that need changing). Once that's done, the choice of `Syntax` can become a runtime parameter, chosen on startup by reading a `~/.unison` file, command line flag, or something similar.

atacratic:
Cool!!

aryairani:
Side note: let's have the user config file and the codebase dirs have different names; they shouldn't both be `.unison`

aryairani:
@pchiusano 
> -- Pretty (ColorText, Ann) means we can add location-based highlighting after the fact

?

pchiusano:
Using Pretty.map, you can map over the segments of the Pretty. If those segments know their location, you can pick different colors for each location. This could be used for highlighting portions of a type in type error messages, for instance.

seagreen:
[...]

PR #508, opened 2019-05-29
Type signatures of builtins not dependent on parser.

TomasMikula:
Motivation:
 - It seems reasonable to not depend on parser for declaration of builtins.
   Even more so if there is going to be pluggable syntax (#499).
 - There was a cyclic definition in Builtin.hs,
   `t` depends on `names` depends on `builtins0` depends on `t`,
   which only worked thanks to laziness. The cycle is now eliminated.

### DIFF
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -13,6 +13,7 @@ import           Control.Applicative            ( liftA2
 import qualified Data.Map                      as Map
 import           Data.Set                       ( Set )
 import qualified Data.Set                      as Set
+import           Data.Text.Internal             ( Text )
 import qualified Text.Megaparsec.Error         as MPE
 import qualified Unison.ABT                    as ABT
 import           Unison.Codebase.CodeLookup     ( CodeLookup(..) )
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -156,140 +160,182 @@ builtinTypeDependents r = Rel.lookupRan r builtinDependencies
...
+      , ("Sequence.snoc", forall1 "a" (\a -> sequence a --> a --> sequence a))
+      , ("Sequence.take", forall1 "a" (\a -> nat --> sequence a --> sequence a))
+      , ("Sequence.drop", forall1 "a" (\a -> nat --> sequence a --> sequence a))
+      , ("Sequence.++", forall1 "a" (\a -> sequence a --> sequence a --> sequence a))
+      , ("Sequence.size", forall1 "a" (\a -> sequence a --> nat))
+      , ("Sequence.at", forall1 "a" (\a -> nat --> sequence a --> optional a))
+
+      , ("Debug.watch", forall1 "a" (\a -> text --> a --> a))
+      , ("Effect.pure", forall2 "a" "e" (\a e -> a --> effect e a)) -- Effect ambient e a
+      , ("Effect.bind", forall4 "e" "a" "ambient" "b" (\e a ambient b -> delayed (effectful e a) --> (a --> effectful ambient b) --> effect e a)) -- Effect ambient e a
```
TomasMikula: Even though the tests pass, I'm not sure I got the type of `Effect.bind` right. What is the full version (with explicit `forall`s) of the original `'{e} a -> (a ->{ambient} b) -> Effect e a`?

pchiusano:
This looks good, actually came out nicer than I thought it would! Thanks. Merging. 

Could you open the same change against wip/branchless, basically applying the same treatment to `Builtin2.hs`?

PR #522, opened 2019-06-20
Some small tweaks to the typechecker code

TomasMikula:
Here are some small edits to the typechecker code that should not change the behavior, but hopefully are improvements to readability.

See each commit's message for description. If you don't find some of them adding any value, then ditch it, no problem.


pchiusano:
Cool! Looks quite reasonable.

PR #563, opened 2019-07-06
Pretty-printer support for colored output

### LINKED ISSUE #287: Change pretty printers to use color
Currently, `Unison.TypePrinter` and `Unison.TermPrinter` produce a `Pretty String`. It would be great if it could produce color, when printing terms to the console, and we can also use this as the basis for HTML rendering (also, having a "copy to HTML" could be nice if you're like, pasting some Unison code into an email and want syntax highlighting).

Goal: change all pretty printers to return a `Pretty ColorText`. Since `ColorText` (reexported by `Unison.Util.Pretty` module I think) has an `IsString` instance, change should be really simple to do. Like it might even just work if you change the signature of the `pretty` functions.

After that, we could add some simple syntax highlighting - I'd suggest adding some constants for different semantic elements (keywords, numeric literals, special operators like `->` and `{}`, `'`).

Since we can map over the `Pretty`, the exact colors chosen doesn't matter too much, as long as distinct colors are used for distinct elements, we can always remap them later.

@atacratic this is all yours if you like, but you can assign to someone else if you're busy with other stuff

atacratic:
Fixes #287 

Produces output that looks like this:
image

Bikeshedding over colors welcome!  I'd recommend actually checking out the branch and playing around with how different colors look, rather than just trying to imagine.  
 - I started from the colors that sublime uses for haskell by default, FWIW.  I'm not especially attached to these.
 - Effectful stuff: uses of requests, and the braces around lists of ability type names in signatures (unfortunately not the ability type names themselves), get hiGreen.
 - Force and delay (! and ') get yellow.
 - Things I'm not super sure about:
    - does so much stuff really want to be green?
    - really red for all the control keywords, and `->`?  
    - what is something just being white meant to signify?  white stuff actually looks quite prominent  (e.g. type parameters)
    - should all the things I've lumped together as control keywords (`let|handle|in|where|case|of|->|if|then|else|and|or`) really be colored the same?  Maybe `let` and `where` should be white or even hiBlack rather than red.  Maybe `->` should be white.
    - have I successfully implemented 'types are blue', and not confused which things are type-level and which things are value-level?
        - In, say, `reverse : [a] -> [a]`, the list brackets are rendered white rather than blue - maybe wrong?

The mapping is in [SyntaxHighlights.hs](https://github.com/unisonweb/unison/blob/c255d4dfd151f998cbc71b3ead5560d96d505dbb/parser-typechecker/src/Unison/SyntaxHighlights.hs).  Some of the way this breaks things down is inspired by the [atom syntax highlighting unison grammar](https://github.com/unisonweb/unison/blob/c255d4dfd151f998cbc71b3ead5560d96d505dbb/editor-support/atom/language-unison/grammars/unison.cson).  Feel free to just change the color scheme yourself before merging.  

The pretty-printer is now free of direct uses of color markup (`hiBlack` etc), and I think should stay that way.  

My intention was that all non-whitespace characters emitted by the pretty-printer are marked up with one of the syntax highlight elements.  (I didn't quite achieve that because things like commas emitted by `Pretty.commas` aren't marked up.)  

I haven't attempted to sync this up with any other syntax highlighting that might be going on elsewhere, e.g. in error message generation.  Is there anything to do there?

This changes some of the pretty-printing functions over from returning `Pretty s` to `Pretty ColorText`.

Also smuggled in with this: a fix to avoid emitting `use () ()` or `use Pair Pair`, which are usually unnecessary due to how tuple syntax works.  Bug in FQN elision spotted by Paul.

### DIFF
--- a/parser-typechecker/src/Unison/SyntaxHighlights.hs
+++ b/parser-typechecker/src/Unison/SyntaxHighlights.hs
@@ -1,0 +1,40 @@
+module Unison.SyntaxHighlights where
+
+import Unison.Util.ColorText
+
+-- The elements of the Unison grammar, for syntax highlighting purposes
+data Element = NumericLiteral
+             | TextLiteral
+             | BooleanLiteral
+             | Blank
+             | Var
+             | Reference
+             | Constructor
+             | Request
+             -- let|alias|handle|in|namespace|type|ability|where|case|of|->|if|then|else|and|or
+             | ControlKeyword 
+             | TypeAscriptionColon
+             -- forall|∀|->
+             | TypeOperator
+             | UseStatement
+             -- , ` [ ] { } @ | =
+             | DelimiterChar
+             | Parenthesis
+
+-- TODO this is currently just inspired by the sublime haskell colors - what would people rather?
+defaultColors :: Element -> ColorText -> ColorText
+defaultColors = \case 
+  NumericLiteral      -> purple
+  TextLiteral         -> yellow
+  BooleanLiteral      -> cyan
+  Blank               -> hiCyan
+  Var                 -> white
+  Reference           -> green
+  Constructor         -> green
+  Request             -> green
+  ControlKeyword      -> red
+  TypeAscriptionColon -> red
+  TypeOperator        -> red
+  UseStatement        -> white
+  DelimiterChar       -> white
+  Parenthesis         -> white
--- a/parser-typechecker/src/Unison/SyntaxHighlights.hs
+++ b/parser-typechecker/src/Unison/SyntaxHighlights.hs
@@ -22,7 +22,7 @@ data Element = NumericLiteral
              | UseKeyword
              | UsePrefix
              | UseSuffix
-             -- ? ! , ` [ ] { } @ | = _
+             -- ? ! , ' ` [ ] { } @ | = _
              | DelimiterChar
              | Parenthesis
 
--- a/parser-typechecker/src/Unison/SyntaxHighlights.hs
+++ b/parser-typechecker/src/Unison/SyntaxHighlights.hs
@@ -12,18 +12,29 @@ data Element = NumericLiteral
              | Reference
              | Constructor
              | Request
-             -- let|alias|handle|in|namespace|type|ability|where|case|of|->|if|then|else|and|or
+             | AbilityBraces
+             -- let|alias|handle|in|namespace|where|case|of|->|if|then|else|and|or
              | ControlKeyword 
              | BindingEquals
-             | TypeAscriptionColon
              -- forall|∀|->
              | TypeOperator
+             | TypeAscriptionColon
+             -- type|ability
+             | DataTypeKeyword
+             | DataTypeParams
+             | DataType
+             -- unique
+             | DataTypeModifier
              -- `use Foo bar` is keyword, prefix, suffix
              | UseKeyword
              | UsePrefix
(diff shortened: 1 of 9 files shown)


atacratic:
I don't know if we should try and match the colours used by the editor syntax highlighting we have, eg atom and vi. Looks like we don't (/can't?) actually specify the colours used though, or at least I didn't see anything for that in the atom integration.

Might be a bit of a shame if people's editors by default showed them completely different colors to what they see in the unison tool.

A completely matching colouring isn't likely though - the fact that editors seem to identify syntax elements by regex would make it too hard/clumsy to exactly match what the pretty-printer does.

parser-typechecker/src/Unison/SyntaxHighlights.hs
```diff
@@ -0,0 +1,69 @@
...
+             -- ? , ` [ ] @ |
+             -- Currently not all commas in the pretty-print output are marked up as DelimiterChar - we miss
+             -- out characters emitted by Pretty.hs helpers like Pretty.commas.
+             | DelimiterChar
+             -- ! '
+             | Parenthesis
+
+defaultColors :: Element -> ColorText -> ColorText
+defaultColors = \case 
+  NumericLiteral      -> purple
```
pchiusano: Just pointing out that you can say `id` for any of these to use the default body text color of the terminal.

pchiusano:
Nice!! Will definitely play with colors after merging. I'd probably err on the side of using fewer colors.

parser-typechecker/src/Unison/TypePrinter.hs
```diff
@@ -45,42 +46,43 @@ import qualified Unison.DataDeclaration as DD
...
 
 pretty
-  :: forall s v a . (IsString s, LL.ListLike s Char, Var v)
+  :: forall v a . (Var v)
   => PrettyPrintEnv
   -> Imports
   -> Int
   -> AnnotatedType v a
-  -> Pretty s
+  -> Pretty ColorText
```
pchiusano: I'm thinking we should work with `Pretty (AnnotatedText SyntaxHighlighting.Element)`, and then `Pretty.map` is used to convert that to `Pretty ColorText` just before rendering to ANSI. For rendering to ANSI terminal you may not use all the colors given how limited the palette is, but it's unfortunate to throw that information away because other targets could use it.

For instance, if you're rendering a `Pretty (AnnotatedText Element)` to HTML or some Javascript-y widget, you'd likely have a CSS class for each `Element` so that the whole thing can be styled with CSS.

atacratic: Ah, good point!! I will make this change, but it will probably take me a week or so to get round to it. I'll still do it even if you merge the PR in the mean time.

atacratic: @pchiusano umm, I just banged the following commit in without asking, doing a couple of renames, hope that's not too much of a transgression given it's trivial - I wanted to minimize any merging.  Remainder of this refactor will get a PR as per proper procedure.  Shout if I should undo/change the renames.

https://github.com/unisonweb/unison/commit/45122cb4a019a99f8afa8f4ca5c7057ef1983b1e

atacratic: (the rest of the refactor is in https://github.com/unisonweb/unison/pull/564)

PR #564, opened 2019-07-08
Refactor most pretty-printer functions to speak SyntaxText instead of ColorText

atacratic:
Refactors most pretty-printer functions to speak `SyntaxText` instead of `ColorText`.

This is coming out of a [comment from Paul](https://github.com/unisonweb/unison/pull/563#discussion_r300846462).

`type SyntaxText = AnnotatedText Util.SyntaxText.Element`
i.e. a markup that identifies different syntactic elements.

The top-level `pretty` and `pretty'` wrappers still return a `ColorText`, but the rest
of the pretty-printer now works using `SyntaxText`.  The mapping from the syntactic
markup to the color markup is `Pretty.syntaxToColor`.

This sets us up better for different pretty-print targets (eg HTML/CSS).

OutputMessages.hs now has quite a lot of calls to `syntaxToColor` peppered through it.

The `defaultColors` mapping now lives in ColorText.hs.


atacratic:
@pchiusano - I've merged this up to latest wip/branchless.  Let me know if there's something else stopping it going in (other than getting time to look at it).

pchiusano:
@atacratic Cool, I think this is good to merge. There are some merge conflicts in TypePrinter which you can resolve in favor of your branch, they all look like they are due to a brittany formatting that @runarorama applied, which we'll try to be more careful about. Haven't looked at `ColorText` conflicts. 

If it gets hairy reconciling the merge, let me or @runarorama know. Thanks!!

PR #571, opened 2019-07-13
Pretty-printer - don't put `use` inside of the `if` condition

### LINKED ISSUE #568: Let's not put `use` inside of the `if` condition
Even if we fix the formatting bug where a multiline condition isn't indented, I just think multiline conditions look pretty weird so we shouldn't insert a `use` statement in the middle of an `if` expression's condition:

Screen Shot 2019-07-12 at 3 15 51 PM

Also maybe related is #518

@atacratic can you take a look at this one?

atacratic:
Fixes #568

Instead of 

image

we now get

image


pchiusano:
Great!

Issue #570, opened 2019-07-12
when updating a type, its new constructors aren't added


aryairani:
```haskell
unique type T = T
type Foo = Foo T
```
![image](https://user-images.githubusercontent.com/538571/61163664-83f35200-a4dd-11e9-92ea-11dd72b1d384.png)
```haskell
type Foo = Foo2 T T
```
![image](https://user-images.githubusercontent.com/538571/61163693-c0bf4900-a4dd-11e9-9ee3-00401d4d2630.png)

Where is `Foo.Foo2`?

PR #572, opened 2019-07-14
Possible fix #548 and fix #544 but needs review

### LINKED ISSUE #544: can't typecheck a unique type with the same name as an existing one
Given .u:
```haskell
unique type MyEnum = X Nat
```
Save the .u file:
```
  ✅ I found and typechecked these definitions in ~/unison2/scratch.u:
  
       Legend:
    +  a new definition with a new name
    
    
    +  unique type MyEnum 
  
.> add

     Legend:
  +  a new definition with a new name
  
  
  +  unique type MyEnum 
  
.> 
```

Save the .u file again:

```
  ✅ I found and typechecked these definitions in ~/unison2/scratch.u:
  
       Legend:
    z  a constructor with the same name as a term
    b  blocked because of one of its dependencies
    
    
    b  unique type MyEnum 
    
    z  MyEnum.X 
  
.>
```

and then `update` also fails.

### LINKED ISSUE #548: can't update a type unless I rename the constructors, and then the old constructors don't get deleted
Given .u:
```
type MyEnum = X Int Int
```
Save, add.

Change .u to:
```
type MyEnum = X Nat Nat
```
Save.
```
  ✅ I found and typechecked these definitions in ~/unison2/scratch.u:
  
       Legend:
    z  a constructor with the same name as a term
    b  blocked because of one of its dependencies
    
    
    b  type MyEnum 
    
    z  MyEnum.X 
  
.> 
```
Also can't 'update'.

pchiusano:
A few things seemed to be busted:

* When updating a type, the constructors of the updated type weren't being added to the namespace. I changed the call to `doSlurpAdds` inside the `update` command to include these new constructors.
* When updating a type, the constructors for the old type with names in the current path weren't being removed from the namespace. I did that, using a helper function `Names3.constructorsForType`.
* There were some things about how the slurp result was being built up that were wrong, like: 
  * constructors of types being updated could still count as a ctor/term collision.
  * duplicates were getting double counted as both duplicates and aliases.

I don't know if the `patch` command correctly removes old constructors from the namespace for types being updated, but it should probably do that.

Lastly, there are a bunch of weird corner cases to think through when the constructors for a type don't live directly under the path of the type itself. Maybe we could think through those, but I'd be in favor of just enforcing this restriction, using a check in `rename.term` and `alias.term`.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -756,6 +756,7 @@ loop = do
         Nothing -> respond NoUnisonFile
         Just uf -> do
           slurpCheckNames0 <- slurpResultNames0
+          currentPathNames0 <- currentPathNames0
           let sr = applySelection hqs uf
                  . toSlurpResult uf
                  $ slurpCheckNames0
@@ -782,6 +783,11 @@ loop = do
                                  ++ Var.nameStr v ++ " but got: "
                                  ++ show otherwise
                   where n = Name.fromVar v
+              termDeprecations :: [(Name, Referent)]
+              termDeprecations =
+                [ (n, r) | (oldTypeRef,_) <- Map.elems typeEdits
+                         , (n, r) <- Names3.constructorsForType0 oldTypeRef currentPathNames0 ]
+
           ye'ol'Patch <- do
             b <- getAt p
             eval . Eval $ Branch.getPatch seg (Branch.head b)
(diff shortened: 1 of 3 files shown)


aryairani:
Does this fix #570 also?

aryairani:
Finger slipped.  Sounds like it does include #570

pchiusano:
Oh, yeah, it does, I just didn't know about that one.

aryairani:
> I don't know if the patch command correctly removes old constructors from the namespace for types being updated, but it should probably do that.

I think the `patch` command is only meant to deal with type-preserving term edits.

pchiusano:
@aryairani hmmm I was thinking that `patch` would apply the changes in the patch to the current namespace and then do the automatic propagation of type preserving stuff. 

So like if you have `foo` in your namespace, and what `foo` refers to has been updated by the patch, you move the name `foo` to point to the new thing. Likewise for types that have been updated by the patch. And for constructors, seems like you should delete old constructors for types that are being updated to their new versions...

If patch isn't doing this, then would we have a separate command or something?

aryairani:
That sounds reasonable if vague.  I’m not exactly sure what patch is or should be doing.

PR #576, opened 2019-07-16
Always print foralls in type signatures.

TomasMikula:
This is a proposal to always print `∀` in type signatures.

#### Current situation

Even if the user provides explicit `∀` in the source code:

```
id : ∀ a. a -> a
id a = a
```

it is not printed back:

```
    
      id : a -> a
```

#### Motivation

I think it is better to err on the side of displaying more information than what is in the source code, not less.

Also, seeing unbound variables can be confusing for beginners (I remember it used to be for me).

#### After this PR

```  
    
      id : ∀ a. a -> a
```

#### Possible improvement

If you find explicit foralls too obtrusive, they could possibly be printed in a lower contrast color.

### DIFF
--- a/parser-typechecker/src/Unison/TypePrinter.hs
+++ b/parser-typechecker/src/Unison/TypePrinter.hs
@@ -97,14 +97,12 @@ prettyRaw n im p tp = go n im p tp
     Effect1' e t ->
       PP.parenthesizeIf (p >= 10) $ go n im 9 e <> " " <> go n im 10 t
     Effects' es         -> effects (Just es)
-    ForallNamed' v body -> if p < 0
-      then go n im p body
-      else
-        paren True
-        $         fmt S.TypeOperator "∀ "
-        <>        fmt S.Var          (PP.text $ Var.name v)
-        <>        fmt S.TypeOperator "."
-        `PP.hang` go n im (-1) body
+    ForallNamed' v body ->
+      paren (p >= 0)
+      $         fmt S.TypeOperator "∀ "
+      <>        fmt S.Var          (PP.text $ Var.name v)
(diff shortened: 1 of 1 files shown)


TomasMikula:
I will fix the tests if there is preliminary agreement.

atacratic:
Maybe at some point in the future this sort of thing could be a matter of personal preference - configure unison to print the code according to your taste.

TomasMikula:
Yeah, or an interactive editor that can expand/collapse these on a very granular basis.

pchiusano:
I feel this is a personal preference, and until we can make it configurable, we gotta pick something. I personally prefer to elide the explicit forall. Sorry! Am going to close this one. However:

- I just noticed a bug with the prettyprinter - it should print out the quantifier if the variable being introduced is uppercase. I'll fix that. With that fix, any lowercase variable in a type signature is universally quantified, and any uppercase variable refers to a concrete type unless bound by an explicit forall.
- I'm okay with the idea of this being a configuration setting. @runarorama recently hooked up [configurator](http://hackage.haskell.org/package/configurator), so if you can read that config file and then plumb that information through everywhere, then you and I can each have our preferred setting for this, which seems like a win. I would probably leave the parser alone if doing this - parser will still generalize over lowercase variables and it's just a prettyprinting setting whether the foralls are shown.

Adding this as a configuration option isn't high priority at the moment for us, but not opposed to it.

pchiusano:
FYI, fixed issue I mentioned above: https://github.com/unisonweb/unison/pull/584

TomasMikula:
Thanks for hints on how to make it configurable! It is not too important for me either, though.

PR #584, opened 2019-07-19
Explicit foralls now get inserted when the quantified variable isn't …

pchiusano:
…a lowercase variable, also, in the event of nested foralls that require an explicit forall, only one `forall` symbol is shown.


parser-typechecker/src/Unison/TypePrinter.hs
```diff
@@ -91,12 +91,12 @@ prettyRaw n im p tp = go n im p tp
     Effect1' e t ->
       PP.parenthesizeIf (p >= 10) $ go n im 9 e <> " " <> go n im 10 t
     Effects' es         -> effects (Just es)
-    ForallNamed' v body -> if (p < 0)
+    ForallsNamed' vs body -> if (p < 0 && all Var.isLowercase vs)
```
pchiusano: @atacratic just calling your attention to this case, using `ForallsNamed` pattern to pull out nested foralls (which also covers the case of just one variable). The existing code would have inserted multiple forall symbols if they were nested.

parser-typechecker/tests/Unison/Test/TypePrinter.hs
```diff
@@ -86,6 +86,9 @@ test = scope "typeprinter" . tests $
   , tc "'(a -> b)"
   , tc "(a -> b) -> c"
   , tc "'a -> b"
+  , tc "∀ A. A -> A"
+  , tc "∀ foo.A. foo.A -> foo.A"
+  , tc "∀ A B. A -> B -> (A, B)"
```
pchiusano: @atacratic and here's a test - quantifier inserted because the variables are uppercase, but we only insert one `∀` symbol.

PR #587, opened 2019-07-22
Some prettprinting tweaks

### LINKED ISSUE #517: Surplus parentheses when if/then/else, handle/in and case/of nested under each other
The pretty-printer sometimes emits code with ugly and unnecessary parentheses, for example:
```
handle bar in
  (if foo then
    use A.X c
    f c c
  else
    use A.Y c
    g c c)
```
As per the comment 'Explanation of precedence handling' above `pretty` in TermPrinter.hs, if/then/else, handle/in, and case/of all use the same precedence level.

But the pretty-printer emits newlines and whitespace in such a way that the parser can pick up the right parenthesization based on layout.

Ideally the pretty-printer would use this fact to suppress some brackets.

pchiusano:
Switched to excluding the `()` and `Pair` not by name, but by Reference, when building up the annotations that form the basis of the `use` insertion. @atacratic could you let me know if this change looks okay? Or does this implementation somehow break something?

Also set ambient precedence to 0 for true and false branches of if, since parser does fine without the parens, even for things like `if true then if false then 0 else 1 else 2`, which looks confusing when all one one line like that but parses okay. I'd say this fixes #517 unless there are some issues with it that I can't see.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -679,7 +679,9 @@ countPatternUsages n p = Pattern.foldMap' f p where
     Pattern.SequenceOpP _ _ _ _   -> mempty
     Pattern.EffectPureP _ _       -> mempty
     Pattern.EffectBindP _ r i _ _ -> countHQ $ PrettyPrintEnv.patternName n r i
-    Pattern.ConstructorP _ r i _  -> countHQ $ PrettyPrintEnv.patternName n r i
+    Pattern.ConstructorP _ r i _  -> 
+      if r == DD.unitRef || r == DD.pairRef then mempty
+      else countHQ $ PrettyPrintEnv.patternName n r i
 
 countHQ :: HQ.HashQualified -> PrintAnnotation
 countHQ hq = fold $ fmap countName (HQ.toName $ hq)
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -35,6 +35,7 @@ import           Unison.NamePrinter             ( styleHashQualified'' )
 import qualified Unison.Pattern                as Pattern
 import           Unison.PatternP                ( Pattern )
 import qualified Unison.PatternP               as PatternP
+import           Unison.Reference               ( Reference )
 import qualified Unison.Referent               as Referent
 import qualified Unison.Util.SyntaxText        as S
 import           Unison.Util.SyntaxText         ( SyntaxText )
@@ -119,7 +120,7 @@ data InfixContext
        3x + 3y + ... 3z
 
      >=2
-       if 2a then 2b else 2c
+       if 0a then 0b else 0c
        handle 2h in 2b
        case 2x of
          a | 2g -> 0b
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -208,7 +208,7 @@ pretty0 n AmbientContext { precedence = p, blockContext = bc, infixContext = ic,
        pt     = branch t
        pf     = branch f
        branch tm = let (im', uses) = calcImports im tm
-                   in uses $ [pretty0 n (ac 2 Block im') tm]
+                   in uses $ [pretty0 n (ac 0 Block im') tm]
     And' x y ->
       paren (p >= 10) $ PP.spaced [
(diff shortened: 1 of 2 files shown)


pchiusano:
@atacratic Hmm, it looks like this didn't do the trick for eliding the `Pair` and `()` types and constructors. Even though the code I added is wrong, is there a way of implementing that sort of idea? We can't skip the `use` for `()` and `Pair` based on their name, since the local name for those terms and types may differ.

pchiusano:
@atacratic disregard, I just missed a spot, in the FQN usage accumulator for term patterns. Everything seems to be working now!

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -644,6 +646,7 @@ suffixCounterTerm n = \case
 suffixCounterType :: Var v => PrettyPrintEnv -> Type v a -> PrintAnnotation
 suffixCounterType n = \case
     Type.Var' v -> countHQ $ HQ.unsafeFromVar v
+    Type.Ref' r | noImportRefs r || r == Type.vectorRef -> mempty
```
atacratic: Why the check on `vectorRef` here but nothing for terms/patterns?

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -765,11 +778,6 @@ calcImports im tm = (im', render $ getUses result)
     -- further down, closer to the use sites.
     narrowestPossible :: Map Name (Prefix, Suffix, Int) -> Map Name (Prefix, Suffix, Int)
     narrowestPossible m = m |> Map.filter (\(p, s, i) -> not $ allInSubBlock tm p s i)
-    -- Don't do `use () ()` or `use Pair Pair`.  Tuple syntax generates ().() and Pair.Pair 
-    -- under the covers anyway.  This does mean that if someone is using Pair.Pair directly, 
-    -- then they'll miss out on FQN elision for that.  
-    exclusions :: Map Name (Prefix, Suffix, Int) -> Map Name (Prefix, Suffix, Int)
-    exclusions m = m |> Map.filterWithKey (\n _ -> notElem (Name.toString n) ["().()", "Pair.Pair"])
```
atacratic: There are plenty of other textual references to `"()"` and `"Pair"` in this file - do we need to get rid of those as well?

parser-typechecker/tests/Unison/Test/TermPrinter.hs
```diff
@@ -151,7 +151,7 @@ test = scope "termprinter" . tests $
   , tc "case e of { a } -> z"
   , pending $ tc "case e of { () -> k } -> z" -- TODO doesn't parse since 'many leaf' expected before the "-> k"
                                               -- need an actual effect constructor to test this with
-  , tc "if a then (if b then c else d) else e"
+  , tc "if a then if b then c else d else e"
```
atacratic: Are you sure the new way is better, in this example?

parser-typechecker/tests/Unison/Test/TermPrinter.hs
```diff
@@ -332,8 +332,8 @@ test = scope "termprinter" . tests $
                  \  go [] a b"
   , tc_breaks 30 "case x of\n\
                  \  (Optional.None, _) -> foo"
-  , tc_breaks 50 "if true then (case x of 12 -> x) else x"  -- re parens around case note #517
-  , tc_breaks 50 "if true then x else (case x of 12 -> x)"  -- re parens around case note #517
+  , tc_breaks 50 "if true then case x of 12 -> x else x"  -- re parens around case note #517
```
atacratic: Should remove the "-- re parens around case note #517" comments.

atacratic: (now I did)

parser-typechecker/src/Unison/TermPrinter.hs
pchiusano: There’s no list/vector/sequence constructor. Doesn’t hurt to check in both spots though if you want to just add to noImportRefs function.

parser-typechecker/tests/Unison/Test/TermPrinter.hs
pchiusano: In this example as a one liner, no. 🥴 If you break it onto multiple lines (which will often happen for more realistic examples) it seems ok to me.

pchiusano:
Am going to merge but agree on keeping 517 open.

PR #594, opened 2019-07-24
25-50% speed improvements to overall file slurp times

pchiusano:
I don't suggest merging this right now, I think I would prefer to keep moving pieces to a minimum. It can go into a future release when there's more time to get ambient testing of it.

Prior to this change, `Base.u` took consistently 6-9s for me; after this change: 3-5s. 

There were a few places in the typechecker where we were doing lookups by searching linearly through the list of typechecking context elements. Which is okay when the context is small, but for larger files those contexts can get big, leading to some quadratic behavior. Now these lookups use `Data.Map`.

Some notes on performance:

* Lexing `Base.u` is fast, basically imperceptible, which you can verify easily in the REPL. I was wondering if that was a problem.
* Not sure about parsing, but I'm curious how long parsing takes. I hypothesize pretty quick given how little backtracking there is in our parser.
* An experiment that would be easy to run - compare slurp times of a version of `Base.u` that uses no TDNR vs the current `Base.u`. Would isolate whether TDNR is a source of slowness.
* Also could try just running with profiling to see if anything jumps out.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -297,11 +297,14 @@ data MEnv v loc = MEnv {
   skipAbilityCheck :: [Type v loc]
 }
 
-newtype Context v loc = Context [(Element v loc, Info v)]
+newtype Context v loc = Context [(Element v loc, Info v loc)]
 
-data Info v =
+data Info v loc =
   Info { existentialVars :: Set v -- set of existentials seen so far
+       , solvedExistentials :: Map v (Monotype v loc) -- `v` is solved to some monotype
+       , unsolvedExistentials :: Set v
        , universalVars :: Set v -- set of universals seen so far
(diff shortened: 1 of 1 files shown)


pchiusano:
Notes for self: confirmed that parsing `Base.u` and resolving the names in it before typechecking is super fast / imperceptible and 99% of the time spent doing typechecking.

A thing I suspect is slowing things down in other cases is evaluation, which currently builds a single self-contained mega-term. It would be better to give the runtime something like a `Map Reference Term` of bindings keyed by their hashes to evaluate, and a `Reference -> m Term` function for loading terms from the codebase. Then the runtime can handle lazily loading transitive dependencies and can do lots of caching internally, keyed off the hashes.

pchiusano:
Okay, merging this after some thought - the change gets hammered by existing test suite, since since it changes how `apply`, `lookupSolved`, and `lookupAnn` work (a `Map.lookup` rather than scanning through a list), and those functions are called all over the place during typechecking, many times.

PR #610, opened 2019-07-30
Some quality of life improvements for `find` command

pchiusano:
This PR makes some tweaks to the `find` command:

* `find` searches both the local namespace and the root but prefers results from the current namespace. Previously it would only search the current namespace, forcing you to `cd` just to execute a search outside your current namespace. This got old real quick, since you are often working within a small namespace but referencing plenty of external definitions.
* `find` results are capped at 15, then it shows `... 28 more`. (For later, make this configurable or have a separate command to show all results). Most of the time, it's not really very useful to have screens full of results, only a few lines of which you are actually interested in. You're better off just refining your search. I feel this is a case where the CLI just can't provide the right experience - you want a search box that refines in real time as you type.
* The autocomplete provider doesn't cap the number of completions, so a typical usage will tab complete until getting the list to a manageable size, then press enter to get detail.
* The matching algorithm is a simple `isInfixOf` check. Of those that match, it prefers prefix matches, suffix matches, infix matches, and lastly infix matches of the wrong case. There's a penalty for names that start with '.' so that local names are preferred.
* Old matching algorithm is still around, just not hooked up. I could see the old algorithm being ok if you have a UI that can refine results as you type, since in that case, the extra fuzziness doesn't get in your way as much as long as the ranking is okay.

I think the find command could still use more work in how it summarizes / aggregates results, like I'd love to see it summarize "the `math` namespace has 45 definitions". Another day. 

I'd call this PR is a definite improvement over what's there now.

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
+++ b/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
@@ -879,7 +879,8 @@ listOfDefinitions' :: Var v
 listOfDefinitions' ppe detailed results =
   if null results then noResults
   else P.lines . P.nonEmpty $ prettyNumberedResults :
-    [formatMissingStuff termsWithMissingTypes missingTypes
+    [ if len <= cap then "" else "... " <> P.shown (len - cap) <> " more" 
+    ,formatMissingStuff termsWithMissingTypes missingTypes
     ,unlessM (null missingBuiltins) . bigproblem $ P.wrap
       "I encountered an inconsistency in the codebase; these definitions refer to built-ins that this version of unison doesn't know about:" `P.hang`
         P.column2 ( (P.bold "Name", P.bold "Built-in")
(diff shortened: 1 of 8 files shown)


aryairani:
image
LGTM

aryairani:
Actually I was expecting the last search to return `.Bytes.take` also, but that's not related to this PR.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -879,7 +879,8 @@ listOfDefinitions' :: Var v
 listOfDefinitions' ppe detailed results =
```
aryairani: Is `find` the only place `listOfDefinitions'` is used?  Guessing we don't generally want to cap every caller's results.  
Also what if there are more than `cap` definitions ending in `take`?  Yes we can refine it if we know the prefix, but what if we don't?  We shouldn't cap results without explaining how to ask for the next page / capful.  e.g. `hoogle` says:
> -- plus more results not shown, pass --count=20 to see more

pchiusano:
Okay, good news, it ended up being straightforward to plumb an extra flag through to indicate whether to show full results (and same idea could be used for showing individual pages of results too, but I didn't implement that). So now you get:

Screen Shot 2019-07-30 at 10 53 18 AM

And then `find.all map` gives:
 
Screen Shot 2019-07-30 at 10 53 55 AM

`find.verbose` also implies `find.all` right now. We'd need some different syntax for command flags if we want to be able to combine these, but I don't want to mess with that for this PR and I'm not sure it's really necessary anyway.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
pchiusano: Good call - it wasn't. Was also used for listing definitions that would be impacted by a delete and maybe something else. Now that we are passing the show all flag explicitly to the `listOfDefinitions'` function, it caught all these places and I just did show all at each call site to preserve their current behavior.

Issue #611, opened 2019-07-30
Parser of `Float` literals should support exponents.


runarorama:
A resolution of this issue should add parsing and pretty-printing for exponents in floating point numbers. For example `1e-3` for `0.003`.

It's currently a syntax error to write `1.6777217e7`, but this should be parsed as the number `16777216.0`. Without the exponent portion, the vast majority of floating point numbers are inaccessible to literals.

The parser should expect floating point literals to have the following form:

``[`-`](decimal.decimal[exponent]|decimal[exponent])``

``where exponent = (`e`|`E`)[`-`]decimal``

pchiusano:
Tip for anyone looking into this, I think on the parsing side, just modifying this one function in the lexer might do it: 

https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Lexer.hs#L498

The `Either Err (String,String)` it is returning is the (unparsed numeric literal, remaining input). The unparsed literal is later converted by the parser to an actual number.

pchiusano:
One more parsing spot - I think this check in term parser here won't be quite right, instead should additionally check for presence of `e` or `E` character maybe?

https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/TermParser.hs#L528

atacratic:
If someone looks at this, it might also be a good moment to consider whether we should be able to parse or pretty-print float literals with values Infinity, -Infinity and NaN.  ([comment](https://github.com/unisonweb/unison/blob/bf943e2df6b33acfcca549dfb8617d99e3988b35/parser-typechecker/src/Unison/TermPrinter.hs#L166))

scott-christopher:
Happy to have a look at this.

pchiusano:
This is closed, thanks to @scott-christopher :)

PR #617, opened 2019-07-31
#611 Adds support for exponents in Float literals

scott-christopher:
Addresses #611 

This change adds support for Float literals to include exponents, such as:
```
1e3
1e+3
1e-3
+1e3
+1e+3
+1e-3
-1e3
-1e+3
-1e-3
1.2e3
1.2e+3
1.2e-3
+1.2e3
+1.2e+3
+1.2e-3
-1.2e3
-1.2e+3
-1.2e-3
```
Both `e` and `E` are supported by the lexer, though only `e` is passed through to the parser - not sure whether this is preferable or not to have the uppercase `E` preserved.

This change also includes a fix for Float literals with a `+` prefix, such as `+3.14` which was breaking when given to `read` in the parser.


pchiusano:
Wow thanks! Could you add yourself to CONTRIBUTORS and we can merge.

PR #622, opened 2019-07-31
nixify the repository

### LINKED ISSUE #592: Open to Nix?
Is Unison open to supporting a Nix build? I'd be open to trying to support it if the team is open to a pull request.

tscholak:
Hi,
I've taken the liberty to write Nix derivations for the two yak projects and the unison-parser-typechecker. This was tested on Linux and MacOS and closes #592.

Here's how you use this:

Evaluating Nix expressions requires the Nix build system. Unless you are on [NixOS](https://nixos.org) or you have the Nix package management system installed on your machine ([available for both Linux and MacOS](https://nixos.org/nix/download.html)), you want to use `docker`.

First, make sure you have recent versions of `docker` installed.

From the root of this repository, start an interactive shell session with the latest `nix` image:

    $ docker run -it \
        --name nixos-unison \
        -v `pwd`:/unison \
        nixos/nix:latest

From inside that shell, run

    # nix-build --attr unison-parser-typechecker /unison/default.nix
    
The command will kick off a series of build steps for the unison-parser-typechecker derivation.

When run for the first time, this command can take an hour or so to execute,
and the container will grow to several GBs in size.
Most time is spent downloading and building dependency artifacts.
Not all dependencies have to be built because the Nix project hosts a binary cache.

The last line of the successful `nix-build` output will be the path to the built derivation.

You can also install the unison-parser-typechecker using

    # nix-env -iA unison-parser-typechecker -f /unison/default.nix

Then you can just call `tests` to run the tests ;)
(If the tests fail, call them with `LC_ALL=C.UTF-8 tests` instead.)

You can also run the `unison` executable:
```
# unison
  ☝️

  No codebase exists here so I'm initializing one in: .unison/v1

   _____     _
  |  |  |___|_|___ ___ ___
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|

  Welcome to Unison!

  I'm currently watching for changes to .u files under ~

  Type help to get help. 😎

.>
```
(If that fails with `unison: <stdout>: commitBuffer: invalid argument (invalid character)`, call `unison` with `LC_ALL=C.UTF-8 unison` instead.)

If you want to hack on the unison-parser-typechecker, you may want to use the development environment provided by `nix-shell`. First, install `cabal` and `wget` with

    # nix-env -i wget cabal-install

Then `cd` into the `parser-typechecker` directory and run `nix-shell`,

    # cd /unison/parser-typechecker
    # nix-shell
    [nix-shell:/unison/parser-typechecker]# cabal v1-configure
    [nix-shell:/unison/parser-typechecker]# cabal v1-repl lib:unison-parser-typechecker
    ...
    Ok, 117 modules loaded.
    *Unison.ABT>

If you want to set up a nixified development environment for working on the unison-parser-typechecker, I recommend using `vscode` and `all-hies`, https://github.com/Infinisil/all-hies.
That gives you many IDE features you may want. Ping me if you want to know more.


tscholak:
I've just added a docker image derivation. Run

    $ nix-build --attr unison-parser-typechecker-image /unison/release.nix

aryairani:
@tscholak Thanks for putting this together!  I want to try to minimize the diff, and was hoping you could walk me through some of the changes:

* `parser-typechecker/.hie-wrapper.sh` Is this actually needed for the nix build, or is it an unrelated convenience (for hacking on Unison haskell code using HIE)
* parser-typechecker/unison-parser-typechecker.cabal Do we need to bump the cabal version & add `default-language` settings, or is that an unrelated improvement?  (Also, is there a way to set the default-language for all 4 build targets at once?
* Can all the `.nix` files go in the `nix/` directory, or do they sort of need to be where they are?
* What are `shell.nix` files?
* Given that none of the core maintainers are using nix, I'm concerned about it getting out of date, e.g. if we need to bump the version for one of our build dependencies.  Can you recommend a good solution for that?  A continuous integration that emails you once the nix build stops working? :)

tscholak:
Hi @aryairani!

Ok, let’s minimize the diff:

* `parser-typechecker/.hie-wrapper.sh` is indeed not needed for `nix-build` or `nix-shell`. Rather, as you suspected, it facilitates development using `hie`, for instance, in vscode. I included it only because I wanted to make it convenient for people to try this out. It’s great!
* `parser-typechecker/unison-parser-typechecker.cabal`: bumping the `cabal` version was indeed necessary for nix. I’m using `cabal2nix` to process the `cabal` file, and that gave me an error when I ran it with your current `cabal` file. Bumping the version made it necessary to add the `default-language` setting, though. That was an unintended side effect. With regards to adding a default section, I’m not sure, but this would yet again require a version bump. I think `default` blocks are only supported since version `2.1` or so.
* I tried to consolidate everything in the `nix` directory. All the logic is in the `nix/shared.nix` file while the remaining `nix/*.nix` files contain override derivations for various haskell packages. `nix` files in other locations are for convenience. The `nix-build` and `nix-shell` commands read the `default.nix` and `shell.nix`, respectively, in the current directory by default. Without them, one would need to specify exactly what expression should be run. Possible, but awkward.
* `shell.nix` files are used by the `nix-shell` command. That command lets you enter a shell environment in which all dependencies are present and transparently so. Only then `cabal v1-build` will find the required packages. Try it!
* Yes, good idea! I will need some time to prototype this with your CI.

bbarker:
When trying to use this in my nix environment, I get e.g.:

```
$ nix-env -iA unison-parser-typechecker -f default.nix
installing 'unison-parser-typechecker-0.1'
building '/nix/store/5bggf2hv6sabb2i02jh9aavg0s36zcbp-cabal2nix-haskeline.drv'...
installing
unpacking...
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
error: program 'tar' failed with exit code 2
error: unable to download '/nix/store/2bwn8jzw4m65ygilfksymbgaj1mf37vv-haskeline': URL using bad/illegal format or missing URL (3)
** need a revision for VCS when the hash is given. skipping.
** need a revision for VCS when the hash is given. skipping.
** need a revision for VCS when the hash is given. skipping.
** need a revision for VCS when the hash is given. skipping.
cabal2nix: user error (Failed to fetch source. Does this source exist? Source {sourceUrl = "/nix/store/2bwn8jzw4m65ygilfksymbgaj1mf37vv-haskeline", sourceRevision = "", sourceHash = Guess "", sourceCabalDir = ""})
builder for '/nix/store/5bggf2hv6sabb2i02jh9aavg0s36zcbp-cabal2nix-haskeline.drv' failed with exit code 1
error: build of '/nix/store/5bggf2hv6sabb2i02jh9aavg0s36zcbp-cabal2nix-haskeline.drv' failed
(use '--show-trace' to show detailed location information)
brandon@brandon-Latitude-7480-DevContainer:~/workspace/unison

```

Have not investigated yet, will post back if I beat you to it.

tscholak:
@bbarker hi, you are missing the contents of the haskeline folder. make sure to initialize the git submodules.

bbarker:
Thanks, the build completed, and now I am wondering if I can use the `nix-build` output (`result`), with `nix-shell` (or alternatively load the resulting environment without having it installed in my default profile). I had this conversation on IRC:

```
<bbarker_home> Is there a way to have a temporary environment create with nix-shell based on the output of a nix-build command? As compared to `nix-build -A libfoo && nix-env -f . -iA libfoo`, which installs the package into one's profile?
<vaibhavsagar> bbarker_home: you can do `nix-shell -E 'with import <nixpkgs> {}; pkgs.mkShell { buildInputs = [ (pkgs.callPackage libfoo/default.nix {}) ]; }'
* ddellacosta has quit (Quit: WeeChat 2.2)
<vaibhavsagar> or alternatively `nix-shell -E 'with import <nixpkgs> {}; pkgs.mkShell { buildInputs = [ (pkgs.callPackage ./. {}).libfoo ]; }'`
<bbarker_home> thanks vaibhavsagar, but with both of those I get: error: attempt to call something which is not a function but a set, at /nix/store/qyr7h072vjyr1yw8bnliihx85fpvbisx-nixpkgs-18.09.2327.37694c8cc0e/nixpkgs/lib/customisation.nix:69:12
* mix has quit (Read error: Connection reset by peer)
<bbarker_home> hmm, maybe it is an issue with the package
* lambda-11235 is now known as de1ebcca-bd71-11
<vaibhavsagar> bbarker_home: adjust to taste, I think something might be up with my `callPackage`
<vaibhavsagar> e.g. `nix-shell -E 'with import <nixpkgs> {}; pkgs.mkShell { buildInputs = [ pkgs.python3 ]; }'` works for me
```

But looks like  other commands like `pkgs.haskellPackages.callPackage` result in different errors.

bbarker:
@pchiusano @aryairani  Just my 2c: I tried it out, and it is working as intended in my Linux container that runs nix. My question above about `nix-shell` was more of a side-curiosity on how to use the output of `nix-build`.

aryairani:
@tscholak Thanks for contributing this patch, but unfortunately we don't feel we can properly support nix in the main repo right now. 

We're all for having as many convenient ways to install Unison as possible however, including Nix, and if you/someone would like to maintain a Nix-ified fork and keep it up to date and cut releases, then that's great and we can add another link to the install guide. 👍 

Side note: If it's your fork, you have the final say in these things, but I'd personally recommend removing the HIE part of the PR and also adding something to `.travis.yml` script to verify that the nix build is always successful.

tscholak:
@aryairani makes sense to me, I didn't feel like there was much interest. Thanks!

aryairani:
(Also sorry that it took us 10 weeks to finally decide.)

tscholak:
@aryairani no problem ;D

bbarker:
I haven't had much time for playing with unison lately, though I'm sure I'll be back to it. When I am, i'll be happy to help keep the branch up-to-date, as I found this quite useful!

aryairani:
@bbarker Yeah, we could also have it as a branch of this repo.

virusdave:
I'd be willing to assist in keeping a nix derivation branch up-to-date as well. @bbarker if you do do this, feel free to include me as a maintainer.  I currently maintain the (rather simple) [derivation for the precompiled binaries](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/unison/default.nix).

matthewess:
I was doing some work on unison with nix-provided ghc + stack, and in case anyone is curious or wants to do it themselves, it's pretty easily doable.

1) Add the following stanza to `stack.yaml`:
```yaml
nix:
  enable: true
  shell-file: stack-shell.nix
```
2) Create a `stack-shell.nix` file with the following contents (adapted from the Idris repo):
```nix
{ ghc }:
with (import <nixpkgs> {});

let
  libs = [
    gmp
    # libffi
    ncurses
    # nodejs
    # perl
    zlib
  ];
  native_libs = lib.optionals stdenv.isDarwin (with darwin.apple_sdk.frameworks; [
    Cocoa
    CoreServices
  ]);

in haskell.lib.buildStackProject {
  inherit ghc;
  nativeBuildInputs = native_libs;
  buildInputs = libs;
  name = "unisonBuildEnv";
  src = if lib.inNixShell then null else ./.;
}
```

ceedubs:
@matthewess thank you for leaving your comment; I was struggling to get an environment working and your suggestion was really helpful. On my Mac I'm still having some issues with Stack's built-in Nix support, but I found that the following could at least set me up to drop into a `nix-shell --pure` with all of the necessary tools to be able to run `stack build && stack exec tests && stack exec unison`:

```nix
let
  nixpkgsRev = "5adf2a6";
in
{ pkgs ? import (builtins.fetchTarball "https://github.com/nixos/nixpkgs/archive/${nixpkgsRev}.tar.gz") {} }:

with pkgs;

let
  libs = [ git less ncurses5 zlib gmp ];

  native_libs =
    if (stdenv.isDarwin) then (with darwin.apple_sdk.frameworks; [Cocoa CoreServices])
    else [autoPatchelfHook];

in haskell.lib.buildStackProject {
  inherit ghc;
  nativeBuildInputs = native_libs;
  buildInputs = libs;
  name = "unisonBuildEnv";
  src = if lib.inNixShell then null else ./.;
}
```

I had to add `git` and `less` dependencies to get the tests running in a pure environment.

PR #666, opened 2019-08-05
ABT.Var: minimal typeclass for variables in ABT

TomasMikula:
Introduce a typeclass `ABT.Var`:

```haskell
class Ord v => Var v where
  freshIn :: Set v -> v -> v
```

which is sufficient for the purposes of ABT.

The `Var` class from `Unison.Var` used so far contains many Unison specifics irrelevant for the ABT.

Considering that in principle ABT could be moved to a separate library, this refactoring makes it easier to do so.

### DIFF
--- a/parser-typechecker/src/Unison/ABT.hs
+++ b/parser-typechecker/src/Unison/ABT.hs
@@ -18,20 +18,16 @@ import Data.List hiding (cycle)
 import Data.Map (Map)
 import Data.Maybe
 import Data.Set (Set)
-import Data.Text (Text)
 import Data.Traversable
 import Data.Vector ((!))
 import Prelude hiding (abs,cycle)
 import Prelude.Extras (Eq1(..), Show1(..), Ord1(..))
 import Unison.Hashable (Accumulate,Hashable1,hash1)
-import Unison.Var (Var)
 import qualified Data.Foldable as Foldable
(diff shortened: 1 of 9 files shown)


pchiusano:
@TomasMikula nice observation!

Before merging this, I'd like to alias `ABT.freshIn` to `Var.freshIn` as well as any other definitions that are now `ABT.Var`-generic, both for convenience and to minimize diff. The resulting PR should just have changes in the file `ABT.hs`, `Var.hs`, and `Symbol.hs` files.

TomasMikula:
I aliased `ABT.freshIn` as `Var.freshIn`, which reduced the diff slightly, but there are changes in more files than just `ABT.hs`, `Var.hs`, and `Symbol.hs`. The additional changes can be attributed to one of:
 - `ABT.v'` used to be an alias for `Var.named`. Now that `ABT.v'` is gone, it has to be changed to `Var.named`.
 - Where there previously was `instance Var foo`, there are now both `instance ABT.Var foo` and `instance Var foo`.
 - Some things were generalized from `Var`-generic to `ABT.Var`-generic and the relaxation of the typeclass constraint results in some diff.
 - I didn't make `ABT.Var` extend `Show` (should I?), so there is an extra requirement for `Show` in some `ABT.Var`-generic places.

PR #675, opened 2019-08-07
Unison.TermPrinter: use unix line endings.

zenhack:
For some reason this file had MS-DOS line endings, unlike everything
else.

I noticed this because my editor "fixed" them automatically when working
on something else; probably better to do that as its own patch.


pchiusano:
@atacratic if you don’t have any outstanding changes to this file that would generate conflicts, can you go ahead and merge this?

atacratic:
@pchiusano yeah, I don't, thanks for checking - will merge

@zenhack thanks :-)  I should really fix my git/sublime config so it doesn't mangle this...

PR #676, opened 2019-08-07
Add Int.mod to builtinsSrc

mitchellwrosen:
Now it appears in `find.all Int` :+1:


runarorama:
You'll also need a runtime representation of `mod`. Check `Rt1.hs`.

mitchellwrosen:
@runarorama I'll have to check when I'm back on my laptop, but the impression I got was much of the functionality was already implemented, it just wasn't appearing in `find`. It works when used in watch expressions already, for example.

runarorama:
You're totally right, it's already implemented in the runtime.

runarorama:
@mitchellwrosen add yourself to https://github.com/unisonweb/unison/blob/master/CONTRIBUTORS.markdown when you get a chance.

PR #678, opened 2019-08-07
Add a builtin in 'Char' type.

zenhack:
Note that this is stacked on top of #675, so the first commit is noise; sorry about that.

Per the commit message, this adds character literals and a new built in character type. Not very useful on its own, since there are no related built in functions, but that comes next.

@pchiusano, the issue I was hitting earlier turned out to be a deserialization failure, apparently the error reporting can't tell the difference between "couldn't find that hashed object" and "couldn't deserialize it into a value." In any case, it's fixed.


zenhack:
Hm, odd the tests failed on travis and not my local machine. It looks like the new char syntax conflicts with some of the existing tests, e.g. `'('bar)` now lexes as `'(',  bar, )` as opposed to `', (, ', bar, )`.

I don't feel strongly about what character syntax should be; open to alternatives. The failing example can still be written though, as `'( 'bar)`, so maybe this isn't a total deal breaker.

runarorama:
Syntax suggestion: use LISP's syntax for characters, which is `#\` followed by the character. So a literal `)` is `#\)`.

pchiusano:
Hash char starts a hash reference now but since \ not a valid hash char this would work out ok. I kinda don’t want to overload the hash symbol any more than I want to overload ‘.

I like idea of it just being one or two leading symbols, no need for a closing quote or anything. Any ideas? Don’t need to confine ourselves to ascii necessarily:

𝕔a 
𝕔\n

I’d also be okay with merging this after fixing that one test to put a space in. We can bikeshed this more later and change the syntax without breaking anyone. (Which is cool!) like we can consider the syntax for char literals provisional and subject to change and not feel like we gotta come up with the best syntax for it here and now.

pchiusano:
Also @zenhack could you add yourself to CONTRIBUTORS.markdown? :)

zenhack:
I decided to steal `$` as a prefix from Erlang.

I'm fairly certain my local `stack test` is not actually running the tests, but I'm not sure why. I usually just work with `cabal v2-*`, so I'm not super familiar with stack. But you've got a crazy old dependency (fsutils) that `v2-*` doesn't like, so couldn't get that working easily; see:

https://github.com/Raynes/fsutils/pull/3

parser-typechecker/src/Unison/Hashable.hs
```diff
@@ -14,6 +14,7 @@ data Token h
   | Bytes !ByteString
   | Int !Int64
   | Text !Text
+  | Char !Char
```
pchiusano: Let's not introduce a new token type. Hashable instance for `Char` can produce `Int` token from the codepoint of the `Char`.

parser-typechecker/tests/Unison/Test/TermParser.hs
```diff
@@ -52,6 +52,8 @@ test1 = scope "termparser" . tests . map parses $
   , "\"forty two\""
   , "[1,2,3]"
   , "\"abc\""
+  , "$x"
+  , "$\\n"
```
pchiusano: What does the `$\\n` parse as? Do you need both `\\`? Why isn't it just `$\n` ?

Is the `$` character written `$$`?

__Edit:__ disregard, I get it. `$\` is the literal character `\`. `$\\n` is the newline character. You need some way of distinguishing the two.

parser-typechecker/src/Unison/Codebase/Serialization/V1.hs
```diff
@@ -479,7 +482,8 @@ getPattern getA = getWord8 >>= \tag -> case tag of
       <*> getPattern getA
       <*> getSeqOp
       <*> getPattern getA
-  12 -> Pattern.TextP <$> getA <*> getText 
+  12 -> Pattern.TextP <$> getA <*> getText
+  13 -> Pattern.CharP <$> getA <*> getChar
```
pchiusano: Good. I'm sure you figured this out, but this is the right move, any new constructors get the next largest available tag word, so that existing serialized ASTs still deserialize fine. We never reassign or reshuffle the tags unless we are prepared to do a codebase format upgrade, which we will try to do as infrequently as possible.

pchiusano:
@zenhack currently `$` is allowed as an operator. I'd be fine with removing that from `Lexer.symbolyIdChars` for now. Can you do that as part of this PR and address my other review comment. Then I think we are good to merge!

FYI: `stack build && stack exec tests` builds and runs the tests. See development.markdown file which has some more common commands... if anything in there out of date let us know or submit PR. :)

runarorama:
Thinking about it, I'm pretty opposed to stealing `$` from identifiers just for character literals which hardly ever get used.

parser-typechecker/tests/Unison/Test/TermParser.hs
zenhack: The extra backslash here is escaping in the haskell string; the literal for a newline in unison would be `$\n`. A literal backslash would be `$\\`, since `\` itself needs to be escaped. `$` doesn't, so yes the literal character would be `$$`.

zenhack:
@runarorama, how about the `?` instead? (some precedent from elixir).

parser-typechecker/src/Unison/Hashable.hs
zenhack: Done.

zenhack:
One oddity with the prefix syntax for chars (whichever prefix we choose) is that spaces won't scan well; perhaps we should add an (optional) escape sequence for that as well...

runarorama:
Elisp uses the `?` prefix for charcter literals. They have a `\s` escape sequence for literal space.

I dislike the `?` for the same reason as `$`: it's a valid identifier and we'd have to decide that it isn't anymore. The `#\` sequence piggybacks on `#` which is already used for certain literals (hash literals), but paired with a character (`\`) that disambiguates.

zenhack:
Looking at the lexer source and experimenting in a scratch file, it doesn't appear to me that `?` is currently valid as part of an identifier, either symboly or wordy; are you sure?

I like `\s` as a space escape; I'll go ahead and add that.

I'm not picky about what we use as a prefix, though I kinda dislike `#\` because we're using backslashes for escapes, and I think `#\\n`, `#\\t` etc. scan poorly, esp. needing three backslashes for a backslash literal. But whatever you guys agree on I'll implement.

runarorama:
Oops, no, `?` is not a valid identifier character. I'll have to change the docs to reflect that. :)

zenhack:
Added the `\s` escape, and swapped in `?` as the prefix (since the objection was based on a misunderstanding, I assume we're good now?)

It probably makes sense to squash merge this.

zenhack:
Hm, looks like the `?` does already have a role somewhere? It's part of a list of chars called "delimiters". The last commit fixes the tests, but I'm not sure what the full implications of this are.

pchiusano:
@runarorama I’m okay with merging as is and we can consider the syntax provisional and subject to more bikeshedding if/when we are up for that. 😀 If you’re on board with this, can you do the merge?

parser-typechecker/src/Unison/Runtime/IR.hs
```diff
@@ -1027,7 +1033,8 @@ constructorId v = case v of
   Requested _ -> 10
   Ref{} -> 11
   Cont _ -> 12
-  UninitializedLetRecSlot{} -> 13
+  C _ -> 13
+  UninitializedLetRecSlot{} -> 14
```
aryairani: While it might be okay in this case, in general, we don't wanna _change_ these numbers. (13 -> 14)

pchiusano: I missed this one. In this case it’s fine since we aren’t persisting runtime values yet anywhere but yeah in general don’t change any of the tag values.

zenhack:
Quoting Paul Chiusano (2019-08-08 00:19:36)

> I missed this one. In this case it's fine since we aren't persisting
> runtime values yet anywhere but yeah in general don't change any of
> the tag values.

Oops. Yeah, I figured out the pattern, don't remember doing this one on
purpose, sorry about that.

aryairani:
@zenhack If you had time to switch it back to `'a'` syntax, that would be super. 😃  We'll just require a space between quoted identifiers: `'a 'b` would mean `(() -> a) (() -> b)`, and `'a'b` would be a type error, since `'a'` is not a function. `abra'c'adabra'

zenhack:
Sounds good to me; I'll find time in the next couple days.

Quoting Arya Irani (2019-08-13 12:50:44)

aryairani:
> Syntax suggestion: use LISP's syntax for character literals, which is `#\` followed by the character. So a literal `)` is `#\)`. A literal newline can be `#\\n`.
> 
> When you think about it, the `'x'` syntax is pretty weird, as it has an end quote even though it's totally unambiguous where the character ends.

Looks like [MIT scheme](https://www.gnu.org/software/guile/manual/html_node/Characters.html) just uses `#\` followed by the character, or the name for nonprintable characters: `#\a` for `a`, `#\\` for `\`, `#\newline` for newline (or `#\nl` in Guile).  `#\ ` or `#\space` for ` `!

PR #687, opened 2019-08-08
New builtins for converting Char <-> Nat.

zenhack:
(where the Nat is the corresponding code point).


zenhack:
I wrote a couple more builtins. Since this PR still hasn't been looked at, I decided to just tack them on here rather than opening a new pr.

pchiusano:
Looks good.

PR #705, opened 2019-08-11
Add Unison.Prelude

mitchellwrosen:
This patch adds a `Unison.Prelude` module with some commonly-imported symbols from `base` and friends. 

Possible next step: re-export bits of `Prelude` from it, and add `NoImplictPrelude` to the `.cabal` file. But for now, it is just a layer on top of `Prelude`.


mitchellwrosen:
I think the biggest quality of life improvement will be always having access to `Debug.Trace`. Soo annoying to import ;)

PR #725, opened 2019-08-13
use template haskell to embed git version during compilation

aryairani:
image

In `release/M2-base-4-g6c53a5c4'`
* `release/M2-base` is the most recent parent tag.
* `4` is the number of commits since the above tag (shown if # new commits > 0)
* `g6c53a5c4` is the commit id (shown if # new commits > 0)
* `'` means there are untracked changes in the build dir.


pchiusano:
When is this `IO` action evaluated? Do I have to do a clean to get an updated version number?

runarorama:
wizardry!

aryairani:
Well, I "upgraded" from d437e27 to e26f1d0 to try to engage cabal's capability for adding the git status into the build dependency calculation, to avoid needing to do a `stack clean` to get the latest revision info.

But, I think `stack` has its own dependency checks, that short-circuit cabal's without honoring them.  Tried upgrading `stack`, but with new stack, my old "stack transplant" doesn't match, so it wants to rebuild the `hfsevents` package but fails because of [this stuff](https://gist.github.com/aryairani/1d6bb6cffb194ea2efdeed6e55fedceb).

Do I install Mojave? 🤔

mitchellwrosen:
@aryairani Have you observed a changed git HEAD not triggering a rebuild? It looks to me like `gitrev` does go through some trouble to call `addDependentFile` on all the right stuff: https://github.com/acfoltzer/gitrev/blob/450a7abf444260e95791649489705e54650c0232/src/Development/GitRev.hs#L61

aryairani:
@mitchellwrosen Admittedly, I am a TemplateHaskell noob, but yeah I am observing build issues, even though the `addDependentFile` stuff looks good to me.  Not sure why it isn't rebuilding as expected.

```
~/unison$ stack build
~/unison$ git rm foo
rm 'foo'
~/unison$ git commit -m"garbage4"
[garbage 0f14d687] garbage4
 1 file changed, 0 insertions(+), 0 deletions(-)
 delete mode 100644 foo
~/unison$ stack build && stack exec unison
Version: release/M2-base-6-ge26f1d0c'
   _____     _             
  |  |  |___|_|___ ___ ___ 
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|
<snip>
~/unison$ git describe --tags --always
release/M2-base-10-g0f14d687
```
😞 

unison still reporting `M2-base-6` instead of `M2-base-10`.

aryairani:
Going to merge the first, simple implementation (d437e27746b526b6cb0b07c47d3cde2e124e8766) as #735 instead for now, and track the remaining work in #736.

Issue #656, opened 2019-08-04
Add `.base.Int.toFloat` and `.base.Int.toText`


runarorama:
And whatever other conversions are missing between the builtins.

PR #728, opened 2019-08-13
Added Nat.fromText, Nat.toText, Int.fromText, Nat.toText, Int.toFloat

runarorama:


parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -151,16 +151,16 @@ pretty0
   -> Pretty SyntaxText
 pretty0 n AmbientContext { precedence = p, blockContext = bc, infixContext = ic, imports = im} term
```
pchiusano: @runarorama are any of these changes relevant to this PR? If no substantive changes in this file, let’s remove it from the PR.

I think a good rule of thumb for doing a linting pass is: do it as a totally separate PR with nothing else in it. Or do it only for code you are already editing for some other reason.

aryairani: Maybe an automated linter on commit (that doesn’t break the code) is the solution.

pchiusano: Maybe... I guess I often find myself disagreeing with the linter so I don’t have high hopes for that being a slam dunk.

There’s a lot to be said for being accepting of minor differences in coding style and not trying to canonicalize every minor decision about that. Like here in this diff, it is pretty arbitrary how you carve up some function application using $, ., ()’s. I think whatever is fine and I’d just leave the code as it is written.

runarorama:
OK, but is this the correct way of adding new builtins?

aryairani:
Closes #656

PR #730, opened 2019-08-14
Text.uncons/unsnoc: use tuples instead of "Pair"

zenhack:
I'd misunderstood the role of `Pair` when I added these in the first place.

It would be good to have a better story re: migrations of builtins, but I found that just updating the definitions that used it (only one, actually) worked ok for unison-parsers.

### DIFF
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -320,8 +320,8 @@ builtinsSrc =
   , B "Text.>=" $ text --> text --> boolean
   , B "Text.<" $ text --> text --> boolean
   , B "Text.>" $ text --> text --> boolean
-  , B "Text.uncons" $ text --> optional (pair char text)
-  , B "Text.unsnoc" $ text --> optional (pair text char)
+  , B "Text.uncons" $ text --> optional (tuple [char, text])
+  , B "Text.unsnoc" $ text --> optional (tuple [text, char])
 
   , B "Char.toNat" $ char --> nat
   , B "Char.fromNat" $ nat --> char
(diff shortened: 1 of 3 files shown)


aryairani:
In general, I’d say you’re not allowed to change anything about any existing builtin, even to fix bugs.  You’d introduce a new one, and possibly adjust names using the `Rename` constructor, possibly deprecate old.  If there was an automated way to enforce this during build or test, that would be ideal.

Since master is already a little bit busted, maybe it’s okay this time.
—-
As a meta comment: I feel like anything in `master` should be considered a release in the sense that someone could reasonably expect to write code with it without breaking their codebase; stable.  If we want to do experimental changes, maybe we use a different branch, and merge to master once it’s a build we have confidence will interop appropriately with the rest of the ecosystem.

As a second meta comment, we should probably require tests for each new language feature; I think that would have caught this in time for the initial merge.

zenhack:
Yeah, I think I agree with @aryairani re: general policy, but given that (1) these haven't hit a release (even an alpha) (2) They've been there only a couple days and (3) the set of users is still *tiny*, it seems like it would be a real shame to be stuck with this.

aryairani:
I'm gonna merge this for now but, for the future, we wouldn't be stuck with anything; we'd just have something like:

```haskell
, B "Text.uncons" $ text --> optional (pair char text)
, B "Text.unsnoc" $ text --> optional (pair text char)
, D "Text.uncons"
, D "Text.unsnoc"
, B "Text.uncons2" $ text --> optional (tuple [char, text])
, B "Text.unsnoc2" $ text --> optional (tuple [text, char])
, Rename "Text.uncons2" "Text.uncons"
, Rename "Text.unsnoc2" "Text.unsnoc"
```

We will definitely have occasion to do this in the future. This time we're going along with the dubious argument that nobody's really gotten this yet — reminiscent of "`git rebase` is okay if I'm really fast", which is probably true except if it's not. :)

zenhack:
Yeah, I think the longer term thing needs to be more scrutiny for anything we can't just change later.

PR #732, opened 2019-08-14
`history` command (alias: `log`) for viewing history of a namespace

pchiusano:
You can now ask for the history of any namespace. Supplying no args gives you the history of the root namespace. It stops when hitting a merge node:

Screen Shot 2019-08-14 at 1 27 29 PM

As the message mentions, you can supply a (complete) hash to view the history of any namespace that isn't named in your current root. This can be used for paging (to see further into the past) or for exploring one of the branches of a merge node - just `history #somehash` to see the history starting from that hash.

Diffs are shown between each hash:

Screen Shot 2019-08-14 at 1 36 15 PM

Currently it shows 10 results at a time, and diff excerpts, but this isn't hardcoded in the command, just in the input pattern, so we could expose a different input pattern or flags to control how far back to look with a few LOC change in `InputPatterns.hs` (haven't done that for this PR).

Screen Shot 2019-08-14 at 1 38 09 PM

Also important is that the `fork` command can now take hashes to identify the source namespace, so you can now use `history` to browse the history, and `fork` to resurrect old versions of your codebase. Here, I forked a version of my codebase before adding the `.stream` namespace, and confirm that the `.stream` namespace is empty in the forked namespace:

Screen Shot 2019-08-14 at 1 45 38 PM

One last thing: d0eff62c adds a fix to `FileCodebase.getBranchForHash`. It was crashing if the provided hash isn't in the codebase. There are cases where you do want to crash if the hash isn't found (like if you are traversing the history of an existing `Branch`, or loading the root), but for loading a user-provided hash like what is done for the `history` command, it's okay to not bomb and just report something to the user ("I don't know about a namespace with that hash"). So that's what's done now.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
@@ -19,6 +19,7 @@ import           Unison.Codebase.Editor.Output
 import           Unison.Codebase.Editor.RemoteRepo
 
 import           Unison.Codebase.Branch         ( Branch )
+import qualified Unison.Codebase.Branch        as Branch
 import           Unison.Codebase.GitError
 import           Unison.Names3                  ( Names, Names0 )
 import           Unison.Parser                  ( Ann )
@@ -108,6 +109,9 @@ data Command m i v a where
   -- Any definitions in the head of the requested root that aren't in the local
   -- codebase are copied there.
   LoadLocalRootBranch :: Command m i v (Branch m)
+
+  -- Like `LoadLocalRootBranch`.
+  LoadLocalBranch :: Branch.Hash -> Command m i v (Branch m) 
 
   LoadRemoteRootBranch ::
     RemoteRepo -> Command m i v (Either GitError (Branch m))
--- a/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
@@ -77,9 +77,9 @@ data Input
     | RemoveTermReplacementI PatchPath Reference Reference
     | RemoveTypeReplacementI PatchPath Reference Reference
   | UndoI
-  -- Maybe (Int,Int) is a start and end index into the results,
-  -- can be used for paging
-  | LogI (Maybe (Int,Int)) (Either Path' Branch.Hash)
+  -- First `Maybe Int` is cap on number of results, if any
+  -- Second `Maybe Int` is cap on diff elements shown, if any
+  | LogI (Maybe Int) (Maybe Int) (Either Path' Branch.Hash)
   -- execute an IO object with arguments
   | ExecuteI String
   | TestI Bool Bool -- TestI showSuccesses showFailures
--- a/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
@@ -28,7 +28,7 @@ data Input
     -- directory ops
     -- `Link` must describe a repo and a source path within that repo.
     -- clone w/o merge, error if would clobber
-    = ForkLocalBranchI Path' Path'
+    = ForkLocalBranchI (Either Branch.Hash Path') Path'
     -- merge first causal into destination
     | MergeLocalBranchI Path' Path'
     | PreviewMergeLocalBranchI Path' Path'
--- a/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
@@ -3,6 +3,7 @@ module Unison.Codebase.Editor.Input
   , Event(..)
   , OutputLocation(..)
   , PatchPath
+  , BranchId, parseBranchId
   ) where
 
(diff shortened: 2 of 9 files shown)


parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -109,6 +110,9 @@ data Command m i v a where
   -- codebase are copied there.
   LoadLocalRootBranch :: Command m i v (Branch m)
 
+  -- Like `LoadLocalRootBranch`.
```
aryairani: Should we rename/modify `LoadLocalRootBranch` to return a `Branch.Hash` instead of a `Branch m`? and then make a helper somewhere that's `LoadLocalRootBranch >>= LoadLocalBranch`

parser-typechecker/src/Unison/Codebase/Editor/Input.hs
```diff
@@ -77,6 +77,9 @@ data Input
     | RemoveTermReplacementI PatchPath Reference Reference
     | RemoveTypeReplacementI PatchPath Reference Reference
   | UndoI
+  -- First `Maybe Int` is cap on number of results, if any
+  -- Second `Maybe Int` is cap on diff elements shown, if any
+  | LogI (Maybe Int) (Maybe Int) (Either Path' Branch.Hash)
```
aryairani: Minor inconsistency, `ForkLocalBranchI` takes a hash on the left and a path on the right; while `LogI` is the opposite.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -411,15 +413,42 @@ renameBranch = InputPattern "move.namespace"
       _ -> Left (I.help renameBranch)
     )
 
+history :: InputPattern
+history = InputPattern "log"
```
aryairani: Recommend making `"history"` be the primary command name, and maybe even eliminate the alias `"log"`.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -279,7 +285,7 @@ getRootBranch root = do
  where
   go single = case hashFromString single of
     Nothing -> failWith $ CantParseBranchHead single
-    Just h  -> branchFromFiles root (RawHash h)
+    Just h  -> branchFromFiles True root (RawHash h)
```
aryairani: Recommend a self-documenting data type rather than `Bool`

parser-typechecker/src/Unison/Codebase/Editor/Command.hs
aryairani: Actually seems like `LoadLocalRootBranch` is not used anywhere?  Unrelated to this PR, but surprising.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -506,6 +508,40 @@ notifyUser dir o = case o of
...
+        "⊙ " <> phash h,
+        ""
+        ]
+    dots = "⠇"
+    go hash diff = P.lines [
+      P.indentN 2 $ prettyDiff cap diff,
+      "",
+      "⊙ " <> phash hash
+      ]
+    ex = "Use" <> IP.makeExample IP.history ["#som3n8m3space"]
```
aryairani: Should be `#som3n4m3space` (`8` -> `4`)

pchiusano:
@aryairani thanks! Your suggestions all sound good to me. I'll take a pass through when I have a minute.

parser-typechecker/src/Unison/Codebase/Editor/Command.hs
pchiusano: This is the only one I didn't address. I guess I wasn't really convinced it mattered one way or the other.

parser-typechecker/src/Unison/Codebase/Editor/Input.hs
pchiusano: Fixed this, added an alias `Input.BranchId` for that either type, and a common parsing function `Input.parseBranchId` which parses it, and which is used in both input patterns that use this type (currently `fork` and `history`)

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
pchiusano: This was a good idea and is done, also added some verbiage on why you would do one mode vs the other.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
pchiusano: Did this one too.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
pchiusano: Did this. I killed `log` as an alias and renamed some types that started with `Log` to start with `History` instead.

pchiusano:
@aryairani merging this, went through those cleanup items, see my comments. Thanks for the review!

PR #746, opened 2019-08-20
Fix #745

### LINKED ISSUE #745: Code wrongly accepted, when using multiple args with abilities
In the following code, `orderServer1` should fail to type check.  But it doesn't.

```haskell
use .base
use .base.io

-- bad code, wrongly accepted... the log call should be an ability check failure 
-- since it is inside the body of an arrow that only provides IO
orderServer1 : ServerConfig ->{Log} '{IO} ()
orderServer1 sc unit = 
  log (ServerConfig.toText sc)
  startServer sc

-- good code (correctly accepted) - each arrow gets its own lambda with 
-- access to the appropriate ability
orderServer2 : ServerConfig ->{Log} '{IO} ()
orderServer2 sc = 
  log (ServerConfig.toText sc)
  '(startServer sc)

-- boring stuff follows
startServer : ServerConfig ->{IO} ()
startServer sc = printLine "started, honest!"
type ServerConfig =
ServerConfig.toText : ServerConfig -> Text
ServerConfig.toText sc = ""
ability Log where
  log : .base.Text -> ()
```

From @pchiusano via slack [here](https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1566135939298700): "So it seems like bug is we are “pushing” onto ambient abilities when we should be “replacing” ambient abilities. I suspect it could be a very small fix in checking / synthesis of lambdas."

pchiusano:
Fixes #745 

Simplified `handle` block synthesis / checking. Previously it had some weird / wrong stuff going on. Now I would call it straightforward. This fix might also clear up some other issues with ability checking.


pchiusano:
Here's screenshot of new error message you get if the handler is of the wrong type:

Screen Shot 2019-08-20 at 11 41 56 AM

pchiusano:
Note on the message: the names for `Effect`/`Request` are chosen from the pretty-print environment, that's not a hard-coded string there.

aryairani:
> Note on the message: the names for `Effect`/`Request` are chosen from the pretty-print environment, that's not a hard-coded string there.

Thanks for clarifying.  Should we open a ticket to update the initial name?

PR #750, opened 2019-08-21
Remove some dead code in PrintError.

TomasMikula:
There were branches testing for

```haskell
Var.typeOf unknownTermV == Var.AskInfo
```
or
```haskell
Var.typeOf unknownTermV == Var.MissingResult
```
where `unknownTermV` was only ever instantiated to `Var.named name`, thus its type can never be `AksInfo` or `MissingResult`.

To prove the point, I even changed the type of `UnknownTerm.unknownTermV` from `v` to `Text`, without any loss of information. I can undo this type change if desired.


aryairani:
Hey @TomasMikula we wanna merge this, but could you change `UnknownTerm.unknownTermV` back to `v` first.

TomasMikula:
@aryairani Done.

aryairani:
@TomasMikula Thanks!

Issue #714, opened 2019-08-12
Implement 3-way merge for `Branch`


pchiusano:
(look back 1 step at a time on both sides to find a match, only 2x worse than optimal) - impact of not having this would be that if there are concurrent edits and deletes to the same package, the deletes will get undeleted! (Doesn't require changing repo format, so easy to alter merge strategy later)

PR #751, opened 2019-08-21
Runarorama/714 implement 3 way merge for branch

runarorama:
Implements #714

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Metadata.hs
+++ b/parser-typechecker/src/Unison/Codebase/Metadata.hs
@@ -15,6 +15,10 @@ type Value = Reference
 -- keys can be terms or types
 type Metadata = Map Type (Set Value)
 
+-- `a` is generally the type of references or hashes
+-- `n` is generally the the type of name associated with the references
+-- `Type` is the type of metadata. Duplicate info to speed up certain queries.
+-- `(Type, Value)` is the metadata value itself along with its type.
 type Star a n = Star3 a n Type (Type, Value)
 
 inserts :: (Ord a, Ord n) => [(a, Type, Value)] -> Star a n -> Star a n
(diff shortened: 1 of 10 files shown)


parser-typechecker/src/Unison/Codebase/Metadata.hs
```diff
@@ -15,6 +15,10 @@ type Value = Reference
 -- keys can be terms or types
 type Metadata = Map Type (Set Value)
 
+-- `a` is generally the type of references or hashes
+-- `n` is generally the the type of name associated with the references
+-- `Type` is the type of metadata. Duplicate info to speed up certain queries.
+-- `(Type, Value)` is the metadata value itself along with its type.
```
aryairani: ❤️

parser-typechecker/src/Unison/Util/Star3.hs
```diff
@@ -25,6 +25,20 @@ toList s = [ (f, x, y, z) | f <- Set.toList (fact s)
                           , y <- Set.toList (R.lookupDom f (d2 s))
                           , z <- Set.toList (R.lookupDom f (d3 s)) ]
 
+-- `difference a b` contains only the facts from `a` that are absent from `b`
+-- or differ along any of the dimensions `d1..d3`.
```
aryairani: Ok.

pchiusano:
@runarorama looking good, but could we add some more tests to `Causal`, I think some property-based tests could be good. Here are some ideas:

* `merge x x == x`, should not add a new head, and also the value at the head should be the same of course
* `merge x mempty == x`, `merge mempty x == x`
* `merge (cons hd tl) tl == cons hd tl`, `merge tl (cons hd tl) == cons hd tl`
* `head (merge (cons hd (cons common _)) (cons hd2 (cons common _))) == patch common (diff hd common <> diff hd2 common)`

https://github.com/unisonweb/unison/tree/master/yaks/easytest has info on how to generate random data with EasyTest, lmk if you have any questions about that, but basically, you just use the monad functions and a few random generators that EasyTest provides:

```haskell
reverseTest :: Test ()
reverseTest = scope "list reversal" $ do
  nums <- listsOf [0..100] (int' 0 99)
  nums `forM_` \nums -> expect (reverse (reverse nums) == nums)
```

PR #760, opened 2019-08-23
Simplify Var.isLowercase

TomasMikula:
Characters 𝕒, 𝕖, ... are already considered lowercase by `isLower`.


aryairani:
I don't think we want to remove the check for synthetic vars, but also: a function called `isLowercase` shouldn't have so many back-door conditions to qualify as "lowercase".  @pchiusano could you review (after STL).

TomasMikula:
Here's my reasoning:

 1. **The check has to be name-based.** Therefore, some more sophisticated methods using `Var.typeOf` are out. _Reason:_ `parse . print` should be identity, so whatever the printer decides is lowercase (and elides `forall`s for) while printing, the parser has to recognize as lowercase (and infer `forall`s). Since the parser only has the name available, the decision has to be name-based.
 2. I pointed out that the check for synthetic vars is redundant, because it so happens that currently `Set.member n syntheticVarNames` implies `(all isLower . take 1 . Text.unpack) n`. If we consider this implication to be just a coincidence, then it could make sense to keep the synthetic var check: so that if a change is made to synthetic vars' names such that `(all isLower . take 1 . Text.unpack) n` no longer holds (e.g. some of them will start with `𝔼` or `𝔽`), they would still be considered "lowercase". The problem with this is that in language reference you **would have to document that variables starting with a lowercase letter _or 𝔼 or 𝔽_ are considered universally quantified.** I think this would be a weird thing to do.

aryairani:
@pchiusano Could you take a look at this to decide the appropriate functionality and name for this function?

aryairani:
Closing in favor of #859 — thanks for bringing this to our attention!

TomasMikula:
Oh well, the inappropriate name was not my main concern. This small PR was part of a gradual elimination of unrelated concerns (like implementation details of type inference) from `Var`.

aryairani:
Sorry about that, the inappropriate name definitely created confusion for me.  

So the check needs to be name based for `parse . print` to be identity, and we need a property test or proof to ensure that synthetic vars that should be universally-quantified-if-free print with a form that would be universally-quantified-if-free even as a user var. @pchiusano

aryairani:
We can delete `Var.syntheticVar` or convert it to docs if it isn't needed.

TomasMikula:
> we need a property test or proof to ensure that synthetic vars that should be universally-quantified-if-free print with a form that would be universally-quantified-if-free even as a user var.

As far as I can tell, a _synthetic variable_ is an auxiliary notion for the typechecking phase. From the perspective of a codebase, there is no distinction between synthetic and user variables. In one of my development branches I eliminated the different `Var.Type`s altogether, so it won't even be possible to treat synthetic vars and user vars differently, by design. Therefore, I would not bother writing a test for that now.

aryairani:
>As far as I can tell, a synthetic variable is an auxiliary notion for the typechecking phase. From the perspective of a codebase, there is no distinction between synthetic and user variables.

Yes, I think so too.

> it so happens that currently Set.member n syntheticVarNames implies (all isLower . take 1 . Text.unpack) n. If we consider this implication to be just a coincidence,

Since it's not a coincidence, but also not enforced with types, my thinking was that a test should confirm this implication, to avoid a regression at some point.

TomasMikula:
Oh, so that would be a test to ensure that variable names used for synthetic vars are lowercase.

I was thinking about a test that would ensure that lowercase synthetic variables were treated (in `parse . print`) the same as lowercase user variables. This one I would like to eventually ensure by design.

aryairani:
👍

TomasMikula:
Test added.

aryairani:
Beautiful, thanks for your work on these.

PR #761, opened 2019-08-24
Add Char.isSpace builtin

mitchellwrosen:


zenhack:
Don't think there's any reason for this to be a builtin.

Note that there's an `.ascii` namespace in my `unison-parsers` codebase that you can pull in if you want; it defines `isSpace` as well as several similar functions. I mentioned to @aryairani that it should probably get added to unisonbase at some point.

mitchellwrosen:
I think hunting down all of the Unicode characters in the "space" category would be a bit difficult, that's why I thought it was appropriate to add as a builtin. Or does your `isSpace` already accomplish that?

zenhack:
No, it's just ascii for now. I'm planning on adding a `.unicode` namespace at some point though.

Hunting down categorization info isn't that difficult:

https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt

zenhack:
FWIW I started working on the unicode stuff in that repo. There's most of a parser for that file.

mitchellwrosen:
Ah, cool. Chasing around and trying to understand Haskell's C implementation of https://github.com/ghc/ghc/blob/master/libraries/base/cbits/WCsubst.c#L5072
led me to believe it's harder than it actually is.

zenhack:
Aha, yeah it looks complicated but ultimately all they're really doing is defining a bunch of static tables corresponding to the data in that file, and looking stuff up in the tables.

aryairani:
@mitchellwrosen Hey I'm gonna close this for the time being. (I do think there may be a place for such a builtin, but it would be in the broader context of Unicode library for Unison—which may or may not be a pure Unison library.)  In the meantime, maybe an `isAsciiSpace` Unison function that checks for ` `, `\n`, `\t` might do the trick?

PR #777, opened 2019-09-06
put builtin term types into codebase (fixes #776)

### LINKED ISSUE #776: Write builtin terms' type to `v1/terms/_builtin/<escaped-name>/type.ub` on ucm startup.
Do we want to do this? Why or why not? /cc @mitchellwrosen 

Pros:
* type info for builtins is visible to external tools

Cons:

mitchellwrosen:
This patch adds builtin terms' types to the codebase during initialization

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -45,6 +45,9 @@ data Codebase m v a =
            , getTypeDeclaration :: Reference.Id -> m (Maybe (Decl v a))
 
            , putTerm            :: Reference.Id -> Term v a -> Type v a -> m ()
+             -- Like 'putTerm', but there's no term to put into the codebase,
+             -- just a type.
+           , putBuiltinTerm     :: Text -> Type v a -> m ()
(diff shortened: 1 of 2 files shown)


aryairani:
Looks good.  I'll note that this only happens during codebase initialization and not at ucm startup, and while that's already an improvement, but it won't necessarily populate for new builtins as we add them.

Maybe we just need to reinitialize the codebase on startup?  Or write missing builtins on startup?

mitchellwrosen:
@aryairani Yah, only happens on the very first init. I'm not totally sure how adding new builtins is meant to work. Want to hash that out before merging, or tackle it in a different ticket/PR?

aryairani:
@mitchellwrosen I'm leaning slightly towards hashing it out before merging, but it wouldn't take much to convince me otherwise.

aryairani:
Per private slack conversation, we're gonna put this on the back burner until we have more clarity.

aryairani:
Gonna close this for now (also it no longer merges cleanly 😞), but we can always revisit.

PR #791, opened 2019-09-11
Fix ABT.freshInBoth and ABT.subst'

TomasMikula:
The problem was that in general, it is not guaranteed that after

```
v' = freshIn vs2 . freshIn vs1 $ v
```

`v'` is not a member of `vs1`. Therefore, freshening cannot be done gradually, but has to be done at once.

A test case of `freshInBoth` that would previously fail is included.


pchiusano:
Thanks!

PR #816, opened 2019-09-24
Unison transcript runner

pchiusano:
`ucm` now takes some optional arguments:

Screen Shot 2019-09-24 at 1 44 17 PM

So you can run standalone scripts now via `ucm mymain.u` (looks for a `main` binding) which is nice.

There's a couple example transcript files along with their output in [this directory](https://github.com/unisonweb/unison/tree/topic/transcripts/unison-src/transcripts). [This file](https://github.com/unisonweb/unison/blob/topic/transcripts/unison-src/transcripts/hello.md) is a good overview of the format, and then [this file](https://github.com/unisonweb/unison/blob/topic/transcripts/unison-src/transcripts/merges.md) has a test of some of the merge operations. Be sure to peep the raw markdown of the source so you can see the different kinds of fenced code blocks and flags for each in use.

All the transcripts in there are run when you do `stack exec transcripts` (which will exit with , and Travis will run through them as well.

### How it's done

Implementation is [here](https://github.com/unisonweb/unison/blob/topic/transcripts/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs). Some notes:

* The transcript parser is written with Megaparsec. That part was easy peasy.
* The runner is [in that same file](https://github.com/unisonweb/unison/blob/topic/transcripts/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs#L98). Warning: Implementation is pretty ugly but self contained, and it shares the exact same code path as `ucm` itself and even calls `CommandLine.main`. This means that everything works like you'd expect.
  - [ ] Todo: document the runner better, at least with some comments
* There is probably a way of refactoring `CommandLine.main` to make it more amenable to having a nicer transcript runner, but I didn't have the stomach for it.

### Other changes

* `execute` (now aliased as `run`) works a bit differently - it looks for a `main` function in the latest typechecked scratch file and calls it with any supplied args and giving it access to `IO`. I thought about introducing another command just to keep the old behavior around, but eventually decided not to for now. What you'd formerly put as the argument to `execute` you can just as easily put in your scratch file or in a `main.u` file.
* So, you can run `IO` stuff without first adding it to the codebase, and your `main` function can still reference preexisting definitions in the codebase.
* A bunch of output message formatting suff. The transcript runner needed to get the output messages before they were printed to the console, so it could splice them into the file.
* The transcript runner also needed to know which messages are considered error messages. That function is given [here](https://github.com/unisonweb/unison/blob/topic/transcripts/parser-typechecker/src/Unison/Codebase/Editor/Output.hs#L171). You can see where that's used in the transcript runner [here](https://github.com/unisonweb/unison/blob/topic/transcripts/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs#L189-L202).
* Fixed a bug in EasyTest

I'm going to do a cleanup pass, will lift the draft status once that's done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Unison
 .unison*/
 .unisonHistory
+test-output
 
 # Haskell
 dist
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,3 +24,4 @@ install:
 
 script:
 - stack --no-terminal exec tests
+- stack --no-terminal exec transcripts
--- a/development.markdown
+++ b/development.markdown
@@ -2,16 +2,17 @@ These are commands that will likely be useful during development.
 
 __General:__ `./scripts/test.sh` compiles and builds the Haskell code and runs all tests. Recommended that you run this before pushing any code to a branch that others might be working on.
 
-_Disclaimer_ If you have trouble getting started, please get in touch via [gitter](https://gitter.im/unisonweb/unison) so we can help.  If you have any fixes to the process, please send us a PR!
+_Disclaimer_ If you have trouble getting started, please get in touch via [Slack](https://unisonweb.org/community) so we can help.  If you have any fixes to the process, please send us a PR!
 
 ## Running Unison
 
-To get cracking with Unison,
+To get cracking with Unison:
+
 * [Install `stack`](https://docs.haskellstack.org/en/stable/README/#how-to-install).
-* Build the project with `stack build`.
+* Build the project with `stack build`. This builds all executables.
 * After building, `stack exec unison` will fire up the codebase editor, create a codebase in the current directory, and watch for `.u` file changes.  If you want to run it in a different directory, just add `unison` to your `PATH`, after finding it with `find .stack-work -name unison -type f`.  (For me, this finds two, they both work, but have different contents.  ¯\\\_(ツ)\_/¯ )
-* Once a file is typechecked, you can do `add` to add it to the codebase,
-* and then `view` to view a definition, or `help` for more ideas.
+* `stack exec tests` runs the tests
+* `stack exec transcripts` runs all the integration tests, found in `unison-src/transcripts`. You can add more tests to this directory.
 
 ### What if you want a profiled build?
 
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -2071,7 +2071,7 @@ addRunMain args uf = let
       a = ABT.annotation tm 
       in
       if Typechecker.isSubtype ty (nullaryMain a) then Just $ let 
-        runMain = Term.app a (Term.var a v) (Term.ref a DD.unitRef) 
+        runMain = DD.forceTerm a a (Term.var a v)
         in UF.typecheckedUnisonFile 
              (UF.dataDeclarations' uf)
              (UF.effectDeclarations' uf)
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -727,7 +727,8 @@ loop = do
         do
           eval . Notify
             $ DisplayDefinitions loc ppe loadedDisplayTypes loadedDisplayTerms
-          eval . Notify . SearchTermsNotFound $ fmap fst misses
+          when (not $ null misses) $
+            eval . Notify . SearchTermsNotFound $ fmap fst misses
           -- We set latestFile to be programmatically generated, if we
           -- are viewing these definitions to a file - this will skip the
           -- next update for that file (which will happen immediately)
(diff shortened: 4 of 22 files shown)


francisdb:
So this won't work for after for example pulling in `.foo.bar` from a remote repo?

```
ucm run .foo.bar.main
```

pchiusano:
Right (although that line also doesn’t work now). You could put “main = .foo.bar.main” in a file foo.u, then do ‘ucm foo.u’.

You could also put that same line in your scratch file then type ‘run’ from ucm.

francisdb:
would be nice if this would work


vi `foo`
```
#!/usr/local/bin/ucm

main = .foo.bar.main
```

```
chmod +x foo
./foo
```

https://en.wikipedia.org/wiki/Shebang_(Unix)

But this would require some central repo in the user home folder

aryairani:
> But this would require some central repo in the user home folder

IMO this will ultimately be the recommended practice.

francisdb:
Want me to create a separate ticket for that?

pchiusano:
We (@aryairani @runarorama and myself) did some workshopping of this today:

### Transcripts

* Goal of the transcript stuff is mostly: 1) Make it easy to write docs and tutorials and 2) Make it easy for us to write integration tests of `ucm` (and for users to submit bug reports that are repeatable).
* `ucm mytranscript.md` by default runs in a new, empty codebase (aka default is sandboxed). We didn't see much value in running a transcript on an existing codebase and will just remove that option for now. We also noted that the transcript can start with a `pull` to bring the codebase to some known state.

### Running I/O programs

* Two cases to consider: interactive development, and batch mode execution.
* For interactive dev, for now, going to start with just `.> run blah`, where `blah : '{IO} ()` is looked for in the most recently typechecked scratch file, with a fallback to looking in the codebase.
  * Decided to nix command args. Basically would like to highly discourage people putting fragments of programs into ucm commands.
* For batch mode, no real consensus, but a few options considered:
  * `ucm run somepackage.main` calls `somepackage.main : '{IO} ()`, found in the existing codebase
  * `ucm run mymain -stdin` loads a `.u` file from standard in, then does `run mymain`.
  * `ucm run mymain -file mymain.u` loads `mymain.u` then does `run mymain`.
  * `ucm run <expr>`

We were going to marinate on the batch mode options a bit, I'll address the other items first.

atacratic:
I love the way the transcript stuff is looking.  A few comments on your last post though.

> We didn't see much value in running a transcript on an existing codebase and will just remove that option for now.

It'll be frustrating if you're iterating on a transcript and every time you run it you need to wait for a pull. 

Also what if you're iterating on a transcript and its 'basis' codebase in parallel?  Do you want to push and pull on every iteration.

> Basically would like to highly discourage people putting fragments of programs into ucm commands.

I would like to be able to write something like "eval 2 + 2" in ucm.  Clearly the watch-based interaction is more scalable, but I think it's good enough that it exists; having it *replace* the option of REPL-style interaction is not necessary to get the benefit.  But I do see where you're coming from, and I'm sure I will be able to live with it!

> Decided to nix command args

Surely we want them for batch-mode though?  Or else how are people going to get information into the program they're going to run?  Not by editing the program before each invocation, or parsing it from a text file, right?  People will want to use batch-mode in hands-off automated applications, so we can't assume they will edit the program with new parameters before each run.

pchiusano:
> It'll be frustrating if you're iterating on a transcript and every time you run it you need to wait for a pull.

Yeah, good call, we'll keep that around but just not make it the default. @atacratic 

> Surely we want them for batch-mode though? Or else how are people going to get information into the program they're going to run?

I think having a way of passing in programs dynamically (including via standard in) will be sufficient here. If you can supply arguments dynamically, you can supply a tiny little bit of Unison text and it has the bonus of not forcing your Unison programs to have a bunch of command line parsing logic.

Not totally sold on the above but I'd like to start with it and see how it goes. If there's huge demand for separate command line parsing support we could always add that later.

atacratic:
Sounds good.

By command args, in my head I thought we were talking about something like:

```
ucm run somepackage.doStuff 1 "hello"
```
in order to call
```
somepackage.doStuff : Nat -> Text ->{IO} ()
```

Might be an option to consider?  Avoids option parsing in the unison code.  And programmatically constructing a unison stub function in order to pipe in doesn't sound super convenient either.  

Suck-it-and-see approach sounds fine though.

aryairani:
Hey @pchiusano, I'm concerned that this approach to command-line arg parsing is going to scale.  If I want to implement #817, how will that fit in?  What if we just did a foldLeft on the args with some state and a config result?

sullyj3:
How many switches and subcommands are there going to be? If unison is anything like most languages, it'll end up with lots, so maybe it's worth going the [optparse-applicative](http://hackage.haskell.org/package/optparse-applicative) route?

pchiusano:
- [ ] Less output + progress bar, use `putStr` and `PrettyTerminal.clearCurrentLine`
- [ ] Sync with master
- [ ] Merge or redo `init` thing

pchiusano:
Hi I’m going to merge this. There’s one more todo, which is porting the init subcommand over.

Issue #820, opened 2019-09-25
pretty-printing glitch in wrapColumn2?


aryairani:
![image](https://user-images.githubusercontent.com/538571/65640268-f4c3e980-dfb7-11e9-9899-4e2016877328.png)

I believe this 👇 should produce 3 lines of output, but it only produces 2? ☝️ 
```
  (P.wrapColumn2
    [ ("`list`", "lists definitions and namespaces at the current level of the current namespace.")
    , ( "`list foo`", "lists the 'foo' namespace." )
    , ( "`list .foo`", "lists the '.foo' namespace." )
    ]
  )
```

Change my mind! :)

pchiusano:
Definitely seems like a bug

PR #826, opened 2019-09-29
Add codebase init command, no longer create codebases implicitly

sullyj3:
~Questions:
Can I use lambdacase in main? (Main.hs:118)
How can I get rid of the parens around the do block in FileCodebase? (FileCodebase.hs:125)
Is it OK to pull out formatAnn to the top level? I'm not sure what it's for. (FileCodebase.hs:155)~

### DIFF
--- a/parser-typechecker/unison/Main.hs
+++ b/parser-typechecker/unison/Main.hs
@@ -112,18 +112,20 @@ runTranscripts inFork mcodepath args = do
   currentDir <- getCurrentDirectory
   transcriptDir <- do
     tmp <- Temp.createTempDirectory currentDir "transcript"
-    when (not inFork) $
+    when (not inFork) $ do
       PT.putPrettyLn . P.wrap $ "Transcript will be run on a new, empty codebase."
+      _ <- FileCodebase.initCodebase tmp
+      pure ()
     when inFork $ FileCodebase.getCodebaseOrExit mcodepath >> do
-      codepath <- case mcodepath of
+      origCodePath <- case mcodepath of
         Just codepath -> pure codepath
         Nothing       -> getHomeDirectory
-      let path = (codepath FP.</> ".unison")
+      let path = (origCodePath FP.</> FileCodebase.codebasePath)
       PT.putPrettyLn $ P.lines [
         P.wrap "Transcript will be run on a copy of the codebase at: ", "",
         P.indentN 2 (P.string path)
         ]
-      Path.copyDir (codepath FP.</> ".unison") (tmp FP.</> ".unison")
+      Path.copyDir path (tmp FP.</> FileCodebase.codebasePath)
     pure tmp
   theCodebase <- FileCodebase.getCodebaseOrExit $ Just transcriptDir
   case args of
--- a/parser-typechecker/unison/Main.hs
+++ b/parser-typechecker/unison/Main.hs
@@ -63,29 +63,34 @@ usage = P.callout "🌻" $ P.lines [
 
 main :: IO ()
 main = do
-  args               <- getArgs
+  args <- getArgs
   -- hSetBuffering stdout NoBuffering -- cool
-  (codepath, args) <- case args of
-    "-codebase" : codepath : args -> pure (codepath, args)
-    _ -> (,) <$> getHomeDirectory <*> pure args
+
+  -- We need to know whether the program was invoked with -codebase for
+  -- certain messages. Therefore we keep a Maybe FilePath - mcodepath
+  -- rather than just deciding on whether to use the supplied path or
+  -- the home directory here and throwing away that bit of information
+  let (mcodepath, restargs) = case args of
+           "-codebase" : codepath : restargs -> (Just codepath, restargs)
+           _                                 -> (Nothing, args)
   currentDir <- getCurrentDirectory
-  case args of
+  case restargs of
     [] -> do
-      theCodebase <- FileCodebase.ensureCodebaseInitialized codepath
+      theCodebase <- FileCodebase.getCodebaseOrExit mcodepath
       launch currentDir theCodebase []
     [version] | isFlag "version" version ->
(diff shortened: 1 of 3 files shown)


pchiusano:
Hi @sullyj3 we ended up needing to shuffle a bunch of things around in `Main`, sorry about that! Would you mind updating this PR to a) add yourself to CONTRIBUTORS file and b) work against master. Here is an updated spec for how it should work:

* `ucm init` creates a new codebase in the user's home directory, (so `~/.unison`), then exits with a nice message "I successfully created a new codebase at <path>" or something. If there's already a codebase there, it prints a friendly message: "I didn't initialize a new codebase at <path> because there is an existing Unison codebase at that location."
* `ucm -codebase /some/path init` creates a codebase in `/some/path/.unison`, then exits. Same behavior if there's already a codebase there.
* Other subcommands should exit with a nice error message like "I didn't find a Unison codebase at the location <path>. Run `ucm -codebase <path> init` [or `ucm init` if user didn't supply `-codebase`] to initialize one, then try again." Notice how this pattern avoids a modal input, but the workflow is still quite efficient - user types `ucm`, then hits up arrow, then types `init<Enter>`. In general, we try to do everything without any modal inputs unless there's a compelling reason not to. Anytime you'd think of doing something with a modal input, just exit with a message about how to complete the command, rather than modally forcing the user to complete the command (and then you need a way to cancel, etc).
* `help` and `version` subcommands don't need a codebase.

You can see how the existing `Main` deals with the `-codebase` flag. Basically, once you have a location for where the codebase should be, you can do a check if one exists there (might need a helper function `FileCodebase.existsAt : FilePath -> IO Bool`).

Sorry for the churn! If you are out of steam on this let us know and someone else could take this over. Cheers!

pchiusano:
One other thing - we also discussed using opt-parse-applicative. I'm not super happy with the help it autogenerates or the error messages it produces, though maybe that can be customized. We could also just use Megaparsec if the ucm argumets / subcommands get more complicated (I really think we should try to keep them simple though). In any case, I'd leave that as a possible refactoring for later and stick with the simple arg parsing via pattern matching that's there right now.

sullyj3:
Cool cool, so `-codebase` must be supplied (if it is supplied at all) before any subcommands?

pchiusano:
@sullyj3 yes that’s right. That -codebase flag already exists and you can see how it’s done if you check current Main in master.

sullyj3:
Ok, seems to mostly be working. Couple of questions:

Firstly, I can't seem to get the Pretty printing to separate lines properly, I'm getting messages that look like:
```No codebase exists at /home/james Run `ucm init` to create one, then try again!```
when I expect `Run` onwards to be on a new line.

 Any ideas?

Secondly, I tried `unison transcript.fork hello.md`. It seemed to terminate successfully:
```
  I've finished running the transcript(s) in this codebase:

    /home/james/Code/sullyj3_unison/unison-src/transcript-b4079f501b0b2d89

  You can run
 `ucm -codebase /home/james/Code/sullyj3_unison/unison-src/transcript-b4079f501b0b2d89` to do more work with it.
```
I followed the suggestion
`unison -codebase transcript-b4079f501b0b2d89`
And it seemed to load successfully. However...
```
.> view x

  ⚠️

    x
```

Is this expected?

**Comment by sullyj3** [comment]
Oops, will get those tests passing

**Comment by sullyj3** [comment]
Fixed it. In the process I altered the transcript runner to use `.unison/v1` codebases rather than just `.unison`, let me know if that's not right.

File: parser-typechecker/unison/Main.hs [hunk 1]
```diff
@@ -63,60 +63,71 @@ usage = P.callout "🌻" $ P.lines [
...
+  args <- getArgs
   -- hSetBuffering stdout NoBuffering -- cool
-  (codepath, args) <- case args of
-    "-codebase" : codepath : args -> pure (codepath, args)
-    _ -> (,) <$> getHomeDirectory <*> pure args
+
+  -- We need to know whether the program was invoked with -codebase for
+  -- certain messages. Therefore we keep a Maybe FilePath - mcodepath
+  -- rather than just deciding on whether to use the supplied path or
+  -- the home directory here and throwing away that bit of information
```
aryairani: Thanks for including this comment.  I was about to ask, and then I thought "what if I try reading..."

aryairani:
Oops, let me take a look at your question I just now spotted.

aryairani:
Re. the pretty-printing: that's odd and I'm not sure.  You've got the `P.lines` call in there, which I believe should be enough.  I might be indicative of a bug related to #820?

And re. https://github.com/unisonweb/unison/pull/826#issuecomment-537510938 did I understand correctly that 1a6ab46 fixed the issue you were seeing?

aryairani:
And yes, feel free to use LambdaCase wherever you like.

sullyj3:
> And re. [#826 (comment)](https://github.com/unisonweb/unison/pull/826#issuecomment-537510938) did I understand correctly that [1a6ab46](https://github.com/unisonweb/unison/commit/1a6ab46a159f30cf402d76edcd86387a3a59afb6) fixed the issue you were seeing?

No, sorry, that's a separate thing I'm still confused about. Is it something to do with needing to qualify the name, maybe?

sullyj3:
> Re. the pretty-printing: that's odd and I'm not sure. You've got the `P.lines` call in there, which I believe should be enough. It might be indicative of a bug related to #820?

Seems likely, but also, the version I have slightly above using `<>` and `P.newline`, rather than P.lines has the same issue.

pchiusano:
So aside from the formatting thing, this is good to go? If so I’m inclined to merge and then investigate the formatting later.

sullyj3:
I think so. Sounds good.

pchiusano:
Yay! Thanks @sullyj3. I’ll work on tracking down the weird formatting issue.

PR #829, opened 2019-09-30
Not a real pull request. Message me. I am interested in this project.

reyesjl:
So I read around the website and took a look at some of the intro things. This sounds very interesting. I want to know more about how this can be used and how practical this is in application.

Thank you

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,19 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+- master
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- script: echo Hello, world!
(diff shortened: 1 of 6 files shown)


pchiusano:
Hi there, I’d come by the Slack for discussions like this. tiny.cc/unisonslack is invite 

Short answer is we are currently still alpha testing and I wouldn’t recommend trying to use Unison for real stuff just yet. But you can definitely play with it in its current form and we will be getting it to a more stable state in the coming months. Cheers!

reyesjl:
Awesome! Sounds like a super cool concept. Still trying to wrap my head around it.

Cheers!

On Sep 30 2019, at 2:18 pm, Paul Chiusano <notifications@github.com> wrote:
Hi there, I’d come by the Slack for discussions like this. tiny.cc/unisonslack is invite

Short answer is we are currently still alpha testing and I wouldn’t recommend trying to use Unison for real stuff just yet. But you can definitely play with it in its current form and we will be getting it to a more stable state in the coming months. Cheers!

PR #831, opened 2019-09-30
Update README.md

joyskmathew:


runarorama:
Hmm, I'm not so sure this is an improvement. The words "purely functional" now appear twice in the same paragraph.

@joyskmathew Can you explain the motivation behind these edits?

joyskmathew:
As a programmer it is our duty to support the new languages if it has the capability.

pchiusano:
Yeah I also am not sure about this change. I feel like README could use a revamp but not sure how. For starters that example with transfer is way out of date.

Going to close this one but will create a new ticket to track README revamp. Sorry @joyskmathew but thanks for the PR. Cheers!

PR #833, opened 2019-10-01
Test serialization on everything in unison-src/tests

TomasMikula:
Tests that everything in `unison-src/tests` can be serialized and deserialized into the original thing.

It duplicates some work of the `Typechecker` test, namely, the files are parsed and typechecked twice: once for the typechecker test and once for the serialization test. An alternative would be to merge these tests into one.

### DIFF
--- a/parser-typechecker/tests/Suite.hs
+++ b/parser-typechecker/tests/Suite.hs
@@ -8,6 +8,7 @@ import           System.IO
 import qualified Unison.Test.ABT as ABT
 import qualified Unison.Test.Codebase.Causal as Causal
 import qualified Unison.Test.Codebase.Path as Path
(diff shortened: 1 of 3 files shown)


TomasMikula:
@pchiusano on Slack:

> that’s all I’d do - in all the typechecking tests, add a round trip serialization. I don’t think it is worth writing a separate test suite for it.

I can incorporate this `Serialization` test into the `Typechecker` test (before or after this PR is merged), but it should then be renamed to something else than `Unison.Test.Typechecker`. Perhaps `Unison.Test.UnisonSources`?

aryairani:
That sounds good.

TomasMikula:
Closing in favor of #841.

Issue #838, opened 2019-10-02
add `FindPatchI` outputs to `LoopState`'s `numberedArgs`


aryairani:
Search `HandleInput.hs` for `numberedArgs` for examples.

pt2121:
I'll have a go at this.

cronokirby:
I think this has been solved?

aryairani:
Thanks @cronokirby

PR #840, opened 2019-10-03
Add FindPatchI outputs to LoopState's numberedArgs

pt2121:
### WIP

re #838

🤞

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -26,3 +26,4 @@ The format for this list: name, GitHub handle, and then optional blurb about wha
 * Ian Jeffries (@seagreen)
 * Silvino Escalona (@sunrisem)
 * James Sully (@sullyj3)
+* Prat Tana (@pt2121)
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -825,12 +825,13 @@ loop = do
           -- next update for that file (which will happen immediately)
           latestFile .= ((, True) <$> loc)
 
-      FindPatchI ->
-        let patches = Set.fromList
+      FindPatchI -> do
+        let patches =
               [ Path.toName $ Path.snoc p seg
               | (p, b) <- Branch.toList0 currentBranch0
               , (seg, _) <- Map.toList (Branch._edits b) ]
-        in respond $ ListOfPatches patches
+        respond $ ListOfPatches $ Set.fromList patches
+        numberedArgs .= fmap Name.toString patches
 
       FindShallowI pathArg -> do
         prettyPrintNames0 <- basicPrettyPrintNames0
--- a/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
+++ b/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
@@ -523,9 +523,15 @@ notifyUser dir o = case o of
           , "You're missing:" `P.hang` P.lines (fmap (P.text . Reference.toText) new)
           , "I'm missing:" `P.hang` P.lines (fmap (P.text . Reference.toText) old)
           ]
-  ListOfPatches patches ->
-    -- todo: make this prettier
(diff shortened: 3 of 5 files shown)


aryairani:
Hi @pt2121, this looks good!  Three questions:
1) Your comment above says “WIP”, is there more you want to do before merging?
2) Have you already added yourself to Contributors.md?
3) Would you be up to adding a transcript test for this functionality (see #816).  maybe the test could create a patch, and then find it using your WIP, then view the patch referenced by number?

pchiusano:
+1 on adding a transcript test for this. See unison-src/transcripts for some examples and do stack exec transcripts to run all the transcripts in that directory. Travis will also run all the transcript tests.

pt2121:
Hi guys,

Thanks for the cool project and your reviews.
I added WIP because I see the `find.patch` result is not numbered and wanted to double check that it's an expect behavior.
I'll add myself in to `Contributors.md` and look into adding the test.

aryairani:
Oops, yes, thanks for pointing that out. The numbers will be important too.  

You can find some somewhat ugly examples of the number rendering at https://github.com/unisonweb/unison/blob/e08581b63bbd5d324993f4c3717ddb1d9f6598ef/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs#L295-L302 and https://github.com/unisonweb/unison/blob/e08581b63bbd5d324993f4c3717ddb1d9f6598ef/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs#L1021-L1022, and then update the `ListOfPatches` case in that same file.  You don't have to use those same methods, but I'm guessing the `P.numbered` function will be helpful.

Thanks for the help!

pt2121:
Updated, patches in the output are numbered now. However, I have an issue create a transcript test.

How can I simulate editing a unison file in the test?

```
.> edit hello
```

Here is an example of what I tried:

[ls-patch.md](https://gist.github.com/pt2121/e01b2b5c580c6ff2688c70ae4e9ef90c)

[ls-patch.output.md](https://gist.github.com/pt2121/c3704ac0800f30ae803034dd71b0a095)

Thanks

pchiusano:
After doing the initial add, just have a second Unison code block with the updated definition. Then do an update as before.

The edit command dumps a definition to a scratch file but since the transcript isn’t working with literal scratch files (only the virtual scratch files you provide via code blocks), I don’t think that will be useful for transcripts. 😀

pt2121:
Thanks for the pointer - I'm totally new to the tools. 

This is getting long so I started a thread in slack.

aryairani:
@pt2121 Hey just wanted to check in to see if there was anything I could do to help with this? 😀

pt2121:
@aryairani, I struggled to write the transcript test for this.

The number command seems to work fine when I tested on a shell:
Screen Shot 2019-10-04 at 11 01 26 AM

Unfortunately, I can't manage to get it to work with the test
Screen Shot 2019-10-04 at 11 17 29 AM

Any thoughts? Thanks!

aryairani:
Doh, I see.

We just encountered [that same issue](#857) yesterday while working on another feature.  The temporary workaround has been "don't use numberedArgs in transcripts", which obviously is not a good workaround in this case!

I would recommend saying `view.patch p1` for now to verify that the rest of the transcript works, then go ahead and switch it back to `view.patch 1` before checking it in.  The PR will continue to fail the continuous integration test, but once we fix #857 we can re-run the build and go from there.

How does that sound?  Thanks!

pt2121:
Sounds great. will update

aryairani:
On second thought, wait — I think I have a fix (#860) for (#857); so let me get that merged, and then you merge `master` hopefully without too much pain 🤞, and then I think it should fix the error you're seeing?

aryairani:
Ok, should be good to go; let me know?

pt2121:
Yeah, the fix works great. I updated find-patch transcript and its output. Thanks again for your helps!

pchiusano:
Very nice, thanks @pt2121 😀 merging now

pt2121:
Thank you both!

PR #848, opened 2019-10-06
support autocompletion of path segments leading up to new names

aryairani:
Previously, there was no assistance in autocompleting new names, but we can help a little because we know the new name will fall somewhere in the existing tree.

image


pchiusano:
Wow, nice!!! I have wanted this for a while.

Issue #695, opened 2019-08-09
Add/Edit roundrip drops parens from patterns


runarorama:
```
use .base
use .Optional
use .Pair

foo : Optional (Pair Nat Nat) -> ()
foo x =
  case x of
    None -> ()
    Some (Pair _ _) -> ()
```

Views back as:

```
foo : Optional (Pair Nat Nat) -> ()
foo x =
  case x of
    Optional.None -> ()
    Optional.Some Pair.Pair _ _ -> ()
```

Which doesn't parse.

zenhack:
Slight correction: It parses ok but then reports a pattern airty mismatch (because you're trying to match Some with three arguments).

virusdave:
I ran into this several times while trying to structurally match certain types of lists: 
```
a +: (b +: rest) -> ...
```

atacratic:
All Unison infix operators associate to the left at the moment, so you will need the parens.  But yes it's bugged that the parens are dropped by the pretty-printer.

PR #854, opened 2019-10-10
Force parens around ctor application patterns

runarorama:


parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -391,7 +391,8 @@ prettyPattern n c@(AmbientContext { imports = im }) p vs patt = case patt of
   l :: IsString s => String -> s
   l = fromString
   patterns vs (pat : pats) =
-    let (printed     , tail_vs      ) = prettyPattern n c (-1) vs pat
+    let (printed     , tail_vs      ) =
+          prettyPattern n c (if Pattern.application pat then 11 else -1) vs pat
```
runarorama: The substantive change is here. The rest is mostly removing unnecessary Haskell code.

atacratic:
I tweaked the fix!  Also made it work for effect patterns like `{MyAbility.blah (Some x) -> k}`.

You'll see I added another test, which I couldn't actually run - I have run it by hand though.  

Thanks for the improvements to the test file!

I didn't delete `Pattern.application`, but you might want to as it's now unused.

Good to merge from my point of view.

atacratic:
Ah, correction, your version already worked for effect patterns - it was just mine that nearly didn't.

Fixes #695

PR #862, opened 2019-10-13
Show aliases in per command help

### LINKED ISSUE #810: Show aliases in per command help
The aliases appear in the overarching help (when entering `help` with no args), but don't appear in the per command help.

Change from this:
```
.> help namespace
`namespace foo.bar`
```

To this:
```
.> help namespace
`namespace foo.bar`
...
aliases: cd or j
```

jdevuyst:
This fixes #810, which is a request to show aliases in the per command help messages.

I figured it's also a bit confusing that when you enter `help j`, all the examples use the original `namespace` name. Hence I thought it might be useful to print both the original name and the aliases in the per command help. And it turns out there's already code that does just that.

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -27,3 +27,4 @@ The format for this list: name, GitHub handle, and then optional blurb about wha
 * Silvino Escalona (@sunrisem)
 * James Sully (@sullyj3)
 * Prat Tana (@pt2121)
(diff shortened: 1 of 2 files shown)


aryairani:
Could you show the new output, and also add yourself to contributors.md?

jdevuyst:
Sure thing. Thanks.

Screen Shot 2019-10-13 at 20 33 34

PR #863, opened 2019-10-13
Force line breaks between rows in wrapColumn2

### LINKED ISSUE #820: pretty-printing glitch in wrapColumn2?

jdevuyst:
This fixes #820

### DIFF
--- a/parser-typechecker/src/Unison/Util/Pretty.hs
+++ b/parser-typechecker/src/Unison/Util/Pretty.hs
@@ -307,7 +307,8 @@ parenthesizeIf False s = s
 parenthesizeIf True s = parenthesize s
 
 lines :: (Foldable f, IsString s) => f (Pretty s) -> Pretty s
-lines = intercalateMap newline id
+lines = intercalateMap (append newline) id where
+  append p = Pretty (delta p) (Append $ Seq.singleton p)
 
 linesNonEmpty :: (Foldable f, IsString s) => f (Pretty s) -> Pretty s
 linesNonEmpty = lines . nonEmpty


jdevuyst:
Oops. I didn't notice the test failures. They don't show up when I run `stack run test` or `stack test` in the root directory.

atacratic:
Could you try `stack exec test`? That's what I type. Let us know!

Shout if you need a hand with the TermPrinter. I think some of the existing behaviour that your change alters is a bit janky. But I think you do need to change the TermPrinter somewhere.

jdevuyst:
`stack exec tests` indeed does the trick.

I updated my PR, which now passes all tests. Not sure if it's the solution you'd want though.

The gist of my patch is that I'm wrapping the lines in an `Append` as this leads to a 'go Right' in `render`, meaning `flow` is not called for the newlines that separate rows.

atacratic:
(`stack run test` seems to be running the tests for yaks/easytest, but I can't work out why.)

Tagging @aryairani for a review and merge?

aryairani:
@atacratic That’s probably just the command to run stack tests for all subprojects, but we want `stack exec tests` not `stack run test`, right
?
@jdevuyst Would you include the new output for the help command, mirroring the buggy output of  #820?

jdevuyst:
Screen Shot 2019-10-13 at 21 47 59

PR #869, opened 2019-10-16
Advertise base library on first launch

jdevuyst:
I'm not entirely sure this is a good idea...

More than once I caught myself (as a new user) being surprised that I couldn't `use` functions and types from `.base`, only to discover I first had to pull in the base library.

Here's a tweak to the welcome message that points out to users that they may want to do this:

Screen Shot 2019-10-16 at 23 34 00

The message is only shown for new codebases. I'm using `Branch.isOne` because `.builtin` would already be available. I'm actually wondering if maybe it is worth pointing out in the message that `.builtin` exists.

Or perhaps a link to the tutorial would be better.

So, as I mentioned already, I'm not entirely sure if this change is worth retaining. :)

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/Main.hs
+++ b/parser-typechecker/src/Unison/CommandLine/Main.hs
@@ -116,6 +116,11 @@ welcomeMessage dir =
          , P.wrap ("Type " <> P.hiBlue "help" <> " to get help. 😎")
          ]
 
+hintFreshCodebase :: P.Pretty P.ColorText
+hintFreshCodebase =
+  P.wrap $ "Enter " <> P.hiBlue "pull https://github.com/unisonweb/base .base"
+    <> "to set up the default base library. 🏗"
+
 main
   :: forall v
   . Var v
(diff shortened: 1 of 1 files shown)


atacratic:
I think this PR fixes a usability hole, which I have fallen into more than once. I'd say merge it and then rethink later if there are ever angry alternate prelude authors. A hint message is an OK privilege I think. At the moment we're 'missing a breadcrumb' in the install process, and from a newbie experience point of view that's broken.

pchiusano:
Sounds good to me. I think it's fine and we can revisit later when there are 50 awesome alternative base libraries. :)

This PR reminds me I might like to do some tweaks to the welcome message, but that could be done later. I feel like `help` command is not super useful as is - it's just a big list of commands.

runarorama:
The welcome message could link to the UCM documentation, which we need to write.

aryairani:
Sold.

aryairani:
This has already paid for itself, as I accidentally `undo`ed my merge of `.base`, and couldn't figure out how to use `.unisonConfig` to set a default path for `.base`.

But when I restarted `ucm`, there was a
> Enter pull https://github.com/unisonweb/base .base to set up the default base library. 🏗

waiting for me.

Issue #870, opened 2019-10-16
function typechecks with nonexistent ctor pattern


aryairani:
Accidentally inspired by #868
```haskell
type Choice = First | Second
type Wrapper = Wrapper Choice

broken = case Wrapper.Wrapper Choice.Second of
    Wrapper.Wrapper Choice.First -> 1
    Wrapper.Wrapper Choid.Second -> 2
    _ -> 3

> broken
```
```
.> 

  I found and typechecked these definitions in ~/unison/u.u. If you do an `add` or `update`, here's
  how your codebase would change:
  
    ⍟ These new definitions are ok to `add`:
    
      type Choice
      type Wrapper
      broken : base.Nat
   
  Now evaluating any watch expressions (lines starting with `>`)... Ctrl+C cancels.


    9 | > broken
          ⧩
          1

.> 
```

Note that **Choid**.Second is not a thing.

aryairani:
Guessing it's the same underlying problem as #868, and that pattern checking is just rawng.

PR #871, opened 2019-10-16
fix #868, #870

### LINKED ISSUE #868: Wrong case matches when pattern matching
In the following case the first case matches (`Wrapper.Wrapper Choice.First`) when matching on `Wrapper.Wrapper Choice.Second`. Pattern matching works correctly without the Wrapper on just the Choice type.

```haskell
type Choice = First | Second
type Wrapper = Wrapper Choice

broken = case Wrapper.Wrapper Choice.Second of
    Wrapper.Wrapper Choice.First -> true
    _ -> false

> broken
-- Should be `false` but evaluates to `true`
```

aryairani:
"Weird, it doesn't seem to recognize a constructor except at the top level"

<Explicitly doesn't allow constructors except at the top level>


parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -183,7 +184,7 @@ parsePattern =
         f patterns vs =
           let loc = foldl (<>) (ann tok) $ map ann patterns
           in (Pattern.Constructor loc ref cid patterns, vs)
-    unzipPatterns f <$> many (leaf <|> seqLiteral)
+    unzipPatterns f <$> many patternCandidates
```
pchiusano, line 187: @aryairani FYI I believe this change introduced #1640 - by having the arguments to the constructor also include the `constructor` parser, you end up with it being right associated, where `Foo Bar a b` is parsed as `Foo (Bar a b)`. The fix I implemented to avoid nullary constructors being treated as variables was to just add a nullary constructor parser and move that to before the variable parser in `leaf`. See this note here: 

https://github.com/unisonweb/unison/pull/1642/files#diff-211d62c2ee7dccd868dc459a4e36136cR159-R164

Sorry for not spotting this during review of this PR!

It was good that we had the #868 test case though, because I did it incorrectly at first and got a failure that led to me figuring this out. :)

Issue #873, opened 2019-10-17
TODO - don't know how to compile this term


jdevuyst:
I entered `baz x = x - 1` in a scratch file and then I got this error:

```
    ⍟ These new definitions are ok to `add`:
    
      baz : base.Nat -> base.Int
   
  Now evaluating any watch expressions (lines starting with `>`)...
  Ctrl+C cancels.


  💔💥
  
  I stopped evaluation after encountering an error:
  
    TODO - don't know how to compile this term:
    #m7uplgfko9
    CallStack (from HasCallStack):
      error, called at src/Unison/Runtime/IR.hs:483:10 in unison-parser-typechecker-0.1-GhEsNfzP1hn5VAyhpU4Q65:Unison.Runtime.IR
```

Note that this seems to only happen after pulling in the base library. Otherwise the editor complains that it doesn't know what `-` is.

noahhaasis:
It looks like a problem with the typechecker. It inferes type `Nat -> Nat -> Nat` for `-` but `Int` for `1`.

```haskell
i : builtin.Nat
i = 1

baz : builtin.Nat -> builtin.Nat
baz x = x - 1
```

![Bildschirmfoto 2019-10-17 um 17 32 34](https://user-images.githubusercontent.com/26230013/67024113-271ebd80-f104-11e9-8111-0b0c4c641e60.png)

Without the type declaration for `baz` it compiles and fails.

pchiusano:
Thanks for reporting @jdevuyst and @noahhaasis, that is a good clue.

pchiusano:
I don't think it's an issue with the typechecker. The `base` library has a version of `-` that works on `Nat` and returns an `Int`:

```
  (base.Nat.-) : base.Nat -> base.Nat -> base.Int
  (base.Nat.-) = base.Nat.sub
```

You can also call `Nat.sub` directly.

This seems to be a regression someplace else. A good clue is that I had an old version of topic/reflog checked out which I just happened to try it on but could not reproduce there, but I can reproduce in current master. I'd probably start by bisecting.

jdevuyst:
The issue appears to have been introduced in 86e82a6faacf90f4c72d69679a903c97541bae3a

aryairani:
@jdevuyst Oof, thank you for tracking that down.

aryairani:
I've reverted the commit that @jdevuyst identified as introducing the bug.

To close this issue, would someone please create a test for this case.

PR #874, opened 2019-10-18
Revert "Is Codebase.makeSelfContained' unnecessary?"

aryairani:
Reverts unisonweb/unison#771

@TomasMikula If you have a chance, could you take a look at #873 in terms of a test that fails without `makeSelfContained'`

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -6,6 +6,7 @@ import Unison.Prelude
 
 import qualified Data.Map                      as Map
 import qualified Data.Set                      as Set
+import qualified Unison.ABT                    as ABT
 import qualified Unison.Builtin                as Builtin
 import           Unison.Codebase.Branch         ( Branch )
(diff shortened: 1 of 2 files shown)


TomasMikula:
Such test would be nice and would definitely save us both some time. I don't know where to start, though - it is not just another test of a kind that's already in the codebase.

Btw, I have a much simpler version of `Var` that is not burdened by the implementation details of the runtime (namely `RefNamed` variables) or type inference (namely `Inference` variables). Removing `makeSelfContained'` was just an attempt to avoid having to modify it so that it does not use `RefNamed` vars. I will have to find a workaround (as I did for other uses of `RefNamed`).

aryairani:
I went ahead and created a transcript test (#873, #816) for it.

TomasMikula:
Nice! I was imagining some tricky programmatic way of initializing a codebase that would then be used for typechecking the term.

aryairani:
@TomasMikula Yeah transcript tests are a big help. 

BTW, I’m all for simplifying and disentangling the current implementation, and appreciate your fresh perspective.

PR #875, opened 2019-10-19
Fix #770: Allow boolean operators in pattern guards

### LINKED ISSUE #770: Boolean operators `and` and `or` don't work in pattern guards
This code: 
``` haskell
use .base

type Foo = Foo Boolean Boolean

f : Foo -> Boolean
f x = case x of
  Foo.Foo a b | and a b -> true
  _ -> false
```
produces:
```
  /home/chris/u4/scratch.u:7:17:
  unexpected and
  expecting infixApp
      7 |   Foo.Foo a b | and a b -> true
```
(where the `and` on the last line is shown in red.)

Similarly `or`.

Workaround: define the following and use it in the guard:
```haskell
and' a b = and a b
```
(M1c, unisonbase master@c4692e)

noahhaasis:


parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -1,7 +1,6 @@
 {-# OPTIONS_GHC -Wno-partial-type-signatures #-}
 
-{-# LANGUAGE ViewPatterns #-}
-{-# LANGUAGE RankNTypes #-}
```
aryairani: Accidentally lost linebreak?

noahhaasis: No it wasn't intentional. Thx for spotting

noahhaasis: the LANGUAGE options shouldn't change

PR #879, opened 2019-10-20
Watch files

jdevuyst:
Two changes:

- Only watch for .u files in the current directory (I believe this was discussed on Slack some time ago and thought to be a good idea)
- On startup scan all .u files in the current directory

What prompted this PR is that I would always find myself calling `touch` because it doesn't come natural to me to launch `ucm` before writing code. I also thought that when the REPL is quit and restarted, it is somewhat annoying to find that `ucm` forgot about earlier scratch files.


aryairani:
I'm thinking we should delete `IO.u` or other .u files from the distro root in this case. Will that cause any problems, @runarorama ?

pchiusano:
No that file is way old and is not read, IOSource embeds its own .u file.

pchiusano:
Great, thanks @jdevuyst, merging...

PR #880, opened 2019-10-20
Infix boolean operators (Fix #530)

### LINKED ISSUE #530: Change syntax of `and` and `or` to not be prefix, which is super annoying (either `and` infix or `&&` infix)
Currently they are builtin prefix syntax and look like functions even though they're not. Possible alternatives:

1. Make these `&&` and `||`, which will be familiar to most programmers. People are more ready to accept that infix operators are builtin syntax.
2. Make them special infix syntax like `foo and bar`. This might be OK and is more friendly to newbies who maybe have never seen a programming language.

noahhaasis:
This implements infix boolean operators using the standard C-style notation.
So `and a b` is now `a && b` and `or a b`, `a || b`.
What I'm not sure about is the label in the parser. Maybe these should use the new notation too but all other labels for symbols use the symbols name instead of the actual symbol

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -372,6 +372,12 @@ lexer0 scope rem =
       '_' : (wordyId -> Right (id, rem)) ->
         let pos' = incBy id $ inc pos
         in Token (Blank id) pos pos' : goWhitespace l pos' rem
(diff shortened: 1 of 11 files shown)


pchiusano:
Great, thanks @noahhaasis this has been on our todo list for a while, thanks for knocking it out!

PR #881, opened 2019-10-21
Add `bug` and `todo` builtin (Close #769)

### LINKED ISSUE #769: Builtins like undefined/error
Ideas from Paul:
> I’d like two versions, one called todo (you’re planning to fill in later) and another called crash (you’re not expecting this situation). They could just be new builtins, each of type Text -> a.
```
todo : Text -> a
crash : Text -> a
```

noahhaasis:


atacratic:
Thanks for the PR! 😁

FWIW I like the idea of having a standardised 'todo' function, although I don't have a view on the right implementation.

I'd ideally want the uses of `todo` to show up in the output of the `todo` ucm command, someday.

(Side note: in general it's nice to put a description in PRs like this so people can easily see the user-facing changes.  For example, calling out that you have a builtin `todo` with an alias `bug`, and saying what the runtime behaviour is.  It makes it easier for people to get involved who don't know the code, and also makes it easier for the maintainers to understand your intention.)

pchiusano:
Thanks @noahhaasis!

I know @aryairani had some ideas for this so you may want to wait on his review, but a couple things I noticed:

- I'd vote for a separate function for `todo` vs `bug`, not just an alias. Could be same implementation for both, but it's useful to distinguish the two sometimes (like the `todo` command could just show dependents of `todo`, not `bug`)
- Can we do a transcript test for this? You can use `unison:error` for a code block that you are expecting to bomb.
- Can we see some screenshots of this in action? Like what does the crash look like? Do we want to provide some nicer rendering of the failure? (Could potentially be done with separate PR if needed)

atacratic:
On seeing it in action, I'd also like to know what happens when you add a definition with a `todo` that is not guarded by a lambda, for example:

```haskell
f : .base.Int
f = todo
```
I'm thinking Unison might try to evaluate `todo`, which we might not want.

noahhaasis:
Currently `bug` and `todo` behave the same.
Bildschirmfoto 2019-10-24 um 13 11 15

Bildschirmfoto 2019-10-24 um 13 11 48

Ideally they should throw an error containing the hash of the function where they are embedded, so that  ucm can catch the error and show the code.

@atacratic Currently it would throw an exception. You could use `undefined : a` maybe?

pchiusano:
We want to think about a couple things before merging:

- Is this the signature we want for these? Only other thought I had was `forall a b . b -> a`, attempts to prettyprint the argument (which might be `Text` or something else).
- Naming? Are we okay with these being top level?
- Some questions about runtime behavior of these - can they be caught in `IO`?

pchiusano:
@noahhaasis gave this some thought and  I’d like to change these to take any value, not just text - it’ll be super nice to not have to do string munging just to get info out of your crashing program 😀. I’m not sure what all the changes are that will be involved but I think the interpreter should be able to return these as an error result, similar to like an uncaught request, and these errors should be prettyprinted nicely in the same way as like watch expression results.

noahhaasis:
@pchiusano The interpreter now returns a (RError String) where the string is a prettyprinted version of the generic argument.
So ucm does display a nice message now.

Bildschirmfoto 2019-11-05 um 21 05 26

A problem that i still have to fix is that `bug` and `todo` should be evaluated lazily.
Because with the current implementation you can't do the following.
```haskell
test = todo "define me"
id x = test
```

noahhaasis:
I looked at the other strict functional languages and it looks like none of them have something like `undefined :: a`, because it needs lazy evaluation. So far I've looked at OCaml, Elm, and PureScript.
So maybe we have to live with the above code example not being possible.

parser-typechecker/src/Unison/Runtime/Rt1.hs
```diff
@@ -211,6 +211,7 @@ data Result
   = RRequest Req
   | RMatchFail Size [Value] Value
   | RDone Value
+  | RError String
```
pchiusano: I'd like to make this `RError ErrorType Value`, basically runtime won't be responsible for converting these errors to strings.

```
type ErrorType = Bug | Todo
```

parser-typechecker/src/Unison/Runtime/Rt1.hs
```diff
@@ -543,6 +545,8 @@ run ioHandler env ir = do
           EQ -> 0
           LT -> -1
           GT -> 1
+      Bug i -> (RError . show) <$> at size i m
+      Todo i -> (RError . show) <$> at size i m
```
pchiusano: No need to show here with the change to `RError`.

parser-typechecker/src/Unison/Runtime/Rt1IO.hs
```diff
@@ -489,6 +489,15 @@ toTermOrError ppe r = case r of
...
       ]
+  Right (RT.RError msg) ->
+    pure . Left . P.callout icon . P.lines $ [
+      P.wrap ("I've encountered a call to a" <> P.red "error function"
+              <> "with the following message:"), "",
+      P.indentN 2 $ P.wrapString msg,
+      "",
+      P.wrap "This happens when calling one of the builtin error functions like `bug` or `todo`.",
+      "", sorryMsg
+      ]
```
pchiusano: With the change I suggested above, you can now give a different error depending on whether the crash is due to a `bug` or `todo` call. You can decompile the `Value` using `IR.decompile`  and  you can use the `TermPrinter` to render the term (not `show`): 

```Haskell
pretty :: Var v => PrettyPrintEnv -> AnnotatedTerm v a -> Pretty ColorText
```

pchiusano:
@noahhaasis I missed that you’d addressed The PR comments, looks great now. Can you do one last thing and run stack exec transcripts locally and check in the output file as well?

noahhaasis:
No prob @pchiusano. I thought that'd you'd maybe wait for location annotations but they are a seperate issue and I haven't gotten around to implementing them yet.

noahhaasis:
I think the transcript fails because it doesn’t run the watch expressions in the same order every time. I’ll fix them later

noahhaasis:
@pchiusano should be ready to merge now

parser-typechecker/src/Unison/Runtime/Rt1IO.hs
```diff
@@ -489,6 +489,17 @@ toTermOrError ppe r = case r of
       P.wrap "This happens when calling a function that doesn't handle all possible inputs.",
       "", sorryMsg
       ]
+  Right (RT.RError t val) -> do
+    msg <- IR.decompile val
+    let errorType = case t of 
+                      RT.ErrorTypeTodo -> "builtin.todo" 
+                      RT.ErrorTypeBug -> "builtin.bug"
```
aryairani: Small caveat: these might not be their names forever; but we don't have a lovely way to supply these names back from the namespace to the runtime.

aryairani:
> I think the transcript fails because it doesn’t run the watch expressions in the same order every time. I’ll fix them later

@pchiusano How do you think this could happen?

pchiusano:
Watch expressions should be run in source order and should be deterministic, if that’s not the case let’s file a ticket. Was that actually what was happening?

This PR looks good though, going to merge!

mitchellwrosen:
Yeah @pchiusano that was actually happening.

```
bug "one"
bug "two"
```
was showing up in the output as "two"

PR #883, opened 2019-10-21
Implement ordering for char literals (fix #850)

### LINKED ISSUE #850: base.Universal.compare always returns 0 for Chars
```
> base.Universal.compare ?a ?a
```

```
    22 | > base.Universal.compare ?a ?a
           ⧩
           +0
```

```
> base.Universal.compare ?a ?c
```

```
    22 | > base.Universal.compare ?a ?c
           ⧩
           +0
```

This means a `Set Char` will always return `true` for `Set.contains`.

noahhaasis:
By adding a case to cyclicOrd, Universal.compare now works for character literals. (fixes #850)


pchiusano:
@noahhaasis great, could we add a test for this? Maybe just add to the existing `universals.u` file: https://github.com/unisonweb/unison/blob/master/unison-src/tests/methodical/universals.u and be sure to update https://github.com/unisonweb/unison/blob/master/unison-src/tests/methodical/universals.ur to match

pchiusano:
Excellent, thanks @noahhaasis merging now...

PR #887, opened 2019-10-23
Fail fast when creating an ill-formed context

TomasMikula:
There doesn't seem to be a good enough reason to allow creation of non-wellformed contexts. Instead of creating a context with a boolean flag `isWellformed`, fail fast with a compiler bug when trying to create a non-wellformed context.

The cases where the discarded part of a context was itself returned as `Context` were easily adapted to just work with a list of the discarded context elements.

Also, better distinguish the different reasons why `Context.solve` might fail. (See the respective commit message for details.)

**Motivation** for this is to get more useful error messages when something goes wrong.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -367,19 +367,18 @@ markThenRetract0 markerHint body = () <$ markThenRetract markerHint body
 replace :: (Var v, Ord loc) => Element v loc -> Context v loc -> Context v loc -> Context v loc
 replace e focus ctx =
   case breakAt e ctx of
-    Just (l, _, r) -> l `mappend` focus `mappend` r
+    Just (l, _, r) -> l `mappend` focus `extendN` r
     Nothing -> ctx
 
 breakAt :: (Var v, Ord loc)
         => Element v loc
         -> Context v loc
-        -> Maybe (Context v loc, Element v loc, Context v loc)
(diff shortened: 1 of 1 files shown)


TomasMikula:
@pchiusano The only place where the `isWellformed` flag was checked was in `wellformedType`. So your question reduces to

_Isn't there a place where we want to report `wellformedType` returning `false` as a type error, even if the reason for it returning `false` was ill-formed context?_

The only place where `wellformedType` is used (except for `extend` where it returning `false` would cause creation of another ill-formed context) is `solve`. There, `wellformedType` is called with a prefix of the context that `solve` is called with. If we assume that `solve` must be called with a wellformed context, then the context passed from `solve` to `wellformedType` is also wellformed, because any prefix of a wellformed context is wellformed. Therefore, the reason for `wellformedType` returning `false` in this case could not be the ill-formedness of the context. So the answer to the above question is "No."

But what about any cases where an invalid context might be created and we don't care about its invalidity? That isn't so easy to analyze. If there is indeed such a case, I think it should better be refactored to not use `Context`, but e.g. a plain list of context elements, so that all the other places can still benefit from detecting unwanted invalid contexts early.

pchiusano:
Okay, am sold, thanks @TomasMikula! 

You know a while ago I had hastily tried a change like this, but I missed that retract can return an invalid context on the right side and you need to do something to not have that bomb (I like your solution of just returning the context elements for the right side). I'm happy that someone worked this out. :)

Issue #513, opened 2019-06-06
Different representation of programs before and after typechecking


TomasMikula:
Currently, the same types `Term` and `Type` are used to represent programs before and after typechecking. I propose to represent typechecked programs differently than programs before typechecking.

### Motivation

 1. Express at the type level that a term is typechecked. Consumers of typed programs would have static guarantees of them being typed, instead of keeping track of that in programmer's head.
 2. Better starting point for pluggable syntax and pluggable typechecker. Different typecheckers could use different term types on input. Now we are at risk that `Term` and `Type` accomodate peculiarities of the default syntax or the default typechecker (for example, `IntroOuter` constructor of `Type` seems like a peculiarity of Haskellish syntax).
3. Have a more "normalized" representation of typed programs to avoid anomalies mentioned below.

### Suggestion

I don't have it all thought out, but for typechecked terms I would suggest to have a required type annotation at each level of a typed term, and the type annotation would contribute to the hash.

Types of typechecked terms should not have the `IntroOuter` case. 

#### Terms differing just by an explicit type annotation are considered different

Given source

```
id1 : Nat -> Nat
id1 x = x

id2 : Nat -> Nat
id2 x = (x: Nat)

> id1 == id2
```

the two functions differ:

```
  ┌
  │  7 | > id1 == id2
  │        ⧩
  │        false
  └
```

If in typechecked terms there was a type annotation at each subterm, the two functions would be the same.

#### Terms with the same source but different types are considered equal

The source

```
idNat: Nat -> Nat
idNat i = i

idInt: Int -> Int
idInt i = i
```

is successfully typechecked

```
  ┌
  │  ✅
  │  
  │  I found and typechecked these definitions in /Users/thomas/projects/unison/source.u:
  │  
  │    idInt : Int -> Int
  │    idNat : Nat -> Nat
  │  
  │  Now evaluating any watch expressions (lines starting with `>`)...
  └
```

but only one of them can be added to the codebase, because they have the same hash:

```
master> add

  ┌
  │  ✅
  │  
  │  I added these definitions:
  │  
  │    idNat : Nat -> Nat
  └
```

If every sub-term had a type annotation that would contribute to the hash, the two functions would have different hashes.

pchiusano:
Hmm, you should be able to add both idNat and idInt even with the current implementation. 

[...]

PR #888, opened 2019-10-23
Make AskInfo a Term instead of a kind of variable.

TomasMikula:
Reason: It has nothing to do with variables. Var operations like freshening do not make sense for `AskInfo`.

Direct representation as a term also avoids the special treatment of the _"`AskInfo` variable applied to a term"_ pattern, which used to encode _"term inside `AskInfo`"_.

---
Also, this is a step towards #899.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Serialization/V1.hs
+++ b/parser-typechecker/src/Unison/Codebase/Serialization/V1.hs
@@ -494,6 +494,8 @@ putTerm putVar putA = putABT putVar putA go where
       -> putWord8 4 *> putText t
     Term.Blank _
       -> error "can't serialize term with blanks"
+    Term.AskInfo _
+      -> error "can't serialize term with AskInfo"
     Term.Ref r
       -> putWord8 5 *> putReference r
     Term.Constructor r cid
--- a/parser-typechecker/src/Unison/Codecs.hs
+++ b/parser-typechecker/src/Unison/Codecs.hs
@@ -170,6 +170,7 @@ serializeTerm x = do
         incPosition
       Blank b -> error $ "cannot serialize program with blank " ++
                          fromMaybe ""  (Blank.nameb b)
+      AskInfo _ -> error "cannot serialize program with AskInfo"
       Handle h body -> do
         hpos <- serializeTerm h
         bpos <- serializeTerm body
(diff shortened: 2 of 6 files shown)


aryairani:
@pchiusano What do you think about this?  Do we even want to keep askinfo?

pchiusano:
Hmm, I don't want to make `AskInfo` a term (if we even keep it around). My feeling is we shouldn't modify the AST just to make it more convenient to get info out of the typechecker. I already think it is a little weird that we have the `Blank` nodes.

Instead, if we really feel the need, we should introduce a separate term / type functor which lets us embed more information into it that is useful for typechecking / learning info from the typechecker.

aryairani:
So do we want to keep it around?

pchiusano:
I'm cool with just getting rid of it for now.

TomasMikula:
> Hmm, I don't want to make AskInfo a term (if we even keep it around). My feeling is we shouldn't modify the AST just to make it more convenient to get info out of the typechecker. I already think it is a little weird that we have the Blank nodes.
> 
> Instead, if we really feel the need, we should introduce a separate term / type functor which lets us embed more information into it that is useful for typechecking / learning info from the typechecker.

@pchiusano I completely agree. #513 also leans in this direction.

However, I don't think this PR makes things any worse than they already are: `Term` already depends on `AskInfo`: not only transitively via `Var`, but `Term` code directly recognizes the special incantation

```haskell
unAskInfo tm = case tm of
  App' t arg | isVarKindInfo t -> ...
```

What this PR improves is that
 1. `Var` no longer depends on `AskInfo`; and
 2. makes the technical debt more explicit instead of burying it deeper and pretending it does not affect the AST.

You can see this PR as an intermediate step in cleaning things up, whether `AskInfo` will eventually be removed or moved to a richer term functor. I can sign up myself for picking that up later.

pchiusano:
I vote for just removing it for now.

Adding it to the term functor means it's part of the codebase format, which I definitely don't want. Right now, all the variable type info is discarded (only the name is retained) so adding or removing variable types doesn't affect the codebase format

TomasMikula:
Let's remove it then! 🎉 

> Adding it to the term functor means it's part of the codebase format

It is treated the same as `Blank` - an error is thrown if someone tries to serialize a term with `AskInfo`. Codebase format is not affected by this PR.

> Right now, all the variable type info is discarded (only the name is retained) so adding or removing variable types doesn't affect the codebase format

Still, adding a term that contained `AskInfo` variable to the codebase would most likely be a bug that would go uncaught. Crashing with an error is better IMO.

(I'm continuing the discussion because I have a similar treatment of `MissingResult` in the pipeline 😉 )

PR #889, opened 2019-10-24
Propagate type edits and non-type-preserving term edits

runarorama:
# Overview

This change expands the `update` propagation algorithm to allow:

1. **Type updates.** Updating a type will replace the old type with the new one in any types that refer to it, as well as any type signatures and type annotations in terms. These updates are confined to the current namespace.
2. **Term updates that don't preserve the type.** Updating a term previously did no propagation if the type of the updated term did not conform to the type of the old term. This change relaxes that restriction and instead attempts to propagate the change to dependents. Each dependent term is typechecked after propagation. Propagation stops if typechecking fails.

[This transcript](https://github.com/unisonweb/unison/blob/1636d8ea32a70d286599052b72a07c4bafcbb71c/unison-src/transcripts/propagate.output.md) demonstrates the new behaviour of `update`.

We also added a `resolve.type` command, which mirrors the existing `resolve.term` command, allowing you to manually add type edits to a patch.

# Code changes

The pith of this change is in [Propagate.hs](https://github.com/unisonweb/unison/pull/889/files#diff-aacd9a8c22b0d8b261f6248cd331504a). The function `propagate` gets called from [the input handler for the `update` command](https://github.com/unisonweb/unison/pull/889/files#diff-a7b61dc605567af930d2e407465ed47cR1336).

The implementation is largely the same as the old implementation which used to be in `HandleInput.hs`.

# Controversial decisions

We made the decision to _do nothing_ if the a dependent doesn't typecheck. You could make the case that it would be better to raise the issue to the user, although we're already sort of doing that via the `todo` command.

We also _do nothing_ to update terms that reference type constructors or patterns of the old type during a type update. We have no idea how (or if) the constructors and patterns of the old type map onto those of the new type, so we can't even try. We could address this later with hygienic macros, term rewriting rules, or some other kind of metaprogramming facility. A workaround is to encourage use of smart constructors and folds for data types that may need to be upgraded later, basically adding a layer of abstraction.

# Loose ends and related issues

1. We can make the propagation faster by _not typechecking_ if the edit is a type-preserving edit in the first place. And even if it's not, the propagation might at some point reach a type-preserving frontier at which we can stop typechecking. This is tracked in #904.

2. Related issue: `resolve.term` sometimes crashes. This is tracked in #903.

3. The `resolve` commands can only take term hashes, making this change a lot less useful than it could be. Should be able to take names as well as data/ability constructor hashes. Tracked in #905.

4. A non-type-preserving edit results in a confusing message: "That patch contains external dependents." Tracked in #906.


aryairani:
Wanted to review this, but not quite sure what I'm looking at, or where to begin.  I read the transcript, though.

Would it be accurate to describe the changes so far as the ability to automatically propagate type changes within term definitions that don't use any of the type's constructors?

I guess we're not quite in a position to be able to write property-transcripts, but how do we tell if it's doing what we want it to do?  (And what do we want it to do?)

pchiusano:
@runarorama I am very excited for this to land but could you spend 30-60 minutes doing a little writeup of this PR? 😀 I think unless the diff is trivial or self-explanatory each PR should have some sort of writeup. Makes it easier for reviewers and means reviews can often be done totally asynchronously which is a win. Here's a proposed structure, wdyt?

1. _Section: High level description of the change._ Possibly with links to transcripts demonstrating the new functionality.
2. _Section: High level description of code changes._ With links to the interesting parts of the implementation. 
3. _Section: "possibly controversial decisions"_. Call out things that might warrant some discussion or that you want to raise awareness of.
4. _Section: Loose ends_ Things not done yet that we may want to address in the PR or create separate issues to track (or not).
5. Also, ProTip, you can create [DRAFT PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/) and populate them as you go. I'm going to start doing this as it's a nice spot to track the todo list you are using for the work.

If you think this is more than say an hour of work you could just timebox it to an hour or less and then we can still do a video call to review.

unison-src/transcripts/redundant.md
```diff
@@ -0,0 +1,15 @@
+The same kind of thing happens with `map`. Are we saying this is incorrect behaviour?
```
pchiusano: @runarorama was this file accidentally committed? Or if not maybe flesh out this file with some more verbiage.

runarorama: I have no idea where this came from. Merge from master?

pchiusano: Here's git blame: https://github.com/unisonweb/unison/blame/ef640acb3ba5fdc91fc36c3a859763c9215b6151/unison-src/transcripts/redundant.md

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -825,13 +831,12 @@ loop = do
...
 
-      FindPatchI -> do
-        let patches =
+      FindPatchI ->
+        let patches = Set.fromList
               [ Path.toName $ Path.snoc p seg
               | (p, b) <- Branch.toList0 currentBranch0
               , (seg, _) <- Map.toList (Branch._edits b) ]
-        respond $ ListOfPatches $ Set.fromList patches
-        numberedArgs .= fmap Name.toString patches
```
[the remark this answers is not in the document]
runarorama: Probably not!

pchiusano:
@aryairani I think for transcripts that we're using for testing / regression, we either should a) not use `unique`, b) not have anything in the transcript output which will change based on the GUID of the `unique` type (so don't use `find.verbose` for instance on a `unique`) or c) supply the GUID explicitly in the `unique` declaration.

aryairani:
Some options: 
1. we always check them in
2. we always ask people to revert such changes to this file
3. we use unique guids in transcripts (which makes them less realistic)
4. we use a nonrandom seed in the transcript runner (seems like best but hardest option)

pchiusano:
Ah, I like option 4 and it might not be too bad actually - the GUID is chosen during parsing, using the parsing environment which currently I think is created in a hardcoded way but seems like a straightforward refactoring to make that customizable, and then the transcript runner can use a deterministic way of picking those GUIDs.

However, for this PR I suggest we do one of the other workarounds and then just create a separate ticket to track the other thing. I don't think usage of `unique` is really essential to the transcript used here. @runarorama can you tweak the transcript and then I think we can merge.

PR #896, opened 2019-10-26
`&&` and `||` now have precedence over infix applications (Close #895)

### LINKED ISSUE #895: Operator precedence for `&&` and `||` doesn't work
The `&&` and `||` operators currently don't have precedence over other operators/functions.
So this code
```haskell
> 1 == 3 || 3 == 3
```
results in a parse error.
Bildschirmfoto 2019-10-26 um 16 09 32

So to work around that you currently have to wrap sub expressions in paretheses.
```haskell
> (1 == 3) || (3 == 3)
```

noahhaasis:
This implements precedence of `&&` and `||` over other terms (#895).
Previously boolean ops and infix applications were at the same level `and <|> or <|> infixApp`.
Now there's an extra level of indirection.


noahhaasis:
Maybe we should think about precedence later because that definitely complicates things.
Currently the parser can't handle multiple boolean operators in sequence (`true || false && true). This is fixed by #898

aryairani:
What happens when you try to parse that?

noahhaasis:
It doesn't parse without either #898 or this pull request:

Bildschirmfoto 2019-10-28 um 05 48 20

Something like `1 == 3 || 3 == 1` does parse with #898 but results in a type error because it's parsed as `((1 == 3) || 3) == 1`.
This pull request would parse it as `(1 == 3) || (3 == 1)`

aryairani:
Got it.

pchiusano:
Going to close this in favor of a real precedence system to be done later.

I know people will have different opinions on this one, but personally, I kinda like that everything has the same precedence right now. Nothing to remember! The cost/benefit of having a bunch of different operators all with difference precedence has never really paid for itself IMO. Precedence info is sort of "non-local" too: you can't just learn the name of a function and be good to go, you have to learn its name, its precedence, and then how that precedence relates to all the precedence of other operators you might use alongside that one.

I had an idea that precedence should just be a parser thing. Everything has default precedence, but you can just locally specify precedence in a block via a parsing directive, in the same way that like `use` clauses are just a parsing directive. There there's nothing to remember when reading the code, but you also save on parens if that is really a bottleneck. Prettyprinter could also potentially be modified to insert these directives automatically if they would massively cut down on parens.

```
blah x = 
  infix right high +:  -- also, just default, high, and low as precedence levels
  x +: x +: x +: [1,2,3]
```

atacratic:
Recording these views about an operator precedence system here, to try not to lose them...

One option on precedence might be to throw an error on unbracketed arithmetic expressions like 1 + 3 * 2 unless they're accompanied by a local precedence declaration. That would avoid having the usual forgettable pile of built-in precedence rules, but also avoid the 'that's just wrong' behaviour for arithmetic. Also then the education about Unison's system happens at the point of first need. 

The pretty-printer could have some rules for when it prefers to emit a local precedence declaration, vs when it just emits some parentheses. The parentheses might look superfluous sometimes, to people expecting the usual arithmetic operator precedence, but if they just delete them, they'll get the error message to explain.

I've committed almost none of the Haskell operator precedence rules to memory, so I use lots of brackets. That means the rules have negative value to me - they make it hard for me to interpret code written by people who do know them, and they make it obvious to others that my code is written by a n00b 🙃 In Unison the latter issue would go away because the pretty-printer would presumably print code assuming expert-level precedence rule knowledge, but that exacerbates the first issue - now I can't even understand my own code any more.

PR #897, opened 2019-10-26
add old nshash to reflog.verbose message

aryairani:
The "old" hash is needed in order to revert the first action, or to revert to a codebase state that was caused out-of-band.

image

Initially I had just planned to add the "old" hash to the `reflog` message, but once there were two hashes on each line, I  didn't know how to populate the LoopState numbered entries (I'm open to suggestions).  I ended up creating a second `.verbose` variant instead, with old and new hashes, but no numbering:

image

Another option would be for `reflog.verbose` to have numbering, but the number refers to the old hash, 
 which would be the hash adjacent to the number, but it's a little weird.

I tried to reuse logic in `OutputMessages.hs`, but it got pretty hairy in the end.   I could split it into totally separate cases if desired.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -354,11 +354,11 @@ loop = do
         stepManyAtM = Unison.Codebase.Editor.HandleInput.stepManyAtM inputDescription
         updateAtM = Unison.Codebase.Editor.HandleInput.updateAtM inputDescription
       in case input of
-      ShowReflogI -> do
-        entries <- fmap reverse $ eval LoadReflog
-        numberedArgs .=
+      ShowReflogI verbose -> do
+        entries <- fmap reverse $ eval LoadReflog        
+        unless verbose $ numberedArgs .=
           fmap (('#':) . Hash.base32Hexs . Causal.unRawHash . Reflog.to) entries
-        respond . ShowReflog $ fmap (shortenReflogEntry sbhLength) entries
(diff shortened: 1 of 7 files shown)


atacratic:
This is awesome :sunglasses: 

I just thought I'd paste in another idea for what it could look like, FWIW.  (This mockup has no colors sorry!)

```
.> reflog

  Here is a log of the root namespace hashes, starting with the most recent, along with
  the command that got us there. Try:
  
    `fork #6s8qaokmoo .old`   to make an old namespace accessible again,
                              
    `reset-root #6s8qaokmoo`  to reset the root namespace and its history to that of the
                              specified namespace.

  1. #8vh0fc10mj
          ↑     add
  2. #6s8qaokmoo 
          ↑     add
  3. #5l295cgiec 
          ↑     add
  4. #ckk7jmqo24 
          ↑     move.term .base.test.internals.v1.Test.tempAnd .base.test.internals.v1.Test.and
  5. #k8en68aoab 
          ↑     move.term .base.test.internals.v1.Test.&& .base.test.internals.v1.Test.tempAnd
  6. #35k6h1o8li

```
Maybe that would mean no need for a separate verbose variant?  Also I find it guides my eye a bit more to what commands made the transition between which states.

aryairani:
> I just thought I'd paste in another idea for what it could look like, FWIW. (This mockup has no colors sorry!)

I like this idea.  There can be discontinuities in the reflog however, but maybe there's some reasonable display heuristic.

aryairani:
Maybe I'll just try to detect the discontinuities, list them as "external changes", and list an initial branch at the end.

PR #898, opened 2019-10-27
Allow boolean operators to be used in sequence

noahhaasis:
Allows boolean operators to be parsed like infix applications.
`true || false && true` does now parse. Before it would throw a parse error.
Boolean operators and infix applications still have the same precedence


pchiusano:
👍

pchiusano:
Thanks! merging..

Issue #899, opened 2019-10-27
Simplify the Var typeclass


TomasMikula:
The current definition of the `Var` typeclass is

```haskell
class (Show v, ABT.Var v) => Var v where
  typed :: Type -> v
  typeOf :: v -> Type
  freshId :: v -> Word64
  freshenId :: Word64 -> v -> v

data Type
  = User Text
  | Inference InferenceType
  | RefNamed Reference
  | MissingResult
  | AskInfo
  | Blank
  | UnnamedWatch WatchKind Text

data InferenceType =
  Ability | Input | Output |
  PatternPureE | PatternPureV |
  PatternBindE | PatternBindV |
  TypeConstructor | TypeConstructorArg |
  Other
```

which is loaded with many implementation details.

My (intermediate) goal is to simplify it to

```haskell
-- | A class for variables. Laws:
--
--   * `name (named n) == n`
--   * `reset . freshIn vs == reset`
--   * `reset . named == named`
class (Show v, ABT.Var v) => Var v where
  named :: Text -> v
  name :: v -> Text
  reset :: v -> v
```

I have achieved this state and am rolling it out in small self-contained steps (on some of which I will need some input from you). I am opening this issue so that I can refer to the bigger picture from individual PRs.

PR #900, opened 2019-10-27
Don't just assume RefNamed vars not to occur in input to Term.unhashComponent

TomasMikula:
`Term.unhashComponent` used to accept some variables and just assumed that they do not appear in the terms of the component. Such assumption is not captured by the type signature of `unhashComponent`.

To drop the assumption:
 1. let `unhashComponent` make up (`RefNamed`) variables itself instead of taking them as part of input;
 2. freshen the made up `RefNamed` variables appropriately (don't just assume that no one else used `RefNamed` variables).

Add a test that previously would have failed.

---
Also, this is a step towards #899 that will later allow to eliminate `RefNamed` vars altogether.

### DIFF
--- a/parser-typechecker/src/Unison/ABT.hs
+++ b/parser-typechecker/src/Unison/ABT.hs
@@ -12,6 +12,7 @@ module Unison.ABT where
 
 import Unison.Prelude
 
+import Control.Monad.State (MonadState,evalState,get,put)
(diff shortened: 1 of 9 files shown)


TomasMikula:
Rebased and applied the same treatment to `DataDeclaration.unhashComponent`, which appeared in the meantime.

PR #901, opened 2019-10-28
First class links / references as Unison values

pchiusano:
Okay, this is ready for review, should be merged after #889 since it builds on that due to it also needing to do stuff with `Term.dependencies` / `Term.updateDependencies` which were touched by #889.  

## Overview

This is being implemented to support documentation which will be represented as Unison terms and which will embed and/or transclude references to definitions. It will also be generally useful for other stuff too. Basically anytime you need a Unison value representing a reference to a definition in a Unison codebase, these types will be used. (Will also come up in distributed programming API for doing code syncing).

Here's some code showing syntax for creating links, which are first class Unison values. `termLink` and `typeLink` are new keywords (they are NOT functions - you should think of this syntax as being similar to like `classOf` in Scala):

```Haskell
natLink : Link.Type
natLink = typeLink Nat

takeLink : Link.Term
takeLink = termLink List.take

dropLink : Link.Term
dropLink = termLink List.drop
```

Syntactically, you can use any hash qualified identifier after the `termLink` or `typeLink` keyword.

## Implementation notes

Main change is adding these two constructors to the `Term` base functor:

https://github.com/unisonweb/unison/blob/958bd1cce229b065d322cc84bb188f2a0f55b7bf/parser-typechecker/src/Unison/Term.hs#L99-L100

Also adding `Link.Term` and `Link.Type` builtin types:

https://github.com/unisonweb/unison/pull/901/files#diff-8117eb68e95042e5ae9d5831b25d93b7R156-R157

And here's the relevant spot in the parser:

https://github.com/unisonweb/unison/pull/901/files#diff-211d62c2ee7dccd868dc459a4e36136cR75-R92

Also notice [these locations in `Term.updateDependencies`](https://github.com/unisonweb/unison/pull/901/files#diff-fec9994c79b3350a1bbb57c6ff494361R866-R867) and [`Term.generalizedDependencies`](https://github.com/unisonweb/unison/pull/901/files#diff-fec9994c79b3350a1bbb57c6ff494361R826-R827). 

Diff should look a lot cleaner after merge of  #889. 

I think everything else was mechanical. One bummer is that we aren't getting much mileage out of pattern coverage checking due to presence of wildcards in various places. It took me a bunch of tries to track all these down.

## Controversial / interesting decisions

- Separate `Link.Term` and `Link.Type` types. In practice we might typically work with `Either Link.Term Link.Type` or some newtype for that.
- Added `termLink` and `typeLink` as new keywords. Originally was thinking `@term` and `@type` but `type` introduces a layout block in lexer and also `@` is used for as patterns already, so decided against that. Note that syntax of links without documentation blocks could be different, we can bikeshed that separately. :)
- You can `termLink Optional.None`, more generally, you can link to any ctor. Seems like a feature.
- A term depends on the types / terms it links to - this has the effect that if a `Doc` has a link to a definition, `foo` and `foo` is modified, the link will get updated to the new `foo`. I consider this an nice feature (no dead links, links are content addressed and follow renames) but you still have to update the verbiage part of the docs if/when things change. 

## Loose ends

- [x] Think carefully through handling of `Term.dependencies` and `Term.updateDependencies` and update these as appropriate.
- [x] Modify `IR` and runtime to support this new kind of values
- [x] Check equals and ordering implementation in runtime to make sure these handle new kind of values
- [x] New builtin types `Link.Term` and `Link.Type`.
- [x] Typechecker synthesize should handle `TermLink` and `TypeLink`
- [x] Some term level syntax for this - maybe `@term factorial` or `@type IO`?
- [x] Prettyprinter updates to be consistent with this
- [x] Better error reporting if can't resolve the type / term reference
- [x] Test cases, can probably just be done with `.u` / `.ur` files, as it's nothing codebase specific
- [ ] Possible optimization - when linking in definitions before execution, don't need to link in definitions that are just dependencies due to `Term.Link` or `Type.Link`, though it's not harmful to include them. Didn't implement this optimization and I think this will go away as a thing when runtime is more "dynamically linked".


atacratic:
Sounds good 👍

I wonder if you'll add something to convert a `Link.Term (Nat -> Nat)` back into a `Nat -> Nat` from Unison code.

I wonder if a Link.Term 'knows' the names associated with the term, if that question even makes sense. (And will the names be accessible from Unison code. Maybe that's some future `Meta` ability.)

pchiusano:
Note: convinced myself that having a typed version of `Term.Link` not useful. We would like to have more general metaprogramming facility later, where you can work with `Code a`, and the typing info is very relevant, but that's not this...

PR #902, opened 2019-10-28
display synthetic reflog entries for initial and external changes

aryairani:
The reflog works by logging the branch the namespace hash before and after each user change.  The current reflog output just displays the "after" hash, under the assumption that the "before" hash will be available on the previous line (i.e. is the previous "after" hash).  When this assumption doesn't hold (e.g. an external change occurs, or when there is no previous line), we miss seeing the relevant "before" hash.

Current output:
image

This PR output:
image

This PR would close / replace PR #897, which lists "before" and "after" hashes in a new `reflog.verbose` command, which looks nice, but creates a hairy situation wrt LoopState numbering, since each line has two hashes on it.  It also doesn't call out the external change, though it might not be important to do so.

Compare to #897 output:
image

I guess we could have two numbers per line, but it's never been done before!!!
```
  2.  #5bnuu08nik -> 1. #0h32n0ihd3 : add
  4.  #5bnuu08nik -> 3. #0h32n0ihd3 : add
  6.  #5bnuu08nik -> 5. #c3k2p9uhgk : delete.namespace .sheepshead.Suit
```
If we want to do that later, we might borrow from the implementation in #897.  

In any case, I think this is ready to merge.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -355,10 +355,35 @@ loop = do
         updateAtM = Unison.Codebase.Editor.HandleInput.updateAtM inputDescription
       in case input of
       ShowReflogI -> do
-        entries <- fmap reverse $ eval LoadReflog
+        entries <- fmap (convertEntries Nothing []) $ eval LoadReflog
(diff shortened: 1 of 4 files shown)


pchiusano:
LGTM! I'd say merge as soon as Travis passes.

Issue #872, opened 2019-10-17
`ucm` shell should cancel input at `ctrl+c` and quit at `ctrl+d`


matthewess:
Most shell/repl environments have lead me to assume that the behavior of keyboard commands is as follows:
- `ctrl + c` : cancel current input and begin a new input line
- `ctrl + c` during evaluation : kill current evaluation and begin a new input line
- `ctrl + d` : kill shell/repl (EOF) [works as expected in `ucm` but mentioning here for completeness]

Currently, it seems like `ucm` responds to `ctrl + c` by immediately shutting down, rather than trapping the signal. Ideally, `ucm` would handle signals like other shells, and simply swallow `ctrl + c` / `SIGINT` signals to cancel the current input line.

pchiusano:
Something I've noticed is that Ctrl+C works to cancel evaluation and exit back to `ucm` prompt... once. But then if you cancel another evaluation, it exits `ucm`. (Does anyone else notice this behavior?) My suspicion is that the second Ctrl+C actually shows up in Haskell as a different exception than the first one. 

Possibly relevant:

http://neilmitchell.blogspot.com/2015/05/handling-control-c-in-haskell.html

Note: we are using `Control.Exception.try` to catch all errors that occur during evaluation here:

https://github.com/unisonweb/unison/blob/95c9673e6d7309b4687f8365f9c22452ba09a9e9/parser-typechecker/src/Unison/Runtime/Rt1IO.hs#L473-L475

I would have thought that would reliably catch signal-based async exceptions too, but maybe not.

Would love help from someone in tracking this one down, and then once we figure out how to do this trapping reliably we can do the same inside the main loop, even when not evaluating. Also, I notice that GHCi does this properly so that might be a place we could look at to see what it's doing.

matthewess:
I'm planning on taking a look at this!

matthewess:
@pchiusano Playing around on master, it looks like ctrl+c _consistently_ works to cancel the evaluation of presently-running unison code. For example, with
```
delayed : '{IO} ()
delayed = '(delay 10000000)
```
then ```
.> run delayed
^C
.> run delayed
^C
.> run delayed
^C
.>```
works consistently. As far as I can tell, the only issue is that before a command line has been entered, ctrl+c will kill the whole process rather than cancelling input and returning you to the prompt.

matthewess:
followup work in #1131 and #1132

PR #912, opened 2019-10-29
Add POSIX compliant ctrl-c handling

matthewess:
first pass at addressing #872 

based on/cribbed from [ghci](https://gitlab.haskell.org/ghc/ghc/blob/581cbc28e143a4ed8e7f794ed1618161222a5646/libraries/ghci/GHCi/Signals.hs)

the main contribution of this feature is more user-friendly signal handling, particularly making it so that `ucm` responds to SIGINT similarly to other shells, like ghci. if you're typing some input and hit `^c`, you will now cancel that input and be given a new line to type on. you can also cancel execution/evaluation of running code in `ucm` the same way. `^d` will emit an EOF that will kill the shell the way `^c` used to.

this only handles posix signals at the moment, but could be feasibly be extended with the same (CPP-based) windows handling that ghci uses.

i haven't made many contributions to codebases this size, so let me know if some of this would be better added somewhere else or packaged up in a more modular way.


aryairani:
Thanks @matthewess, LGTM.  Didn't notice it was ready for review.

For all of us who haven't looked closely at the issue and maybe haven't tried it out, could you add a comment in the PR to describe:
* How specifically does it change the behavior of `ucm` & how can we tell it's doing what it's expected to: e.g., is there a reproducible case that it handles better?
* The PR description calls it a "first-pass" — what would you say is left to do?  Just the Windows port, eventually, or something more?

Thanks!

mitchellwrosen:
LGTM, I think this is ready to merge once conflicts are resolved.

@aryairani, in a nutshell, it changes the default behavior of the GHC runtime (translate the first SIGINT into a `UserInterrupt` delivered to the main thread) into ghci-like behavior (translate every SIGINT into a `UserInterrupt` delivered to the main thread).

Elsewhere in the commandline main you're catching `UserInterrupt` and re-polling for user input (hence why you currently get the nice re-prompt behavior, but only once).

parser-typechecker/src/Unison/CommandLine/Main.hs
```diff
@@ -184,13 +185,15 @@ main dir initialPath initialInputs startRuntime codebase = do
...
+        (case i of
           h:t -> writeIORef initialInputsRef t >> pure h
           [] ->
             -- Race the user input and file watch.
             Async.race (atomically $ Q.peek eventQueue) getInput >>= \case
               Left _ -> Left <$> atomically (Q.dequeue eventQueue)
-              x      -> pure x
+              x      -> pure x) `catch` interruptHandler
+      interruptHandler (asyncExceptionFromException -> Just UserInterrupt) = awaitInput
+      interruptHandler _ = pure $ Right QuitI
```
mitchellwrosen: I think it would probably be more correct to re-throw non `UserInterrupt` exceptions rather than just quit.

aryairani: Should we wait on this?

mitchellwrosen: I think it could be fixed up post-merge, it's good enuf ;)

matthewess:
added a bit more context but @mitchellwrosen has got it. @aryairani yes, the remaining work is Windows compatibility, I believe. but as mitchell called out there should probably be more delicate handling of non `UserInterrupt` exceptions, as well.

aryairani:
Great!  Thanks!

aryairani:
If you wouldn’t mind creating tickets to track the remaining work, that would be awesome. :)

PR #914, opened 2019-10-30
`Doc` type with some special rendering for it in UCM

pchiusano:
## Overview

This adds a Unison type to be used for writing documentation that links to and/or includes the source of other definitions. It builds on #901 (docs contain links to terms and/or types).

[Here is a transcript demonstrating the usage](https://github.com/unisonweb/unison/blob/topic/docs/unison-src/transcripts/docs.output.md). I would read this first.

## Implementation notes

I marked up the diff below.

* I included an `evaluate` splice for including live examples. It was easy to add and I think it's really cool.
* I added a special-purpose `docs` command even though it's just implemented in terms of `link`/`links` command.
* There's a bit of wrapping of text done on display - anything between newlines is `wrap`'d. I went back and forth on algorithms for this or whether to do it at all. I think this algo might be okay and is easy to change. Basically, avoid hard line breaks unless you want those to show up in the output.
* The `Doc` type is unstructured, basically just text with a few different kinds of things you can splice into that text. I'm still not sure I'm sold on this decision.
 
## Todo

- [x] Helper function for converting a `Doc` to a `Pretty`
- [x] New `display` comand
- [x] Tweak `links` to present a numbered menu of results, similar to `find` results and which can be passed to `display`. As in: `links Doc`, then `display 3`.
- [x] Tip about using `display` on output of `links` command.
- [x] Simplify `Evaluate`, can just be `Evaluate Term.Link`, user can combine `Source`  with `Evaluate` if they wish.
- [x] Lexer changes for docs. Looks like `[: a bunch of text :]`, which is parsed with:
  - [x] `@[source] List.map`
  - [x] `@[include] someOtherDoc`
  - [x] `@[evaluate] examples.List.map`
  - [x] `@[signature] List.map` (just show the signature)
  - [x] `@List.map` - link to another definition
  - [x] Everything else parses as text
- [x] Parser changes for docs
- [x] `display`/ paragraphyText tweaks - currently removes the newline after the opening triple backticks. 
- [x] Some transcript tests / examples
- [x] Do we want to something about rewrapping of text or not? Thought: preserve newlines, segments between newlines gets wrapped in `wrap` function. All splices get a `group` put around them.
- [ ] Give `display` command access to the definitions in the file (tracked in #937)

## Dev journal

* Note to self: we can axe the transclude segment type, transclusion is just evaluation. But need to combine `Doc` and `Segment` types into one.


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -695,19 +698,16 @@ loop = do
...
-            --     types <- filterM (eval . IsType) allRefs
-            --     traverse (\r -> (r,) <$> loadTypeDisplayThing r) types
-            ppe <- prettyPrintEnv =<< makePrintNamesFromLabeled' (deps termDisplays)
-            respond $ DisplayLinks ppe allMd' mempty termDisplays
-          where
-            deps :: Map Reference (DisplayThing (Term v Ann)) -> Set LabeledDependency
-            deps dts = Set.map LD.termRef (Map.keysSet dts)
-                    <> foldMap Term.labeledDependencies
-                               (mapMaybe DT.toMaybe $ Map.elems dts)
+                results :: Set Reference = Set.unions $ Map.elems allMd'
```
pchiusano: Note: simplified logic here - `links` command just shows a menu of results, doesn't print them out. You use `display` or `view` to look at the metadata.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -787,6 +787,34 @@ loop = do
...
+            loadTerm (Reference.DerivedId r) = eval $ LoadTerm r
+            loadTerm r = pure Nothing
+            loadDecl (Reference.DerivedId r) = eval $ LoadType r
+            loadDecl _ = pure Nothing 
+          rendered <- DisplayValues.displayTerm ppe loadTerm loadTypeOfTerm evalTerm loadDecl tm 
+          respond $ DisplayRendered loc rendered
+          -- We set latestFile to be programmatically generated, if we
+          -- are viewing these definitions to a file - this will skip the
+          -- next update for that file (which will happen immediately)
+          latestFile .= ((, True) <$> loc)
```
pchiusano: This is the implementation of the new `display` command.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1533,20 +1561,10 @@ collateReferences (toList -> types) (toList -> terms) =
...
--- Foo, want Foo#123 and Foo#890
--- Foo#1, want Foo#123
--- #567, want bar -- what goes in the SR.name?
--- blah, cat, want blah (with comment about cat)?
-
--- #567 :: Int
--- #567 = +3
+resolveTermName :: Names -> HQ.HashQualified -> Set Referent
+resolveTermName names query = 
+  Names3.lookupHQTerm query names
```
pchiusano: This function turned out a lot simpler than expected... will inline it away. :)

parser-typechecker/src/Unison/CommandLine/DisplayValues.hs
```diff
@@ -0,0 +1,102 @@
...
+    Just typ -> pure $ TypePrinter.prettySignatures ppe [(PPE.termName ppe r, typ)]
+  prettyTerm terms r = case r of
+    Referent.Ref (Reference.Builtin _) -> prettySignature r 
+    Referent.Ref ref -> terms ref >>= \case
+      Nothing -> pure $ "😶  Missing term source for: " <> termName ppe r
+      Just tm -> pure . P.syntaxToColor $ TP.prettyBinding ppe (PPE.termName ppe r) tm
+    Referent.Con r _ _ -> prettyType r 
+  prettyType r = types r >>= \case
+    Nothing -> pure $ "😶  Missing type source for: " <> typeName ppe r
+    Just ty -> pure . P.syntaxToColor $ DP.prettyDecl ppe r (PPE.typeName ppe r) ty
```
pchiusano: This is the function responsible for rendering a `Doc`. Uses the pattern synonyms, so this is actually quite readable.

parser-typechecker/src/Unison/CommandLine/DisplayValues.hs
```diff
@@ -0,0 +1,102 @@
...
+```
+
+@ docs.List.map
+@ pchiusano
+@ licenses.MIT
+List.map f x = 
+  ...
+
+-}
+
```
pchiusano: Some notes to self, will delete.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -676,38 +680,21 @@ loop = do
...
-          where
-            deps :: Map Reference (DisplayThing (Term v Ann)) -> Set LabeledDependency
-            deps dts = Set.map LD.termRef (Map.keysSet dts)
-                    <> foldMap Term.labeledDependencies
-                               (mapMaybe DT.toMaybe $ Map.elems dts)
+      LinksI src mdTypeStr -> getLinks input src (Right mdTypeStr) >>= \case
+        Left e -> respond e
+        Right (ppe, out) -> do
+          numberedArgs .= fmap (HQ.toString . view _1) out
+          respond $ ListOfLinks ppe out
```
pchiusano: Note - moved bulk of `LinksI` implementation out into helper function so it could be used by both `docs` command and `links` commands.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -787,6 +774,15 @@ loop = do
...
             respond $ CantDelete input ppe failed failedDependents
 
+      DisplayI outputLoc s@(HQ.unsafeFromString -> hq) -> do 
+        parseNames <- (`Names3.Names` mempty) <$> basicPrettyPrintNames0
+        let results = Names3.lookupHQTerm hq parseNames 
+        if Set.null results then 
+          respond $ SearchTermsNotFound $ [hq]
+        else if Set.size results > 1 then
+          respond $ TermAmbiguous input (Right hq) results
+        else doDisplay outputLoc parseNames (Set.findMin results)
```
pchiusano: Note - moved most of `display` logic out to `doDisplay` function, which is used by both `display` and `docs` commands.

parser-typechecker/src/Unison/DataDeclaration.hs
```diff
@@ -479,6 +496,24 @@ builtinDataDecls = case hashDecls $ Map.fromList
...
+    , ((), v "Doc.Evaluate", Type.termLink() `arr` var "Doc")
+    , ((), v "Doc.Join", Type.app() (Type.vector()) (var "Doc") `arr` var "Doc")
+    ]
+  link = DataDeclaration
+    (Unique "a5803524366ead2d7f3780871d48771e8142a3b48802f34a96120e230939c46bd5e182fcbe1fa64e9bff9bf741f3c04")
+    ()
+    []
+    [ ((), v "Link.Term", Type.termLink () `arr` var "Link")
+    , ((), v "Link.Type", Type.typeLink () `arr` var "Link")
+    ]
```
pchiusano: I had to declare these types in Haskell code because the parser needs them and it would be a circular dependency to put them in `IOSource` (since `IOSource` needs the parser). It would be possible to abstract the parser somewhat to break this cycle but that was more work than I was willing to do.

parser-typechecker/src/Unison/DataDeclaration.hs
```diff
@@ -487,6 +522,26 @@ pattern TupleType' ts <- (unTupleType -> Just ts)
...
+pattern DocBlobId <- ((== docBlobId) -> True)
+pattern DocLinkId <- ((== docLinkId) -> True)
+pattern DocSourceId <- ((== docSourceId) -> True)
+pattern DocEvaluateId <- ((== docEvaluateId) -> True)
+pattern DocJoinId <- ((== docJoinId) -> True)
+pattern LinkTermId <- ((== linkTermId) -> True)
+pattern LinkTypeId <- ((== linkTypeId) -> True)
+pattern LinkRef <- ((== linkRef) -> True)
+pattern LinkTerm tm <- Term.App' (Term.Constructor' LinkRef LinkTermId) tm 
+pattern LinkType ty <- Term.App' (Term.Constructor' LinkRef LinkTypeId) ty
```
pchiusano: Here's where those pattern synonyms are introduced.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -345,6 +345,9 @@ lexer0 scope rem =
       '?' : c : rem ->
         let end = inc $ inc pos in
         Token (Character c) pos end : goWhitespace l end rem
+      '[' : ':' : rem -> 
+        let end = inc . inc $ pos in
+        Token (Open "[:") pos (inc . inc $ pos) : lexDoc l end rem
```
pchiusano: Here's the lexer change  - `[:` switches to the `lexDoc` state...

parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -274,8 +293,54 @@ termLeaf =
...
+                    l 
+  link'' = either ty t <$> link' where
+    t tok = Term.app (ann tok) 
+                     (Term.constructor (ann tok) DD.linkRef DD.linkTermId) 
+                     (Term.termLink (ann tok) (L.payload tok))
+    ty tok = Term.app (ann tok) 
+                      (Term.constructor (ann tok) DD.linkRef DD.linkTypeId) 
+                      (Term.typeLink (ann tok) (L.payload tok))
+  link = d <$> link'' where
+    d tm = Term.app (ann tm) (Term.constructor (ann tm) DD.docRef DD.docLinkId) tm
```
pchiusano: Term parser changes. Straightforward.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -476,6 +479,55 @@ lexer0 scope rem =
...
+
+    docType :: Pos -> String -> Maybe (Token Lexeme, Pos, String) 
+    docType pos rem = case rem of
+      -- this crazy one liner parses [<stuff>]<whitespace>, as a pattern match
+      '[' : (span (/= ']') -> (typ, ']' : (span isSpace -> (spaces, rem)))) -> 
+         -- advance past [, <typ>, ], <whitespace>
+         let pos' = incBy typ . inc . incBy spaces . inc $ pos in
+         -- the reserved token doesn't include the `[]` chars
+         Just (Token (Reserved typ) (inc pos) (incBy typ . inc $ pos), pos', rem)
+      _ -> Nothing
```
pchiusano: Here's the lexer changes for the inside of those `:]` blocks.

atacratic:
Awesome 😎  This is so nice.

Do you think ideally one day the prettyprinter would somehow special-case the printing of `Doc` literals to produce the `[: :]` markup syntax? I think yes otherwise docs will be hard to edit. How can the prettyprinter spot a Doc literal I wonder, as opposed to something that evaluates to a Doc? Does it need metadata on the top level Doc term? 

I wonder if it would be nice to have an option one day to store transcripts in the codebase as Docs, to get the benefit of nothing breaking over renames.

pchiusano:
Yes you totally could have prettyprinter reverse this and we should track that as an issue. It is pretty straightforward - you just check if the term corresponds to a fully evaluated doc, and if so use the same smart patterns that already exist to pretty-print the doc using the same syntax that the parser supports.

Agree that is important for being able to edit the docs conveniently so that should be high priority!

pchiusano:
There's a weird issue noticed by @atacratic that I'd like to get to the bottom of before merging this one. To reproduce, do `ucm init` with latest released `ucm` version. Then switch to this branch, do `stack build && stack exec unison`. Then `builtins.merge` then `view builtin.Doc`:

```
.> builtins.merge

  Done.

.> ls builtin.Doc

  1. ++        (type missing)
  2. Blob      (type missing)
  3. Evaluate  (type missing)
  4. Join      (type missing)
  5. Link      (type missing)
  6. Signature (type missing)
  7. Source    (type missing)

.> view builtin.Doc

  -- The name builtin.Doc is assigned to the reference #v00j3, which is missing from the codebase.
  Tip: You might need to repair the codebase manually.

.> 
```

pchiusano:
Pushed fix for that weird issue [mentioned above](https://github.com/unisonweb/unison/pull/914#issuecomment-550105001) - it was unrelated to this PR, would have shown up for any new builtin type.

I think this is ready for review / merge. I'd like to address #936 but that could be done separately.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -787,6 +774,15 @@ loop = do
                 (foldMap SR'.labeledDependencies $ failed <> failedDependents)
             respond $ CantDelete input ppe failed failedDependents
 
+      DisplayI outputLoc s@(HQ.unsafeFromString -> hq) -> do 
+        parseNames <- (`Names3.Names` mempty) <$> basicPrettyPrintNames0
+        let results = Names3.lookupHQTerm hq parseNames 
+        if Set.null results then 
+          respond $ SearchTermsNotFound $ [hq]
```
runarorama: Use one fewer `$` here.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2221,3 +2266,13 @@ executePPE unisonFile =
...
+
+loadTypeOfTerm :: Referent -> Action m i v (Maybe (Type v Ann))
+loadTypeOfTerm (Referent.Ref r) = eval $ LoadTypeOfTerm r 
+loadTypeOfTerm (Referent.Con (Reference.DerivedId r) cid _) = do
+  decl <- eval $ LoadType r
+  case decl of
+    Just (either DD.toDataDecl id -> dd) -> pure $ DD.typeOfConstructor dd cid
+    Nothing -> pure Nothing
+loadTypeOfTerm (Referent.Con r cid _) = error $
+  reportBug "924628772" "Attempt to load a type declaration which is a builtin!"
```
runarorama: Where does this bugId come from? Is it just arbitrarily selected?

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
```diff
@@ -81,7 +82,7 @@ data Output v
   | TypeHasFreeVars Input (Type v Ann)
   | TermAlreadyExists Input Path.Split' (Set Referent)
   | TypeAmbiguous Input Path.HQSplit' (Set Reference)
-  | TermAmbiguous Input Path.HQSplit' (Set Referent)
+  | TermAmbiguous Input (Either Path.HQSplit' HQ.HashQualified) (Set Referent)
```
runarorama: Clearly we need to refactor these types.

runarorama:
Is there any enforcement that documentation is of type `Doc`? Is that just lexically enforced?

parser-typechecker/src/Unison/DataDeclaration.hs
aryairani: Maybe into a separate Haskell module if DataDeclaration.hs gets out of control.

parser-typechecker/src/Unison/DataDeclaration.hs
aryairani: Like this. 😄

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
aryairani: `Path.HQSplit'` and `HQ.HashQualified` can overlap.  Did you mean `Either Path.HQSplit' ShortHash`?

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -797,6 +801,29 @@ displayDefinitions' ppe types terms = P.syntaxToColor $ P.sep "\n\n" (prettyType
...
+    prependToFile pp path'
+    pure (message pp path')
+    where
+    prependToFile pp path = do
+      existingContents <- do
+        exists <- doesFileExist path
+        if exists then readFile path
+        else pure ""
+      writeFile path . Text.pack . P.toPlain 80 $
+        P.lines [ pp, "", P.text existingContents ]
```
aryairani: Would it make sense to share this definition with the copy that's local to `displayDefinitions`?  Maybe `scratchAndDisplay` too?

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -1036,6 +1036,17 @@ prettyParseError s = \case
     "I expected a non-empty watch expression and not just \">\""
   go (Parser.UnknownAbilityConstructor tok _referents) = unknownConstructor "ability" tok
   go (Parser.UnknownDataConstructor    tok _referents) = unknownConstructor "data" tok
+  go (Parser.UnknownId               tok referents references) = Pr.lines
+    [ if missing then 
+        "I couldn't resolve the reference " <> style ErrorSite (HQ.toString (L.payload tok)) <> "."
+      else
+        "The reference " <> style ErrorSite (HQ.toString (L.payload tok)) <> " was ambiguous."
```
aryairani: Did you meant to print the referents / references here?

parser-typechecker/src/Unison/TermParser.hs
aryairani: Do you want parser tests for any of these, or are you just catching it in the transcript?

parser-typechecker/src/Unison/Util/ColorText.hs
```diff
@@ -95,6 +96,7 @@ toANSI (AnnotatedText chunks) =
     HiCyan   -> [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Cyan]
     HiWhite  -> [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.White]
     Bold     -> [ANSI.SetConsoleIntensity ANSI.BoldIntensity]
+    Underline -> [ANSI.SetUnderlining ANSI.SingleUnderline]
```
aryairani: I didn't remember ANSI underlining was a thing.  Can you add a screen shot?

parser-typechecker/src/Unison/Util/Pretty.hs
```diff
@@ -141,6 +144,16 @@ wrapImpl (p:ps) = wrap_ . Seq.fromList $
 wrapString :: (LL.ListLike s Char, IsString s) => String -> Pretty s
 wrapString s = wrap (lit $ fromString s)
 
+-- 0. Preserve all leading and trailing whitespace 
+-- 1. Preserve all newlines
+-- 2. Wrap all text in between newlines
+paragraphyText :: (LL.ListLike s Char, IsString s) => Text -> Pretty s
```
aryairani: Great!

parser-typechecker/src/Unison/Util/Pretty.hs
```diff
@@ -341,6 +354,13 @@ numbered
   -> Pretty s
 numbered num ps = column2 (fmap num [1 ..] `zip` toList ps)
 
+numberedColumn2
+  :: (Foldable f, LL.ListLike s Char, IsString s)
+  => (Int -> Pretty s)
+  -> f (Pretty s, Pretty s)
+  -> Pretty s
+numberedColumn2 num ps = numbered num (align $ toList ps)
```
aryairani: How is this different from `numbered`?  Maybe an example in comment?

parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -88,15 +88,22 @@ data Codebase m v a =
...
+-- | Write all of the builtins types and IO types into the codebase. Returns the names of builtins
+-- but DOES NOT add these names to the namespace.
+initializeBuiltinCode :: forall m. Monad m => Codebase m Symbol Parser.Ann -> m Names.Names0
+initializeBuiltinCode c = do
+  let uf = (UF.typecheckedUnisonFile (Map.fromList Builtin.builtinDataDecls)
+                                     (Map.fromList Builtin.builtinEffectDecls)
+                                     mempty mempty)
+  addDefsToCodebase c uf
+  addDefsToCodebase c IOSource.typecheckedFile
+  pure $ Builtin.names0 <> UF.typecheckedToNames0 IOSource.typecheckedFile
```
aryairani: It might be clearer if this didn't do anything with Names?  
```haskell
-- | Write all of the builtin term and type definitions into the codebase.
-- (Doesn't do anything with namespaces.)
writeBuiltinDefinitions :: forall m. Monad m => Codebase m Symbol Parser.Ann -> m ()
```
and then move the return value back into `initializeCodebase` or into a separate top-level definition. (The value isn't derived from earlier execution in this function, so it seems random.)

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
aryairani: (It's fine if you don't want to.)

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
aryairani: I dug into this a bit more.  First, the data in `TermAmbiguous` doesn't ever get displayed, so it's all kinda moot.  Second, the `Either` here is to overload `TermAmbiguous` for a different case when the ambiguous thing you have is a `HashQualified` @runarorama

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
pchiusano: Right

parser-typechecker/src/Unison/Util/Pretty.hs
pchiusano: It's like `column2`, but with numbers next to each.

parser-typechecker/src/Unison/TermParser.hs
pchiusano: Right.

parser-typechecker/src/Unison/PrintError.hs
pchiusano: I deliberately didn't bother but I think it could be nice! It's not too terrible as is, user can always `view` the identifier that's ambiguous.

parser-typechecker/src/Unison/Util/ColorText.hs
pchiusano: It looks like an underline. :) Here's a screenshot https://i.stack.imgur.com/Kw4Yf.png

parser-typechecker/src/Unison/Util/Pretty.hs
pchiusano: added an example in the docs and explanation

parser-typechecker/src/Unison/Codebase.hs
pchiusano: good idea, did this

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
pchiusano: didn't end up doing this one...

pchiusano:
> Is there any enforcement that documentation is of type Doc? Is that just lexically enforced?

1. The parser can only create `Doc` values via the `[:` and `:]` syntax.
2. If you try to `@[include] blah` where `blah` is something that isn't a `Doc` you'd get a regular type error.
3. Docs are associated with defs via the `link` command so in that sense `Doc` is  nothing special, you can link `MyBetterDoc` to your definitions instead. What is special - the `display` command knows how to render `Doc` values, and the `docs` command looks for links of type  `Doc`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
pchiusano: thanks, fixed

pchiusano:
OK MERGING!! Went through the PR comments.

parser-typechecker/src/Unison/Util/ColorText.hs
aryairani: Lol, thanks, but I was interested in a screenshot of the ucm output that uses it — currently the only output samples are the transcript which doesn’t include the ansi sequences.

parser-typechecker/src/Unison/Codebase.hs
aryairani: the remaining `pure ()` is redundant

aryairani:
NBD this time, but I'd generally recommend checking for feedback on the interpretation of the feedback before merging; in this case I don't think there were any explicit approvals.

Good work overall though!

PR #925, opened 2019-11-02
Remove RefNamed variable type

TomasMikula:
`RefNamed` variables used to be assumed fresh, i.e. that no one has inserted the same `RefNamed` variable into the term at hand before.

After even `RefNamed` variables are appropriately freshened (#900, #924, which this PR builds on), there is no more need for the special variable type—an appropriately freshened ordinary variable is sufficient for all purposes where `RefNamed` variable was used previously.

---


pchiusano:
Cool!

PR #932, opened 2019-11-04
Fix a bug in isSubtype

TomasMikula:
Checking that

```
∀a. a  <:  a_2
```

where `a_2` is a variable with name `"a"` and id `2` (not a variable with name `"a_2"`) used to incorrectly return `false`.

The issue was that `Context.freshenVar` could have returned as fresh a variable that was already present in the context.

The implemented fix is to _"reserve"_ variables before adding them to the context if the variables were not themselves returned by `freshenVar`. Reserving a variable ensures it can't be returned from `freshenVar` later.

The actual fix is 5 lines of code: the line `reserveAll (TypeVar.underlying <$> vars)` in `isSubtype'` plus 4 lines of `reserveAll` implementation. However, it was quite hard to track down the issue. Therefore, I took additional measures to detect such bugs earlier. The commits go as follows:
 1. Add a pending test demonstrating the issue.
 2. Instead of `isSubtype` returning `false`, make the compiler crash because of creating an ill-formed context (#887 bearing fruit, after a small tweak 😎). Had to also tweak `Context.succeeds` to not swallow `CompilerBug`s.
 3. Be even more cautious: even if the context extension does not immediately result in an ill-formed context, crash if the added vars are not "reserved".
 4. Implement the fix.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -467,6 +467,13 @@ universals = universalVars . info
 existentials :: Ord v => Context v loc -> Set v
 existentials = existentialVars . info
 
+-- | "Reserves" the given variables in this typechecking environment,
+-- i.e. ensures that they won't be returned from `freshenVar` as fresh.
+reserveAll :: (Var v, Foldable t) => t v -> M v loc ()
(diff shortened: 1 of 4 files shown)


pchiusano:
Wow, good catch! Thanks @TomasMikula

PR #935, opened 2019-11-05
support remote Unison paths in push/pull

### LINKED ISSUE #931: push/pull should support specifying a remote path
Tracking in #935.

aryairani:
This PR closes #931 by updating the `push` and `pull` commands to accept an expanded git repo url format:

```
pull <url>[:gitbranch][:[#shortbranchhash][.path]]
push <url>[:gitbranch][:.path]
```
e.g.
```
.> pull http://github.com/runarorama/unison:.multiset .libs.runar.multiset
```
or
```
.multiset> push git@github.com:runarorama.unison:.multiset
```
or
```
.> pull http://github.com/runarorama/unison:#a123c41.multiset .libs.runar.oldMultiset
```

I also added some prefix autocompletes for the `push` and `pull` command url arguments; so `pull gh<tab>` will expand to `pull https://github.com/` and you can then finish typing the `org/repo` and hit enter.  Here are the abbreviations that are currently supported (the 's' suffix means 'scp-style'):
https://github.com/unisonweb/unison/blob/f42c95826255925739a53e5253dc181d337d1b29/parser-typechecker/src/Unison/CommandLine/InputPatterns.hs#L1082-L1090

```
.> help push

  push
  The `push` command merges a local namespace into a remote
  namespace.
  `push remote local`  merges the contents of the local namespace
                       `local` into the remote namespace `remote`.
  `push remote`        publishes the current namespace into the
                       remote namespace `remote`
  `push`               publishes the current namespace into the
                       remote namespace configured in `.unisonConfig`
                       with the key `GitUrl.ns` where `ns` is the
                       current namespace

  where `remote` is a git repository, optionally followed by `:` and
  an absolute remote path, such as:
    https://github.com/org/repo
    https://github.com/org/repo:.some.remote.path

.> help pull

  pull
  The `pull` command merges a remote namespace into a local
  namespace.
  `pull remote local`  merges the remote namespace `remote` into the
                       local namespace `local`.
  `pull remote`        merges the remote namespace `remote` into the
                       current namespace
  `push`               merges the remote namespace configured in
                       `.unisonConfig` with the key `GitUrl.ns` where
                       `ns` is the current namespace, into the
                       current namespace

  where `remote` is a git repository, optionally followed by `:` and
  an absolute remote path, such as:
    https://github.com/org/repo
    https://github.com/org/repo:.some.remote.path
```

### Details: 
* Added a [`UriParser` module](https://github.com/unisonweb/unison/blob/topic/931-push-pull-remote-path/parser-typechecker/src/Unison/Codebase/Editor/UriParser.hs) which parses most git addresses with an optional `[:gitbranch][:[#shortbranchhash][.path]]` attached to them.  It's used by the updated `push`/`pull` inputparsers, and also when reading `GitUrl.` entries from `.unisonConfig`.

* Added the `LoadRemoteShortBranch` command, which searching a repo for a `ShortBranchHash` and returns the branch if there is a unique match; otherwise it returns `NoRemoteNamespaceWithHash` or `RemoteNamespaceHashAmbiguous`.

* Modified `RemoteRepo.GitRepo` to make the treeish `Maybe Text` instead of testing for `Text.null`, and removed the default of `"master"` because a rare repo may have a different default branch than that.

* Delegated `Git.pullGitRootBranch` to a new `Git.pullGitBranch` which takes a `Maybe ShortBranchHash`.  It will load a unique matching Branch (or one of the failures above), or the `_head` branch when passed `Nothing`.

* Reworked the `PullRemoteBranchI`/`loadRemoteBranchAt` and `PushRemoteBranchI` implementations to use a supplied remote unison path.  `syncRemoteRootBranch` (the push counterpart to `loadRemoteBranchAt` didn't need changes, because we ultimately are syncing a new root back to the remote repo, with its children modified.

### Notes:
* I reworked `Git.hs` to use `git -C` instead of `System.Directory.{get,set}CurrentDirectory`, using these helper functions:
```haskell
gitIn :: FilePath -> [Text] -> IO ()
gitTextIn :: FilePath -> [Text] -> IO Text
```
- You'll get an error if you configure a `GitUrl.` config entry that hard-codes a unison branch hash; the error messages tells you why you are bad and should feel bad.
- skipped #940 for now
- git branches for Unison codebases can't start with `#` or `.` (I think that's fine).
- I didn't add a transcript test, because it uses the network.

### Controversial?
* I think there's a case to be made for dropping support for git branches.  This would also allow Unison paths to be provided without the leading dot.  WDYT?
* I didn't document git branch syntax or read-only namespace syntax in the `help`, I think there's a case for leaving them out of the simple story.  Only caveat in my mind is how easy it might be to leave off the `.` and reference a git branch by mistake, especially when you don't even know that's a possibility.


runarorama:
Please do simplify the help if you can. There's way too much here! Maybe reference a "topic" for the git url help. Like "see `help giturl` for more information on the format of these URLs"

aryairani:
@runarorama Yeah — I didn't really _see_ the help until I made the screenshot for the PR summary.

aryairani:
I updated the help (shown in the summary).  I'm inclined to drop support for git branches altogether, so that the remote unison path can be specified without the leading `.` (and avoid accidentally getting a git branch instead of a remote namespace if the dot is accidentally omitted).  WDYT?

aryairani:
The help text is still a little busted, pending fixes or workarounds for #949, #950.

pchiusano:
Can try working around the printing error with `fmap P.grouped . P.lines`

pchiusano:
@aryairani LGTM, am excited to try this out!

aryairani:
I guess this was merged as a "squash" commit, which creates a more beautiful point object in the git log (bf631d2fcdbe43d2f9b59a5eda518eb2832bdec6), but it leaves the original branch (89b989aa87e03ae47cb956adf5e32f43c93d3681 / ad30ff98748a69a4c319a423326160c4d294dc35) orphaned in the Git log, with no obvious indication that they've been merged and are safe to delete.  (In this case, approximately 2 hrs to rediscover.)

I recommend we not use "squash" for stuff that's already been published (i.e. it's okay to use before you make your git history public, and then only briefly afterwards).  It's easy enough to make the git history appear equivalently beautiful just using git client display configuration: "Show First Parent Only" in SourceTree, / `git log --first-parent` at the command line.

aryairani:
Also, I had published intermediate commit hashes in the docs as a way to dig up a deleted implementation if we needed it again:
image
they are not part of the squashed commit at all.  😅 So even though the active implementation would survive if I delete the old orphaned branch, the referenced history wouldn't, so I guess I need to remember not to delete it, but don't have an easy way to do that.

Maybe I can still merge it into master without too much trouble.

PR #948, opened 2019-11-08
Fix #947

### LINKED ISSUE #947: Subtype check problem in typechecker
This program should not typecheck, because `forall e . Nat ->{Log,e} Nat` is not a subtype of `forall e . Nat ->{e} Nat`:

```
ability Log where
  log : Text -> ()

foo : (∀ e . Nat ->{e} ()) -> Nat
foo _ = 0

bar : (∀ e . Nat ->{e,Log} ()) -> Nat
bar f = foo f
```

But it does typecheck currently, which is wrong!

pchiusano:
This was an amazing bug. Given an ability check for `{e, Other}`, and ambient of `{e2}`, we were bailing with a successful ability check after just the first `e` request, but not even checking the `Other` request against ambient. `e` had to be an unsolved existential to trigger this.

See the note below.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -1555,12 +1555,13 @@ abilityCheck' ambient0 requested0 = go ambient0 requested0 where
       -- 2b. If no:
       Nothing -> case r of
         -- It's an unsolved existential, instantiate it to all of ambient
-        Type.Existential' b v ->
+        Type.Existential' b v -> do
           let et2 = Type.effects (loc r) ambient
           -- instantiate it to `{}` if can't cover all of ambient
-          in instantiateR et2 b v `orElse`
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -1555,12 +1555,13 @@ abilityCheck' ambient0 requested0 = go ambient0 requested0 where
...
+        Type.Existential' b v -> do
           let et2 = Type.effects (loc r) ambient
           -- instantiate it to `{}` if can't cover all of ambient
-          in instantiateR et2 b v `orElse`
-             instantiateR (Type.effects (loc r) []) b v `orElse`
-             die1
+          instantiateR et2 b v 
+            `orElse` instantiateR (Type.effects (loc r) []) b v 
+            `orElse` die1
+          go ambient rs
```
pchiusano: This recursive call `go ambient rs` was straight up missing. `rs` is the remainder of the requested abilities.

PR #952, opened 2019-11-08
Quality-of-life improvements for type refactoring

### LINKED ISSUE #906: Confusing message during type upgrades.
After an `update` to a type, UCM displays the message "that patch contains external dependents". It's not clear what this means, or if this is a problem.

runarorama:
Adds `todo` output to `numberedArgs` so we can refer to them by number.

Fixes #906

Depends on #946


pchiusano:
@runarorama I wanna give this one more thought. Current logic is not correct (it's too strict I think, and the error message isn't good), but just having the patch application silently noop in the cases where it would fail is no good either.

aryairani:
Also we’d want to rebase e466 onto the new #946 or probably really onto master, but I too would like to remember what that message was about first.  We also have to be careful not to lose the refactor in 14bb, but I’m guessing the add and delete of the transcript weren’t meant to be here.

runarorama:
The logic is already nooping before this change. This change just makes it silently noop instead of noisily.

The message can occur in two cases:

1. Something outside the current namespace (the scope of the propagation) is referencing the thing that the propagation is updating. This should be a common occurrence with things like utility libraries.
2. You've deleted some of the names for things that the propagation algorithm is updating. This happens every time you update a type, for example (the old data constructors are in the frontier, and they depend on the old type, which no longer has a name).

I suggest merging this for now and opening a separate issue to track the work of thinking about whether the propagation should do anything special in these (or other related) cases.

aryairani:
Sounds okay to me.

PR #953, opened 2019-11-08
Remove Var type Blank

TomasMikula:
Its only use can be safely replaced by `Var.named "_"`.

---


pchiusano:
Closing, see: https://github.com/unisonweb/unison/pull/956#issuecomment-563647343

PR #954, opened 2019-11-08
Make MissingResult a Term instead of a kind of variable

TomasMikula:
`MissingResult` has nothing to do with variables. `Var` operations like freshening do not make sense for it.

Strictly speaking, `MissingResult` should not be a `Term` either, at least not the term that represents a program stored in the codebase. It could be part of a term that is the output of parser and input to typechecker, which should be a different type than the `Term` stored in the codebase. We don't have this distinction at the moment.

Still, this is an improvement in the following sense: `Term` depends on `Var` anyway, so any accidental garbage in `Var` is a burden for `Term` anyway. Moving `MissingResult` _up_ to `Term` makes it a little less entrenched.

Note: Later, I would like to clean up `Term` from implementation details of the typechecker as well. In this pass, though, I focus on `Var`.


TomasMikula:
Some [previous discussion](https://github.com/unisonweb/unison/pull/888#issuecomment-547060476) about `AskInfo` applies equally to `MissingResult`. Here is the relevant part (`s/AskInfo/MissingResult/`):

@pchiusano:
> Adding it to the term functor means it's part of the codebase format

@TomasMikula:
> It is treated the same as `Blank` - an error is thrown if someone tries to serialize a term with `AskInfo`. Codebase format is not affected by this PR.

@pchiusano:
> Right now, all the variable type info is discarded (only the name is retained) so adding or removing variable types doesn't affect the codebase format

@TomasMikula:
> Still, adding a term that contained `AskInfo` variable to the codebase would most likely be a bug that would go uncaught. Crashing with an error is better IMO.

pchiusano:
Sorry for the long delay on this one. I don’t want to merge this or any other PRs that adds other cases to the Term base functor, unless those cases are really part of the language. 

The right thing to do, which is a much bigger project, would be to have a separate parse tree and syntax tree (ideally sharing some code). The parse tree would have things like Blank and MissingResult, the AST would not. That might be worth brainstorming on in a separate ticket if you like.

PR #955, opened 2019-11-09
Remove Var type UnnamedWatch.

TomasMikula:
An appropriately named ordinary variable is sufficient.

This is a step towards [simpifying the `Var` typeclass](https://github.com/unisonweb/unison/issues/899).

PR #956, opened 2019-11-09
Simplify the Var typeclass

### LINKED ISSUE #899: Simplify the Var typeclass

TomasMikula:
Resolves #899. `Var` typeclass is simplified to

```haskell
class (Show v, ABT.Var v) => Var v where
  named :: Text -> v
  name :: v -> Text
  reset :: v -> v
```

This includes #926 , #953, #954, #955 and a few other things:
 - Remove `Var` type `Inference`.
   - Ordinary variable with a non-ordinary name is just as good.
 - Remove `Var.freshenId`.
   - Combination of `freshIn` and `reset` turns out to be sufficient for all purposes.
 - Remove `Var.freshId`.
   - Turned out to be no longer used.


pchiusano:
@TomasMikula Hmm. I prefer tracking the variable type in an explicit way and not just using naming conventions. Though we aren't doing this now, the variable type info can also be used for generating richer error messages (like a type mismatch between an inferred type variable and something else could get treated differently).

I could maybe see using a multiparameter typeclass, something like `class Var typ v`. But I think that's overkill and probably wouldn't bother with that route either.

Basically, I think it is okay that `Var` is a more Unison-specific typeclass. 

If you have compelling arguments for this one I'm open to hearing them. :) Otherwise I'd probably close this one out.

TomasMikula:
> the variable type info can also be used for generating richer error messages (like a type mismatch between an inferred type variable and something else could get treated differently)

That's all good, though I think something like that should be done _locally_ to the aspect of Unison it relates to (typechecker in your example), instead of adding complexity to `Var` for everyone. Typechecker, for example, is already locally using `data TypeVar b v = Universal v | Existential b v` and `instance Var v => Var (TypeVar b v)`. A similar datatype could be used to distinguish between user-provided and inferred variables, and reporting of type errors could work with terms still lifted to this custom type of variables. Also, `Var` instances for more custom datatypes are admissible with this simplified typeclass, because each variable no longer has to fit into one of the predefined types.

Another compelling argument is the reduction of cognitive load. When I started looking at the source code, early on I asked: _"So what is a `Var`?"_ The answer

```haskell
class (Show v, ABT.Var v) => Var v where
  typed :: Type -> v
  typeOf :: v -> Type
  freshId :: v -> Word64
  freshenId :: Word64 -> v -> v

data Type
  = User Text
  | Inference InferenceType
  | RefNamed Reference
  | MissingResult
  | AskInfo
  | Blank
  | UnnamedWatch WatchKind Text

data InferenceType =
  Ability | Input | Output |
  PatternPureE | PatternPureV |
  PatternBindE | PatternBindV |
  TypeConstructor | TypeConstructorArg |
  Other
```

was just daunting. And even if in the end you find out that you don't need to care about most of those things most of the time, it takes a lot of work to realize that you don't care. So I thought, if I was going to cut through this, I might as well simplify things for those who come after me.

Finally, this is not an end in itself. I originally wanted (and still want) to implement Λ-abstraction and type application. For that, I would argue (and will when the time comes), it would be most natural if term and type variables lived in the same ABT. For that, if we want to keep separate namespaces for term and type variables, some changes to `Var` are required. I just could not add more things to `Var` with any confidence of correctness, or in good conscience 😉.

TomasMikula:
Now that there is the [`unison-core` package](https://github.com/unisonweb/unison/pull/1022), it is even more desirable to be able to talk about programs without referring to syntax or type inference.

pchiusano:
Hi @TomasMikula apologies for the long delay. I’m going to close this and suggest that if you’re interested in continuing to pursue, we open an issue where we workshop possible approaches to breaking up the Var class and what the benefits and drawbacks are. I’d like to get totally aligned on that before we do any more PRs in this area.

In general I think it’s a good idea to have some workshopping in advance for refactorings to things like Var which are used pervasively throughout the compiler. The discussion can give type signatures and whatnot so everyone can get a feel for the refactoring without anyone having to modify any code. 😀

I think part of the reason this PR has languished for a long time is that reviewing it actually is difficult and time consuming (because of how many things it touches), and if we aren’t totally on the same page about the changes then I find it’s hard to really give feedback to get it to a mergeable state.

aryairani:
I still want to pursue simplifying / cleaning up this fundamental stuff, reducing cognitive load, even if it means [revisiting the parser/typechecker](https://github.com/unisonweb/unison/pull/954#issuecomment-563612393).  

How can we get this going again?  😅 

@mitchellwrosen and I have been pondering the question "What is a `Var`?" as part of #1095, and I guess I still don't know the answer to my complete satisfaction.  #499 is a great opportunity to simplify the parser.

I know these refactors isn't our top priority, but I hate to lose the momentum too.

PR #959, opened 2019-11-10
Make Term and Type not depend on TypeVar

TomasMikula:
`TypeVar` is just an auxiliary data type for typechecking. `Term` and `Type` should not depend on it.

Also, move it to the `Typechecker` module.


TomasMikula:
After resolving conflicts, this now moves `TypeVar` back to `parser-typechecker` package.

TomasMikula:
Don't know why GitHub hasn't picked up the result from Travis.

pchiusano:
This looks good, I’m going to try closing and reopening to kickstart Travis (anyone know a better way to achieve that?).

pchiusano:
Great, thanks, merging.

PR #961, opened 2019-11-12
Propagate always succeeds

runarorama:


parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -115,175 +115,160 @@ propagate patch b = case validatePatch patch of
...
-                                   newTerms'
-                                   typeEdits
-                                   typeReplacements
-                                   newTypes
-                    , seen'
+    order <- sortDependentsGraph initialDirty
+    let
+      getOrdered :: Set Reference -> Map Int Reference
+      getOrdered rs =
+        Map.fromList [ (i, r) | r <- toList rs, Just i <- [Map.lookup r order] ]
```
pchiusano: @runarorama is this the code that ends up filtering the updates to just this branch? Can you explain why it works? As I was reading it wasn't clear to me that `getOrdered` will only return a submap of the references which are found in the current branch.

pchiusano:
@runarorama LGTM but can you answer that one question above, and maybe add a comment in the code to clarify?

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
runarorama: Looks like the filtering actually happens here:

https://github.com/unisonweb/unison/blob/e62a1809fe10889573419eb3e2817d3edc290eb0/parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs#L401

runarorama: So I think what ends up happening is that the patch is populated with all the dependents in the codebase, then the patch gets applied to the current branch.

runarorama: Since this behaviour is a little weird, I've made a change so that they patch only considers the current branch.

pchiusano:
@runarorama cool - can you confirm that `todo` doesn't need revisiting as well? it computes the transitive dependents of the patch for reporting the "you have N dependents left to upgrade" stuff

PR #962, opened 2019-11-12
Create a `flush` builtin

sullyj3:
```
use .builtin.io IO

prog : '{IO} ()
prog = 'let
  use base.Text ++
  print "> "
  flush stdout
  line = !readLine
  printLine ("You entered '" ++ line ++ "'")

print t = putText stdout t
```

Todo: probably need to add a test or something? Will have a look tomorrow.


sullyj3:
Looking around, I'm assuming I should just add a test transcript?

sullyj3:
See #974 974

PR #963, opened 2019-11-12
Parsing and pretty-printing use shortest unique suffix to identify definitions

### LINKED ISSUE #552: Allow any unique suffix to identify a definition
Example: you can use `Optional` unqualified, without an import, as long as there's no other `Optional` type in your tree. 

If you've got two `Optional` types, `.foo.bar.Optional` and `.qux.zing.Optional`, then `zing.Optional` is a suffix of name segments that uniquely identifies the type you want and can be used without any imports.

I think this goes a long way. Like no need to `use Optional None Some`, since `None` and `Some` are unique suffixes (as are many constructor names for other data types).

More generally, you can refer to a thing without imports using any suffix of name segments that has just one corresponding `Reference`/`Referent`.

Idea for implementation: a function `Names0 -> Names0`. Any referent `r` with just one name gets all the suffixes of that name inserted as aliases for `r` as long as these suffixes aren't taken. Example, if `foo.bar.baz` uniquely maps to `r`, we also insert `(bar.baz, r)` and `(baz, r)`. If `baz` or `bar.baz` already exist as mappings, we skip inserting these aliases.

pchiusano:
This closes #552 and massively cuts down on import boilerplate, both when parsing and when prettyprinting. Also any of the commands that do parsing (like type-based search) benefit from this as well. Overall a nice quality of life improvement.

[Here's an example](https://github.com/unisonweb/unison/blob/fa780146fdf5e09a2b49cdfa8a970a2adf1b4df5/unison-src/tests/suffix-resolve.u) that I added as a test case and [here is a transcript](https://github.com/unisonweb/unison/blob/00d0fc8c56d36ffb1c12cb314139b5fb0dd1bb03/unison-src/transcripts/suffixes.output.md).

## Interesting decisions

* The `names` and `find` commands show full names. So if you are looking at some code and want to know the full name for something, just pass that name to one of those commands.
* `view` shows full names for the declaration being shown. So if you do `view foo.bar.baz` it shows up as `foo.bar.baz = ...` and not `baz = ...` even if `baz` happens to be a unique suffix. This seemed like the right call.

## Implementation notes:

See markup below. Changes in `HandleInput` and `OutputMessages` are plumbing changes and not very interesting.

### DIFF
--- a/parser-typechecker/src/Unison/Name.hs
+++ b/parser-typechecker/src/Unison/Name.hs
@@ -11,6 +11,7 @@ module Unison.Name
   , sortNamed
   , stripNamePrefix
   , stripPrefixes
+  , suffixes
   , toString
   , toVar
   , unqualified
@@ -29,7 +30,7 @@ import qualified Unison.Hashable               as H
 import           Unison.Var                     ( Var )
 import qualified Unison.Var                    as Var
 import qualified Data.RFC5051                  as RFC5051
-import           Data.List                      ( sortBy )
+import           Data.List                      ( sortBy, tails )
 
 newtype Name = Name { toText :: Text } deriving (Eq, Ord)
 
@@ -102,6 +103,12 @@ parent (Name txt) = case unsnoc (Text.splitOn "." txt) of
   Nothing -> Nothing
   Just ([],_) -> Nothing
   Just (init,_) -> Just $ Name (Text.intercalate "." init)
+
+suffixes :: Name -> [Name]
+suffixes (Name n) = 
+  fmap up . tails . dropWhile (== "") $ Text.splitOn "." n
+  where
+  up ns = Name (Text.intercalate "." ns)
 
 unqualified' :: Text -> Text
 unqualified' = last . Text.splitOn "."
--- a/parser-typechecker/src/Unison/Names3.hs
+++ b/parser-typechecker/src/Unison/Names3.hs
@@ -41,14 +41,17 @@ type ResolutionResult v a r = Either (Seq (ResolutionFailure v a)) r
 -- of that name [[foo.bar.baz], [bar.baz], [baz]]. Insert these suffixes
 -- into a multimap map along with their corresponding refs. Any suffix
 -- which is unique is added as an entry to `ns`.
-suffixify :: Names0 -> Names0
-suffixify ns = ns <> suffixNs 
+suffixify0 :: Names0 -> Names0
+suffixify0 ns = ns <> suffixNs 
   where
   suffixNs = names0 (R.fromList uniqueTerms) (R.fromList uniqueTypes) 
   terms = List.multimap [ (n,ref) | (n0,ref) <- R.toList (terms0 ns), n <- Name.suffixes n0 ]
   types = List.multimap [ (n,ref) | (n0,ref) <- R.toList (types0 ns), n <- Name.suffixes n0 ]
   uniqueTerms = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList terms ]
   uniqueTypes = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList types ]
+
+suffixify :: Names -> Names
+suffixify ns = Names (suffixify0 (currentNames ns)) (oldNames ns)
 
 filterTypes :: (Name -> Bool) -> Names0 -> Names0
 filterTypes = Unison.Names2.filterTypes
--- a/parser-typechecker/src/Unison/Names3.hs
+++ b/parser-typechecker/src/Unison/Names3.hs
@@ -1,10 +1,12 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE ViewPatterns #-}
 
 module Unison.Names3 where
 
 import Unison.Prelude
 
+import Data.List.Extra (nubOrd)
 import Unison.HashQualified (HashQualified)
 import qualified Unison.HashQualified as HQ
 import qualified Unison.HashQualified' as HQ'
@@ -45,8 +47,8 @@ suffixify ns = ns <> suffixNs
   suffixNs = names0 (R.fromList uniqueTerms) (R.fromList uniqueTypes) 
   terms = List.multimap [ (n,ref) | (n0,ref) <- R.toList (terms0 ns), n <- Name.suffixes n0 ]
   types = List.multimap [ (n,ref) | (n0,ref) <- R.toList (types0 ns), n <- Name.suffixes n0 ]
-  uniqueTerms = [ (n,ref) | (n, [ref]) <- Map.toList terms ]
-  uniqueTypes = [ (n,ref) | (n, [ref]) <- Map.toList types ]
+  uniqueTerms = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList terms ]
+  uniqueTypes = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList types ]
 
 filterTypes :: (Name -> Bool) -> Names0 -> Names0
 filterTypes = Unison.Names2.filterTypes
--- a/parser-typechecker/src/Unison/Names3.hs
+++ b/parser-typechecker/src/Unison/Names3.hs
@@ -15,9 +15,11 @@ import Unison.Referent (Referent)
 import Unison.Referent as Referent
 import Unison.Util.Relation (Relation)
 import qualified Data.Set as Set
+import qualified Data.Map as Map
 import qualified Unison.Name as Name
 import qualified Unison.Names2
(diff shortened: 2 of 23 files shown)


parser-typechecker/src/Unison/Names3.hs
```diff
@@ -33,6 +37,22 @@ data ResolutionFailure v a
...
+-- into a multimap map along with their corresponding refs. Any suffix
+-- which is unique is added as an entry to `ns`.
+suffixify0 :: Names0 -> Names0
+suffixify0 ns = ns <> suffixNs 
+  where
+  suffixNs = names0 (R.fromList uniqueTerms) (R.fromList uniqueTypes) 
+  terms = List.multimap [ (n,ref) | (n0,ref) <- R.toList (terms0 ns), n <- Name.suffixes n0 ]
+  types = List.multimap [ (n,ref) | (n0,ref) <- R.toList (types0 ns), n <- Name.suffixes n0 ]
+  uniqueTerms = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList terms ]
+  uniqueTypes = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList types ]
```
pchiusano: This is the main interesting function responsible for adding new name aliases for unique  suffixes. Thought: we may want to move to representing `Names` as a suffix trie or some such if we want this kind of operation to be more efficient. NBD for now though.

parser-typechecker/src/Unison/Parser.hs
```diff
@@ -240,12 +240,13 @@ rootFile :: Ord v => P v a -> P v a
...
 run' :: Ord v => P v a -> String -> String -> ParsingEnv -> Either (Err v) a
-run' p s name =
+run' p s name env =
   let lex = if debug
             then L.lexer name (trace (L.debugLex''' "lexer receives" s) s)
             else L.lexer name s
       pTraced = traceRemainingTokens "parser receives" *> p
-  in runParserT pTraced name (Input lex)
+      env' = env { names = Names.suffixify (names env) } 
+  in runParserT pTraced name (Input lex) env'
```
pchiusano: This was the one spot in the parser I needed to modify to add the "suffixification" logic. We just suffixify the names that are passed in before invoking the parser.

parser-typechecker/src/Unison/PrettyPrintEnv.hs
```diff
@@ -34,10 +36,43 @@ instance Show PrettyPrintEnv where
...
 
 fromNames :: Int -> Names -> PrettyPrintEnv
-fromNames length names = PrettyPrintEnv terms' types' where
-  terms' r = safeHead . Set.map HQ'.toHQ $ (Names.termName length r names)
-  types' r = safeHead . Set.map HQ'.toHQ $ (Names.typeName length r names)
-
+fromNames len names = PrettyPrintEnv terms' types' where
+  terms' r = shortestName . Set.map HQ'.toHQ $ (Names.termName len r names)
+  types' r = shortestName . Set.map HQ'.toHQ $ (Names.typeName len r names)
+  shortestName ns = safeHead $ sortOn (length . HQ.toString) (toList ns)
```
pchiusano: Important: we pick the shortest name that exists if there are multiple rather than (arbitrarily) picking the first name alphabetically.

atacratic:
Nice :smile: I love how this doesn't interact in the code with the use statement insertion in the pretty-printer.  The two just seem to work together nicely.  I've had a play around trying to find some tricky interaction but haven't. 

Code looks a heck of a lot cleaner now!

parser-typechecker/src/Unison/PrettyPrintEnv.hs
aryairani: For `shortestName`, you don't wanna pick `.a.b.c.d` over `computePath` — in the absence of a better metric for semantic closeness, I suggest sorting on the number of segments of the suffix name?  Or I dunno, this is tricky and needs some thought.

parser-typechecker/src/Unison/PrettyPrintEnv.hs
```diff
@@ -34,10 +36,43 @@ instance Show PrettyPrintEnv where
...
+  suffixifiedPPE :: PrettyPrintEnv
+  } deriving Show
+
+-- declarationPPE uses the full name for references that are
+-- part the same cycle as the input reference, used to ensures
+-- recursive definitions are printed properly, for instance:
+--
+-- foo.bar x = foo.bar x
+-- and not
+-- foo.bar x = bar x
```
aryairani: — the full, probably-relative name, right?  Are these always relative names btw?

unison-src/transcripts/docs.output.md
```diff
@@ -153,14 +153,14 @@ Now that documentation is linked to the definition. We can view it if we like:
...
   🔽
-  List.take.ex1 = []
+  ex1 = []
   
   
-  List.take.ex2 = builtin.List.take 2 [1, 2, 3, 4, 5]
+  ex2 = builtin.List.take 2 [1, 2, 3, 4, 5]
   🔽
-  List.take.ex2 = [1, 2]
+  ex2 = [1, 2]
```
aryairani: This looks like the LHS is using the short name when it should use the full name?

unison-src/transcripts/fix942.output.md
```diff
@@ -94,7 +94,7 @@ test> t1 = if z == 3 then [Fail "nooo!!!"] else [Ok "great"]
   
     ⍟ These new definitions are ok to `add`:
     
-      t1 : [builtin.Test.Result]
+      t1 : [Result]
```
aryairani: I have to keep reminding myself that this shortening is only happening because there's only one `Result` type in the namespace. (Right?)

parser-typechecker/src/Unison/Name.hs
```diff
@@ -103,6 +104,12 @@ parent (Name txt) = case unsnoc (Text.splitOn "." txt) of
   Just ([],_) -> Nothing
   Just (init,_) -> Just $ Name (Text.intercalate "." init)
 
+suffixes :: Name -> [Name]
+suffixes (Name n) = 
+  fmap up . tails . dropWhile (== "") $ Text.splitOn "." n
+  where
+  up ns = Name (Text.intercalate "." ns)
+
```
aryairani: This just becomes `tails` with a `[NameSegment]`-based `Name` implelmentation.

parser-typechecker/src/Unison/Names3.hs
```diff
@@ -33,6 +37,22 @@ data ResolutionFailure v a
...
+
+suffixify :: Names -> Names
+suffixify ns = Names (suffixify0 (currentNames ns)) (oldNames ns)
```
aryairani: Do we want to suffixify `oldNames`?

parser-typechecker/src/Unison/PrettyPrintEnv.hs
```diff
@@ -34,10 +36,43 @@ instance Show PrettyPrintEnv where
...
+fromNamesDecl len names = 
+  PrettyPrintEnvDecl (fromNames len names) (fromSuffixNames len names)
+
+-- A pair of PrettyPrintEnvs:
+--   - suffixifiedPPE uses the shortest unique suffix
+--   - unsuffixifiedPPE uses the shortest full name
+data PrettyPrintEnvDecl = PrettyPrintEnvDecl {
+  unsuffixifiedPPE :: PrettyPrintEnv,
+  suffixifiedPPE :: PrettyPrintEnv
+  } deriving Show
```
aryairani: @pchiusano and I discussed looking into an as-of-yet-undetermined `PPE` interface would be clearer to use than the mixed current `PPE` / `PrettyPrintEnvDecl`.  A spec for the tool and pretty printer outputs would probably inform this design, or we could work backwards from the use-site calls we want to the model that neatly supports them.

unison-src/transcripts/docs.output.md
aryairani: This is in the documentation printer.

aryairani:
I'd like to go over and understand https://gist.github.com/atacratic/f8850c1f1872bbe69a15f8ef3efc91c6 together before merging too.

unison-src/transcripts/fix942.output.md
pchiusano: That's right.

unison-src/transcripts/docs.output.md
pchiusano: Yeah, this is a bug... taking a look.

parser-typechecker/src/Unison/Names3.hs
pchiusano: I gave it some thought and decided no.

unison-src/transcripts/docs.output.md
pchiusano: Okay fixed.

Issue #929, opened 2019-11-04
a pushed namespace may be missing dependencies / dependency names unknown at receiver


aryairani:
Option 1: just copy dependencies on push / pull, may be missing names
Option 2: copy dependency names on push into `_deps` namespace
  - `_deps` probably should be some name that "looks" reserved, like `_deps`.
  -  and deprioritize these `_deps` names during view so that some more maintained pull of them is shown instead whenever available (filter these out before pretty printing)
  * we might warn the user that we're going to push definitions with missing dependency names, but that is the common case (e.g. depending on `.base` and no we don't want to special-case it)

Other ideas for how to prioritize name selection
* prioritize according to graph distance from the definition(s) being displayed
* deprioritize for one-offs like `_deps` or `_startsWithUnderscore`
* "priority" metadata on namespace
* prioritize names that have metadata indicating current user had some hand in selecting them
* on a per-namespace basis, give a list of priority namespaces

atacratic:
Worth noting that in the story for PRs in #867, there are lots of copies of names being pulled in to namespaces like `.pulls.base.contributor1.topic`.  So maybe not enough to deprioritize just based on `_` or `_deps`.  

Another variation:
- each name has an 'am I a copy' metadata flag, and non-copies are prioritized
- the user can configure a list of namespaces acting as a priority order for name selection.

aryairani:
Yeah, I'm thinking we should have some heuristic (e.g. something based on the relationship between the current path, the definition being viewed, and the available dependency names) and some optional configuration.

aryairani:
We could have an optional arg that tells ucm how to copy in external dependency names; e.g. "copy everything in <specified namespace> into `dependencies`".  When the provided namespace is omitted or not inclusive enough, we could compute and propose a nearest common ancestor for the user.

PR #981, opened 2019-11-19
Add support for multiple IO types

runarorama:
## Overview 

This adds support for an `IO` type that evolves over time. The issue is that we want `ucm` to be able to execute `IO` programs written with old versions of `base`. This PR establishes a method for upgrading `IO` as follows:

1. Add a new `IOn.hs` with the source of the new version of the `IO` type, where `n` is some number.
2. Add a reference to the source in `IOSource.hs` by adding it to the end of the `ioSources` list.
3. Update `base.io.IO` and push to http://github.com/unisonweb/base

## Implementation notes

The key to making this work is that there's a `Map` of all the `IO` types, indexed by hash. This is probably super inefficient, but it does work. When the runtime in `RT1IO` sees a request with one of the `IO` hashes it knows about, it passes the appropriate `IO` dictionary (a `TypecheckedUnisonFile`) into the handler.

## Controversial decisions

Since the different `IO` types _are different_, they are not compatible with one another. So programs written for different versions of `IO` cannot interoperate (you'll get a cryptic type error at the moment).

The interim solution is that before you call code that uses old versions of `IO`, you need to upgrade it in your codebase by using the latest patch from `base`.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -107,8 +107,8 @@ import qualified Unison.Codebase.TermEdit as TermEdit
 import qualified Unison.Codebase.TypeEdit as TypeEdit
 import qualified Unison.Typechecker as Typechecker
 import qualified Unison.PrettyPrintEnv as PPE
-import           Unison.Runtime.IOSource       ( isTest, ioReference )
-import qualified Unison.Runtime.IOSource as IOSource
+import           Unison.Runtime.IOSources.IOSource ( isTest )
+import qualified Unison.Runtime.IOSources.IOSource as IOSource
 import qualified Unison.Util.Star3             as Star3
 import qualified Unison.Util.Pretty            as P
 import           Unison.Util.Monoid (foldMapM)
@@ -1253,7 +1253,7 @@ loop = do
 
       MergeBuiltinsI -> do
           let names0 = Builtin.names0
-                       <> UF.typecheckedToNames0 IOSource.typecheckedFile
+                       <> foldMap UF.typecheckedToNames0 IOSource.ioSourceFiles
           let b0 = BranchUtil.addFromNames0 names0 Branch.empty0
           let srcb = Branch.one b0
           _ <- updateAtM (Path.consAbsolute "builtin" currentPath') $ \destb ->
@@ -2198,15 +2198,16 @@ basicNames0' = do
   pure (parseNames00, prettyPrintNames00)
 
 -- {IO} ()
-ioUnit :: Ord v => a -> Type v a
-ioUnit a = Type.effect a [Type.ref a ioReference] (Type.ref a DD.unitRef)
+ioUnit :: Ord v => a -> IOSource.SourceFile -> Type v a
+ioUnit a src =
+  Type.effect a [Type.ref a $ IOSource.ioReference src] (Type.ref a DD.unitRef)
 
 -- '{IO} ()
-nullaryMain :: Ord v => a -> Type v a
-nullaryMain a = Type.arrow a (Type.ref a DD.unitRef) (ioUnit a)
+nullaryMain :: Ord v => a -> IOSource.SourceFile -> Type v a
+nullaryMain a src = Type.arrow a (Type.ref a DD.unitRef) (ioUnit a src)
 
 mainTypes :: Ord v => a -> [Type v a]
-mainTypes a = [nullaryMain a]
+mainTypes a = nullaryMain a <$> IOSource.ioSourceFiles
 
 -- Given a typechecked file with a main function called `mainName`
 -- of the type `'{IO} ()`, adds an extra binding which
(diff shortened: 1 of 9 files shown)


unison-src/transcripts/updateio.md
```diff
@@ -0,0 +1,240 @@
...
+testio.foo = '(io.printLine "Hello")
+```
+
+Let's see if it works:
+
+```ucm
+.> run testio.foo
+```
+
+Great! Now let's add an old version of `IO`:
```
aryairani: Is there any reasonable, tidier way of referencing the old supported versions of `.builtin.io`?  I feel like `.builtin` should include names for whatever the `ucm` binary understands; including old versions of IO.

Option 1: make `.builtin` a branch `Branch m`, with its own history, and then reference the abilities hash-qualified?  
Option 2: add old versions of `IO` as `.builtin.io.v1`, `.builtin.io.v2`, ..., and `.builtin.io` includes all of `vN` at the top level, where `vN` is the latest?

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1246,7 +1246,7 @@ loop = do
 
       MergeBuiltinsI -> do
           let names0 = Builtin.names0
-                       <> UF.typecheckedToNames0 IOSource.typecheckedFile
+                       <> foldMap UF.typecheckedToNames0 IOSource.ioSourceFiles
```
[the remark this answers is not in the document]
pchiusano: @runarorama I also wondered about this

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2258,34 +2259,45 @@ addRunMain mainName Nothing = do
...
+                  $ let runMain = DD.forceTerm a a (Term.var a v)
+                    in
+                      UF.typecheckedUnisonFile
+                        (UF.dataDeclarations' uf)
+                        (UF.effectDeclarations' uf)
+                        (  UF.topLevelComponents' uf
+                        <> [(v2, runMain, ) <$> mainTypes a]
+                        )
+                        (UF.watchComponents uf)
+              else Nothing
```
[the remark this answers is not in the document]
pchiusano: Yeah this is not great. Not only is it formatted poorly, in a way that no human would ever format code, but I had to stare at this for a while just to figure out what the actual diff was.

pchiusano:
@runarorama Before reviewing this PR in detail, can we discuss whether we think the runtime needs to support multiple versions of `IO`? What do you think of just doing the following:

* To create a new version of `IO`, just modify `IOSource`. Publish a patch in `.builtin` (or base?) for (partially) converting old `IO` to new `IO`. Do this whenever `IO` changes. Maybe `io.patches.v0tov1`, `io.patches.v1tov2` or whatever convention.
* Write a couple pure Unison functions to handle old `IO` into new `IO`. That is:
  * `io.handleV0 : Request IO#old a ->{IO} a`.
  * `io.runV0 : '{IO#old} a ->{IO} a` (defined as `runV0 io = handle handleV0 in !io`)

So the user might just apply the relevant patch and be done. Otherwise, the user can wrap their top level `main` function(s) that uses old `IO` to first get handled into new `IO`. This does not seem too bad.

Advantage is less machinery on the runtime side and the conversion code is just pure Unison. Minor point you touched on, not really relevant now is there might be some efficiency advantages to not needing to support multiple IO's in the runtime. If you support multiple `IO` then you have to dispatch on which one you have.

unison-src/transcripts/updateio.md
pchiusano: I think `builtin` will include (historical) names for all old versions of `IO`, even without us doing anything. If I'm writing new code, it uses the latest `IO`. If I acquire old code that uses old `IO`, that also pulls in the history which will give the old `IO` nice hash qualified names.

I don't really think old and new IO both need nice names in the current namespace by default, but users could always `alias.type` the old one if they wish.

aryairani:
> @runarorama Before reviewing this PR in detail, can we discuss whether we think the runtime needs to support multiple versions of `IO`? What do you think of just doing the following:
> 
> * To create a new version of `IO`, just modify `IOSource`. Publish a patch in `.builtin` (or base?) for (partially) converting old `IO` to new `IO`. Do this whenever `IO` changes. Maybe `io.patches.v0tov1`, `io.patches.v1tov2` or whatever convention.
> * Write a couple pure Unison functions to handle old `IO` into new `IO`. That is:
>   
>   * `io.handleV0 : Request IO#old a ->{IO} a`.
>   * `io.runV0 : '{IO#old} a ->{IO} a` (defined as `runV0 io = handle handleV0 in !io`)
> 
> So the user might just apply the relevant patch and be done. Otherwise, the user can wrap their top level `main` function(s) that uses old `IO` to first get handled into new `IO`. This does not seem too bad.
> 
> Advantage is less machinery on the runtime side and the conversion code is just pure Unison. Minor point you touched on, not really relevant now is there might be some efficiency advantages to not needing to support multiple IO's in the runtime. If you support multiple `IO` then you have to dispatch on which one you have.

Oof unfortunate, but this sounds better to me too.

aryairani:
Not totally clear on the "patch" part (except that it should go in base), since we don't have a complete story for type refactors (or do we? I don't actually know), but I do like the handler idea, and I imagine chains of known handlers can be JITed away.

I was a little worried about dispatching based on `TypecheckedUnisonFile`; e.g. would the approach still work later when we want to be fast, and if not, should we just come up with something different right away?

unison-src/transcripts/updateio.md
aryairani: (recapping our discussion) Acquiring old code that uses old IO doesn't currently necessarily pull in a historical name for `IO`, though that could change in the future (#929).

runarorama:
Yeah the patch is only going to work for certain programs and not others, since we can’t automatically migrate code that uses patterns or calls the constructors directly.

Given that, does Paul’s idea still work?

runarorama:
For the handlers, we need n such handlers where n is the number of previous IO versions

pchiusano:
Doing a patch upgrade is preferred since it's producing definitions that directly call the right functions. Yeah, it may require manual intervention if you are pattern matching on `IO` or directly calling its constructors (I suspect this will be less common but it's possible). The handler proxying covers the rest of the cases, basically if you don't feel like doing the full upgrade for some pattern-matchy-y `IO` code.

> For the handlers, we need n such handlers where n is the number of previous IO versions

Well, you can just have a handler from the previous version to the latest version. You can get from v1 to vN by composing the 1-step translations, and there can maybe be some [handler fusion stuff](http://hackage.haskell.org/package/fused-effects-1.0.0.0#readme) happening here at a future time that makes this still efficient. Even without this, people can still write direct translations "v3 to v42" on an ad hoc basis, and they could do that at any time and publish as a regular library.

pchiusano:
Okay, after [some discussion in Slack](https://unisonlanguage.slack.com/archives/CLG4P0GSD/p1574375497131600), we decided to close this in favor of the [basic approach outlined above](https://github.com/unisonweb/unison/pull/981#issuecomment-557230976). Summary of discussion: 

To update `IO`:

1. Mutate IOSource
2. Update base generating a patch for `IO`. Use some standard naming convention for these patches.
3. Make a handler from old `IO`  into new `IO` (and optionally a handler in the other direction, maybe partial if new operations are added that have no analogue in old `IO`)
4. base should merge in latest `builtin` I think so it always shares history?

Can get from any IO version to any other by composing these 1 step handlers. A fancier runtime might do some [fusion of effects](http://hackage.haskell.org/package/fused-effects-1.0.0.0#readme). People can write direct translations from version X to version Y on an ad hoc basis, as needed. These are just pure Unison functions, nothing fancy.

One question that came up: what if you update `base` but not your version of Unison? You just won't be able to run any of the new `IO` programs (typechecker won't let you). The error message won't be great, will just be ability check failure. Would be nice if we could signal out of band somehow what version of Unison code is meant to work with.

@aryairani and @runarorama you can pipe in with any extra detail you wanna record here for posterity. :)

PR #984, opened 2019-11-20
messed up stack.yaml as a test -- don't merge

aryairani:
verifying that travis reports a failure if an install step fails


aryairani:
If this fails, then that should mean #983 is ok to merge.

aryairani:
Successfully failed.

PR #988, opened 2019-11-21
resolve.{term,type} => replace.{term,type}

pchiusano:
This PR renames the commands `resolve.term` to `replace.term` and `resolve.type` to `replace.type`. Motivation for this is I'm doing some docs for the refactoring process (as a Unison transcript actually) and want to improve the nomenclature. The "resolve" nomenclature was overly specific to one use case of those commands, so that nomenclature has now been sacked. 😀 You can resolve edit conflicts by using the `replace.*` commands, but you can also just issue `replace` commands to establish that the patch should replace `<old>` with `<new>`. "Replace" is more neutral terminology that still reflects what's happening.

This PR also has a fix: previously, transcripts with invalid commands would just silently noop. Now they fail the transcript as expected. Transcripts and output have been updated too.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -294,11 +294,11 @@ loop = do
           DeleteBranchI opath -> "delete.namespace " <> ops' opath
           DeletePatchI path -> "delete.patch " <> ps' path
           ReplaceTermI srcH targetH p ->
-            "resolve.term " <> SH.toText srcH <> " "
+            "replace.term " <> SH.toText srcH <> " "
                             <> SH.toText targetH <> " "
                             <> opatch p
           ReplaceTypeI srcH targetH p ->
-            "resolve.type " <> SH.toText srcH <> " "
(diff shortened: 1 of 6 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -293,11 +293,11 @@ loop = do
...
-          ResolveTermI srcH targetH p ->
+          ReplaceTermI srcH targetH p ->
             "resolve.term " <> SH.toText srcH <> " "
                             <> SH.toText targetH <> " "
                             <> opatch p
-          ResolveTypeI srcH targetH p ->
+          ReplaceTypeI srcH targetH p ->
             "resolve.type " <> SH.toText srcH <> " "
                             <> SH.toText targetH <> " "
                             <> opatch p
```
aryairani: We want to update the reflog nomenclature too ^

aryairani: I updated in f1ef35d

pchiusano:
@aryairani great, thanks, missed that one. Merging now.

PR #994, opened 2019-11-24
Added pretty-printer support for Doc literals

### LINKED ISSUE #936: Add pretty-printer support for Doc literals
As of #914 Unison accepts code like
```
use .builtin
docs.List.take = [:

`@List.take n xs` returns the first `n` elements of `xs`. 

## Example:

@[source] List.take.ex1
:]
```

But at the moment, doing a view/edit on it returns
```

  docs.List.take : builtin.Doc
  docs.List.take =
    use builtin.Doc Blob
    use builtin.Link Term
    builtin.Doc.Join
      [ Blob "`",
        builtin.Doc.Link (Term termLink builtin.List.take),
        Blob " n xs` returns the first `n` elements of `xs`. \n\n## Example:\n\n",
        builtin.Doc.Source (Term termLink List.take.ex1),
        Blob "" ]

```

We should add support for the `[: :]` sugar to the pretty-printer, so that Docs are output in code as nicely as they are input.  Otherwise they will be hard to edit.  

I will implement this.  (It might be a small handful of weeks before it's done.)

@pchiusano can you give me an extra steer about how I should go about determining in the pretty-printer whether a term corresponds to a fully evaluated doc?  Am I checking that it's composed solely of calls to Doc's constructors (note to self: cf DataDeclaration.pairCtorRef), `[]`, text literals, and uses of termLink?  Any other function call (even a `$`) means we give up a revert to non-sugared printing, at least at that level of the tree.  Right?

(note to self: maybe pre-annotate the tree with the 'render with sugar' decision, to avoid quadratic behaviour in the no-sugar case - maybe overkill?)

atacratic:
Fixes #936.  

### New behavior
[Transcript diffs showing changes for users to be aware of.](https://github.com/unisonweb/unison/pull/994/files#diff-d4af4aa1aef7ad0bc7cedefbe6651266R24)  (I'll PR a docs update to the [langref here](https://www.unisonweb.org/docs/language-reference#documentation-literals) for this, sometime soon.)

[Transcript showing minor fiddly details of the behavior.](https://github.com/unisonweb/unison/pull/994/files#diff-d7f8b990921939b6dabb6c5c37fc5f44R1)

Doing view/edit on a doc now looks like this:

image

This PR also changes how paragraph wrapping for doc literals is handled. Previously, all text was treated as paragraphs to be reflowed/wrapped.  Now it's only text that is not indented with respect to the rest of the doc.  So indenting can be used to cause part of the literal to have its whitespace faithfully preserved - useful for code snippets and tables.  (Code snippets might well be better as Unison terms to be brought in with `@[include]`, but let's not assume everyone will always want to do that.)  

In a similar vein, sequences of multiple spaces within a paragraph are now preserved, rather than being collapsed into one space.  

A bit more detail on the paragraph handling, quoting TermParser.hs:

```haskell
-- Doc literal normalization
--
-- - reduces the whitespace after all newlines so that at least one of the
--   non-initial lines has zero indent (important because the pretty-printer adds
--   indenting when displaying doc literals)
--
-- - removes trailing whitespace from each line
--
-- - removes newlines between any sequence of non-empty zero-indent lines
--   (i.e. undo line-breaking within paragraphs).
--
-- This normalization allows the pretty-printer and doc display code to do
-- indenting, and to do line-wrap of paragraphs, but without the inserted
-- newlines being then frozen into the text for ever more over subsequent
-- edit/update cycles.
```
One minor behavior change is that newlines at the start and end of the doc are now preserved.  Previously they were removed.  (We still remove opening/trailing spaces on the first and last lines though.)  This is because the normalization's unindenting code needs to tell the difference between the case where the first non-whitespace text is sharing a line with the opening `[:`, and the case where it's not (and I didn't want to work out how to plumb this info out of the lexer).  

Finally, and slightly randomly, the PR extends the escaping syntax within doc literals, so you can type `\:]` to get a `:]` inside your doc, instead of ending the doc.  This means its possible to give example doc definitions within a doc.  This isn't a complete and regular story for escaping - you still can't put a `\@` or a `\:]` in your doc, since those are themselves escape sequences.  Probably we don't care (and we might not want to do the obvious thing and make people double up any backslash).  

### Notable implementation stuff

It was more painful than you might think necessary...

Take a look at the [comment above `docNormalize`](https://github.com/unisonweb/unison/pull/994/files#diff-211d62c2ee7dccd868dc459a4e36136cR361) which gives a flavor as to why.  

### Stop reading here -- old notes from draft PR follow

@pchiusano could I get your help with the following snag?

- At the moment a round-trip edit+slurp cycle is not a no-op on docs, because each time extra leading spaces get added.  You can see this in docs.output.md.  Should the doc parser discard n leading spaces, where n is the minimum number of leading spaces following a newline?

- [x] todo - fix this indentation issue
  - [x] debug regressions in docs.output.md
  - [x] get rid of superfluous 'trim initial spaces on first line' in docUnIndent

Notes to self - other tests I still need to do (just as one-offs, not as regression scripts):

- [x] references to symbolic names >> don't parse :(  raised #1054
- [x] escaping of :], if supported; handling of escaped at symbols
- [x] printing non-literal docs without the sugar (i.e. when `isDocLiteral` returns false)
- [x] hit each case of prettyDoc
- [x] various types of reference including data and effect constructors
- [x] impact of nobbling FQN elision on Doc and Link 

Possible discussion points:
- placement of `[:` with respect to the doc text - maybe should always be on its own line?  
- green coloring - I think justified since this is a separate embedded syntax?
- hmm, edit+slurp is not a no-op on some hand-crafted docs, e.g. the following both get rendered as `[:  :]`... Any interesting hand-crafted doc won't render using the sugar anyway though.  Let's live with this...
```
doc = Join [ Blob "", Blob ""]
doc' = Join [ Blob ""] 
```

### Issue with paragraphyText
`Pretty.paragraphyText` reflows paragraphs, which is great, but it also (a) combines any sequence of spaces into one space, and (b) inserts newlines into the output.  (a) means people can't write nested bullets or any other space-sensitive formatting in their docs.  (b) is OK except that after an `edit`, the user's changes to the text are not reflowed, because Unison now sees it as a series of one-line paragraphs rather than one big one.  Probably the answer is to treat double-newlines as paragraph separators.  Probably needs some spec review before coding.  [done]

- [x] fix paragraph reflow to work after `edit` and be less brutal to spacing.
  - [x] handle TODOs in doc-formatting.md and extend tests
  - [x] TODO in termprinter
  - [x] brittany
  - [x] code cleanup and commenting, review diff
  - [x] PR explanation


pchiusano:
@atacratic nice! 

I see, the round trip fails because the doc block is nested in another block so gets indented.

It seems reasonable to trim whitespace from start of each line. It might be just as good / easier to do it by looking for column of first non-whitespace character after the `[:` and strip up to that column for each subsequent line in the block.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -301,9 +322,10 @@ pretty0 n AmbientContext { precedence = p, blockContext = bc, infixContext = ic,
     _ -> False
 
   nonForcePred :: AnnotatedTerm3 v PrintAnnotation -> Bool
-  nonForcePred = \case
+  nonForcePred = \case  -- rename
```
atacratic: (fix)

pchiusano:
@atacratic how is this one going? Anything I can help with? 😀 I consider this an important one to make the docs stuff usable... unless the plan is to tell people to always write perfect docs on the first try. 😀😀

atacratic:
Yes I know!  Sorry to 'help' and then be real slow...  I reckon I will get back to it and crack it this weekend.  So far I've had a session reading the lexer/parser trying to work out the best way to attack it.

atacratic:
@pchiusano please check out my latest commit message (the one on `776af61`) and shout if you don't like the sound of it.  Otherwise am planning to implement it on Wednesday.

pchiusano:
Exciting!! Will review tomorrow

unison-src/transcripts/doc-formatting.md
```diff
@@ -0,0 +1,215 @@
+This transcript explains a few minor details about doc parsing and pretty-printing, both from a user point of view and with some implementation notes.  The later stuff is meant more as unit testing than for human consumption.  (The ucm `add` commands and their output are hidden for brevity.)
```
pchiusano: Excellent, thanks for putting this together.

unison-src/transcripts/doc-formatting.output.md
```diff
@@ -0,0 +1,469 @@
...
+  doc2 : Doc
+  doc2 =
+    [: hello
+      - foo
+      - bar
+    and the rest. :]
+
+```
+```unison
+doc3 = [: When Unison identifies a paragraph, it removes any newlines from it before storing it, and then reflows the paragraph text to fit the display window on display/view/edit.
```
pchiusano: @atacratic wow this reflowing is great!! I had wanted to do that but found it was more complicated than I was willing to take on at the time so it's cool that you figured it out.

PR #996, opened 2019-11-25
Fix type propagation so that it does something

runarorama:
This PR fixes type propagation in applying patches and updates in UCM. Previously the propagation was not actually writing any new types to the codebase. Now we're doing that.

In addition, this adds deduplication for "needs alias" warnings in the case where a term/type has both an absolute and relative name.

I've also added the dumbest possible data constructor mapping which does what you want in the most common case, where you're editing a data type by adding, removing, or changing the type of a constructor. This makes the experience quite similar to what you'd expect when refactoring by mutating text files.


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1796,12 +1796,27 @@ applySelection hqs file = \sr@SlurpResult{..} ->
...
-toSlurpResult :: forall v. Var v => UF.TypecheckedUnisonFile v Ann -> Names0 -> SlurpResult v
-toSlurpResult uf existingNames =
-  Slurp.subtractComponent (conflicts <> ctorCollisions) $
-  SlurpResult uf mempty adds dups mempty conflicts updates
-              termCtorCollisions ctorTermCollisions termAliases typeAliases
-              mempty
+toSlurpResult
+  :: forall v
+   . Var v
+  => Path.Absolute
```
runarorama: Passing the current path so we can deduplicate names that have both absolute and relative versions in the `names`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1856,19 +1878,29 @@ toSlurpResult uf existingNames =
...
+      ]
 
   -- alias (n, r) if (n' /= n, r) exists in names0
   termAliases :: Map v (Set Name)
   termAliases = Map.fromList
     [ (var n, aliases)
     | (n, r@Referent.Ref{}) <- R.toList $ Names.terms fileNames0
-    , aliases <- [Set.delete n $ R.lookupRan r (Names.terms existingNames)]
+    , aliases               <-
+      [ Set.map (Path.unprefixName currentPath) . Set.delete n $ R.lookupRan
```
runarorama: Unprefix the names so they're relative to the current path if possible. This has the effect of deduplicating absolute and relative names.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -72,6 +75,27 @@ propagateAndApply patch branch = do
   f     <- applyPropagate edits
   pure $ Branch.step (f . applyDeprecations patch) branch
 
+-- Creates a mapping from old data constructors to new data constructors
+-- by looking at the original names for the data constructors which are
+-- embedded in the Decl object because we carefully planned that.
+generateConstructorMapping
```
runarorama: This uses the names in the original `Decl` to figure out which constructors of the old type should get replaced with which constructors of the new type. This lets us reassign those names to constructors of the new type.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -204,14 +233,20 @@ propagate patch b = case validatePatch patch of
             seen' = seen <> Set.fromList (view _1 . view _2 <$> joinedStuff)
             writeTypes =
               traverse_ (\(Reference.DerivedId id, tp) -> eval $ PutDecl id tp)
-          writeTypes (Map.elems componentMap)
```
runarorama: This was wrong. `componentMap` has the _old types_, so we were actually just writing those to the codebase.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -204,14 +233,20 @@ propagate patch b = case validatePatch patch of
...
+            constructorMapping =
+              constructorReplacements
+                <> generateConstructorMapping componentMap hashedComponents'
+          traceM $ "Better not be empty: " <> show constructorMapping
+          traceM $ "Better not be empty either: " <> show hashedComponents'
+          writeTypes $ Map.toList newNewTypes
```
runarorama: This value `newNewTypes` now has all the brand new types we want to actually write to the codebase at this point.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -56,10 +57,12 @@ data Edits v = Edits
   , typeEdits :: Map Reference TypeEdit
   , typeReplacements :: Map Reference Reference
   , newTypes :: Map Reference (Decl v Ann)
+  , constructorReplacements :: Map (Reference, Int, ConstructorType)
```
runarorama: The `Edits` type now tracks which constructors of old types are replaced with which constructors of new types. The mapping is based on the same thing as term and type replacements: on there being a common name between the two. The name we use is somewhat more stable though as it comes from the original `Decl`.

pchiusano:
@runarorama we just merged #1022 which relocates `DataDeclaration` to the core package and we were hoping git would handle this properly, can you take a look and see what conflicts it is reporting?

Also, other than that conflict, is this PR ready for review / merge?

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -72,6 +75,27 @@ propagateAndApply patch branch = do
...
+  => Map v (Reference, Decl v _)
+  -> Map v (Reference, Decl.DataDeclaration' v _)
+  -> Map
+       (Reference, Int, ConstructorType)
+       (Reference, Int, ConstructorType)
+generateConstructorMapping oldComponent newComponent = Map.fromList
+  [ let t = Decl.constructorType oldDecl in ((oldR, oldC, t), (newR, newC, t))
+  | (v1, (oldR, oldDecl)) <- Map.toList oldComponent
+  , (v2, (newR, newDecl)) <- Map.toList newComponent
+  , v1 == v2
```
pchiusano: I think it's fine given the relatively small number of constructors in data types written by humans, but this is O(n^2) when it could be O(n log n) - just iterate through the first map and do lg n lookups into the second map for each entry. There's an [`intersectionWith` function on `Map`](https://hackage.haskell.org/package/containers-0.6.2.1/docs/Data-Map-Internal.html#v:intersectionWith).

pchiusano:
Nice, am excited to try this out! Thanks @runarorama

PR #997, opened 2019-11-25
Add flush and make mode unique

runarorama:


aryairani:
@runarorama The build bombed because the transcript output changed unexpectedly — would you rerun transcripts on your machine, double-check that the transcript output diffs look right to you, and then check in the updated transcript outputs too?

aryairani:
Is this waiting for review?  I know no review has been requested, so maybe that means no; but it’s not a draft PR, so maybe that means yes? :)

runarorama:
It is awaiting review, but it has conflicts now

aryairani:
I’m confused:
* How does the transcript work?
  * At a high level, what is it doing?
  * How is it able to reference `.base`?  I didn’t think standard transcripts could, without a `pull`, but I don’t see one.
  * Where is the input transcript, I only see the output in the PR.
* Do we need to coordinate this with an update to `unisonweb/base`?  I’m guessing not, if old versions are still supported, but I didn’t understand that.  Maybe it was thanks to an earlier PR that may or may not have gotten merged?  Where did we land with our high level strategy for supporting multiple versions of `IO`?

Thanks

pchiusano:
Closing as obsolete with new IO type.

PR #1002, opened 2019-11-26
Use a short hash in the `names` output

aryairani:
## Before
image

## After
image


aryairani:
@pchiusano for a nicer `refactoring.md`

PR #1008, opened 2019-11-29
Fixed pretty printing of Char when it is part of a pattern

alvaroc1:
Fixed the pretty-printing of Char when it is part of a pattern (mainly just copy-pasted code from above, seems to do the right thing)


pchiusano:
Looks great, thanks for quick turnaround. Could you add yourself to CONTRIBUTORS.markdown? 🙏

alvaroc1:
Done :D

PR #1012, opened 2019-12-02
Add implementation for Nat.toFloat (Fix #1011)

### LINKED ISSUE #1011: Error when trying to use .base.Nat.toFloat - B8920912182
Error when doing:
```
> .base.Nat.toFloat 4
```

```
    The program being compiled referenced this definition ##Nat.toFloat
```
Bug reference: B8920912182

noahhaasis:
.base.Nat.toFloat was declared and had a type but was missing a implementation


aryairani:
Whoops.  Thanks, @noahhaasis.

PR #1019, opened 2019-12-04
Switch to (hopefully) more robust debouncing logic for file events

### LINKED ISSUE #625: `I loaded scratch.u and didn't find anything` message, but then it works
It seems to take ucm two goes to load the scratch file it created for me, after I edit it.

I follow the quickstart, run `edit quickstart.dsort`, and open the scratch file it creates.  I make an edit to add a watch expression, then save.  I see the output below.  First it claims not to find anything, then immediately it evidently does find the contents OK.

```
  I loaded /home/chris/u1/scratch.u and didn't find anything.

.> 

  
  ✅
  
  I found and typechecked these definitions in ~/u1/scratch.u. If you do an
  `add` or `update` , here's how your codebase would change:
  
    ⍟ These new definitions will replace existing ones of the same name and are
      ok to `update`:
    
      quickstart.dsort : (a
                         ->{quickstart.Remote} a
                         ->{quickstart.Remote} base.Boolean)
                         -> [a]
                         ->{quickstart.Remote} [a]
  
  Ctrl+C cancels.

    1 | > runLocal '(quickstart.dsort (<) [8,2,3,1,4,5,6,7])
          ⧩
          [1, 2, 3, 4, 5, 6, 7, 8]

.> 
```
M1a, Ubuntu 18.04.2 fresh install.

pchiusano:
This should fix #625 which has to do with finicky behavior with file watch events sometimes being dropped. Rather than relying on FSNotify to do debouncing, which it seems to do incorrectly (it seems to drop later events rather than earlier events), this PR uses FSNotify "un-debounced" and adds the debouncing logic on the Haskell side (we already had a function for it `collectUntilPause`).

In my usage, this works exactly the same as before but I think it should better handle cases where a user's text editor issues multiple writes in rapid succession, one of which is a write that clears the file. 

@bbarker and Pete Tsamouris if you have a chance, would greatly appreciate you testing out this fix on your setup, since I wasn't able to reproduce the original behavior.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Watch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Watch.hs
@@ -6,17 +6,19 @@ module Unison.Codebase.Watch where
 
 import Unison.Prelude
 
+import qualified UnliftIO                       as UnliftIO
 import           UnliftIO.Concurrent            ( forkIO
                                                 , threadDelay
                                                 , killThread
                                                 )
 import           UnliftIO                       ( MonadUnliftIO
-                                                , withRunInIO )
+                                                , withRunInIO
+                                                , unliftIO )
 import           UnliftIO.Directory             ( getModificationTime
                                                 , listDirectory
                                                 )
 import           UnliftIO.MVar                  ( newEmptyMVar, takeMVar
-                                                , tryTakeMVar, putMVar )
+                                                , tryTakeMVar, tryPutMVar, putMVar )
 import           UnliftIO.STM                   ( atomically )
 import           UnliftIO.Exception             ( catch, IOException)
 import           UnliftIO.IORef                 ( newIORef
@@ -29,12 +31,11 @@ import qualified Data.Text.IO
 import           Data.Time.Clock                ( UTCTime
                                                 , diffUTCTime
                                                 )
-import           System.FSNotify                ( Event(Added, Modified)
-                                                , watchDir
-                                                , withManager
-                                                )
+import           System.FSNotify                ( Event(Added, Modified))
+import qualified System.FSNotify               as FSNotify
 import           Unison.Util.TQueue             ( TQueue )
 import qualified Unison.Util.TQueue            as TQueue
+import qualified Control.Concurrent.STM        as STM
 
 untilJust :: Monad m => m (Maybe a) -> m a
 untilJust act = act >>= maybe (untilJust act) return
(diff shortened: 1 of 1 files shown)


atacratic:
I was consistently seeing the broken behavior, and now with this PR I'm consistently not seeing it :+1: :smile:

BenFradet:
same :fire:

pchiusano:
Great, thanks for testing! @runarorama or @aryairani would you like to review before merging? Should be quick, I don’t think anything controversial in implementation.

pete-ts:
Indeed much better. Not 100% fixed, at least on my setup.

1. I am adding a pedantic example, on a build from latest master.
```
f: Int -> Int -> Int
  x/scratch.u:2:21:
  unexpected <outdent>
  expecting -> or [
      2 | f: Int -> Int -> Int
```
:heavy_check_mark: 

2. I implement the naive function
```
f: Int -> Int -> Int
f x y = x + y
  I found and typechecked these definitions in ~/unison/advent-19/scratch.u. If you do an
  `add` or `update`, here's how your codebase would change:
  
      f : Int -> Int -> Int
   
```
:heavy_check_mark: 

Try and break it by being careless/thoughtless/I didn't reat the manual: 
```
> f
    4 | > f
          ⧩
          x y -> x Int.+ y
```
:heavy_check_mark: 

```
f: Int -> Int -> Int
f x y = x + y

> f 3
 I loaded x/scratch.u and didn't find anything.
```
:heavy_multiplication_x: 

Goes in a bad state, and does 3 iterations of "didn't find anything", after the initial change of focus, I try and save it manually then finally:
```
.> 
  I loaded x/scratch.u and didn't find anything.
.> 
  I loaded x/scratch.u and didn't find anything.
.> 
  I loaded x/scratch.u and didn't find anything.
.> 
  The 1st argument to the function f is base.Nat, but I was expecting base.Int:
  
      1 | f: Int -> Int -> Int
      2 | f x y = x + y
      3 | 
      4 | > f 3 3
  
  Tip: Use the syntax +3 to produce an Int.
```

* Previously, it was verging on the unusable from time to time (but my exact same setup on macosx was not really exhibiting any breakage). 
* It does seems to be much improved now.
* In the right direction, but cannot comment with 100% certainty at this time.

pchiusano:
@pete-ts It sounds like this is at least not *worse* than before. What editor / platform etc are you using? A screen recording might be super helpful, I wasn't totally sure how to interpret some of your transcript above.

Just a guess - we have a theory that the debouncing period of 50ms might not be enough, could you try making this number like 100k or 200k and see if that fixes it for you? https://github.com/unisonweb/unison/blob/5c40135c5c1c2595bb278061a999728172bb5f59/parser-typechecker/src/Unison/Codebase/Watch.hs#L139

pchiusano:
@pete-ts we are going to merge this since it seems to be an improvement for other folks, but let's keep working on getting it to 100%

pete-ts:
It is an improvement for me too, just not complete.

pete-ts:
After updating my settings, it works perfectly. :+1:

atacratic:
Hi @pete-ts, what was your config change - would be good to know the old value, new value, and default.  Is it likely to be something other VSCode users need to change too?

pete-ts:
VSCode has an `Auto Save` feature, and when set to `afterDelay`, also uses the `Auto Save Delay` field. https://code.visualstudio.com/docs/editor/codebasics#_save-auto-save. The default is 1000msec. 

After reading on this, I am using `onFocusChange`, so any value in `Auto Save Delay` should be disregarded, thus no change in configuration. Upon closer examination I don't think I have a fix, will keep investigating.

PR #1020, opened 2019-12-04
Fix #987

### LINKED ISSUE #987: Minimize term cycles before hashing
Thanks to @isovector for surfacing this, he was asking if there were any cases where hash based cycle normalization can cause definitions with different effect orders to hash the same. I replied:

> let rec blocks are not allowed to use abilities (the bindings in the block may be functions that use abilities when called, but that is different), and same goes for top level definitions, precisely because it’s a can of worms re: ordering of effects. So you shouldn’t be able to construct a typechecking Unison program where the cycle order normalization leads two semantically different programs to hash the same.

> That said, while looking at this to verify I noticed that we are failing to do cycle minimization before hashing, which is a bug that makes your concern still valid. :upside_down_face:  I’ll create an issue to track

Fix is either:

* Modify the term hashing to have a call to cycle minimization before doing the hashing.
* Modify the term parser to do cycle minimization upon parsing the term (not sure this will work - could this depend on name resolution which isn't done until after parsing?)
  * If we go this route, we should make sure prettyprinter does fine with nested let / let rec blocks.

For those curious, cycle minimization is converting one big `let rec` block to nested `let` / `let rec` blocks. [Here is the current implementation](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Typechecker/Components.hs#L19).

pchiusano:
## The problem being fixed

Currently, these two programs have the same hash, because they are parsed as one big `let rec` and hashing normalizes binding order of `let rec`:

```Haskell
spaceAttack1 x = 
  y = IO.launchMissiles saturn
  z = IO.launchMissiles neptune
  "All done"

spaceAttack2 x = 
  z = IO.launchMissiles neptune
  y = IO.launchMissiles saturn
  "All done"
```

This PR doesn't change the hashing algorithm, but parses these blocks as nested `let`, which are ordered and which will get different hashes. 

See [this transcript](https://github.com/unisonweb/unison/blob/109fbdbc7ad9d1ad8157d1a9af889350b3c923da/unison-src/transcripts/blocks.output.md)

@aryairani - we had discussed just disallowing all forward references unless they are being used to form a cycle, but I decided against that because it would mean you need to list all functions in dependency order, which is kind of annoying. Instead, as the linked transcript shows, forward references just have to be guarded.

##  Interesting decisions

* Doing this during parsing seems like an interesting decision. Why not do this normalization just before hashing? Because that would result in definitions getting a different hash after this PR - like you read an old definition, rehash it, and it suddenly has a different hash. This PR doesn't affect the hashing algorithm, only the input passed to that algorithm, so it only affects new definitions going forward. Any change to the hashing algorithm requires a full codebase upgrade.
* This doesn't attempt to use typechecking information to decide when it is safe to reorder. That seemed like a can of worms. The algorithm is purely syntactic.

Still todo:

- [x] Looks like the `use` insertion logic doesn't properly handle nested `let` / `let rec`. I'm guessing that it floats the `use` out to the nearest block, but during printing, that block doesn't correspond to a syntactic block so it doesn't work?
- [x] Add a transcript test that would fail with the old codebase, using [this example](https://github.com/unisonweb/unison/issues/987#issuecomment-561849242) as inspiration. Like try to add `spaceAttack1` and `spaceAttack2` and have it bomb with a "needs alias".
- [x] Fix to term cycle minimization as discussed [here](https://github.com/unisonweb/unison/issues/987#issuecomment-562204546)


pchiusano:
Here's the current errors, first one looks fine, test should just be updated, the others need a fix.

```
termprinter.let
  x = 1
  x = 2
  f x x EXCEPTION!!!: FancyError (SourcePos {sourceName = "", sourceLine = Pos 4, sourceColumn = Pos 8} :| []) (fromList [ErrorCustom (DuplicateTermNames [(User "x",[Ann {start = Pos 2 7, end = Pos 2 8},Ann {start = Pos 3 7, end = Pos 3 8}])])])
  error, called at tests/Unison/Test/Common.hs:49:16 in main:Unison.Test.Common
termprinter.let
  x = 1
  x = 2
  f x x EXCEPTION!!!: FancyError (SourcePos {sourceName = "", sourceLine = Pos 4, sourceColumn = Pos 8} :| []) (fromList [ErrorCustom (DuplicateTermNames [(User "x",[Ann {start = Pos 2 7, end = Pos 2 8},Ann {start = Pos 3 7, end = Pos 3 8}])])])
  error, called at tests/Unison/Test/Common.hs:49:16 in main:Unison.Test.Common
💥  termprinter.let
  x = 1
  x = 2
  f x x
💥  termprinter.let
  x = 1
  x = 2
  f x x
🦄  termprinter.let
  x = (1 : Int)
  (x : Int)
🦄  (2) termprinter.let
  x = (1 : Int)
  (x : Int)
🦄  termprinter.let
  x = if a then b else c
  if x then y else z
🦄  (2) termprinter.let
  x = if a then b else c
  if x then y else z
🦄  termprinter.let
  go acc a b =
    case List.at 0 a of
      Optional.None -> 0
      Optional.Some hd1 -> 0
  go [] a b
🦄  (2) termprinter.let
  go acc a b =
    case List.at 0 a of
      Optional.None -> 0
      Optional.Some hd1 -> 0
  go [] a b
🦄  termprinter.let
  delay = 'isEven
  ()
🦄  (2) termprinter.let
  delay = 'isEven
  ()
🦄  termprinter.let
  a = ()
  b = ()
  c = (1, 2)
  ()
🦄  (2) termprinter.let
  a = ()
  b = ()
  c = (1, 2)
  ()
🦄  termprinter.let
  a =
    use A x
    if foo then f x x else g x x
  bar
🦄  (2) termprinter.let
  a =
    use A x
    if foo then f x x else g x x
  bar
🦄  termprinter.let
  use A x
  f x x
  c = g x x
  h x x
🦄  (2) termprinter.let
  use A x
  f x x
  c = g x x
  h x x
🦄  termprinter.let
  c =
    use A x
    f x x
  g c
🦄  (2) termprinter.let
  c =
    use A x
    f x x
  g c
🦄  termprinter.let
  a =
    handle foo in
      use A x
      f x x
  bar
🦄  (2) termprinter.let
  a =
    handle foo in
      use A x
      f x x
  bar
🦄  termprinter.let
  a =
    b =
      use A x
      f x x
    foo
  bar
🦄  (2) termprinter.let
  a =
    b =
      use A x
      f x x
    foo
  bar
🦄  termprinter.let
  a =
    case x of
      () ->
        use A x
        f x x
  bar
🦄  (2) termprinter.let
  a =
    case x of
      () ->
        use A x
        f x x
  bar
🦄  termprinter.let
  a =
    use A x
    b = f x x
    c = g x x
    foo
  bar
🦄  (2) termprinter.let
  a =
    use A x
    b = f x x
    c = g x x
    foo
  bar
expected:
let
  a =
    use A p q r
    f p p
    f q q
    f r r
  foo
actual:
let
  a =
    use A p
    f p p
    f
      A.q
      A.q
    f
      A.r
      A.r
  foo
show(input)  : (let (let Var User "f" Var User "A.p" Var User "A.p" in (User "_1". (let Var User "f" Var User "A.q" Var User "A.q" in (User "_2". Var User "f" Var User "A.r" Var User "A.r")))) in (User "a". Var User "foo"))
prettyprint  : Group
  (Group
    [ (Lit "let"),
      (Lit "\n"),
      (Group
      [ (Lit "  "),
        (Group
        (Group
          [ (Group [(Lit "a"), Empty, (Lit " =")]),
            (Lit ""),
            (Lit "\n"),
            (Lit "  "),
            (Lit ""),
            Empty,
            (Group
            [ (Lit "  "),
              (Lit "use "),
              (Lit "A"),
              Empty,
              (Lit " "),
              (Lit "p"),
              Empty,
              Empty,
              Empty,
              (Lit ""),
              (Lit ""),
              (Lit "\n"),
              (Lit "  "),
              (Lit ""),
              (Lit "  "),
              (Lit ""),
              (Group
              [ (Group
                (Group
                  (OrElse
                    [ (Lit "f"),
                      (Lit " "),
                      Empty,
                      (Group
                      [ (Lit "p"),
                        (OrElse
                        (Lit " ")
                        [ (Lit ""),
                          (Lit ""),
                          (Lit "\n"),
                          (Lit "  "),
                          (Lit ""),
                          (Lit "  "),
                          Empty,
                          (Lit "")]),
                        (Lit "p"),
                        Empty,
                        Empty])]
                    [ (Lit "f"),
                      (Lit ""),
                      (Lit ""),
                      (Lit "\n"),
                      (Lit "  "),
                      (Lit ""),
                      (Lit "  "),
                      (Lit ""),
                      Empty,
                      (Group
                      [ (Lit "  "),
                        (Lit "p"),
                        (OrElse
                        (Lit " ")
                        [ (Lit ""),
                          (Lit ""),
                          (Lit ""),
                          (Lit "\n"),
                          (Lit "  "),
                          (Lit ""),
                          (Lit "  "),
                          (Lit ""),
                          (Lit "  "),
                          Empty,
                          (Lit "")]),
                        (Lit "p"),
                        Empty,
                        Empty])]))),
                (Lit ""),
                (Lit ""),
                (Lit "\n"),
                (Lit "  "),
                (Lit ""),
                (Lit "  "),
                (Lit ""),
                (Group
                (Group
                  (OrElse
                    [ (Lit "f"),
                      (Lit " "),
                      Empty,
                      (Group
                      [ (Lit "A.q"),
                        (OrElse
                        (Lit " ")
                        [ (Lit ""),
                          (Lit ""),
                          (Lit "\n"),
                          (Lit "  "),
                          (Lit ""),
                          (Lit "  "),
                          Empty,
                          (Lit "")]),
                        (Lit "A.q"),
                        Empty,
                        Empty])]
                    [ (Lit "f"),
                      (Lit ""),
                      (Lit ""),
                      (Lit "\n"),
                      (Lit "  "),
                      (Lit ""),
                      (Lit "  "),
                      (Lit ""),
                      Empty,
                      (Group
                      [ (Lit "  "),
                        (Lit "A.q"),
                        (OrElse
                        (Lit " ")
                        [ (Lit ""),
                          (Lit ""),
                          (Lit ""),
                          (Lit "\n"),
                          (Lit "  "),
                          (Lit ""),
                          (Lit "  "),
                          (Lit ""),
                          (Lit "  "),
                          Empty,
                          (Lit "")]),
                        (Lit "A.q"),
                        Empty,
                        Empty])]))),
                (Lit ""),
                (Lit ""),
                (Lit "\n"),
                (Lit "  "),
                (Lit ""),
                (Lit "  "),
                Empty,
                (Lit ""),
                (Group
                (Group
                  (Group
                    (OrElse
                      [ (Lit "f"),
                        (Lit " "),
                        Empty,
                        (Group
                        [ (Lit "A.r"),
                          (OrElse
                          (Lit " ")
                          [ (Lit ""),
                            (Lit ""),
                            (Lit "\n"),
                            (Lit "  "),
                            (Lit ""),
                            (Lit "  "),
                            Empty,
                            (Lit "")]),
                          (Lit "A.r"),
                          Empty,
                          Empty])]
                      [ (Lit "f"),
                        (Lit ""),
                        (Lit ""),
                        (Lit "\n"),
                        (Lit "  "),
                        (Lit ""),
                        (Lit "  "),
                        (Lit ""),
                        Empty,
                        (Group
                        [ (Lit "  "),
                          (Lit "A.r"),
                          (OrElse
                          (Lit " ")
                          [ (Lit ""),
                            (Lit ""),
                            (Lit ""),
                            (Lit "\n"),
                            (Lit "  "),
                            (Lit ""),
                            (Lit "  "),
                            (Lit ""),
                            (Lit "  "),
                            Empty,
                            (Lit "")]),
                          (Lit "A.r"),
                          Empty,
                          Empty])])))),
                Empty,
                Empty]),
              Empty])])),
        (Lit ""),
        (Lit "\n"),
        (Lit "  "),
        (Lit ""),
        (Group (Lit "foo")),
        Empty,
        Empty,
        Empty])])
termprinter.let
  a =
    use A p q r
    f p p
    f q q
    f r r
  foo FAILURE actual != expected CallStack (from HasCallStack):
  crash, called at tests/Unison/Test/TermPrinter.hs:47:13 in main:Unison.Test.TermPrinter
💥  termprinter.let
  a =
    use A p q r
    f p p
    f q q
    f r r
  foo
💥  termprinter.let
  a =
    use A p q r
    f p p
    f q q
    f r r
  foo
🦄  termprinter.let
  a =
    use A B.x
    b =
      use A.B x
      f x x
    c =
      g B.x B.x
      h A.D.x
    foo
  bar
🦄  (2) termprinter.let
  a =
    use A B.x
    b =
      use A.B x
      f x x
    c =
      g B.x B.x
      h A.D.x
    foo
  bar
expected:
let
  use A x
  use A.T.A T1
  g = T1 +3
  h = T1 +4
  i : T -> T -> Int
  i p q =
    g' = T1 +3
    h' = T1 +4
    +2
  if true then x else x
actual:
let
  use A.T.A T1
  g = T1 +3
  h = T1 +4
  i : T -> T -> Int
  i p q =
    g' = T1 +3
    h' = T1 +4
    +2
  if true then A.x else A.x
show(input)  : (let Var User "A.T.A.T1" +3 in (User "g". (let Var User "A.T.A.T1" +4 in (User "h". (let (λ (User "p". (λ (User "q". (let Var User "A.T.A.T1" +3 in (User "g'". (let Var User "A.T.A.T1" +4 in (User "h'". +2)))))))):Var User "T" -> Var User "T" -> Var User "Int" in (User "i". if true then Var User "A.x" else Var User "A.x"))))))
prettyprint  : Group
  (Group
    [ (Lit "let"),
      (Lit "\n"),
      (Group
      [ (Lit "  "),
        (Lit "use "),
        (Lit "A"),
        (Lit "."),
        (Lit "T"),
        (Lit "."),
        (Lit "A"),
        Empty,
        (Lit " "),
        (Lit "T1"),
        Empty,
        (Lit ""),
        (Lit "\n"),
        (Lit "  "),
        (Lit ""),
        (Group
        (Group
          (OrElse
            [ (Group [(Lit "g"), Empty, (Lit " =")]),
              (Lit " "),
              Empty,
              (Group
              [ (Group
                (Group
                  (OrElse
                    [ (Lit "T1"),
                      (Lit " "),
                      Empty,
                      (Group [(Lit "+"), Empty, (Lit "3"), Empty])]
                    [ (Lit "T1"),
                      (Lit ""),
                      (Lit "\n"),
                      (Lit "  "),
                      (Lit ""),
                      Empty,
                      (Group
                      [(Lit "  "), (Lit "+"), Empty, (Lit "3"), Empty, Empty])]))),
                Empty,
                Empty])]
            [ (Group [(Lit "g"), Empty, (Lit " =")]),
              (Lit ""),
              (Lit "\n"),
              (Lit "  "),
              (Lit ""),
              Empty,
              (Group
              [ (Lit "  "),
                (Group
                (Group
                  (OrElse
                    [ (Lit "T1"),
                      (Lit " "),
                      Empty,
                      (Group [(Lit "+"), Empty, (Lit "3"), Empty, Empty])]
                    [ (Lit "T1"),
                      (Lit ""),
                      (Lit ""),
                      (Lit "\n"),
                      (Lit "  "),
                      (Lit ""),
                      (Lit "  "),
                      (Lit ""),
                      Empty,
                      (Group
                      [ (Lit "  "),
                        (Lit "+"),
                        Empty,
                        Empty,
                        (Lit "3"),
                        Empty,
                        Empty])]))),
                Empty,
                Empty,
                Empty])]))),
        Empty,
        (Lit ""),
        (Lit "\n"),
        (Lit "  "),
        (Lit ""),
        (Group
        (Group
          (OrElse
            [ (Group [(Lit "h"), Empty, (Lit " =")]),
              (Lit " "),
              Empty,
              (Group
              [ (Group
                (Group
                  (OrElse
                    [ (Lit "T1"),
                      (Lit " "),
                      Empty,
                      (Group [(Lit "+"), Empty, (Lit "4"), Empty])]
                    [ (Lit "T1"),
                      (Lit ""),
                      (Lit "\n"),
                      (Lit "  "),
                      (Lit ""),
                      Empty,
                      (Group
                      [(Lit "  "), (Lit "+"), Empty, (Lit "4"), Empty, Empty])]))),
                Empty,
                Empty])]
            [ (Group [(Lit "h"), Empty, (Lit " =")]),
              (Lit ""),
              (Lit "\n"),
              (Lit "  "),
              (Lit ""),
              Empty,
              (Group
              [ (Lit "  "),
                (Group
                (Group
                  (OrElse
                    [ (Lit "T1"),
                      (Lit " "),
                      Empty,
                      (Group [(Lit "+"), Empty, (Lit "4"), Empty, Empty])]
                    [ (Lit "T1"),
                      (Lit ""),
                      (Lit ""),
                      (Lit "\n"),
                      (Lit "  "),
                      (Lit ""),
                      (Lit "  "),
                      (Lit ""),
                      Empty,
                      (Group
                      [ (Lit "  "),
                        (Lit "+"),
                        Empty,
                        Empty,
                        (Lit "4"),
                        Empty,
                        Empty])]))),
                Empty,
                Empty,
                Empty])]))),
        (Lit ""),
        (Lit "\n"),
        (Lit "  "),
        (Lit ""),
        (Group
        [ (Lit "i"),
          Empty,
          (Group
          (OrElse
            [ (Lit " :"),
              (Lit " "),
              Empty,
              (Group
              [ (Lit "T"),
                (OrElse (Lit " ") [(Lit ""), (Lit "\n"), (Lit "  "), (Lit "")]),
                (Lit "->"),
                Empty,
                Empty,
                (Lit " "),
                (Lit "T"),
                (OrElse (Lit " ") [(Lit ""), (Lit "\n"), (Lit "  "), (Lit "")]),
                (Lit "->"),
                Empty,
                Empty,
                Empty,
                (Lit " "),
                (Lit "Int"),
                Empty])]
            [ (Lit " :"),
              (Lit ""),
              (Lit "\n"),
              (Lit "  "),
              (Lit ""),
              Empty,
              (Group
              [ (Lit "  "),
                (Lit "T"),
                (OrElse
                (Lit " ")
                [ (Lit ""),
                  (Lit ""),
                  (Lit "\n"),
                  (Lit "  "),
                  (Lit ""),
                  (Lit "  "),
                  Empty,
                  (Lit "")]),
                (Lit "->"),
                Empty,
                Empty,
                (Lit " "),
                Empty,
                (Lit "T"),
                (OrElse
                (Lit " ")
                [ (Lit ""),
                  (Lit ""),
                  (Lit "\n"),
                  (Lit "  "),
                  (Lit ""),
                  (Lit "  "),
                  Empty,
                  (Lit "")]),
                (Lit "->"),
                Empty,
                Empty,
                Empty,
                (Lit " "),
                (Lit "Int"),
                Empty])]))]),
        (Lit ""),
        (Lit "\n"),
        (Lit "  "),
        (Lit ""),
        (Group
        (Group
          (Group
            [ (Group
              [ (Group
                (OrElse
                  [ (Lit "i"),
                    (Lit " "),
                    Empty,
                    (Group
                    [ (Lit "p"),
                      (OrElse
                      (Lit " ") [(Lit ""), (Lit "\n"), (Lit "  "), (Lit "")]),
                      (Lit "q"),
                      Empty,
                      Empty])]
                  [ (Lit "i"),
                    (Lit ""),
                    (Lit "\n"),
                    (Lit "  "),
                    (Lit ""),
                    Empty,
                    (Group
                    [ (Lit "  "),
                      (Lit "p"),
                      (OrElse
                      (Lit " ")
                      [ (Lit ""),
                        (Lit ""),
                        (Lit "\n"),
                        (Lit "  "),
                        (Lit ""),
                        (Lit "  "),
                        Empty,
                        (Lit "")]),
                      (Lit "q"),
                      Empty,
                      Empty])])),
                Empty,
                (Lit " =")]),
              (Lit ""),
              (Lit "\n"),
              (Lit "  "),
              (Lit ""),
              Empty,
              (Group
              [ (Lit "  "),
                (Group
                [ (Group
                  (Group
                    (OrElse
                      [ (Group [(Lit "g'"), Empty, Empty, (Lit " =")]),
                        (Lit " "),
                        Empty,
                        (Group
                        [ (Group
                          (Group
                            (OrElse
                              [ (Lit "T1"),
                                (Lit " "),
                                Empty,
                                (Group
                                [(Lit "+"), Empty, (Lit "3"), Empty, Empty])]
                              [ (Lit "T1"),
                                (Lit ""),
                                (Lit ""),
                                (Lit "\n"),
                                (Lit "  "),
                                (Lit ""),
                                (Lit "  "),
                                (Lit ""),
                                Empty,
                                (Group
                                [ (Lit "  "),
                                  (Lit "+"),
                                  Empty,
                                  Empty,
                                  (Lit "3"),
                                  Empty,
                                  Empty])]))),
                          Empty,
                          Empty,
                          Empty])]
                      [ (Group [(Lit "g'"), Empty, Empty, (Lit " =")]),
                        (Lit ""),
                        (Lit ""),
                        (Lit "\n"),
                        (Lit "  "),
                        (Lit ""),
                        (Lit "  "),
                        (Lit ""),
                        Empty,
                        (Group
                        [ (Lit "  "),
                          (Group
                          (Group
                            (OrElse
                              [ (Lit "T1"),
                                (Lit " "),
                                Empty,
                                Empty,
                                (Group
                                [(Lit "+"), Empty, (Lit "3"), Empty, Empty])]
                              [ (Lit "T1"),
                                (Lit ""),
                                (Lit ""),
                                Empty,
                                (Lit ""),
                                (Lit "\n"),
                                (Lit "  "),
                                (Lit ""),
                                (Lit "  "),
                                (Lit ""),
                                (Lit "  "),
                                (Lit ""),
                                Empty,
                                (Group
                                [ (Lit "  "),
                                  (Lit "+"),
                                  Empty,
                                  Empty,
                                  (Lit "3"),
                                  Empty,
                                  Empty])]))),
                          Empty,
                          Empty,
                          Empty])]))),
                  (Lit ""),
                  (Lit ""),
                  (Lit "\n"),
                  (Lit "  "),
                  (Lit ""),
                  (Lit "  "),
                  (Lit ""),
                  (Group
                  (Group
                    (OrElse
                      [ (Group [(Lit "h'"), Empty, Empty, (Lit " =")]),
                        (Lit " "),
                        Empty,
                        (Group
                        [ (Group
                          (Group
                            (OrElse
                              [ (Lit "T1"),
                                (Lit " "),
                                Empty,
                                (Group
                                [(Lit "+"), Empty, (Lit "4"), Empty, Empty])]
                              [ (Lit "T1"),
                                (Lit ""),
                                (Lit ""),
                                (Lit "\n"),
                                (Lit "  "),
                                (Lit ""),
                                (Lit "  "),
                                (Lit ""),
                                Empty,
                                (Group
                                [ (Lit "  "),
                                  (Lit "+"),
                                  Empty,
                                  Empty,
                                  (Lit "4"),
                                  Empty,
                                  Empty])]))),
                          Empty,
                          Empty,
                          Empty])]
                      [ (Group [(Lit "h'"), Empty, Empty, (Lit " =")]),
                        (Lit ""),
                        (Lit ""),
                        (Lit "\n"),
                        (Lit "  "),
                        (Lit ""),
                        (Lit "  "),
                        (Lit ""),
                        Empty,
                        (Group
                        [ (Lit "  "),
                          (Group
                          (Group
                            (OrElse
                              [ (Lit "T1"),
                                (Lit " "),
                                Empty,
                                Empty,
                                (Group
                                [(Lit "+"), Empty, (Lit "4"), Empty, Empty])]
                              [ (Lit "T1"),
                                (Lit ""),
                                (Lit ""),
                                Empty,
                                (Lit ""),
                                (Lit "\n"),
                                (Lit "  "),
                                (Lit ""),
                                (Lit "  "),
                                (Lit ""),
                                (Lit "  "),
                                (Lit ""),
                                Empty,
                                (Group
                                [ (Lit "  "),
                                  (Lit "+"),
                                  Empty,
                                  Empty,
                                  (Lit "4"),
                                  Empty,
                                  Empty])]))),
                          Empty,
                          Empty,
                          Empty])]))),
                  (Lit ""),
                  (Lit ""),
                  (Lit "\n"),
                  (Lit "  "),
                  (Lit ""),
                  (Lit "  "),
                  Empty,
                  (Lit ""),
                  (Group [(Lit "+"), Empty, Empty, (Lit "2")]),
                  Empty,
                  Empty]),
                Empty,
                Empty,
                Empty])]))),
        Empty,
        (Lit ""),
        (Lit "\n"),
        (Lit "  "),
        (Lit ""),
        (Group
        (Group
          [ (Group
            (OrElse
              [(Lit "if"), (Lit " "), Empty, (Group (Lit "true"))]
              [ (Lit "if"),
                (Lit ""),
                (Lit "\n"),
                (Lit "  "),
                (Lit ""),
                Empty,
                (Group [(Lit "  "), Empty, (Lit "true")])])),
            (Group
            (OrElse
              [ (Lit " then"),
                (Lit " "),
                Empty,
                (Group [(Lit "A.x"), Empty, Empty])]
              [ (Lit " then"),
                (Lit ""),
                (Lit "\n"),
                (Lit "  "),
                (Lit ""),
                Empty,
                (Group [(Lit "  "), (Lit "A.x"), Empty, Empty, Empty])])),
            (OrElse (Lit " ") [(Lit ""), (Lit "\n"), (Lit "  "), (Lit "")]),
            (Group
            (OrElse
              [ (Lit "else"),
                (Lit " "),
                Empty,
                (Group [(Lit "A.x"), Empty, Empty])]
              [ (Lit "else"),
                (Lit ""),
                (Lit "\n"),
                (Lit "  "),
                (Lit ""),
                Empty,
                (Group [(Lit "  "), (Lit "A.x"), Empty, Empty, Empty])])),
            Empty])),
        Empty,
        Empty])])
termprinter.let
  use A x
  use A.T.A T1
  g = T1 +3
  h = T1 +4
  i : T -> T -> Int
  i p q =
    g' = T1 +3
    h' = T1 +4
    +2
  if true then x else x FAILURE actual != expected CallStack (from HasCallStack):
  crash, called at tests/Unison/Test/TermPrinter.hs:47:13 in main:Unison.Test.TermPrinter
💥  termprinter.let
  use A x
  use A.T.A T1
  g = T1 +3
  h = T1 +4
  i : T -> T -> Int
  i p q =
    g' = T1 +3
    h' = T1 +4
    +2
  if true then x else x
💥  termprinter.let
  use A x
  use A.T.A T1
  g = T1 +3
  h = T1 +4
  i : T -> T -> Int
  i p q =
    g' = T1 +3
    h' = T1 +4
    +2
  if true then x else x


  15 passed
  3 FAILED (failed scopes below)
    "termprinter.let"
    "termprinter.let"
    "termprinter.let"

    EasyTest.rerun 744709729785416746
    EasyTest.rerunOnly 744709729785416746 "termprinter.let
  a =
    use A p q r
    f p p
    f q q
    f r r
  foo"
```

pchiusano:
@atacratic I could use a steer on fixing the pretty-printing. Here's some background info - with this fix, the parser will now generate nested cycle-minimized blocks, so when you have a block like:

```
x = 42
y = 19
fib n = if n < 2 then n else fib (n - 1) + fib (n - 2)
"Done"
```

It actually converts this to a nested `let` / `let rec` block (`let x = 42 in let y = 19 in let rec fib n = ... in "Done"`).

There's a helpful pattern in `TermPrinter`, `TermPrinter.LetBlock` which discards this nesting. However, it seems like it still is messing with `use` insertion, see the attached failures above - https://github.com/unisonweb/unison/pull/1020#issuecomment-561855745. Any ideas where to look for this one?

atacratic:
@pchiusano My guess is it's the `LetRecNamed'`/`Lets'` patterns in `TermPrinter.immediateChildBlockTerms` which want to be switched over to `LetBlock`.  There are comments above `allInSubBlock` and `narrowestPossible` which should hopefully give some clue as to what it's doing.  You'll see there's some tracing you can enable in `allInSubBlock`.  

I wonder how that existing logic to determine if a term is a 'blockterm' (in the parser's sense) is going to interact with your `unLetBlock` code that makes decisions about whether to start a new block.

pchiusano:
@atacratic I could really use your help with this one.

I tried replacing the usage in `immediateChildBlockTerms` to no avail. I also tried enabling those trace statements, but wasn't really sure what I was looking for - I'm not familiar enough with the use insertion logic to debug this effectively. :)

atacratic:
OK I will take a look this weekend!  The use insertion is fiddly...

atacratic:
@pchiusano hmm, well after 3 hours of trying to remember how it works and looking through traces, I've come up with a fix of removing `[tm] ++` from `immediateChildBlockTerms`.  Seems to make the tests work and I can't find any problems with it.  It does make me slightly worry that there could be some case involving let blocks where we don't put the use statement quite as far down the tree as we could, but I guess that might not be a total disaster...

Overall the trickiness here is because we're trying to traverse the surface syntax (to analyse what the parser calls blockterms) but all we have to go on is the underlying ABT, so it gets hairy when they diverge.  I've tried to improve the comments at least.

pchiusano:
@atacratic thanks for the fix!

unison-core/src/Unison/ABT.hs
```diff
@@ -558,6 +558,24 @@ instance (Foldable f, Functor f, Ord1 f, Var v) => Ord (Term f v a) where
...
+  go cur deps rem = case findIndex isDep rem of
+    Nothing -> reverse cur : let (hd,tl) = splitAt 1 rem
+                             in go hd (depsFor hd) tl
+    Just i  -> go (reverse newMembers ++ cur) deps' (drop (i+1) rem) 
+               where deps' = deps <> depsFor newMembers
+                     newMembers = take (i+1) rem
+    where
+    depsFor = foldMap (freeVars . snd)
+    isDep (v, _) = Set.member v deps
+
```
pchiusano: This is a key new algorithm for converting a list of bindings into a minimal list of components, preserving their order. /cc @TomasMikula

unison-src/transcripts/fix987.md
```diff
@@ -0,0 +1,33 @@
+
```
pchiusano: This transcript used to fail on master, but passes now.

unison-src/transcripts/merges.output.md
```diff
@@ -120,13 +120,13 @@ We can also delete the fork if we're done with it. (Don't worry, it's still in t
   Note: The most recent namespace hash is immediately below this
         message.
   
-  ⊙ #qb1es1ffsc
+  ⊙ #6s4tc161c3
```
pchiusano: The hash of root namespace changed because one of the definitions in `IOSource` is a block which is parsed into something different than before. @aryairani what if anything do we need to do about this?

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -873,3 +877,33 @@ immediateChildBlockTerms = \case
...
+        Just (bindings ++ innerBindings, innerBody)
+      _ -> Just (bindings, body)
+  nonrec t = case unLet t of
+    Nothing -> Nothing
+    Just (bindings0, body) -> 
+      let bindings = [ (v,b) | (_,v,b) <- bindings0 ] in 
+      case rec body of
+        Just (innerBindings, innerBody) | dontIntersect bindings innerBindings -> 
+          Just (bindings ++ innerBindings, innerBody)
+        _ -> Just (bindings, body)
```
pchiusano: Used by the prettyprinter to "undo" the nesting of let/let rec.

parser-typechecker/src/Unison/Typechecker/Components.hs
```diff
@@ -43,7 +49,7 @@ minimize (Term.LetRecNamedAnnotatedTop' isTop ann bs e) =
   in  if not $ null dupes
         then Left $ Nel.fromList dupes
         else
-          let cs0            = ABT.components bindings
+          let cs0 = if isTop then unordered bindings else ordered bindings
```
pchiusano: Top level definitions can be reordered - the order of definitions in a file isn't semantically meaningful, but the order of definitions within a function body might be so we conservatively don't do any reordering.

unison-core/src/Unison/ABT.hs
```diff
@@ -576,6 +575,26 @@ orderedComponents tms = go [] Set.empty tms
...
+  tweak :: Var v => [(v,Term f v a)] -> [[(v,Term f v a)]]
+  tweak bs@(_:_:_) = case takeWhile isCyclic (components bs) of
+    [] -> [bs]
+    cycles -> cycles <> orderedComponents rest 
+      where
+      rest = [ (v,b) | (v,b) <- bs, Set.notMember v cycleVars ]
+      cycleVars = Set.fromList (fst <$> join cycles)
+  tweak bs = [bs] -- any cycle with < 2 bindings is left alone
+  isCyclic [(v,b)] = Set.member v (freeVars b)
+  isCyclic bs      = length bs > 1
```
pchiusano: @aryairani here's implementation of that tweak.

unison-src/transcripts/merges.output.md
aryairani: Let's get rid of it, per #1056.  In the meantime, it will only affect people who `use .builtin` with a codebase initialized after this PR.

PR #1022, opened 2019-12-04
Carve out unison-core package

mitchellwrosen:
This patch adds a `unison-core` package with:

- `Unison.ABT`
- `Unison.DataDeclaration`
- `Unison.Term`
- `Unison.Type`

and all of their dependencies.

Part of #779


aryairani:
@mitchellwrosen You told me, but I forgot the specifics, sorry — do you remember why `Names2` and `Names3` were needed?

mitchellwrosen:
@arya not off the top of my head but I think
```
rg '^import.*Unison.Names2' unison-src
```
would answer.

mitchellwrosen:
(Short answer, if it wasn't obvious, is something pulled them in)

pchiusano:
Okay, I just played with this some and everything in our current workflow (using `ghcid`, `stack exec tests`, `stack ghci`) all work fine. So going to go ahead and merge this.

One thing we weren't sure about - we want to make sure that concurrent renames/moves and edits are always merged correctly by git. 

@aryairani is skeptical that it will be 100% accurate since Git is just using some heuristics to detect renames and those heuristics might get confused depending on what you are merging with (this seems crazy to me but it wouldn't shock me if it weren't handled correctly).

PR #1031, opened 2019-12-07
Read .unisonConfig from where the codebase is

### LINKED ISSUE #917: Read the `.unisonConfig` file from wherever the codebase is.
Right now `.unisonConfig` is read from the working directory (wherever `ucm` is launched). It makes more sense for it to be read from wherever the codebase is. So for example if you start `ucm` with no arguments, the codebase is in the user's home directory. The `.unisonConfig` file should be read from there as well.

pt2121:
Fix #917

This PR basically passes the config file path to the functions where we watch it.


pchiusano:
Great, thanks!

aryairani:
We should think this through again for #1113

PR #1032, opened 2019-12-08
Output fewer blank lines when only whitespace is input

mitchellwrosen:
Before (when pressing Enter at the prompt):
```
.>



.>
```
After:
```
.>
.>
```


pchiusano:
Nice!

Little fixes like this are hugely appreciated.

PR #1037, opened 2019-12-08
Improve delete term/type feedback

mitchellwrosen:
This patch improves the delete term/type feedback a bit. Here's what it looks like at the moment (feedback welcome of course):

2019-12-08-143548_1205x311_scrot

I did leave one case untouched: `display`ing an ambiguous hash. It still just says `That term is ambiguous.` like before. This is because it actually takes a `HashQualified` rather than a `HQSplit'`, so it didn't quite fit the new mold. Doable, for sure, but I thought I'd leave it for another patch because it brings up a separate question: why _can't_ I `delete.type #byhash`? :)

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -270,8 +270,9 @@ loop = do
         patchExists s = respond $ PatchAlreadyExists input s
         typeNotFound = respond . TypeNotFound input
         termNotFound = respond . TermNotFound input
-        typeConflicted src = respond . TypeAmbiguous input src
-        termConflicted src = respond . TermAmbiguous input (Left src)
+        nameConflicted src tms tys = respond (NameAmbiguous hqLength input src tms tys)
+        typeConflicted src = nameConflicted src Set.empty
+        termConflicted src tms = nameConflicted src tms Set.empty
         hashConflicted src = respond . HashAmbiguous input src
         branchExists dest _x = respond $ BranchAlreadyExists input dest
         branchExistsSplit = branchExists . Path.unsplit'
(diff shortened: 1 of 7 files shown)


aryairani:
`delete.type #byhash` is just undefined.  We can define it, but what should it do?  Delete at the current level?  Delete recursively below the current level?  Delete recursively from the root?  My thinking at the time was that deleting a name without specifying which name to delete was tough. :)

For feedback, do you mind shortening the displayed hashes according to the `CodebaseHashLength` command (with `HashQualified.take`).

mitchellwrosen:
@aryairani I guess I didn't think about it too hard, maybe it doesn't make sense. My guess is: since you haven't supplied a name, your current level is irrelevant, so perhaps it should (attempt to) delete all names associated with that hash from the root.

aryairani:
@mitchellwrosen I suppose whatever it does could prompt for confirmation.  I think current namespace and below is a reasonable interpretation, and that all namespaces is a reasonable interpretation.  Or current and below, list the ones that will be deleted on confirmation, along with those that won’t be deleted but could be if you issued the command from the root? 😕

aryairani:
At any rate, this is okay to merge with or without the `.take` change.

pchiusano:
@mitchellwrosen this one has conflicts now, I'm guessing due to #1040 that I just merged. :)

mitchellwrosen:
2019-12-09-220923_712x328_scrot

mitchellwrosen:
The last push I made was not just a rebase, but also a small rework to account for the new `NameAmbiguous` thing that landed in another patch, so I will not merge immediately... please re-review if you'd like. But it's also not very interesting stuff, probably not worth yet another review round... so if I don't hear anything after a day or so I'll go ahead and merge. I give myself an LGTM :)

pchiusano:
LGTM I’d say merge once Travis finishes. @mitchellwrosen do stack exec transcripts to regenerate the transcripts of you are expecting them to change their output (which in this case you are). Then commit the updated output file.

mitchellwrosen:
@pchiusano Yeah, working on that now. The `unique[abc]` syntax doesn't seem to be working for me (?), but I can rewrite the transcript to use a structural type as an example instead.

pchiusano:
😎👍

PR #1038, opened 2019-12-08
Add COMPLETE pragma to Reference

mitchellwrosen:
Removes two calls to `error`

### DIFF
--- a/unison-core/src/Unison/Reference.hs
+++ b/unison-core/src/Unison/Reference.hs
@@ -48,6 +48,8 @@ data Reference
 
 pattern Derived h i n = DerivedId (Id h i n)
 
+{-# COMPLETE Builtin, Derived #-}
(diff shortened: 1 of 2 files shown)


mitchellwrosen:
Huh... ok... CI failure looks like a GHC bug to me? It thinks this is now inaccessible.
```
Pattern.ConstructorP _ UnitRef 0 [] -> Just []
```

aryairani:
Hehe, yeah :-/  see the two lines below the newly added COMPLETE pragma.

pchiusano:
Should we close this? I’d love to have this work but @mitchellwrosen is now the second person to discover this GHC feature is buggy at least in the GHC version we are using currently.

mitchellwrosen:
Think so :man_shrugging:

PR #1040, opened 2019-12-08
Add 'delete' command

### LINKED ISSUE #662: should I really have to specify delete.term vs delete.type?
I have to write `delete.term` or `delete.type` according to what I'm trying to delete.  But should I really have to, in the common case where the name does not exist both as term and type, but only as one of the two?

Ditto `alias` and `move`.

I appreciate it's hard to get too excited about this one.

(`delete` and `move` have the patch and namespace variants to think about as well.)

mitchellwrosen:
This patch adds `delete`, which sort of combines `delete.term` and `delete.type`. If there are any ambiguities, it bails. Transcript test included!

Fixes #662

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,5 @@ install:
 - stack --no-terminal exec tests
 - stack --no-terminal exec transcripts
 # fail if running transcripts modified any versioned files
-- x=`git status --porcelain -uno` bash -c 'if [[ -n $x ]]; then echo "$x" && false; fi' 
+- git diff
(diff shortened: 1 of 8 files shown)


mitchellwrosen:
I would hope so! (Pretty sure I did not regress here... double checking)

mitchellwrosen:
Yeah, nope, `delete` just doesn't give feedback yet.

aryairani:
Ok well then FF to merge if you’re ready. :)

pchiusano:
Nice! I like the transcript test.

PR #1057, opened 2019-12-12
ucm run: do not set up interactive environment

int-index:
This patch changes the way `ucm run` works, so that it doesn't set up an interactive environment, does not watch the current directory, does not print any messages to the user. As such, now it can be used to run applications implemented in Unison.

Before this patch, running `main = '(putText stdout "Hello, World!\n")` via `ucm run` resulted in the following output:

```
   _____     _             
  |  |  |___|_|___ ___ ___ 
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|
  Welcome to Unison!
  I'm currently watching for changes to .u files under ~/Projects/unison
  Type help to get help. :sunglasses:
Hello, World!
```

Now it just prints `Hello, World!` as expected.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Execute.hs
+++ b/parser-typechecker/src/Unison/Codebase/Execute.hs
@@ -1,0 +1,50 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | Execute a computation of type '{IO} () that has been previously added to
+-- the codebase, without setting up an interactive environment.
+--
+-- This allows one to run standalone applications implemented in the Unison
+-- language.
+
+module Unison.Codebase.Execute where
+
+import Unison.Prelude
+
+import           Unison.Codebase.MainTerm      ( getMainTerm )
+import qualified Unison.Codebase.MainTerm      as MainTerm
+import qualified Unison.Codebase               as Codebase
+import           Unison.Parser                 ( Ann )
+import qualified Unison.Codebase.Runtime       as Runtime
+import           Unison.Codebase.Runtime       ( Runtime )
+import           Unison.Var                    ( Var )
+import qualified Unison.PrettyPrintEnv         as PPE
+import qualified Unison.Names3                 as Names3
+import qualified Unison.Codebase.Branch        as Branch
+import           System.Exit (die)
+import           Control.Exception (finally)
+
+execute
+  :: Var v
+  => Codebase.Codebase IO v Ann
+  -> Runtime v
+  -> String
+  -> IO ()
+execute codebase runtime mainName =
+  (`finally` Runtime.terminate runtime) $ do
+    root <- Codebase.getRootBranch codebase
+    let parseNames0 = Names3.makeAbsolute0 (Branch.toNames0 (Branch.head root))
+        loadTypeOfTerm = Codebase.getTypeOfTerm codebase
+    mt <- getMainTerm loadTypeOfTerm parseNames0 mainName
+    case mt of
+      MainTerm.NotAFunctionName s -> die ("Not a function name: " ++ s)
+      MainTerm.NotFound s -> die ("Not found: " ++ s)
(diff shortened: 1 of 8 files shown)


aryairani:
I haven't run this personally, but it sounds good and looks straighforward.  What do y'all think about the new error messages' text?  How much do we care about flowery language here?

parser-typechecker/src/Unison/Codebase/Execute.hs
```diff
@@ -0,0 +1,69 @@
...
+              Just typ | Typechecker.isSubtype typ (nullaryMain a) -> do
+                let tm = DD.forceTerm a a (Term.ref a ref)
+                    codeLookup = Codebase.toCodeLookup codebase
+                    uf = UF.UnisonFile mempty mempty mempty
+                           (Map.singleton UF.RegularWatch [(Var.nameds "result", tm)])
+                selfContained <- Codebase.makeSelfContained' codeLookup uf
+                _ <- Runtime.evaluateWatches codeLookup ppe Runtime.noCache runtime selfContained
+                return ()
+              _ -> die (mainName ++ " is not of type '{IO} ()")
+          _ -> die ("Not found: " ++ mainName)
```
pchiusano: Hmm so I guess this is a bit of a copy paste (but not exactly) of the existing code. How do we feel about this?

Other option might be to just add a flag to the existing codepath to control whether to print out the welcome, etc.

int-index: The existing codepath does much more than printing the welcome message. It creates unneeded mutable references:

```
    rootRef                  <- newIORef root
    pathRef                  <- newIORef initialPath
    initialInputsRef         <- newIORef initialInputs
    numberedArgsRef          <- newIORef []
```

It starts watching for configuration, file system, and branch updates:

```
    (config, cancelConfig)   <-
      catchIOError (watchConfig configFile) $ \_ ->
        die "Your .unisonConfig could not be loaded. Check that it's correct!"
    cancelFileSystemWatch    <- watchFileSystem eventQueue dir
    cancelWatchBranchUpdates <- watchBranchUpdates (Branch.headHash <$>
                                                      readIORef rootRef)
                                                   eventQueue
                                                   codebase
```

It uses a queue to drive an event loop:

```
  eventQueue <- Q.newIO
      awaitInput = do
        -- use up buffered input before consulting external events
        i <- readIORef initialInputsRef
        case i of
          h:t -> writeIORef initialInputsRef t >> pure h
          [] ->
            -- Race the user input and file watch.
            Async.race (atomically $ Q.peek eventQueue) getInput >>= \case
              Left _ -> Left <$> atomically (Q.dequeue eventQueue)
              x      -> pure x
```

And then there are some free monad to organize all of that.

How would one hide all of that behind a flag? A little bit of duplication is better, in my opinion. If we want to remove duplication, the solution would be not a flag, but a proper abstraction that would be used both by this runner and the interactive loop.

int-index:
@pchiusano I removed the duplication by factoring out common bits of code. What do you think?

parser-typechecker/src/Unison/Codebase/MainTerm.hs
```diff
@@ -0,0 +1,64 @@
...
+import qualified Unison.Type                   as Type
+import           Unison.Type                    ( Type )
+import qualified Unison.Typechecker as Typechecker
+import           Unison.Runtime.IOSource        ( ioReference )
+
+data MainTerm v
+  = MtNotAFunctionName String
+  | MtNotFound String
+  | MtBadType String
+  | MtSuccess HQ.HashQualified (Term.AnnotatedTerm v Ann) (Type v Ann)
```
pchiusano: Minor thing - could you remove the `Mt` prefix from constructor names. Convention we try to use is to import the module qualified so referencing the constructors is done via `MainTerm.NotAFunctionName`, etc. If we're using the names of a module a lot, we'll also sometimes give it shorter 1-2 letter qualified import, like just `MT`.

pchiusano:
@int-index I like it! Left one minor comment for you then I think this is good to merge. Can you also add yourself to CONTRIBUTORS.markdown file?

parser-typechecker/src/Unison/Codebase/Execute.hs
pchiusano: Also - appreciate this comprehensive reply that saved me from needing to dig into the code (that I don't remember super well). :)

PR #1064, opened 2019-12-14
Diff.namespace

aryairani:
Sample output:

```
.> diff.namespace ns1 ns2

  Resolved name conflicts:
  
    1.  ┌ fromJust#hs2i9lcgkd : Text
    2.  └ fromJust#jk19sm5bf8 : Nat
        ↓
    3.  fromJust#1o1iq26cq7 : Nat
        + 4.  ns1.b : Nat
  
  Updates:
  
    5.  b : Nat
        ↓
    6.  b : Text
    
    7.  c : Nat
        + 8.  c : Nat
    
    9.  fromJust' : Nat
        ↓
    10. fromJust' : Nat
        + 11. ns1.b : Nat
  
    There were 1 auto-propagated updates.
  
  Added definitions:
  
    12. unique type Y a b
    13. Y.Y  : a -> b -> Y a b
    14. ┌ d  : Nat
    15. └ d' : Nat
    16. e    : Nat
    17. f    : Nat (+1 metadata)
  
    18. patch patch (added 2 updates)
  
  Name changes:
  
    Original  Changes
    19. X     20. X' (added)
    
    21. A    22. A' (added)
```

The output is numbered and you can refer to elements of the diff by number in subsequent commands, as in `view 21` or `docs 14`.

---
- [x] Mark propagated updates as such with metadata in implementation of `propagate`
- [ ] boxLeft + 2 col + multiple lines on the RHS = 😞 
image
- [x] Audit/replace spots that use Names.diff as diff output with the new branch diff (`history`, `merge`) (started this in `topic/consolidate-diff-commands`)
  - verify diff outputs for
    - [ ] after delete
    - [ ] after undo
    - [ ] after merge
    - [ ] after merge preview
    - [ ] after pull
  - [ ] convert 3-way merge to use new branch diff
- [x] Change diff output categories to present tense to flow better in preview scenarios
- [x] Rename `Branchdiff.talladds` etc (I don't particularly care about this —arya)
- [x] Do sorting in `Output.BranchDiff.toOutput`
- [x] Add `HandleInput` implementation of `DiffNamespace`
- [x] populate the numbered args thing in handleinput
- [x] Add `Output` type for namespace diffing
- [x] Add `OutputMessages` logic for displaying the diff
- [ ] Write a test to see what existing `merge` functionality does if Alice and Bob concurrently tweak metadata associated with a definition. The existing `BranchDiff` data structure seems like it might not allow for correct handling of this.x
- [ ] kill Name2.numHashChars and its callers who are wrong
- [x] make numberedArgs use absolute names (currently they are relative to root)
- [x] eliminate solitary `[` bracket for updates because it looks weird; like a mismatched parenthesis
- [ ] should support namespace hashes as arguments — what do you do about patches, which don't have a way of being referenced by oldname / hash?

## Design issues
* For things that exist in both the old and new namespaces, the `numberedArgs` model breaks down a little.  It's okay for `view`, but suppose I have an updated patch, assigned number 7.  If I say `move.patch 7 somewhere` or `delete.patch 7`, what should happen?  How do can we specify whether we meant to change the old or new namespace?  We can just hard-code one arbitrarily(*current implementation), or display both so the user can choose by number.
 * Current thinking is: we're trying to fixup the "new" namespace to yield the diff that we want, so in these cases we'll point to the new namespace.

## Loose ends
- [ ] Try getting rid of existing `BranchDiff` and replacing it with the new version. This will require revisiting the 3 way merge implementation, which uses the diff type. Luckily we have okay tests for this.
  - [ ] Will need a `Branch.diff : Branch0 m -> Branch0 m -> BranchDiff` and an `apply : Monad m => BranchDiff -> Branch0 m -> m (Branch0 m)` (LCA calculation is unchanged)


aryairani:
Moved remaining work into #1151.

PR #1066, opened 2019-12-14
Make names0ToNames more generic

mitchellwrosen:
This function was pinned to `Name`s but it doesn't have to me. (Most functions in this module are generic over `n`)


aryairani:
It might have been intentional to give more context at the call site.  I would hold off on making this general unless we have a caller that wants to pass a different `n`.

mitchellwrosen:
Good point!

PR #1067, opened 2019-12-14
Move Unison.Codebase.NameSegment into unison-core

mitchellwrosen:
... and rename it to `Unison.NameSegment`.

# Why

There are other `Name` things in `unison-core`, so this seems like it belongs there, too.


aryairani:
Maybe just change modules and hold off on rename?

mitchellwrosen:
@aryairani Good call, done

PR #1068, opened 2019-12-15
Show branch diff after delete

mitchellwrosen:
Now deleting types/terms shows a namespace diff.

Screen Shot 2019-12-14 at 9 56 46 PM

Unfortunately when definitions with the same name, `prettyDiff` currently only reports one name:

Screen Shot 2019-12-14 at 10 01 17 PM

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -388,8 +388,11 @@ loop = do
             (failed, failedDependents) <-
               getEndangeredDependents (eval . GetDependents) toDelete rootNames
             if failed == mempty then do
-              stepManyAt . fmap (BranchUtil.makeDeleteTermName resolvedPath) . toList $ tms
-              stepManyAt . fmap (BranchUtil.makeDeleteTypeName resolvedPath) . toList $ tys
+              let makeDeleteTermNames = fmap (BranchUtil.makeDeleteTermName resolvedPath) . toList $ tms
+              let makeDeleteTypeNames = fmap (BranchUtil.makeDeleteTypeName resolvedPath) . toList $ tys
+              stepManyAt (makeDeleteTermNames ++ makeDeleteTypeNames)
+              root'' <- use root
+              respond $ ShowDiff input (Branch.namesDiff root' root'')
             else do
               failed <-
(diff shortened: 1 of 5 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -388,8 +388,11 @@ loop = do
             (failed, failedDependents) <-
               getEndangeredDependents (eval . GetDependents) toDelete rootNames
             if failed == mempty then do
-              stepManyAt . fmap (BranchUtil.makeDeleteTermName resolvedPath) . toList $ tms
-              stepManyAt . fmap (BranchUtil.makeDeleteTypeName resolvedPath) . toList $ tys
+              let makeDeleteTermNames = fmap (BranchUtil.makeDeleteTermName resolvedPath) . toList $ tms
+              let makeDeleteTypeNames = fmap (BranchUtil.makeDeleteTypeName resolvedPath) . toList $ tys
+              stepManyAt (makeDeleteTermNames ++ makeDeleteTypeNames)
```
mitchellwrosen: I think this fixes a small bug - the branch was stepped twice before (so two history/reflog entries)

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -388,8 +388,11 @@ loop = do
...
+              root'' <- use root
+              respond $ ShowDiff input (Branch.namesDiff root' root'')
```
mitchellwrosen: Is grabbing the new root and diffing expensive? I noticed it's done differently in `delete.namespace`:
```haskell
            let deletedNames =
                  Names.prefix0
                    (Path.toName' (Path.unsplit' p))
                    (Branch.toNames0 b)
                diff = Names3.diff0 deletedNames mempty
            respond $ ShowDiff input diff
```

aryairani: Guessing the latter was just reusing ShowDiff without thinking.  It shouldn’t be particularly expensive.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: Weird; great.

PR #1069, opened 2019-12-15
Show term/type hashes (for adds/removes) in diff output, if ambiguous

mitchellwrosen:
Before (with #1068 applied):

Screen Shot 2019-12-14 at 11 06 57 PM

After:

Screen Shot 2019-12-14 at 11 03 18 PM

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
+++ b/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
@@ -1303,21 +1303,28 @@ prettyDiff diff = let
   orig = Names.originalNames diff
   adds = Names.addedNames diff
   removes = Names.removedNames diff
-  addedTerms = [ n | (n,r) <- R.toList (Names.terms0 adds)
-                   , not $ R.memberRan r (Names.terms0 removes) ]
-  addedTypes = [ n | (n,r) <- R.toList (Names.types0 adds)
-                   , not $ R.memberRan r (Names.types0 removes) ]
-  added = Name.sortNames . nubOrd $ (addedTerms <> addedTypes)
 
-  removedTerms = [ n | (n,r) <- R.toList (Names.terms0 removes)
-                     , not $ R.memberRan r (Names.terms0 adds)
-                     , Set.notMember n addedTermsSet ] where
-    addedTermsSet = Set.fromList addedTerms
-  removedTypes = [ n | (n,r) <- R.toList (Names.types0 removes)
-                     , not $ R.memberRan r (Names.types0 adds)
-                     , Set.notMember n addedTypesSet ] where
-    addedTypesSet = Set.fromList addedTypes
-  removed = Name.sortNames . nubOrd $ (removedTerms <> removedTypes)
+  addedTerms = [ (n,r) | (n,r) <- R.toList (Names.terms0 adds)
+                       , not $ R.memberRan r (Names.terms0 removes) ]
+  addedTypes = [ (n,r) | (n,r) <- R.toList (Names.types0 adds)
+                       , not $ R.memberRan r (Names.types0 removes) ]
+  added = HQ'.sort (hqTerms ++ hqTypes)
+    where
+      hqTerms = [ Names.hqName adds n (Right r) | (n, r) <- addedTerms ]
+      hqTypes = [ Names.hqName adds n (Left r)  | (n, r) <- addedTypes ]
+
+  removedTerms = [ (n,r) | (n,r) <- R.toList (Names.terms0 removes)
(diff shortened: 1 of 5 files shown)


mitchellwrosen:
Looking over the transcript output, I now see the point of `ordNub`bing the list of names. With this patch, if you do an update, you'll see it rendered as an add and a delete:

Screen Shot 2019-12-15 at 8 51 30 AM

🤔 😞

mitchellwrosen:
Better now.

aryairani:
I'm having trouble following what's changing, relative to the thread comments and force-pushes throwing away history.  What's the current status? :)

aryairani:
> Looking over the transcript output, I now see the point of `ordNub`bing the list of names.

Good case for having a comment in future code for stuff like why `ordNub` was being called; I definitely don't remember.

aryairani:
> > Looking over the transcript output, I now see the point of `ordNub`bing the list of names.
> 
> Good case for having a comment in future code for stuff like why `ordNub` was being called; I definitely don't remember. 😅

I'm not sure if `ordNub` was doing this; it looks like `Set.notMember n addedTermsSet` was doing this.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1303,21 +1303,28 @@ prettyDiff diff = let
...
+
+  removedTerms = [ (n,r) | (n,r) <- R.toList (Names.terms0 removes)
+                         , not $ R.memberRan r (Names.terms0 adds)
+                         , Set.notMember n addedTermsSet ] where
+    addedTermsSet = Set.fromList (map fst addedTerms)
+  removedTypes = [ (n,r) | (n,r) <- R.toList (Names.types0 removes)
+                         , not $ R.memberRan r (Names.types0 adds)
+                         , Set.notMember n addedTypesSet ] where
+    addedTypesSet = Set.fromList (map fst addedTypes)
+  removed = HQ'.sort (hqTerms ++ hqTypes)
```
aryairani: Ok, so the summary of this change is to remember the references associated with the added/removed names.

mitchellwrosen: Yes, that's correct. The "adds/updates" and "deletes" parts of a diff are now strictly more informative. Prior to this patch, they did not really nicely handle either of these two cases:

- Actually conflicting terms or conflicting types
- "Visually" conflicting (meaning, where term and type namespaces collide)

And now, in both of the above cases you'd see a hash qualified name.

unison-core/src/Unison/Names2.hs
```diff
@@ -216,6 +217,26 @@ addType n r = (<> fromTypes [(n, r)])
...
+--
+-- This is useful in contexts such as printing branch diffs. Example:
+--
+--     - Deletes:
+--
+--       foo
+--       foo
+--
+-- We want to append the hash regardless of whether or not one is a term and the
+-- other is a type.
```
aryairani: No, I think we want:
```
   - Deletes:

     foo (Type)
     foo (Term)
```
rather than hashes, right?  Actually, probably all of the deletes should indicate or be grouped by term/type.

aryairani: I'm also okay with merging this as-is but with a separate ticket to track this— this would be the first instance where hash-qualifying is used as substitute for distinguishing Term from Type, which I think is the wrong direction.

mitchellwrosen:
> I'm having trouble following what's changing, relative to the thread comments and force-pushes throwing away history. What's the current status? :)

Sorry, the force pushing is just because the changes have not been that interesting ("fix typo", "re-generate transcript files"). The summary of the current status is:

- "Adds/updates" and "deletes" summaries of a diff are more informative

unison-core/src/Unison/Names2.hs
aryairani: Thinking it over some more, I dunno.  Hash-qualified is an improvement over indistinguishable "name-only"s, so this is fine for now; I dunno what the answer is in the long term.  My guess is it's "not have multiple parallel namespaces".

mitchellwrosen: I agree having separate "Deleted terms" and "Deleted types" stanzas would improve this situation. The current pretty-printer actually just space-separates things, so this might look a bit weird:

```
- Deletes:

    foo (Type) foo (Term) bar
```

aryairani: I see, yeah.

aryairani: Let's just — I dunno whether in a ticket or in comments in some appropriate place — remember that this is something we might be able to improve later.

PR #1076, opened 2019-12-15
Make `delete` tab completion also list types

### LINKED ISSUE #1074: `delete` tab completion is wrong
I have two types `Magic` and `Memory`.

If I write `delete.type M` and then tab-complete, I see:

Screen Shot 2019-12-15 at 12 12 26 PM

which is correct. But if I instead tab-complete `delete M` I get:

Screen Shot 2019-12-15 at 12 13 40 PM

(so it missed `Magic` somehow)

mitchellwrosen:
Fixes #1074


pchiusano:
Good call, merging...

PR #1077, opened 2019-12-16
Better duplicate definitions error message

mitchellwrosen:
Example output, first add is

```
a = 1
a = 2
```

and the second is

```
a = 1
a = 2
b = 3
b = 4
```

Screen Shot 2019-12-15 at 11 35 08 PM

It could be improved (like a lot)... but is better than nothing. I tried putting some source info alongside the names; verdict: didn't look that great and the spans were kind of inaccurate. But feedback welcome on how to style this message, of course.


pchiusano:
Great! Any little improvement like this is a win.

Issue #1028, opened 2019-12-06
Add ucm command to preview adds/updates 


anovstrup:
I find myself touching my scratch file at times just to get ucm to echo the output that it produces when a scratch file is saved. It would be nice if there were a ucm command that would produce this output on demand, without having to save/touch a scratch file. (It can be useful when the output from other intervening ucm commands has "hidden" the output produced after the most recent save or the codebase/namespace has changed in ways that alter the interpretation of the current scratch code.)

This [was discussed](https://unisonlanguage.slack.com/archives/CLKV43YE4/p1575669593027300) briefly on the Slack channel, and some possible command names were thrown around: `load`, `status`, `preview`/`add.preview`/`update.preview`.

aryairani:
Hi @anovstrup @pchiusano, let's decide exactly what this would do

Option 1. Preserve and re-display the typechecking results from last time (shows you "hidden" output again)

Option 2: re-parse/typecheck the previously typechecked file (takes into account namespace changes)

anovstrup:
I'd prefer Option 2, as that was part of my motivation for this suggestion in the first place. We could potentially avoid reparsing/type-checking when it really isn't necessary, but that's an implementation detail that should be invisible to the user, IMO.  I want to know what's going to happen to my codebase if I `add` or `update` _now_ (even if that's changed since the last time I touched/saved my scratch file).

aryairani:
Just to clarify: as things currently stand, I believe that if you save your scratch file and ucm typechecks and prints the status, then you change the namespace, and then you `add`/`update`, the changed namespace doesn't affect the outcome of the `add`/`update`.  Does this change anything for you?

anovstrup:
Well, it would change my mental model. 😂

anovstrup:
But, actually, I don't think that's correct. Consider:

In scratch.u:
```
add1 x = x + 1
```

In ucm:
```
add
cd childns
```

In scratch.u:
```
add2 x = (add1 . add1) x
```
(ucm reports errors, because it can't resolve `add1`)

In ucm, `cd..` (notionally) then `add`.

(ucm successfully adds `add2`)

anovstrup:
And, for what it's worth, this behavior is exactly what I would expect—that `add` or `update` applies my changes to the _current_ state of the codebase/namespace, regardless of the state it was in when I saved/touched the last scratch file.

pchiusano:
[...]

Issue #1027, opened 2019-12-06
don't use `less` for typechecking output


aryairani:
(i.e. commands that are initiated externally/without focus)

atacratic:
We'd still want a way to view it with less though, right? Maybe that's the same command as the 'preview add/update' command that people requested (`load`?)

anovstrup:
I didn't quite understand this one. What's the real problem it's trying to address?

I've had the experience of saving a scratch file when ucm was in `less` mode, which is admittedly confusing—I think the behavior is that after you `q` `less`, ucm will then render the output from loading the scratch file (potentially putting you back into `less`). To address that, my preference would be for ucm to drop out of `less` automatically when a scratch file is touched.

aryairani:
@anovstrup Yeah I think that was the actual motivation, and I should’ve done a better job of writing it up.

But should it kill any `less` or just a typechecking `less`?  Maybe any is okay.  It would be nice if there were some signal to have less finish its output before exiting, but maybe it doesn’t matter.

anovstrup:
@aryairani I guess my expectation is that if I save a scratch file, I'll see the resulting output in ucm right away regardless of whether another command's output was being displayed in less (assuming I've explicitly decided to save my scratch file; no aggressive auto-save doing it for me!). So I'd say the right path is to always kill less, but I agree that it would be best if we can have less dump the remaining output to the buffer first.

anovstrup:
@aryairani Question: is ucm actually spawning a `less` process? Or is it just emulating less with a library or something?

aryairani:
@anovstrup That seems like a reasonable expectation.

Yes, ucm is actually spawning a `less` process; I couldn't find a library that emulates it.  This happens in `Less.hs`.

I guess you'd somehow need to return a callback to kill `less`, and thread it back through the Output command (or a copy) — maybe Output returns a cancelOutput command, that you can call to interrupt a modal display? :-\

anovstrup:
@aryairani I spent some time on this over the weekend and can now appreciate the motivation for the simpler solution you proposed. 😛

I got stuck on a couple problems:

[...]

PR #1087, opened 2019-12-21
Don’t use `less` to display ucm output triggered by external events (#1027)

anovstrup:
This PR implements a fix for #1027 as originally specified (i.e., avoiding the use of `less` for ucm output triggered by external events).  

It would arguably be preferable to use `less` for all long output but kill any running `less` process (and cancel any pending outputs) when an external event occurs. However, I ran into various snags implementing that solution (see my comments on #1027 for details). This fix is thus a temporary workaround. Combined with #1028 (which would provide a way to view the typechecking output in `less` on demand), it may suffice for now.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
@@ -61,6 +61,9 @@ data Command m i v a where
   -- Presents some output to the user
   Notify :: Output v -> Command m i v ()
 
+  -- Presents output to the user (non-modal/paged)
+  Notify' :: Output v -> Command m i v ()
+
   -- literally just write some terms and types .unison/{terms,types}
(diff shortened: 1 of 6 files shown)


parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -61,6 +61,9 @@ data Command m i v a where
   -- Presents some output to the user
   Notify :: Output v -> Command m i v ()
 
+  -- Presents output to the user (non-modal/paged)
+  Notify' :: Output v -> Command m i v ()
```
aryairani: Would you mind renaming this to maybe `NotifyNonpaged`?  I think we are already suffering too many `'`-suffixed definitions. 😅

anovstrup: Fixed!

Issue #1029, opened 2019-12-06
lag on `todo`; are we eagerly fetching historical names for things that aren't going to be displayed anyway?


aryairani:
We can maybe ask @seagreen for a copy of his hardtack repo if the problem isn't immediately obvious from one we make up.

mitchellwrosen:
Here's the slow line: https://github.com/unisonweb/unison/blob/67254cc36f17c4715750640d004ae4b5a9cf410b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs#L2190

pchiusano:
That’s helpful thanks @mitchellwrosen 

Fetching historical names should be fast for things in the recent past, but the slowness we were seeing seemed like a full history search.

aryairani:
possibly related to #781

aryairani:
I suppose the very idea of the historical search is flawed (because even in cases where we actually need to search through the whole history, we don't want to); however, maybe we can identify something that will help in this specific case.

aryairani:
@mitchellwrosen Assuming that `Branch.findHistoricalRefs` isn't broken (and I'm guessing it isn't), then either 
1. `deps` includes something that it ought not include, and we should figure out how to remove the thing, or
2. everything in `deps` does need to be there, and the search is just gonna be slow.

The ticket topic supposes it's #1, in which case we need a more appropriate value for `deps`.

mitchellwrosen:
@aryairani Are you able to reproduce this in any of your namespaces? If not I would be happy to publish mine somewhere for you to poke at

aryairani:
@mitchellwrosen No, I can't reproduce in mine.  If you could publish yours, that would be great.

mitchellwrosen:
@aryairani Here ya go https://github.com/mitchellwrosen/unison-slow-todo-branch

runarorama:
This is coming up in testing the PR workflow. Running `todo` takes literally minutes, likely related to #781

PR #1088, opened 2019-12-21
Speed up todo output if there is nothing to do

mitchellwrosen:
This patch speeds up `todo` output when there is nothing to do.

- `showTodoOutput` now takes an action that computes a `PrettyPrintEnv` rather than a `PrettyPrintEnv` directly (because it's slow to compute, and only needed if there are things to do)
- `NothingTodo` output message was reused for this output case (rather than making a new `NoTodoOutput` or similar)
- Calculations for whether or not a `TodoOutput` has any conflicts/edits were moved to the `TodoOutput` so they could be called from more than one place.

Maybe mostly addresses https://github.com/unisonweb/unison/issues/1029?

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -1450,7 +1450,7 @@ propagatePatch input inputDescription patch scopePath = do
 
 -- | Show todo output if there are any conflicts or edits.
 showTodoOutput
-  :: _
+  :: Input
   -> Action' m v PPE.PrettyPrintEnvDecl
      -- ^ Action that fetches the pretty print env. It's expensive because it
      -- involves looking up historical names, so only call it if necessary.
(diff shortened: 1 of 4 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -472,7 +472,7 @@ loop = do
           if b then do
             respond (ShowDiff input (Branch.namesDiff destb merged))
             patch <- getPatchAt defaultPatchPath
-            void $ propagatePatch inputDescription patch dest
+            void $ propagatePatch input inputDescription patch dest
```
mitchellwrosen: `input` was piped around here and elsewhere for displaying the `NothingTodo` case.

mitchellwrosen:
Happy consequence: `update`, which also seems to call `todo` for you at the end, is now much faster when your edit frontier is empty (which, in my experience, is most of the time)

aryairani:
What do you think about `NoTodo` in general?  It bugs me a little that we have to thread `input` through.  Should `propagatePatch` even generate output messages, or should it be part of the return value?

mitchellwrosen:
I guess I'm not a fan of `NoTodo` as opposed to just having different constructors for different purposes. Agreed that feeding the input to the output, just so the output can switch on it and tweak the display, is not great. Want to see that changed here, or in a different patch?

aryairani:
Let's do it here if you don't mind, and reduce the diff — I think `NoTodo` was a mistake; we can address its old cases in a separate patch.

mitchellwrosen:
@aryairani done!

PR #1094, opened 2019-12-24
Import NameSegment from Name

mitchellwrosen:
- Import `NameSegment` from `Name` rather than the other way around (to define `Name` as a list of `NameSegment`)
- Move `HQSegment` definition into `HashQualified'` because of circular imports


aryairani:
Oh, heh.  I thought this was merging into master, and I was confused about why CI didn't run.  Though, I wonder why CI didn't run.

PR #1097, opened 2019-12-26
Fill in a few name functions

mitchellwrosen:
Along the way, I deleted some unused stuff.

### DIFF
--- a/unison-core/src/Unison/Name.hs
+++ b/unison-core/src/Unison/Name.hs
@@ -1,4 +1,7 @@
+{-# LANGUAGE BlockArguments      #-}
 {-# LANGUAGE OverloadedStrings   #-}
+
+{-# OPTIONS_GHC -Wwarn #-} -- temporary
 
 module Unison.Name
   ( Name
@@ -8,18 +11,14 @@ module Unison.Name
   , joinDot
   , makeAbsolute
   , parent
-  , segments
-  , sortNames
   , sortNamed
   , sortNamed'
   , stripNamePrefix
-  , stripPrefixes
   , suffixes
   , toString
   , toText
   , toVar
   , unqualified
-  , unqualified'
   , unsafeFromText
   , unsafeFromString
   , fromVar
@@ -29,9 +28,11 @@ where
 import Unison.Prelude
 
 import           Control.Lens                   ( unsnoc )
-import           Data.List.NonEmpty             ( NonEmpty )
+import qualified Data.List                     as List
+import           Data.List.NonEmpty             ( NonEmpty((:|)) )
+import qualified Data.List.NonEmpty            as NonEmpty
 import qualified Data.Text                     as Text
-import           Unison.Codebase.NameSegment    ( NameSegment(NameSegment) )
+import           Unison.Codebase.NameSegment    ( NameSegment )
 import qualified Unison.Codebase.NameSegment   as NameSegment
 import qualified Unison.Hashable               as H
 import           Unison.Var                     ( Var )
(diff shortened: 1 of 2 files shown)


unison-core/src/Unison/Name.hs
```diff
@@ -46,16 +47,12 @@ data Name
 
 toText :: Name -> Text
 toText (Name name) = name
+toText (Name' names) =
```
mitchellwrosen: This is kinda sketchy but I think it preserves how "name to text" currently works. We just puts dots between the name segments. I guess a `NameSegment` must not contain a `.` character? (If so that should probably be written down somewhere :))

unison-core/src/Unison/Name.hs
```diff
@@ -87,38 +84,33 @@ toString :: Name -> String
...
 -- stripNamePrefix . .Nat = Just Nat
 stripNamePrefix :: Name -> Name -> Maybe Name
-stripNamePrefix prefix name =
+stripNamePrefix prefix@Name{} name@Name{} =
   Name <$> Text.stripPrefix (toText prefix <> mid) (toText name)
   where
   mid = if toText prefix == "." then "" else "."
-
--- a.b.c.d -> d
-stripPrefixes :: Name -> Name
```
mitchellwrosen: This was deleted because it's also exported as `unqualified` (and no one was calling this ver)

unison-core/src/Unison/Name.hs
```diff
@@ -130,15 +122,16 @@ parent (Name txt) = case unsnoc (Text.splitOn "." txt) of
...
 suffixes :: Name -> [Name]
 suffixes (Name n) =
   fmap up . tails . dropWhile (== "") $ Text.splitOn "." n
   where
   up ns = Name (Text.intercalate "." ns)
-
-unqualified' :: Text -> Text
-unqualified' = last . Text.splitOn "."
+suffixes (Name' names) =
+  map (Name' . NonEmpty.fromList) (NonEmpty.init (NonEmpty.tails names))
```
mitchellwrosen: `NonEmpty.fromList` is safe here because `NonEmpty.init . NonEmpty.tails` contains no empty lists.

unison-core/src/Unison/Name.hs
aryairani: Side note: Not sure where this gets used, but it seems syntax-specific (#499).  

IMO, `NameSegment` should be allowed to have a `.`, but we'd need a way to escape / unescape the `NameSegment` according to the syntax in use.

unison-core/src/Unison/Name.hs
```diff
@@ -87,38 +84,33 @@ toString :: Name -> String
...
+Name' a `isPrefixOf` Name' b = toList a `NonEmpty.isPrefixOf` b
 
 -- stripNamePrefix a.b  a.b.c = Just c
 -- stripNamePrefix a.b. a.b.c = undefined, "a.b." isn't a valid name IMO
 -- stripNamePrefix x.y  a.b.c = Nothing, x.y isn't a prefix of a.b.c
 -- stripNamePrefix "" a.b.c = undefined, "" isn't a valid name IMO
 -- stripNamePrefix . .Nat = Just Nat
 stripNamePrefix :: Name -> Name -> Maybe Name
-stripNamePrefix prefix name =
+stripNamePrefix prefix@Name{} name@Name{} =
```
aryairani: Do we need the other two combinations too?

aryairani: I guess not since we aren't trying to run anything, just test out compilation with the new constructor.

unison-core/src/Unison/Name.hs
mitchellwrosen: Totally agree.

unison-core/src/Unison/Name.hs
mitchellwrosen: Yeah, the other two cases seem less useful to get down since the original constructor is to be deleted by the time this branch is ready.

unison-core/src/Unison/Name.hs
aryairani: Right now, a tangle of parsers enforces the "no `.` in `NameSegment`" rule.

PR #1099, opened 2019-12-27
Add load command

anovstrup:
This PR implements `load`, `add.preview`, and `update.preview` commands (see issue #1028). It also reimplements the fix for #1027, moving the responsibility for managing whether output is paged into Main.hs.

**Implementation notes**
* The `load` command is implemented in terms of an action `SourceName -> IO LoadSourceResult` that is passed in to `HandleCommand.commandLine`.  For `CommandLine/Main.hs`, this action loads the given file from disk (failing with a InvalidSourceNameError if the filename does not have a ".u" extension or the file does not exist, or with a LoadError if any other IOException occurs while reading the file). For the transcript parser, the action reevaluates the most recent unison block of the given name.
* The implementation adds additional state to `CommandLine/Main.hs` to manage whether output is paged. Usually, it is paged for command line inputs and not for external events. ~~However, since `load` uses the file watch system, this ordinary behavior must be overridden for the first external event after a file is touched for a `load` so that output triggered by `load` _is_ paged.~~
* ~~The implementation uses `System.Posix.touchFile`, requiring a build dependency on `unix`.  I tried to implement `touchFile` in terms of `getCurrentTime`, `setAccessTime` and `setModificationTime`, but this resulted in a linker issue on my system, apparently stemming from [an upstream issue](https://stackoverflow.com/questions/56029761/undefined-symbols-for-architecture-x86-64-utimensat-referenced-from-cazw). We probably want to avoid this additional dependency and use the alternative implementation (which I commented out).~~
* The two preview commands show the results for the most recent (successfully) typechecked file. As with `add` and `update`, an intervening parsing/typechecking failure does _not_ invalidate the cached typechecked result. Not sure whether that's a feature or a bug, but it could be confusing and probably at least warrants some kind of warning.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
+++ b/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
@@ -111,6 +111,7 @@ run dir configFile stanzas codebase = do
     numberedArgsRef          <- newIORef []
     inputQueue               <- Q.newIO
     cmdQueue                 <- Q.newIO
+    unisonFiles              <- newIORef Map.empty
     out                      <- newIORef mempty
     hidden                   <- newIORef False
     allowErrors              <- newIORef False
@@ -132,15 +133,15 @@ run dir configFile stanzas codebase = do
       awaitInput = do
         cmd <- atomically (Q.tryDequeue cmdQueue)
         case cmd of
-          Just Nothing -> do
+          Just (Right Nothing) -> do
             output "\n```\n" -- this ends the ucm block
             writeIORef hidden False
             awaitInput
-          Just (Just p@(UcmCommand path lineTxt)) -> do
+          Just (Right (Just p@(UcmCommand path lineTxt))) -> do
             curPath <- readIORef pathRef
             numberedArgs <- readIORef numberedArgsRef
             if curPath /= path then do
-              atomically $ Q.undequeue cmdQueue (Just p)
+              atomically $ Q.undequeue cmdQueue (Right $ Just p)
               pure $ Right (SwitchBranchI (Path.absoluteToPath' path))
             else case (>>= expandNumber numberedArgs)
                        . words . Text.unpack $ lineTxt of
@@ -156,6 +157,9 @@ run dir configFile stanzas codebase = do
                       output $ P.toPlain 65 (P.indentN 2 msg <> P.newline <> P.newline)
                       die
                     Right input -> pure $ Right input
+          Just (Left event) ->
+            pure $ Left event
+
           Nothing -> do
             errOk <- readIORef allowErrors
             hasErr <- readIORef hasErrors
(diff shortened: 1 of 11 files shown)


pchiusano:
@anovstrup Thanks for this and for the writeup. 

I like idea of this, but my concern with this one is that I don't want to incur a unix dependency (even though UCM is currently busted on Windows #598, I don't want to add more debt here). Can you come up with something platform independent?

There's definitely a more direct implementation of `load` that doesn't involve touching files, but it might require more knowledge of how `HandleInput` is structured. But basically, it would be factoring out the logic that's currently triggered in response to that file change event, so it can be invoked explicitly. @anovstrup if there's no easy platform independent way to fix your implementation, would you like to investigate if you can implement the command more directly?

anovstrup:
@pchiusano There's a (sort of) platform independent fix for my implementation (using `setAccessTime` and `setModificationTime` instead of the unix-dependent `touchFile`)—it's actually commented out in the PR.  The caveat is that it introduces a linker error on older Macs (Sierra or older), like mine, due to an upstream issue (the standard library uses a system call (for `getCurrentTime`, I think) that's not available on these systems), although maybe that's fixable by using an older version of GHC on such systems. I could look into that possibility. Do we know the minimum GHC version that Unison requires?

I considered the alternative of not relying on the file watch system for this, but as you say it would definitely require some refactoring.  I think I've got a decent handle on how `HandleInput` works; I just hoped to avoid a more complex change if it wasn't necessary, but I can look into it further if this touch-based implementation doesn't pan out.

aryairani:
I'd also vote to skip "touch" altogether, and just load the file directly instead of rube-goldberging it :)

parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
```diff
@@ -111,6 +112,7 @@ run dir configFile stanzas codebase = do
     numberedArgsRef          <- newIORef []
     inputQueue               <- Q.newIO
     cmdQueue                 <- Q.newIO
+    unisonFiles              <- newIORef Map.empty
```
aryairani: Could you explain `unisonFiles` and how this works?  Which commands depend on it?

anovstrup:
This transcript demonstrates how the load command works with transcripts:
[test.md](https://github.com/unisonweb/unison/files/4028866/test.txt)

parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
anovstrup: Explain in code comments? 

`unisonFiles` stores the contents of each of the unison stanzas, keyed by "filename", so that the `load` command works as expected in transcripts.

aryairani: Oh wow. Good thinking.  I didn't notice that it was specific to the transcript runner.

aryairani:
@anovstrup Thanks for putting this together!

PR #1100, opened 2019-12-28
Fill out Name API a bit

mitchellwrosen:
This patch tries to reduce the number of calls to `Name.toText` by using other bits of the `Name` API instead. I couldn't get rid of all `Name.toText` calls - for some stuff like pretty-printing it seems perhaps necessary (for now anyway), but there are also some remaining calls that I was unsure what to do with in some typechecker code.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -1800,11 +1800,8 @@ applySelection hqs file = \sr@SlurpResult{..} ->
   selection = SlurpComponent selectedTypes selectedTerms
   closed = SC.closeWithDependencies file selection
   selectedTypes, selectedTerms :: Set v
-  selectedTypes = Set.map var $ R.dom (Names.types selectedNames0)
-  selectedTerms = Set.map var $ R.dom (Names.terms selectedNames0)
-
-var :: Var v => Name -> v
-var name = Var.named (Name.toText name)
+  selectedTypes = Set.map Name.toVar $ R.dom (Names.types selectedNames0)
+  selectedTerms = Set.map Name.toVar $ R.dom (Names.terms selectedNames0)
 
 toSlurpResult
   :: forall v
@@ -1831,8 +1828,8 @@ toSlurpResult currentPath uf existingNames =
   fileNames0 = UF.typecheckedToNames0 uf
 
   sc :: R.Relation Name Referent -> R.Relation Name Reference -> SlurpComponent v
-  sc terms types = SlurpComponent { terms = Set.map var (R.dom terms)
-                                  , types = Set.map var (R.dom types) }
+  sc terms types = SlurpComponent { terms = Set.map Name.toVar (R.dom terms)
+                                  , types = Set.map Name.toVar (R.dom types) }
 
   -- conflict (n,r) if n is conflicted in names0
   conflicts :: SlurpComponent v
@@ -1851,7 +1848,7 @@ toSlurpResult currentPath uf existingNames =
   -- r is Ref and r' is Con
   termCtorCollisions :: Set v
   termCtorCollisions = Set.fromList
-    [ var n
+    [ Name.toVar n
     | (n, Referent.Ref{}) <- R.toList (Names.terms fileNames0)
     , [r@Referent.Con{}]  <- [toList $ Names.termsNamed existingNames n]
     -- ignore collisions w/ ctors of types being updated
@@ -1872,7 +1869,7 @@ toSlurpResult currentPath uf existingNames =
   -- what if (n,r) and (n,r' /= r) exists in names and r, r' are Con
   ctorTermCollisions :: Set v
   ctorTermCollisions = Set.fromList
-    [ var n
+    [ Name.toVar n
     | (n, Referent.Con{}) <- R.toList (Names.terms fileNames0)
     , r                   <- toList $ Names.termsNamed existingNames n
     -- ignore collisions w/ ctors of types being updated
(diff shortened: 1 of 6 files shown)


aryairani:
> for some stuff like pretty-printing it seems perhaps necessary (for now anyway)

Yeah!  `Name.toText` will probably need to be part of the pluggable syntax interface or implementation.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -271,7 +271,7 @@ uncons p = case p of
 -- todo: fromName needs to be a little more complicated if we want to allow
 --       identifiers called Function.(.)
 fromName :: Name -> Path
-fromName = fromList . fmap NameSegment . Text.splitOn "." . Name.toText
+fromName = fromList . Name.segments
```
aryairani: 🏆

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -709,14 +708,7 @@ countHQ hq = fold $ fmap countName (HQ.toName $ hq)
...
 countName n = let f = \(p, s) -> (s, Map.singleton p 1)
-              in PrintAnnotation { usages = Map.fromList $ map f $ splitName n}
-
-splitName :: Name -> [(Prefix, Suffix)]
-splitName n = let ns = splitOn "." (Name.toText n)
-              in dropEnd 1 ((inits ns) `zip` (map dotConcat $ tails ns))
--- > splitName "x" == [([], "x")]
--- > splitName "A.x" == [(["A"], "x")]
--- > splitName "A.B.x" == [(["A"], "B.x"), (["A.B"], "x")]
+              in PrintAnnotation { usages = Map.fromList $ map f $ Name.oldSplits n}
```
aryairani: Thanks, git diff.

unison-core/src/Unison/Name.hs
```diff
@@ -134,9 +144,33 @@ suffixes (Name' names) =
   map (Name' . NonEmpty.fromList) (NonEmpty.init (NonEmpty.tails names))
 
 makeAbsolute :: Name -> Name
-makeAbsolute n | toText n == "."                = Name ".."
-               | Text.isPrefixOf "." (toText n) = n
-               | otherwise                      = Name ("." <> toText n)
+makeAbsolute n | toText n == "." = Name ".."
+               | isAbsolute n    = n
+               | otherwise       = Name ("." <> toText n)
```
aryairani: This could just be setting the flag, though I actually wonder how it's being used, and whether it should be used at all.

unison-core/src/Unison/Name.hs
```diff
@@ -134,9 +144,33 @@ suffixes (Name' names) =
...
+splits =
+  map (map NameSegment *** unsafeFromText) . oldSplits
+
+-- > oldSplits "x" == [([], "x")]
+-- > oldSplits "A.x" == [(["A"], "x")]
+-- > oldSplits "A.B.x" == [(["A"], "B.x"), (["A.B"], "x")]
+oldSplits :: Name -> [([Text], Text)]
+oldSplits (Name n) = let ns = Text.splitOn "." n
+                     in dropEnd 1 (inits ns `zip` (map dotConcat $ tails ns))
+  where dotConcat = Text.concat . intersperse "."
```
aryairani: I wonder where these are used.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1800,11 +1800,8 @@ applySelection hqs file = \sr@SlurpResult{..} ->
   selection = SlurpComponent selectedTypes selectedTerms
   closed = SC.closeWithDependencies file selection
   selectedTypes, selectedTerms :: Set v
-  selectedTypes = Set.map var $ R.dom (Names.types selectedNames0)
-  selectedTerms = Set.map var $ R.dom (Names.terms selectedNames0)
-
-var :: Var v => Name -> v
-var name = Var.named (Name.toText name)
+  selectedTypes = Set.map Name.toVar $ R.dom (Names.types selectedNames0)
+  selectedTerms = Set.map Name.toVar $ R.dom (Names.terms selectedNames0)
```
aryairani: This code isn't actually being used (although it might be in the future, if there is demand for being able to add individual or few definitions from a .u file, rather than slurping them all at once.)  In other words, I think line 1790 is always hit, and not 1791.

aryairani: Moreover, it looks like these are only dealing with `Name` because `UF.typecheckedToNames0` uses `Name` because `Names.filterByHQs` uses `Name`.  However, I don't think there's anything about the filtering that necessarily involves `Name`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1831,8 +1828,8 @@ toSlurpResult currentPath uf existingNames =
   fileNames0 = UF.typecheckedToNames0 uf
 
   sc :: R.Relation Name Referent -> R.Relation Name Reference -> SlurpComponent v
-  sc terms types = SlurpComponent { terms = Set.map var (R.dom terms)
-                                  , types = Set.map var (R.dom types) }
+  sc terms types = SlurpComponent { terms = Set.map Name.toVar (R.dom terms)
+                                  , types = Set.map Name.toVar (R.dom types) }
```
aryairani: Same here: maybe we can skip `Name`, but we'd have to dig into the rest of the function.

aryairani: Maybe we would benefit from using `Var v => Names' v` rather than `Names' Name` aka `Names2.Names0` in these functions; it's never been done before!  I don't think `Names'` existed when this code was written.

unison-core/src/Unison/Name.hs
mitchellwrosen: `oldSplits` is used in the term printer. I initially moved it into `Name` and called it `splits`, which (IMO) has a more correct type, but left `oldSplits` that just returns `([Text], Text)` so I didn't have to refactor that term printer code. I think `oldSplits` should be replaced by `splits` eventually.

unison-core/src/Unison/Name.hs
mitchellwrosen: It does seem to be used, maybe not a lot, but I couldn't easily refactor the call sites without thinking :)

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
mitchellwrosen: > This code isn't actually being used (although it might be in the future, if there is demand for being able to add individual or few definitions from a .u file, rather than slurping them all at once.)

Can't you apply a filter with e.g. `add <term>`?

aryairani: Oops, you sure can.  I was thinking it wasn't fully implemented.

PR #1103, opened 2019-12-29
Make NameSegment an abstract type

mitchellwrosen:


aryairani:
This all seems reasonable, but what was the motivation?

mitchellwrosen:
I think I was just chasing down how `NameSegment` was being constructed in preparation for `Name = NonEmpty NameSegment`. I wouldn't say this is an important patch to merge but I think newtypes with invariants (in `NameSegment`'s case, they are syntax-specific) are nice to make abstract with some "unsafe" construction functions.

PR #1104, opened 2019-12-29
Remove Name.fromString, Name.unsafeFromString

mitchellwrosen:
The goal here is to make `Name.unsafeFromText` the one unsafe function to construct a name. I imagine it should only be called by `Name` parsers, and everyone else should work with the abstract type.


mitchellwrosen:
> Did I understand correctly that it's not that we don't want *string* convenience functions, but that we're just minimize the API for now.?

Yeah, that's what I was thinking. I think it's better to just convert to `Text` before making a name rather than proliferate `unsafeFrom*` functions. And yeah - "for now" - exactly. I think it could go back before this patch is completed but having one symbol to grep for makes it easier to see how `Name` is being used right now.

PR #1105, opened 2019-12-30
Remove more calls to Name.toVar

mitchellwrosen:
I'm not very confident about this one because I don't really understand the relationship between `Name` and `Var v => v` especially as it relates to pluggable syntax.

This patch is based on the assumption that calling `Name.toVar` is weird/smelly:

- "Names" start out life as `User` vars.
- However, we are often parsing directly as `Name` and then calling `Name.toVar` as necessary, especially around the "unison file" boundary where we are cross-referencing data structures full of `Name`s (like `Names0`) and the freshly parsed/typechecked code.
- It seems better to delay conversion from `Var v => v` to `Name` until as late as possible, and treat the other direction (`Name` to `Var v => v` via `Name.toVar`) as a code smell that's currently everywhere simply due to convenience (and lack of a second syntax to worry about)

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -54,7 +54,7 @@ import qualified Unison.Util.Pretty as Pr
 import Unison.Util.Pretty (Pretty, ColorText)
 import qualified Unison.Names3 as Names
 import qualified Unison.Name as Name
-import Unison.HashQualified (HashQualified)
+import Unison.HashQualified (HashQualified')
 import Unison.Type (Type)
 import Unison.NamePrinter (prettyHashQualified0)
 
@@ -931,6 +931,7 @@ prettyParseError s = \case
     , ").\n"
     ]
   go' (P.ErrorCustom e) = go e
+  errorVar :: v -> Pretty ColorText
   errorVar v = style ErrorSite . fromString . Text.unpack $ Var.name v
   go :: Parser.Error v -> Pretty ColorText
   -- | UseInvalidPrefixSuffix (Either (L.Token Name) (L.Token Name)) (Maybe [L.Token Name])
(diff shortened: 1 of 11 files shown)


parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -951,14 +952,14 @@ prettyParseError s = \case
...
           Nothing -> useExamples
           Just parent -> Pr.wrap $
             "You can write" <>
             Pr.group (Pr.blue $ "use " <> Pr.text (Name.toText parent) <> " "
-                                       <> Pr.text (Name.toText (Name.unqualified (L.payload tok)))) <>
-            "to introduce " <> Pr.backticked (Pr.text (Name.toText (Name.unqualified (L.payload tok)))) <>
-            "as a local alias for " <> Pr.backticked (Pr.text (Name.toText (L.payload tok)))
+                                       <> Pr.text (Name.toText (Name.unqualified (Name.fromVar (L.payload tok))))) <>
+            "to introduce " <> Pr.backticked (Pr.text (Name.toText (Name.unqualified (Name.fromVar (L.payload tok))))) <>
+            "as a local alias for " <> Pr.backticked (Pr.text (Name.toText (Name.fromVar (L.payload tok))))
```
aryairani: Note to self: this is an error message for when you say `use Nat.++`, it suggests you say `use Nat ++`. 🤔

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -1062,34 +1063,34 @@ prettyParseError s = \case
...
+        "I couldn't find a type for " <> style ErrorSite (HQ.toString (Name.fromVar <$> L.payload tok)) <> "."
       else
-        "The type reference " <> style ErrorSite (HQ.toString (L.payload tok)) <> " was ambiguous."
+        "The type reference " <> style ErrorSite (HQ.toString (Name.fromVar <$> L.payload tok)) <> " was ambiguous."
     , ""
-    , tokenAsErrorSite s $ HQ.toString <$> tok
+    , tokenAsErrorSite s $ HQ.toString . fmap Name.fromVar <$> tok
     , if missing then "Make sure it's spelled correctly."
       else "Try hash-qualifying the type you meant to reference."
     ]
```
aryairani: I understand that this code is converting some `v`s into `Name`s (as they were `Name`s here before the PR), but I already forget what that means.  

What's the relationship between `Name`, `v`, custom syntax, and what should get displayed here?

If `Name` is just "thing that represents what's stored in the `Names`pace", then we maybe don't need to convert to `Name` here (although it's fine as an incremental change, regardless), since we're not dealing with `Names`.  Or if the pluggable syntax tells us how to render `Name`, then maybe that's a fine type to use here.

But is there a difference between the syntax's rendered `Name` and the original `v`?

parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -489,12 +489,12 @@ namespaceBlock :: Var v => P v (BlockElement v)
...
   _ <- reserved "namespace"
   -- need a version of verifyRelativeName that takes a `Token Name`
-  name <- verifyRelativeName importWordyId
+  name <- verifyRelativeVarName importWordyId
   let statement = (Binding <$> binding) <|> namespaceBlock
   _ <- openBlockWith "where"
   elems <- sepBy semi statement
   _ <- closeBlock
-  pure $ Namespace (Name.toString $ L.payload name) elems
+  pure $ Namespace (Text.unpack . Var.name $ L.payload name) elems
```
aryairani: What's the diff between `Var.name` and `Name.fromVar` again?

unison-core/src/Unison/Term.hs
```diff
@@ -171,8 +172,8 @@ bindSomeNames
 --                   || traceShow e False
 --                   = undefined
 bindSomeNames ns e = bindNames keepFree ns e where
-  keepFree = Set.difference (freeVars e)
-                            (Set.map Name.toVar $ Rel.dom (Names.terms0 ns))
```
aryairani: I thought I had an answer for this before, but I wonder if there could be any synthetic names like `Symbol 3 foo` that _are_ supposed to be kept free, but would now wouldn't be, since we don't consider the `id` in the new `Set.difference`?

Good question for @pchiusano.

parser-typechecker/src/Unison/PrintError.hs
aryairani: @mitchellwrosen Would you add it as a comment in the code? 🙏

aryairani:
The fact that tests are still passing is pretty amazing.

parser-typechecker/src/Unison/PrintError.hs
mitchellwrosen: Good questions. My guess is many of these don't have to be converted to `Name`, they are just reusing printing/convenience functions that already operate on `Name` like `HQ.toString`.

parser-typechecker/src/Unison/TermParser.hs
mitchellwrosen: `Var.name` takes a `Var v => v` to a `Text`. For vars of type `User`, that's just the `Text` contained within (plus a number if it's not 0).

Example:

```
Var.name (Symbol 0 (User "foo")) = "foo.bar"
Var.name (Symbol 1 (User "foo")) = "foo1"
```

And `Name.fromVar` just takes that output `Text` and calls `unsafeFromText` on it.

unison-core/src/Unison/Term.hs
mitchellwrosen: @aryairani Good question. I think you've identified how the behavior might have changed here (but note the `id` _is_ considered in the new `Set.difference` - it's in `Symbol`'s `Ord` instance.

Before:

```haskell
{ Symbol 1 "foo" } `difference` { Name.toVar "foo1" }
{ Symbol 1 "foo" } `difference` { Symbol 0 "foo1 }
{ Symbol 1 "foo" }
```

After:

```haskell
{ Name.fromVar (Symbol 1 "foo") } `difference` { "foo1" }
{ "foo1" } `difference` { "foo1" }
{ }
```

😢

PR #1107, opened 2019-12-30
Simple cabal.project for building with cabal.

dolio:
- Unison doesn't seem to work with GHC 8.8 right now, so e.g.
    `cabal v2-configure --with-ghc=ghc-8.6.5`
  to set up if 8.6 isn't the `ghc` you have in your path. This
  doesn't go in the global project, though, I think, because it's
  specific to the local machine, and I'm not aware of cabal
  managing multiple GHC versions automatically (although I don't
  pay close attention).


aryairani:
Heya,

Could you speak to the pros/cons of building with `cabal` vs `stack`?  Why add this?  Thanks.

If we do want to commit to keeping this file up to date as the project evolves, we should add something for it into CI too.

mitchellwrosen:
I can speak to a couple of them -

Big advantage:

- Contributors who don't have `stack` don't have to use it; `cabal build && cabal run tests` would just work

Small advantages (features `stack` doesn't have):

- Backpack
- Multiple library components in one package
- Better caching (though stack is improving here)

Keeping it up to date is a concern since none of the core developers use `cabal`. I think just putting a disclaimer like "we use `stack` but the project should build with `cabal`. The `cabal.project` and other metadata is maintained by the community; if it doesn't work you may have to do some debugging" is an ok compromise

alanz:
Also, the project.cabal currently in unison does not build, it needs the `allow-newer` directives in this PR

dolio:
I'm not suggesting you switch the project over to cabal or anything. The advantage for me is that I was already set up to do everything with cabal, and know how to use it. So I can just do that in most cases, and only worry about making sure it also builds with stack when it comes to pushing/CI. If anyone else is in my situation, they can get spun up a little more quickly.

Most of the work was already done, because stack still requires the per-package .cabal files if I'm not mistaken. The cabal.project file just has the information for building/running the multiple packages as a single project, similar to your stack.yaml (it doesn't fill in any version constraints, though).

aryairani:
Is there any benefit to using stack, compared to cabal? (Stuff we’re taking for granted.)

dolio:
I don't think I'm qualified to answer that. I think the big difference is that they have lists of package versions that you're using instead of manually putting all the numbers in the .cabal file for each project. I think cabal can also do that (maybe with cabal.freeze files or something), but I'm not sure if there's an interface to that where you can just write `lts-n` in a file and have some published list used. I also don't know if the --allow-newer thing (which you're also using in stack) overrides that somehow, though.

The other thing I noticed it does is download GHC. But I have many versions of GHC already installed, so that's just wasting space for me. :)

alanz:
> Is there any benefit to using stack, compared to cabal? (Stuff we’re taking for granted.)

Not really, but some people prefer it, and there is not much pain in supporting it.

pchiusano:
This seems pretty painless... unlike the Nix stuff which seemed a lot more involved. I'm okay with merging.

@dolio as part of this PR, could you update https://github.com/unisonweb/unison/blob/master/development.markdown, add a section "Building with cabal" that has these instructions and notes?

dolio:
Sorry about the delay. Been busy. How does this look?

pchiusano:
@dolio could you also add yourself to CONTRIBUTORS at some point, would like you acknowledged there, could be in this PR or you can add it in your runtime branch.

aryairani:
Also, you guys decide if we want to officially support the cabal build file; if so, let's add something appropriate to CI for it to make sure it's always working as intended.  If not, then no need to slow CI — I assume it would mean doubling the build time to build with stack and then cabal.

aryairani:
I'm also imagining various Haskell dev tools arbitrarily picking up the cabal project vs the stack one, so we should maintain that they are equally usable.

pchiusano:
@dolio if it's possible, can we move this into a `/contrib` subdirectory? For now I think we want to continue just having one official build process, which at the moment is Stack (though open to revisiting that). Sorry for the churn.

development.markdown
```diff
@@ -25,3 +25,19 @@ Again you can leave off the flag. To run an executable with profiling enabled, d
...
+contrib/ so you'll need to specify its location on the command line.
+
+* To configure the build, you can use e.g. `cabal v2-configure
+  --project-file=contrib/cabal.project --with-ghc=ghc-8.6.5` if you have
+  multiple versions of GHC installed
+* To build all projects use `cabal v2-build --project-file=contrib/cabal.project all`
+* Tests can be run with e.g. `cabal v2-run --project-file=contrib/cabal.project
+  parser-typechecker:tests`
+* The executable can be installed with `cabal v2-install
+  --project-file=contrib/cabal.project unison` the install directory can be
```
aryairani: Maybe a missing sentence split before "the install directory".

PR #1115, opened 2020-01-03
Add Placement to Name

mitchellwrosen:
This patch adds an enum at the value level to track whether a `Name` is absolute or relative. It has revealed to me that a few of the existing `Name` functions don't make a ton of sense, so I put in calls to `error`.

Example: what does it mean to `joinDot foo .bar`? Do we throw away that `.bar` is absolute and just return a relative `foo.bar`?

Proposed longer-term solution: audit all calls to these questionable functions and see if they can be refactored to use variants of `Name` that encode absolute/relative at the type level.


aryairani:
> Example: what does it mean to `joinDot foo .bar`?

My guess is that `joinDot : Name -> Name -> Name` doesn't make semantic sense to begin with.   Like what are those two args, really?  We're not putting together the names of two terms, e.g. (`.base.Nat.++.tmp.lib.math.sine`). So are they really a Path and a NameSegment?  A .u namespace declaration and a local identifier? 

In any case, as a first stab, I might try something like normal absolute/relative resolution, like:
```
join : Name -> Name -> Name
join l r = case (l, r) of
  (Abs l, Rel r) -> Abs (l <> r)
  (Rel l, Rel r) -> Rel (l <> r)
  (_    , Abs r) -> Abs r
```

unison-core/src/Unison/Name.hs
```diff
@@ -103,20 +118,26 @@ stripNamePrefix prefix@Name{} name@Name{} =
...
+    error ("stripNamePrefix (" ++ show n1 ++ ") (" ++ show n2 ++ ")")
 
 joinDot :: Name -> Name -> Name
 joinDot prefix@Name{} suffix@Name{} =
   if toText prefix == "." then Name (toText prefix <> toText suffix)
   else Name (toText prefix <> "." <> toText suffix)
-joinDot (Name' prefix) (Name' suffix) = Name' (prefix <> suffix)
+joinDot n1@(Name' p1 prefix) n2@(Name' p2 suffix) =
+  case p2 of
+    Absolute -> error ("joinDot (" ++ show n1 ++ ") (" ++ show n2 ++ ")")
```
aryairani: despite my earlier comment, it will be interesting to see if and where this actually occurs.

unison-core/src/Unison/Name.hs
```diff
@@ -103,20 +118,26 @@ stripNamePrefix prefix@Name{} name@Name{} =
...
+    Relative -> Name' p1 (prefix <> suffix)
 
 unqualified :: Name -> Name
 unqualified (Name name) = unsafeFromText (last (Text.splitOn "." name))
-unqualified (Name' names) = Name' (NonEmpty.last names :| [])
+unqualified (Name' _ names) = Name' Relative (NonEmpty.last names :| [])
```
aryairani: This function's existence seems fishy to me; like an intermediate step in something more meaningful.

unison-core/src/Unison/Name.hs
```diff
@@ -128,16 +149,18 @@ parent (Name txt) = case unsnoc (Text.splitOn "." txt) of
...
 suffixes (Name n) =
   fmap up . tails . dropWhile (== "") $ Text.splitOn "." n
   where
   up ns = Name (Text.intercalate "." ns)
-suffixes (Name' names) =
-  map (Name' . NonEmpty.fromList) (NonEmpty.init (NonEmpty.tails names))
+suffixes name@(Name' p names) =
+  case p of
+    Absolute -> error ("suffixes (" ++ show name ++ ")")
+    Relative -> map (Name' p . NonEmpty.fromList) (NonEmpty.init (NonEmpty.tails names))
```
aryairani: Probably `Absolute` is okay here?  e.g. if the Name `.foo.bar.Baz` exists, suffix-based resolution should find it with as `Baz` or as `bar.Baz`.  Worth noting that the result of this suffixes operation is all `Relative` names.

aryairani: Can probably just ignore `p`.

unison-core/src/Unison/Name.hs
```diff
@@ -103,20 +118,26 @@ stripNamePrefix prefix@Name{} name@Name{} =
...
-  fmap Name' do
-    suffix <- List.stripPrefix (toList prefix) (toList name)
-    NonEmpty.nonEmpty suffix
+stripNamePrefix n1@(Name' p1 prefix) n2@(Name' p2 name) =
+  if p1 == p2 then
+    fmap (Name' Relative) do
+      suffix <- List.stripPrefix (toList prefix) (toList name)
+      NonEmpty.nonEmpty suffix
+  else
+    error ("stripNamePrefix (" ++ show n1 ++ ") (" ++ show n2 ++ ")")
```
aryairani: I don't really know when this one is used.

unison-core/src/Unison/Name.hs
```diff
@@ -89,9 +100,13 @@ fromVar = unsafeFromText . Var.name
...
 
-isPrefixOf :: Name -> Name -> Bool
+isPrefixOf :: HasCallStack => Name -> Name -> Bool
 Name a `isPrefixOf` Name b = a `Text.isPrefixOf` b
-Name' a `isPrefixOf` Name' b = toList a `NonEmpty.isPrefixOf` b
+n1@(Name' pa a) `isPrefixOf` n2@(Name' pb b) =
+  if pa == pb then
+    toList a `NonEmpty.isPrefixOf` b
+  else
+    error ("isPrefixOf (" ++ show n1 ++ ") (" ++ show n2 ++ ")")
```
aryairani: Note that the original implementation would just return false if `pa` and `pb` mismatched.

unison-core/src/Unison/Name.hs
mitchellwrosen: 👍

unison-core/src/Unison/Name.hs
mitchellwrosen: Ok, so you mean this?
```haskell
suffixes .foo.bar.baz = [foo.bar.baz, bar.baz, baz]
```
I like that. I wasn't sure if that first entry should be absolute or relative, so I put an error.

unison-core/src/Unison/Name.hs
mitchellwrosen: Agreed

unison-core/src/Unison/Name.hs
mitchellwrosen: Same! I plan to audit all of the functions that currently call `error` and get rid of them long term... but for now this just rules out

```
stripNamePrefix .foo foo.bar
stripNamePrefix foo .foo.bar
```
which both seem sketchy (since they ignore one arg or the other's absolute/relative quality)

unison-core/src/Unison/Name.hs
mitchellwrosen: Yeah, good point. I think this was a defensive `error` to try to rule out bad calls that mix name types. Want to just return `False` here, or leave the error for investigating later?

unison-core/src/Unison/Name.hs
aryairani: Yeah; I think that's what the current stringy implementation does, right?  `dropWhile (== "")` effectively making everything relative first.

unison-core/src/Unison/Name.hs
aryairani: Oh, I see.  I think looking at where `isPrefixOf` is called (period) will be more helpful than seeing if the error gets thrown.

aryairani: So, I dunno; what do you think?

mitchellwrosen: Yeah, I agree investigating calls to `isPrefixOf` is the move here. The `error`s are not that easy to trip at the moment anyway, probably - I'm not trying to actually use `ucm` yet on this patch :P

They're more like reminders that something is off with this API, to be refactored away soon. So I'm leaning towards leaving it in for this patch.

aryairani: Ok sounds good to me.

PR #1116, opened 2020-01-03
Derive instance Show Name

mitchellwrosen:
This patch switches `Name`'s `Show` instance from

```haskell
instance Show Name where
  show = toString
```

to

```haskell
deriving instance (Show)
```

The current instance is serving two purposes: pretty-printing a name, and for using `show` in `error` when handling some "impossible" case. `Show` is very useful for the latter, but the derived instance is fine here. The former is easily handled by calling `Name.toText` directly (and `Name`'s `Show` instance shouldn't pin it to any particular syntax, anyway).


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2111,8 +2111,8 @@ parseSearchType input typ = fmap Type.removeAllEffectVars <$> parseType input ty
 parseType :: (Monad m, Var v)
   => Input -> String -> Action' m v (Either (Output v) (Type v Ann))
 parseType input src = do
-  -- `show Input` is the name of the "file" being lexed
-  (names0, lexed) <- lexedSource (Text.pack $ show input) (Text.pack src)
+  -- `inputName input` is the name of the "file" being lexed
+  (names0, lexed) <- lexedSource (inputName input) (Text.pack src)
```
aryairani: Is this nicer?  Why not stick with `show input` here?

parser-typechecker/src/Unison/Codebase/Editor/Input.hs
```diff
@@ -120,6 +123,63 @@ data Input
...
+  FindShallowI{} -> "FindShallowI"
+  FindPatchI{} -> "FindPatchI"
+  ShowDefinitionI{} -> "ShowDefinitionI"
+  ShowDefinitionByPrefixI{} -> "ShowDefinitionByPrefixI"
+  ShowReflogI{} -> "ShowReflogI"
+  UpdateBuiltinsI{} -> "UpdateBuiltinsI"
+  MergeBuiltinsI{} -> "MergeBuiltinsI"
+  DebugBranchHistoryI{} -> "DebugBranchHistoryI"
+  QuitI{} -> "QuitI"
+
```
aryairani: 😓

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
mitchellwrosen: Good call, since it's just a stub impl it seems fine to `show input` still (showing uglier names)

PR #1118, opened 2020-01-04
Add NEW_NAME env switch for constructing new names

mitchellwrosen:
This patch adds a switch on the existence of a `NEW_NAME` env var for constructing new names (`Name'`) instead of old names (`Name`) in functions like `Name.unsafeFromText`.

### DIFF
--- a/unison-core/src/Unison/Name.hs
+++ b/unison-core/src/Unison/Name.hs
@@ -46,6 +46,8 @@ import           Data.List.Extra                ( dropEnd )
 import           Data.List.NonEmpty             ( NonEmpty((:|)) )
 import qualified Data.List.NonEmpty            as NonEmpty
 import qualified Data.Text                     as Text
+import           System.IO.Unsafe               ( unsafePerformIO )
+import           System.Environment             ( lookupEnv )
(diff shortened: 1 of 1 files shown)


unison-core/src/Unison/Name.hs
```diff
@@ -78,7 +104,10 @@ sortNamed' by by2 as = let
 
 unsafeFromText :: Text -> Name
 unsafeFromText t =
-  if Text.any (== '#') t then error $ "not a name: " <> show t else Name t
+  if useNewName then
+    undefined
+  else
+    if Text.any (== '#') t then error $ "not a name: " <> show t else Name t
```
aryairani: This should probably also support some syntax-specific escaping to allow '#' characters IMO.

mitchellwrosen: Yeah, agreed; by the time you are calling `unsafeFromText` you should know this is a valid `Name` per the syntax.

mitchellwrosen:
Yeah, for a couple functions I didn't want to just start returning `Name'` instead of `Name`... cause things will probably break somewhere. But I wanted to get the new implementations down somehow.

PR #1119, opened 2020-01-04
Fill out Name API a bit more

mitchellwrosen:


unison-core/src/Unison/Codebase/NameSegment.hs
```diff
@@ -27,6 +29,10 @@ unsafeFromText = NameSegment
 isEmpty :: NameSegment -> Bool
 isEmpty ns = toText ns == mempty
 
+-- | Does this name segment contain only lowercase letters?
+isLower :: NameSegment -> Bool
+isLower = coerce (Text.all Char.isLower)
```
aryairani: This is weird right, coercing a `Text -> Bool` into a `NameSegment -> Bool`?

Do we like `Text.all Char.isLower . coerce` any better?  (This is new to me.)

mitchellwrosen: I think the latter has a small runtime cost for `.` :)

aryairani: Does calling `NameSegment.toText` actually have a runtime cost that's greater than `coerce`?

aryairani: I like the idea of avoiding syntactic newtype cruft. ¯\_(ツ)_/¯

aryairani: Let's back out `coerce`, and then merge?

mitchellwrosen:
@aryairani Whoops, missed that comment. Done!

PR #1121, opened 2020-01-04
Remove Name.countDots

mitchellwrosen:
It was only called once, and better implemented with the existing `Name` API (namely `Name.segments`).

### DIFF
--- a/unison-core/src/Unison/HashQualified.hs
+++ b/unison-core/src/Unison/HashQualified.hs
@@ -9,6 +9,7 @@ import           Data.Maybe                     ( fromJust
                                                 )
 import qualified Data.Text                     as Text
 import           Prelude                 hiding ( take )
+import           Unison.Codebase.NameSegment    ( NameSegment )
 import           Unison.Name                    ( Name )
 import qualified Unison.Name                   as Name
 import           Unison.Reference               ( Reference )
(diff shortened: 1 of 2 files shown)


unison-core/src/Unison/HashQualified.hs
```diff
@@ -51,9 +52,16 @@ toName = \case
 --   [.foo.bar, foo.bar] -> [foo.bar, .foo.bar]
 sortByLength :: [HashQualified] -> [HashQualified]
 sortByLength hs = sortOn f hs where
-  f (NameOnly n) = (Name.countDots n, 0, Left n)
-  f (HashQualified n _h) = (Name.countDots n, 1, Left n)
-  f (HashOnly h) = (maxBound, 0, Right h)
+  -- comparator: (num segments, is hash qualified, is absolute, name itself)
+  f :: HashQualified -> Either (Int, Bool, Bool, [NameSegment]) ShortHash
```
mitchellwrosen: Reworked this comparator but its behavior should be roughly the same. The only thing that I intended to change was more accurately ordering absolute/relative.

Before: an absolute name's leading dot just counted against it as if it was a name segment.
After: a relative name is preferred to an absolute name

aryairani: Seems reasonable.

PR #1122, opened 2020-01-05
Adds punctuation to error messages

pete-ts:
There was no fullstop added after the first sentence in these cases:
- running `ucm` with no codebase initialised in $HOME
- running `ucm -codebase x` with no codebase initialised in "x"
- running `ucm init` a second time

### DIFF
--- a/parser-typechecker/src/Unison/Util/Pretty.hs
+++ b/parser-typechecker/src/Unison/Util/Pretty.hs
@@ -449,8 +449,8 @@ num n = fromString (show n)
 string :: IsString s => String -> Pretty s
 string = fromString
 
-endSentence :: IsString s => String -> Pretty s
-endSentence p = group (fromString p <> ".")
+endSentence :: IsString s => Pretty s -> Pretty s
+endSentence p = group (p <> ".")
 
 shown :: (Show a, IsString s) => a -> Pretty s
 shown = fromString . show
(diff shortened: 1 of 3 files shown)


pete-ts:
A first and obvious solution would be to add `<> "." <>` and be done with it. 

However, due to how `P.string dir` gets combined with subsequent `Pretty` elements, this always results in a space appearing between the directory and the full stop. So the full stop needs to be appended to the directory from `inside` the `Pretty`.

If there is a different suggestion for how to proceed, please feel free to mention it.

parser-typechecker/src/Unison/Util/Pretty.hs
```diff
@@ -448,6 +449,9 @@ num n = fromString (show n)
 string :: IsString s => String -> Pretty s
 string = fromString
 
+endSentence :: IsString s => String -> Pretty s
+endSentence str = fromString (str ++ ".")
```
pchiusano: @pete-ts I'd do this with:

```Haskell
endSentence :: IsString s => Pretty s -> Pretty s
endSentence p = group (p <> ".")
```

The `group` here will not get broken up when spaces are inserted by `wrap`.

pete-ts: @pchiusano I have a fix with the signature `IsString s => Pretty s -> Pretty s`, but that means that the call to `P.string` has to happen in the do block and it might look a bit "noisier" than `IsString s => String -> Pretty s`, which can get applied without the need to `endSentence . string $ dir`.

pchiusano:
Looks great! Can you add yourself to CONTRIBUTORS.markdown as part of this PR?

parser-typechecker/src/Unison/Util/Pretty.hs
pchiusano: Understood, and I think that's fine.

pchiusano:
😎 merging

PR #1123, opened 2020-01-05
Show aliases when searching by type

mitchellwrosen:
This patch makes `find :` work like it suggests it should in a comment - by showing aliases on separate lines.

Screen Shot 2020-01-05 at 4 44 26 PM


unison-core/src/Unison/Names2.hs
```diff
@@ -27,8 +27,6 @@ module Unison.Names2
   , prefix0
   , restrictReferences
   , refTermsNamed
-  , termName
```
mitchellwrosen: These are no longer used (and partial), so I deleted them.

Issue #1025, opened 2019-12-06
Add getChar function


seagreen:
Something like:
```
.base.io.getChar : '{IO} Char
```

seagreen:
I'm interested in doing this, but am not sure exactly what functions we want.

For `getLine` we have:

```
ability io.IO where
  getLine_ : io.Handle -> (Either io.Error Text)

namespace io where
  readLine : '{IO} Text
  readLine = '(getLine stdin)

  getLine : Handle ->{IO} Text
  getLine h = rethrow (io.IO.getLine_ h)
```

What do we want to have for `getChar`? The equivalent of that would be:
```
ability io.IO where
  getChar_ : io.Handle -> (Either io.Error Char)

namespace io where
  readChar : '{IO} Char
  readChar = '(getChar stdin)

  getChar : Handle ->{IO} Char
  getChar h = rethrow (io.IO.getChar_ h)
```

pete-ts:
@seagreen I thought I would look into something along these lines (add getChar or getWhatever in the IO ability), more to understand how it all comes together than to actually fix it (at the end of the day, not due to lack of confidence, I might plain not be able to fix it.)

Hope you don't mind?

seagreen:
Hey @pete-ts, I don't mind at all. However (and I should have updated this issue to say this) the current blocker is that upgrading IO abilities causes each Unison user to have to do manual work upgrading their codebase. Our thinking is that we need to either improve the codebase manager so that work happens automatically, or bundle as many IO changes we can think of into one release so that users have to do as little manual work as possible.

This was the thinking as of a couple months ago at least: the Unison devs can chime in if I'm giving out-of-date info.

pete-ts:
Hi @seagreen 

I asked in #contrib and realised after reading the answers how disruptive the need for updating the IO ability is. You are correct and (imo) with good reason updates to IO are not in scope for the time being.

Thanks :+1:

aryairani:
Maybe we should create a branch that accumulates updates for builtin, and we can merge to that while we wait, instead of killing the workstreams (like #1134); or tag the PRs or keep track of them somehow.

PR #1134, opened 2020-01-10
adds readChar to the IO ability, following along the lines of readLine

pete-ts:
I was reading up on how IO is implemented, and came up with a solution for [this issue](https://github.com/unisonweb/unison/issues/1025).

Feel free to close the PR if you feel that there is no need for `readChar` in the `IO` ability.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/IOSource.hs
+++ b/parser-typechecker/src/Unison/Runtime/IOSource.hs
@@ -150,8 +150,8 @@ type Either a b = Left a | Right b
 
 type Optional a = None | Some a
 
-d1 Doc.++ d2 = 
-  use Doc 
+d1 Doc.++ d2 =
+  use Doc
   case (d1,d2) of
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -150,8 +150,8 @@ type Either a b = Left a | Right b
 
 type Optional a = None | Some a
 
-d1 Doc.++ d2 = 
-  use Doc 
+d1 Doc.++ d2 =
+  use Doc
```
pete-ts: I could have sworn I saw something on slack about trailing whitespace, my vim config does this, sorry.

pete-ts:
I am also aware that I broke the build on travis, but have no idea how to go about fixing that. 
`stack build && stack exec tests && stack exec unison` + manual testing on a fresh codebase was all I thought to test.

PR #1135, opened 2020-01-10
add catalog of Unison libraries in development

anovstrup:
This patch adds a section to the project's README.md for cataloging Unison libraries under development and is intended as a quick, short-term solution.

Long-term, it may be better to store the catalog in a web-accessible database, as [suggested by @theduke](https://unisonlanguage.slack.com/archives/CLKV43YE4/p1578682688029300?thread_ts=1578682456.026500&cid=CLKV43YE4). The database could be as simple as a github repo containing a json "database", like [Nim's package database](https://github.com/nim-lang/packages). A simple web client, similar to Nim's [nimble.directory](https://nimble.directory/) package viewer, would make it possible to view/browse the database in a developer-friendly way.

A more Unison-y long-term solution would be to store the package database as a value in a Unison codebase [as exemplified here](https://github.com/anovstrup/unison-packages) or in a typed, distributed data store (once such a Unison-native capability has been developed).


aryairani:
I see we've got this wiki page: https://github.com/unisonweb/unison/wiki/Unison-libraries

Maybe we can just use that for now?

anovstrup:
@aryairani I wasn't aware of that. Let's close this PR, and I can add an issue to track ideas for a more manageable database.

anovstrup:
@aryairani Opened [this issue](https://github.com/unisonweb/unisonweb-org/issues/79) in the unisonweb-org project (that seemed like the more appropriate place). I guess it may also be a good idea to add a link from the README or from unisonweb.org to this catalog, at least when you're ready to encourage more active library development.

aryairani:
> @aryairani I wasn't aware of that. Let's close this PR, and I can add an issue to track ideas for a more manageable database.

Actually, I wasn't aware of it either until today ;-)

Issue #1021, opened 2019-12-04
unison won't start if there's a broken symlink in the directory


seagreen:
```
$ mkdir test
$ cd test
$ unison -codebase . init
$ unison
<this works>
$ ln -s foo this-does-not-exist
$ unison

   _____     _
  |  |  |___|_|___ ___ ___
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|

  Welcome to Unison!

  I'm currently watching for changes to .u files under ~/test

  Type help to get help. 😎

  Enter pull https://github.com/unisonweb/base .base to set up the default base library. 🏗

unison: this-does-not-exist: getModificationTime:getFileStatus: does not exist (No such file or directory)
```

@aryairani reports that this is a regression

EDIT: In case it wasn't clear, `unison` crashes after the last command

int-index:
I can't run `unison` in `$HOME` on macOS because of this:

```
ucm: Applications: getModificationTime:getFileStatus: does not exist (No such file or directory)
```

aryairani:
@int-index Is that also a broken symlink for you?

int-index:
Yes

aryairani:
We want to find a fix for this, but as a workaround, can you fix or delete the broken link?

int-index:
Sure, `rm Applications` did the trick.

aryairani:
I suspect it’s a bug in the file watching library we’re using, but haven’t had a chance to dig into it :/

pete-ts:
The bug only materializes if the current working directory being watched by unison contains a broken symlink, I have managed to reproduce it by adding a broken synlink in a directory, then running unison from inside it.

https://github.com/unisonweb/unison/blob/9f81ec91a9539f7baeb9a5810d2e843fe20ca214/parser-typechecker/src/Unison/Codebase/Watch.hs#L94

[listDirectory](https://hackage.haskell.org/package/directory-1.3.5.0/docs/System-Directory.html#v:listDirectory) claims that it will fail if a directory does not exist.

[getModificationTime](https://hackage.haskell.org/package/directory-1.3.5.0/docs/System-Directory.html#v:getModificationTime) also claims that it will fail if a directory does not exist.

What actually happens: `listDirectory` returns a broken symlink in the `[String]`, and does not seem to throw (at least on ubuntu 18.04), then when `getModificationTime` actually tries to get the modification time for the broken symlink, an exception is raised.

pete-ts:
[...]

PR #1144, opened 2020-01-15
filters listed directories preventing getModificationTime from throwing

pete-ts:
[This comment](https://github.com/unisonweb/unison/issues/1021#issuecomment-574742378) describes the failure mechanism, or at least what I think is the cause of the failure.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Watch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Watch.hs
@@ -16,6 +16,7 @@ import           UnliftIO                       ( MonadUnliftIO
                                                 , unliftIO )
 import           UnliftIO.Directory             ( getModificationTime
                                                 , listDirectory
+                                                , doesPathExist
                                                 )
 import           UnliftIO.MVar                  ( newEmptyMVar, takeMVar
                                                 , tryTakeMVar, tryPutMVar, putMVar )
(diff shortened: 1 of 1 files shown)


aryairani:
Hey thanks for looking into this!  Since the PR description is sparse, could you confirm that ucm successfully ignores vs complains about broken symlinks exactly when this patch is applied?

Thanks again!

pete-ts:
Original behaviour: `ucm` encounters a broken symlink and throws an exception after starting
```
$ ./unison 

   _____     _             
  |  |  |___|_|___ ___ ___ 
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|
  
  Welcome to Unison!
  
  I'm currently watching for changes to .u files under
  ~/unisonweb/unison/.stack-work/install/x86_64-linux/fd8751afd3efe7a4a550a8daa061e9a3c7a8fb0b2c155e4a5a01b36dc1720450/8.6.5/bin
  
  Type help to get help. 😎

unison: 2: getModificationTime:getFileStatus: does not exist (No such file or directory)
```
After the fix, the broken symlink is filtered as "non existent", and `ucm` skips it (along with any other non-existent paths), so it can watch the dir without issues:

```
./unison 

   _____     _             
  |  |  |___|_|___ ___ ___ 
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|
  
  Welcome to Unison!
  
  I'm currently watching for changes to .u files under
  ~/unisonweb/unison/.stack-work/install/x86_64-linux/fd8751afd3efe7a4a550a8daa061e9a3c7a8fb0b2c155e4a5a01b36dc1720450/8.6.5/bin
  
  Type help to get help. 😎

.> 
```

aryairani:
Thanks @pete-ts !

Issue #1141, opened 2020-01-14
eliminate references to `Input` from `Output`


aryairani:
To the extent that `Input` and `Output` are separate to support easily porting the shell of UCM from one UI (CLI) to another (HTML), it's worth noting that the `Input` message of one is generally not meaningful in another.

If there are specific values in the `Input` constructor that are relevant to the output, just copy them into the `Output` constructor instead of passing through an arbitrary `Input`, which must be matched on in `OutputMessages` with mostly impossible cases.

If the `Output` constructor is meant to handle many similar cases that could arise from various `Input`s, split that `Output` constructor into several for those several specific cases.

My very bad for introducing the first such instance: `data Output v = Success Input | ...`.  In this specific case, we don't actually inspect the `Input` field, so we can just drop it.

PR #1154, opened 2020-01-17
Remove Input from Output

mitchellwrosen:
#1141


mitchellwrosen:
This is probably ok to merge, actually; it's just the easy no-brainer ones and doesn't fully close #1141, but it's something :) @aryairani wdyt?

aryairani:
Seems okay to merge!

aryairani:
Let's see what happens with CI.

Issue #975, opened 2019-11-18
let the default `ucm transcript` behavior delete a codebase that the transcript runner created


aryairani:
The resulting transcript `.output.md` would still be preserved, and the resulting codebase should be preserved with a flag, e.g. `-preserve`.

Rationale: Since transcript running so far is just for running tests or generating `.output.md`, the resulting codebases are uninteresting accumulated garbage.

See: `Main.runTranscripts`

aryairani:
Just spent over 10 minutes trying to clone my unison dev working directory before cancelling; it was partially due to many test-output subdirs.

pete-ts:
Quick question, are you envisioning the default behaviour to delete the temporary folder for both `transcript` and `transcript.fork`?

pete-ts:
Further question: if a transcript is syntactically correct, but causes an error, like calling an invalid ucm command for example:
```
🛑

Transcript failed due to the message above.
Codebase as of the point of failure is in
```
should the preserve flag affect the cleanup of the temporary dir?

aryairani:
I think this is done?

PR #1155, opened 2020-01-17
`ucm transcript` default behavior deletes the codebase that the transcript runner created

pete-ts:
[#975](https://github.com/unisonweb/unison/issues/975)

Running `ucm -transcript` or `ucm -transcript.fork` will cleanup the temporary `transcript-hash` directory used for running the transcript in the following circumstances:
1. transcript exists successfully
2. transcript file does not exist
3. command not recognised.

The change does not affect transcripts that fail inside the call to the `TranscriptParser.run` method.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
+++ b/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
@@ -13,6 +13,7 @@ import Control.Exception (finally)
 import Control.Monad.State (runStateT)
 import Data.IORef
 import Prelude hiding (readFile, writeFile)
+import System.Directory ( doesFileExist )
 import System.Exit (die)
 import System.IO.Error (catchIOError)
 import Unison.Codebase (Codebase)
@@ -88,8 +89,12 @@ instance Show Stanza where
 
 parseFile :: FilePath -> IO (Either Err [Stanza])
 parseFile filePath = do
-  txt <- readUtf8 filePath
-  pure $ parse filePath txt
+  exists <- doesFileExist filePath
+  if exists then do
+    txt <- readUtf8 filePath
+    pure $ parse filePath txt
+  else
+    pure $ Left $ show filePath ++ " does not exist"
 
 parse :: String -> Text -> Either Err [Stanza]
 parse srcName txt = case P.parse (stanzas <* P.eof) srcName txt of
--- a/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
+++ b/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
@@ -232,9 +232,8 @@ run dir configFile stanzas codebase = do
         output "\n```\n\n"
         transcriptFailure out $ Text.unlines [
           "\128721", "",
-          "Transcript failed due to the message above.",
-          "Codebase as of the point of failure is in:", "",
-          "  " <> Text.pack dir ]
+          "Transcript failed due to the message above.", "",
+          "Run `ucm -codebase" <> Text.pack dir <> "` " <> "to do more work with it."]
 
       loop state = do
         writeIORef pathRef (HandleInput._currentPath state)
--- a/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
+++ b/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
@@ -233,7 +233,7 @@ run dir configFile stanzas codebase = do
         transcriptFailure out $ Text.unlines [
           "\128721", "",
           "Transcript failed due to the message above.", "",
-          "Run `ucm -codebase" <> Text.pack dir <> "` " <> "to do more work with it."]
+          "Run `ucm -codebase " <> Text.pack dir <> "` " <> "to do more work with it."]
 
       loop state = do
         writeIORef pathRef (HandleInput._currentPath state)
--- a/parser-typechecker/unison/Main.hs
+++ b/parser-typechecker/unison/Main.hs
@@ -54,10 +54,10 @@ usage = P.callout "🌻" $ P.lines [
         <> "Multiple transcript files may be provided; they are processed in sequence"
         <> "starting from the same codebase.",
   "",
-  P.bold "ucm transcript -preserve mytranscript.md",
+  P.bold "ucm transcript -save-codebase mytranscript.md",
   P.wrap $ "Executes the `mytranscript.md` transcript and creates"
-        <> "`mytranscript.output.md` if successful. Exits after completion, and maintains"
-        <> "the temporary directory created."
+        <> "`mytranscript.output.md` if successful. Exits after completion, and saves"
+        <> "the resulting codebase to a new directory on disk."
         <> "Multiple transcript files may be provided; they are processed in sequence"
         <> "starting from the same codebase.",
   "",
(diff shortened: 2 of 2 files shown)


parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
```diff
@@ -88,8 +89,12 @@ instance Show Stanza where
...
 parseFile :: FilePath -> IO (Either Err [Stanza])
 parseFile filePath = do
-  txt <- readUtf8 filePath
-  pure $ parse filePath txt
+  exists <- doesFileExist filePath
+  if exists then do
+    txt <- readUtf8 filePath
+    pure $ parse filePath txt
+  else
+    pure $ Left $ show filePath ++ " does not exist"
```
pete-ts: originally a failure due to file not found in `readUtf8` resulted in an exception

parser-typechecker/unison/Main.hs
```diff
@@ -59,6 +67,13 @@ usage = P.callout "🌻" $ P.lines [
         <> "Multiple transcript files may be provided; they are processed in sequence"
         <> "starting from the same codebase.",
   "",
+  P.bold "ucm transcript.fork -preserve mytranscript.md",
+  P.wrap $ "Executes the `mytranscript.md` transcript in a copy of the current codebase"
+        <> "and creates `mytranscript.output.md` if successful. Exits after completion,"
+        <> "and maintains the temporary directory created."  
+        <> "Multiple transcript files may be provided; they are processed in sequence"
+        <> "starting from the same codebase.",
+  "",
```
pete-ts: Suggestions on the wording more than welcome. The "-codebase" flag is supported for transcripts and does take "precedence" before "-preserve", should that be mentioned?

parser-typechecker/unison/Main.hs
```diff
@@ -164,16 +201,32 @@ runTranscripts inFork mcodepath args = do
...
           P.lines [
             "I've finished running the transcript(s) in this codebase:", "",
             P.indentN 2 (P.string transcriptDir), "",
             P.wrap $ "You can run"
                   <> P.backticked ("ucm -codebase " <> P.string transcriptDir)
                   <> "to do more work with it."])
-    [] -> do
+
+  unless completed $ do 
+      unless keepTemp $ cleanup transcriptDir
```
pete-ts: I broke off two smaller functions off of the original `runTranscripts` with no changes in functionality. Deletion of the temporary directory happens right before the end.

parser-typechecker/unison/Main.hs
aryairani: I was just thinking that `-preserve` makes it sound like we stop short of deleting the resulting codebase (which is true!)  I think I'd rather ignore the fact that the temporary codebase exists at all; as it's just an implementation detail after all.  Suppose that the temporary codebase only existed in memory, unless the user requested to have it saved on disk at the end of the run.  What could we call the flag in that case?  Maybe `-preserve` still makes sense, but I wouldn't be surprised if there was something clearer.  `-save` or `-write` come to mind, just want it to be clear enough that the resulting codebase is the thing that's being saved, as opposed to raising a question about whether the transcript output is saved.

Along the same vein, instead of "maintains the temporary directory created", I'd rather say something like "saves the resulting codebase into a new directory on disk".

pete-ts: I agree that ideally we should be hiding the implementation detail, and communicating the end result, which is you end up with a codebase stored in a path. `-save-codebase` or `-store-codebase`? `-preserve` feels closely coupled to the "I am starting by creating a directory and then..." underlying implementation, your point is valid.

pete-ts:
Will update the wording (in the very near future) and request another review.

parser-typechecker/unison/Main.hs
```diff
@@ -122,40 +137,62 @@ main = do
...
-    "transcript" : args -> runTranscripts False mcodepath args
-    "transcript.fork" : args -> runTranscripts True mcodepath args
+    "transcript" : args' -> 
+      case args' of
+      "-preserve" : transcripts -> runTranscripts False True mcodepath transcripts
+      _                         -> runTranscripts False False mcodepath args'
+    "transcript.fork" : args' -> 
+      case args' of
+      "-preserve" : transcripts -> runTranscripts True True mcodepath transcripts
+      _                         -> runTranscripts True False mcodepath args'
```
aryairani: Hehe already feeling a little Boolean-blind here, though I'm not feeling creative in coming up with a short, clear type name we could use instead of Bool; just an idea, not a requirement.

parser-typechecker/unison/Main.hs
```diff
@@ -122,40 +137,62 @@ main = do
...
+    Path.copyDir path (tmp FP.</> FileCodebase.codebasePath)
+
+  pure tmp
+
+cleanup :: FilePath -> IO ()
+cleanup transcriptDir = do
+  PT.putPrettyLn $ P.lines [
+    P.wrap "Deleting temporary directory: ", "",
+    P.indentN 2 (P.string transcriptDir)
+    ]
```
aryairani: Similarly, I think we can eliminate this message.

parser-typechecker/unison/Main.hs
```diff
@@ -164,16 +201,32 @@ runTranscripts inFork mcodepath args = do
...
+    [] -> 
+      pure False
+
+runTranscripts :: Bool -> Bool -> Maybe FilePath -> [String] -> IO ()
+runTranscripts inFork keepTemp mcodepath args = do
+  transcriptDir <- prepareTranscriptDir inFork mcodepath
+  completed <- runTranscripts' mcodepath transcriptDir args
+  when completed $ do
+    unless keepTemp $ cleanup transcriptDir
+    when keepTemp $ PT.putPrettyLn $
```
aryairani: If I understood correctly, if the transcript fails, then the work directory is automatically preserved, right?  In that case, we still want it to explain what happened + how to `ucm -codebase <transcriptDir>` for the user to poke around in it to debug.

parser-typechecker/unison/Main.hs
aryairani: Yeah, maybe `-save-codebase`?

parser-typechecker/unison/Main.hs
pete-ts: My thinking behind originally was to make it "observable" so I could test with a script containing all flag combinations. Removing it, as it exposes underlying implementation detail. Subsequently there's no point to the wrapper, so calling `removeDirectoryRecursive` directly

parser-typechecker/unison/Main.hs
pete-ts: In various places in the run method 
https://github.com/unisonweb/unison/blob/e42267fc671004970683aac3488b511be3aa8cdd/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs#L99

Errors exit prematurely with `die`

https://github.com/unisonweb/unison/blob/e42267fc671004970683aac3488b511be3aa8cdd/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs#L235-L241

I think this can be accomplished by 
1. updating the `dir` passed to `run` to be the `transcriptDir`, so the error points to the correct temporary path
2. adding the output message for how to debug.

aryairani: Ok, great.  That `dir` on line 241 in your excerpt above, is that not already the `transcriptDir`?

pete-ts: No, that is not the transcript dir, have taken care of it in the latest commit.

aryairani: Out of curiosity; which dir was it?  Sorry I’m just on github mobile site, not a full IDE in front of me

pete-ts: It was the current directory . 

Originally a parse failure would fail with (when running from inside "test"):
```
🛑

Transcript failed due to the message above.

Run `ucm -codebase /.../test` to do more work with it
```

For `transcript` and `transcript.fork`, the new codebase has been initialised in the temporary directory, and copied in the temporary directory respectively, so the message was not correct.

Currently:
```
🛑

Transcript failed due to the message above.

Run `ucm -codebase /.../test/transcript-72911f80dd3db7c8` to do more work with it.
```

I am still missing a space in the string concatenation, just noticed.

aryairani: Whoops, this is much better, thanks!

aryairani:
This is great, @pete-ts! I'm gonna merge it.

PR #1158, opened 2020-01-18
Adds handle .. with syntax

bontaq:
This allows `handle something with` in addition to `handle something in` syntax.

I saw it mentioned in the slack and it sounded nice.  Also happy to update the reference page.


pchiusano:
@bontaq nice! However I’d like to allow a block after the ‘handle’ keyword. Take a look at how if/then is parsed for inspiration.

anovstrup:
Hi, @bontaq! I just ran across your PR after creating my own (#1159). AFAICT, there are some key differences between the two patches:
* Mine obliterates the `handle ... in ...` form. Do we want to hang on to that form? I personally find `handle ... in` confusing—you don't "handle a handler in a computation", you "handle a computation _with_ a handler" (or, more precisely, you "handle effects in an effectful computation with an effects-handler")—so I think I'm happy to get rid of it.
* It looks like your patch treats `with` as an alias for `in`, whereas mine swaps the order of the handler and the effectful computation. That is, my `handle computation with handler` is equivalent to the old `handle handler in computation`.
* My patch allows both `handle` and `with` to open a new block, whereas the old syntax did not allow the handler to be given in a block-expression.

pchiusano:
Wow funny that two people near simultaneously put together a PR for this. I think I’m going to close in favor of #1159 @bontaq

PR #1159, opened 2020-01-18
Handle-with syntax

anovstrup:
This patch replaces the existing handle-in syntax (e.g., `handle handler in body`) with handle-with syntax (e.g., `handle body with handler`), per [this thread](https://unisonlanguage.slack.com/archives/CLKV43YE4/p1579212933102300). 

In the new syntax, both the `handle` and `with` keywords open a block. As a result, handlers can be defined at the point of use (if desired). For example,
```
> handle loggingFactorial 10 with 
    h msgs req = case req of
      { Log.log m -> k } ->
        handle !k with h (msgs :+ m)
      { a } -> (msgs, a)
    h []

-- OR

> handle loggingFactorial 10 with
    ignore req = case req of 
      { Log.log _ -> k } -> handle !k with ignore
      { a } -> a
    ignore
```

When merging this PR, the documentation will also need to be updated to reflect the new syntax. 

Note: because the change is purely syntactic, existing codebases should not be affected. (Thanks, Unison!)

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -195,11 +195,12 @@ pretty0
       elideFQN im $ PrettyPrintEnv.termName n (Referent.Con ref i CT.Data)
     Request' ref i -> styleHashQualified'' (fmt S.Request) $
       elideFQN im $ PrettyPrintEnv.termName n (Referent.Con ref i CT.Effect)
-    Handle' h body -> let (im', uses) = calcImports im body in
+    Handle' h body -> let (imBody, usesBody) = calcImports im body
+                          (imH, usesH) = calcImports im h in
       paren (p >= 2)
-        $ (fmt S.ControlKeyword "handle" `PP.hang` pretty0 n (ac 2 Normal im doc) h)
+        $ (fmt S.ControlKeyword "handle" `PP.hang` usesBody [pretty0 n (ac 2 Block imBody doc) body])
         <> PP.softbreak
-        <> (fmt S.ControlKeyword "in" `PP.hang` uses [pretty0 n (ac 2 Block im' doc) body])
+        <> (fmt S.ControlKeyword "with" `PP.hang` usesH [pretty0 n (ac 2 Block imH doc) h])
     App' x (Constructor' DD.UnitRef 0) ->
       paren (p >= 11) $ (fmt S.DelayForceChar $ l "!") <> pretty0 n (ac 11 Normal im doc) x
     LamNamed' v x | (Var.name v) == "()" ->
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -195,12 +195,22 @@ pretty0
       elideFQN im $ PrettyPrintEnv.termName n (Referent.Con ref i CT.Data)
     Request' ref i -> styleHashQualified'' (fmt S.Request) $
       elideFQN im $ PrettyPrintEnv.termName n (Referent.Con ref i CT.Effect)
-    Handle' h body -> let (imBody, usesBody) = calcImports im body
-                          (imH, usesH) = calcImports im h in
-      paren (p >= 2)
-        $ (fmt S.ControlKeyword "handle" `PP.hang` usesBody [pretty0 n (ac 2 Block imBody doc) body])
-        <> PP.softbreak
-        <> (fmt S.ControlKeyword "with" `PP.hang` usesH [pretty0 n (ac 2 Block imH doc) h])
+    Handle' h body -> paren (p >= 2) $
+      if height > 0 then PP.lines [
+        (fmt S.ControlKeyword "handle") `PP.hang` pb,
+        (fmt S.ControlKeyword "with") `PP.hang` ph
+       ]
+      else PP.spaced [
+        (fmt S.ControlKeyword "handle") `PP.hang` pb
+          <> PP.softbreak
+          <> (fmt S.ControlKeyword "with") `PP.hang` ph
+      ]
+      where
+        height = PP.preferredHeight pb `max` PP.preferredHeight ph
+        pb = pblock body
+        ph = pblock h
+        pblock tm = let (im', uses) = calcImports im tm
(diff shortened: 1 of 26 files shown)


parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -195,11 +195,12 @@ pretty0
...
       elideFQN im $ PrettyPrintEnv.termName n (Referent.Con ref i CT.Effect)
-    Handle' h body -> let (im', uses) = calcImports im body in
+    Handle' h body -> let (imBody, usesBody) = calcImports im body
+                          (imH, usesH) = calcImports im h in
       paren (p >= 2)
-        $ (fmt S.ControlKeyword "handle" `PP.hang` pretty0 n (ac 2 Normal im doc) h)
+        $ (fmt S.ControlKeyword "handle" `PP.hang` usesBody [pretty0 n (ac 2 Block imBody doc) body])
         <> PP.softbreak
-        <> (fmt S.ControlKeyword "in" `PP.hang` uses [pretty0 n (ac 2 Block im' doc) body])
+        <> (fmt S.ControlKeyword "with" `PP.hang` usesH [pretty0 n (ac 2 Block imH doc) h])
```
anovstrup: I'm not confident about these changes. It seems to work, but I didn't look that closely at the `im`/`calcImports` stuff.

anovstrup:
I should have posted this as a draft. ... and I should have run the tests locally first. :sigh: 

Code review would still be appreciated. Hopefully the test failures can be fixed just by updating to the new syntax.

parser-typechecker/tests/Unison/Test/TermPrinter.hs
```diff
@@ -248,12 +248,13 @@ test = scope "termprinter" . tests $
                  \  baz.f : Int -> Int\n\
                  \  baz.f x = x\n\
                  \  13"
-  , pending $ tcBreaks 90 "handle foo in\n\
+  , tcBreaks 90 "handle\n\
                  \  a = 5\n\
                  \  b =\n\
                  \    c = 3\n\
                  \    true\n\
-                 \  false"  -- TODO comes back out with line breaks around foo
```
anovstrup: This test passes after the patch (but maybe it was already passing).

parser-typechecker/tests/Unison/Test/TermPrinter.hs
```diff
@@ -160,13 +160,13 @@ test = scope "termprinter" . tests $
   , pending $ tc "case e of { () -> k } -> z" -- TODO doesn't parse since 'many leaf' expected before the "-> k"
                                               -- need an actual effect constructor to test this with
   , tc "if a then if b then c else d else e"
-  , tc "handle foo in (handle bar in baz)"
+  , tc "handle handle foo with bar with baz"
```
anovstrup: stacked handler syntax is kinda weird after the patch... As a user, I'd probably use parens.

unison-src/tests/console.u
```diff
@@ -29,24 +29,26 @@ namespace Console where
...
+x = handle
+      handle
         use Console read write
         use Optional Some None
         write "What's your name?"
         case read of
           Some name -> write ("Hello" ++ name)
           None -> write "Fine, be that way."
+      with Console.simulate
+    with Console.state ([],[])
```
anovstrup: Here's another example of stacked handlers. Doesn't seem quite as odd with indents.

unison-src/tests/handler-stacking.u
```diff
@@ -1,9 +1,11 @@
...
 use Writer tell
 
-> handle stateHandler "hello" in
-  handle writerHandler [] in
-  replicate 5 main
+> handle
+    handle replicate 5 main
+    with writerHandler []
+  with stateHandler "hello"
+
```
anovstrup: yet another stacked handler...

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -195,11 +195,22 @@ pretty0
...
+        (fmt S.ControlKeyword "handle") `PP.hang` pb
+          <> PP.softbreak
+          <> (fmt S.ControlKeyword "with") `PP.hang` ph
+      ]
+      where
+        height = PP.preferredHeight pb `max` PP.preferredHeight ph
+        pb = pblock body
+        ph = pblock h
+        pblock tm = let (im', uses) = calcImports im tm
+                    in uses $ [pretty0 n (ac 0 Block im' doc) tm]
```
anovstrup: I'm least confident about these changes... Improvements are more than welcome.

parser-typechecker/tests/Unison/Test/TermPrinter.hs
mitchellwrosen: It looks ok to me. Not unlike `if x then if y then z else p else q`

unison-src/tests/console.u
mitchellwrosen: I wonder if this syntactic sugar would make sense?

```
  handle
    expression
  with
    handler1
    handler2
```

... probably not

anovstrup: I [was thinking](https://unisonlanguage.slack.com/archives/CLKV43YE4/p1579212213100800?thread_ts=1579210507.097500&cid=CLKV43YE4) the maybe the block following `with` could be a term that either evaluates to a single handler function or a list of handler functions. So, e.g., `handle exp with [handler1, handler2]` would be allowed. But I guess that wouldn't just be a sugar if it requires evaluation.

pchiusano: You can already handle multiple abilities in a single handler so I’d say no need for further special syntax (yet).

pchiusano:
LGTM! Merging. @anovstrup if you like you can submit PR for the website docs too. If not don’t sweat since I am working on revised abilities tutorial and can roll this in.

unison-src/tests/console.u
anovstrup: @pchiusano Oh, I didn't realize you could do that! Cool!

aryairani: @pchiusano Do we have a way to compose handlers yet?  
`handle exp with (handler1 and handler2)`

PR #1164, opened 2020-01-20
Use namespace diff in link output

mitchellwrosen:
Here's a screenshot of linking, then undoing (previously `link` would just say `Done.`):

Screen Shot 2020-01-20 at 2 29 27 PM

I'm a bit confused by the output... note that `contains` is called `lib.contains` when the link is undone. 🤷‍♂

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -580,8 +580,8 @@ loop = do
                     (Branch.toNames0 b)
                 diff = Names3.diff0 deletedNames mempty
             diffHelper b Branch.empty0 >>=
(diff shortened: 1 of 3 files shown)


aryairani:
Cool, yeah sorry for the confusion.  The unfortunate reason that is happening is because the diff function takes two namespaces to compare.  For `link` it’s using the old and new namespace at path `parent`, as you’ve defined it; for `undo` it’s using the old and new root, so `lib` shows up in the output.  Not sure what else we could do here for an `undo` in general, short of resolving absolute names to maybe-relative ones later than the current pre-baked `Name` model supports.

mitchellwrosen:
Gotcha. The `link` diff could diff the roots, too - but I agree a better fix is to use absolute names everywhere until final rendering

PR #1167, opened 2020-01-20
Add ":hide:all" option for unison and ucm fences in transcripts

atacratic:
Add "unison:hide:all" option for fences in transcripts.

Hides not only the output, but also suppresses the echo of the input as well.  Indistinguishable from the fenced block not being there, except for any effect that fenced code has on the codebase's state.

Useful for behind-the-scenes tinkering with the codebase during a transcript - say to add supporting dependencies that aren't an important part of the story being told.

This should help me finish https://github.com/unisonweb/unisonweb-org/pull/26 (an ability tutorial).

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
+++ b/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
@@ -3,7 +3,7 @@
 {-# Language ViewPatterns #-}
 
 module Unison.Codebase.TranscriptParser (
-  Stanza(..), FenceType, ExpectingError, HideOutput, Err, UcmCommand(..),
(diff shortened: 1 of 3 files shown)


pete-ts:
Question on the `ucm` block handling, on a slight tangent to the PR.
Is `ucm:hide:error` supported, or not really a sensible operation at all?

atacratic:
@pete-ts I just checked: `ucm:hide:error` does what you'd expect.  Yes maybe not so useful though as hopefully failed ucm commands don't affect any state.

`unison:hide:all:error` works too, but also shouldn't be useful for the same reason.

pchiusano:
@atacratic cool, I have sometimes wanted this sort of thing when prepping transcripts, especially for "boring" ucm blocks that I don't want to keep looking at when developing the transcript.

PR #1170, opened 2020-01-22
`ls` wasn't displaying patches, + cleanup

aryairani:
Primarily wanted to get eliminate `{-# OPTIONS_GHC -Wno-unused-local-binds #-}` (towards #1111), and found this dead code + bug in the process.

Sample output:
image


aryairani:
Side note: I think we originally envisioned listing the number of replacements in the patch or something, but then got lazy because it required loading the patch monadically. 😬

PR #1177, opened 2020-01-24
fix/1160

pete-ts:
I am aware that @atacratic has [a PR](https://github.com/unisonweb/unison/pull/1167) for the transcript parser already. This can wait for 1167 to be merged, and I will revisit after a rebase.

In the first commit you can see how the TranscriptParser behaves at the moment: an attempt to get `fenceType` is not firing correctly, and as a result `fenceType` is an empty string, meaning the word directly next to the triple quote ends up being part of the body, which is why it gets printed in the next line after the triple quote is `shown`, in the excerpt below:

https://github.com/unisonweb/unison/blob/260e18296f323240bbcb73cf01a9cd6ba3970707/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs#L84-L87

If there is no need to support `ucm<space>:error`, `lineToken` can be removed too.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
+++ b/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
@@ -225,13 +225,13 @@ run dir configFile stanzas codebase = do
         when (Output.isFailure o) $
           if errOk then writeIORef hasErrors True
           else die
-          
+
       printNumbered o = do
         let (msg, numberedArgs) = notifyNumbered o
         errOk <- readIORef allowErrors
         let rendered = P.toPlain 65 (P.border 2 msg)
         output rendered
-        when (Output.isNumberedFailure o) $ 
+        when (Output.isNumberedFailure o) $
           if errOk then writeIORef hasErrors True
           else die
         pure numberedArgs
@@ -293,7 +293,7 @@ ucmCommand = do
 fenced :: P Stanza
 fenced = do
   fence
-  fenceType <- lineToken (word "ucm" <|> word "unison" <|> lineUntilSpace)
+  fenceType <- lineToken(word "ucm" <|> word "unison" <|> language)
   stanza <-
     if fenceType == "ucm" then do
       hideOutput <- hideOutput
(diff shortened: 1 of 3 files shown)


aryairani:
> If there is no need to support `ucm<space>:error`, `lineToken` can be removed too.

I don't think it's necessary to support the separating space.

parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
```diff
@@ -370,8 +370,8 @@ expectingError = isJust <$> optional (word ":error")
 untilSpace1 :: P Text
 untilSpace1 = P.takeWhile1P Nothing (not . Char.isSpace)
 
-lineUntilSpace :: P Text
-lineUntilSpace = P.takeWhileP Nothing (\ch -> ch `elem` (" \t" :: String))
+language :: P Text
+language = P.takeWhileP Nothing (\ch -> Char.isDigit ch || Char.isLower ch || ch == '_' )
```
aryairani: This might be overly conservative; e.g. should it have to be all lowercase?

aryairani: Is there some list or spec somewhere online we can use as reference?  I've been looking but haven't found one.

aryairani: Well maybe this is fine. 😅

pete-ts: I mentioned that through some digging, it looks like for language highlighting to work (as supported by github at least), the `fenceType` of the block must match the names of the files in this repo. https://github.com/ajaxorg/ace/tree/master/lib/ace/mode 

From what I can tell, there are only lowercase names, with very few that contain a digit or an underscore. Definitely no symbols or punctuation. In reality, most people writing a blog via the transcript parser will probably care to compare with a "haskell/agda/scala" etc block.

pchiusano:
Just going to bounce this to trigger Travis, other than that is ready to merge.

PR #1181, opened 2020-01-27
Fix/1042 big list crashes

pete-ts:
Checks if all the elements in the stack are needed, in which case it returns the stack unchanged. This means that the arithmetic for finding the first element can never underflow to a `-1` index for slicing.


pete-ts:
Closing while reconsidering the implementation, this is not quite right at this point.

aryairani:
Hang onto the nice transcripts though :)

So with my proposed fix (back out the `if` and change `0` to `-1` here https://github.com/pete-ts/unison/blob/29aec1b9e94cfc22f00e65443c2d13a9e1bc13d7/parser-typechecker/src/Unison/Runtime/Rt1.hs#L463), I still get a crash with

`gc size=129 maxSlot=129`, leading to `MV.slice start=-1 len=130`.  `MV.slice` doesn't like the `-1`, but I think the real issue here is that `maxSlot` = `size`.  If `maxSlot` is an array index, it should always be strictly less than `size`.

PR #1186, opened 2020-02-01
split help-topic(s) into its own command

### LINKED ISSUE #1083: split help-topic(s) into its own command, which lists the topics if none is supplied
Or split them off to a website.

pt2121:
fix #1083

This is what it looks like:
Screen Shot

Let me know if it matches what you have in mind.


pchiusano:
Thanks! Merging.

PR #1192, opened 2020-02-04
Lambda-case `cases` syntax

anovstrup:
This PR provides support for syntax like the following:
```
store : a -> Request {Store a} v -> v
store init = cases
  { x }                -> x
  { Store.get   -> k } -> handle k init with store init
  { Store.put v -> k } -> handle !k with store v
```

Changes:
- [x] Parse lambda-case syntax (`cases` keyword)
- [x] Pretty-print using the new syntax when appropriate
- [x] Update tests 
- [x] Update editor support

Before the next release:
- [ ] Update docs

### DIFF
--- a/parser-typechecker/src/Unison/TermParser.hs
+++ b/parser-typechecker/src/Unison/TermParser.hs
@@ -39,7 +39,7 @@ import qualified Unison.HashQualified as HQ
 import qualified Unison.Lexer as L
 import qualified Unison.Name as Name
 import qualified Unison.Names3 as Names
-import qualified Unison.Parser as Parser (seq)
+import qualified Unison.Parser as Parser (seq, uniqueName)
 import qualified Unison.PatternP as Pattern
 import qualified Unison.Term as Term
 import qualified Unison.Type as Type
@@ -77,7 +77,7 @@ term3 = do
     Just y -> Term.ann (mkAnn t y) t y
 
 keywordBlock :: Var v => TermP v
-keywordBlock = letBlock <|> handle <|> ifthen <|> match
+keywordBlock = letBlock <|> handle <|> ifthen <|> match <|> lamCase
 
 typeLink' :: Var v => P v (L.Token Reference)
 typeLink' = do
@@ -239,13 +239,23 @@ lam p = label "lambda" $ mkLam <$> P.try (some prefixDefinitionName <* reserved 
   where
     mkLam vs b = Term.lam' (ann (head vs) <> ann b) (map L.payload vs) b
 
-letBlock, handle, ifthen :: Var v => TermP v
+letBlock, handle, lamCase, ifthen :: Var v => TermP v
 letBlock = label "let" $ block "let"
 
 handle = label "handle" $ do
   b <- block "handle"
   handler <- block "with"
   pure $ Term.handle (ann b) handler b
+
+lamCase = do
+  start <- openBlockWith "cases"
+  cases <- sepBy1 semi matchCase
+  -- TODO: Add error for empty match list
+  _ <- closeBlock
+  lamvar <- Parser.uniqueName 10
+  let lamvarTerm = Term.var (ann start) (Var.named lamvar)
+      matchTerm = Term.match (ann start <> ann (last cases)) lamvarTerm cases
+  pure $ Term.lam (ann start <> ann (last cases)) (Var.named lamvar) matchTerm
 
 
 ifthen = label "if" $ do
--- a/unison-core/src/Unison/Term.hs
+++ b/unison-core/src/Unison/Term.hs
@@ -443,6 +443,7 @@ pattern LamsNamed' vs body <- (unLams' -> Just (vs, body))
 pattern LamsNamedOpt' vs body <- (unLamsOpt' -> Just (vs, body))
 pattern LamsNamedPred' vs body <- (unLamsPred' -> Just (vs, body))
 pattern LamsNamedOrDelay' vs body <- (unLamsUntilDelay' -> Just (vs, body))
+pattern LamsNamedMatch' vs branches <- (unLamsMatch' -> Just (vs, branches))
 pattern Let1' b subst <- (unLet1 -> Just (_, b, subst))
 pattern Let1Top' top b subst <- (unLet1 -> Just (top, b, subst))
 pattern Let1Named' v b e <- (ABT.Tm' (Let _ b (ABT.out -> ABT.Abs v e)))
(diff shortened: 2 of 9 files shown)


pchiusano:
@anovstrup Discussion on Slack is for more general lambda case mechanism. Dan suggesed `cases` to introduce such a lambda: 

```
cases
  { x } -> x
  { Store.get   -> k } -> ...
  { Store.put v -> k } -> ...

-- proposal to have it work for n-ary lambdas too (can be added later)
fix cases
  self { x } -> x
  self { Store.get   -> k } -> ...
  self { Store.put v -> k } -> ...
```

I like this proposal. You need some keyword or token to introduce such a lambda. Should the keyword be `cases`? Just `case`? Something else?

I think it has to be syntactically distinct from the keyword that introduces a regular pattern matching, otherwise ambiguity.

parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -238,14 +238,24 @@ lam p = label "lambda" $ mkLam <$> P.try (some prefixDefinitionName <* reserved
...
 
+lamCase = do
+  _ <- openBlockWith "matches"
+  cases <- sepBy1 semi matchCase
+  -- TODO: Add error for empty match list
+  _ <- closeBlock
+  lamvar <- Parser.uniqueName 10
+  let lamvarTerm = Term.var (ann (head cases)) (Var.named lamvar)
+      matchTerm = Term.match (ann (head cases) <> ann (last cases)) lamvarTerm cases
+  pure $ Term.lam (ann (head cases) <> ann (last cases)) (Var.named lamvar) matchTerm
```
anovstrup: Are the annotations right? Is there a way to test/examine them?

anovstrup:
@pchiusano Can you recommend someone to review this PR? I'm not sure who's best able to review parser/lexer/pretty-printer changes. It's not ready to merge, yet, but I'd appreciate feedback before going further, when somebody's got time.

pchiusano:
@anovstrup - could we actually do `cases` instead of `matches` for the lambda case intro? I think `matches` reads a little weird (I keep reading it as "this matches that" rather than "here's a bunch of cases to match against"). Also just think `cases` is a bit better since the various cases of a pattern matching expression are usually called "cases" (where each case consists of a pattern on the left side and a right hand side).

I read the `match ... with` as "match this expression against these cases". So just the cases is introduced via `cases`.

I'll take a look at the lexer / parser changes.

@atacratic could you review the pretty-printer changes?

pchiusano:
@anovstrup honestly, seeing some of the complexity in the lexer makes me wonder if we should punt on the `match ... with` change until we can rewrite the lexer (I'd like lexer to just be Megaparsec) to make these sorts of changes easier.

Like just do the lambda  case `cases` syntax. That should be easy to get working and merged and would be pretty nice to have!!

WDYT?

anovstrup:
@pchiusano 😬 You're killing me! :) 

Although the lexer changes _did_ end up being more of a bear than desirable, they're already done. If review is too difficult, my preference would be to add code/review comments rather than throw the changes away. That said, I could cherry-pick the `match .. with` changes into a separate PR so that lambda-case can be finished up sooner. No reason to couple these changes if the lambda-case doesn't use a similar keyword, and I agree that `cases` reads better than `matches` (I kept going back and forth on that).

pchiusano:
@anovstrup lol, okay, sorry! I misinterpreted your "It's not ready to merge, yet, but I'd appreciate feedback before going further, when somebody's got time." thinking you still had stuff that needed working out.

If you were doing this from scratch it would have been better to separate the lambda cases into a separate PR but I think it's okay not to split them up now and I'm guessing it'd be a hassle.

I'll review the parser and lexer changes, hopefully this weekend sometime.

anovstrup:
@pchiusano Splitting off a PR would be easy, actually. The commits for lambda-case are separate from ones for match-with, so it's a simple cherry-pick. There's still stuff to do (for both parts; see the checkboxes in the description), but the hardest stuff is done (I think).

anovstrup:
@pchiusano @atacratic I split off the `match .. with` changes into a separate PR (#1214). Those are the trickier changes, but I still wouldn't mind feedback on this PR.

pchiusano:
@anovstrup for the pretty-printer, I think you could check here inside `TermPrinter.prettyBinding0`. You're looking for `f v1 v2 ... vN = match vN with ...` where `vN` is not a free variable in the `match` (you can use `Term.freeVars` to get the set of free variables of a term).

```Haskell
    LamsNamedOrDelay' vs body ->
      let (im', uses) = calcImports im body'
          -- In the case where we're being called from inside `pretty0`, this
          -- call to printAnnotate is unfortunately repeating work we've already
          -- done.
          body'       = printAnnotate env body
      in  PP.group
            $         PP.group (defnLhs v vs <> fmt S.BindingEquals " =")
            `PP.hang` uses [pretty0 env (ac (-1) Block im' doc) body']
```

You can use the `Term.Match'` pattern to detect if the body is a `match` expression.

Let me know how it goes!

anovstrup:
@pchiusano Thanks for the tip. I've starting looking at it -- hopefully I'll have something working this weekend.

anovstrup:
@pchiusano Turns out it's a bit more complicated, I think. In the cases we're interested in, `vN` _is_ (of course) a free variable in the `match` term, since it's the scrutinee. We need to check that it's _not_ free in the cases themselves, but the MatchCases are not themselves terms (nor are the patterns, which may need to be checked, too, in case they reference the variable?). So I think I need to check each guard and case RHS term with `freeVars`, but I'm not sure what to do with the Patterns.

pchiusano:
@anovstrop - cool, yeah that makes sense. You don’t need to do anything for the patterns, those don’t contain variable references, just the RHS and guards is sufficient. You can take my word for this for now, but I’ll dig up link re: how pattern matching is encoded with ABTs if you are curious.

anovstrup:
@pchiusano Progress! I'm probably not dealing with imports correctly, though — I'd appreciate if you or @atacratic could review my WIP when you get a chance. In the meantime, I'm going to work on adding more test cases and pretty-printing `cases` outside of let bindings.

anovstrup:
Any idea why this would fail as a test case in the TermParser tests, while a similar form parses and typechecks just fine in ucm?

```
"handle foo with cases\n" ++
" { a -> k } -> b\n" ++ 
" { x } -> x"
```

It fails with: 
```
🦄  termparser.handle foo with cases
2:6:
unexpected ->
    2 |  { a -> k } -> b

💥  termparser.handle foo with cases
termparser.handle foo with cases FAILURE 2:6:
unexpected ->
    2 |  { a -> k } -> b
  crash, called at tests/Unison/Test/TermParser.hs:225:7 in main:Unison.Test.TermParser
```

This, on the other hand, parses, typechecks, and evaluates correctly in ucm:
```
ability Abort where abort : a

> handle
    Abort.abort
    1
  with cases
    { Abort.abort -> k } -> 0
    { x } -> x
```

anovstrup:
I made the Abort example a test case, and it fails with this:
```
I don't know about any ability constructor named
Abort.abort. Maybe make sure it's correctly spelled and
that you've imported it:

    5 |   { Abort.abort -> k } -> 0
```

So I guess that explains the other failure... I didn't realize that the parser knew about existing definitions/names.

pchiusano:
@anovstrup yes pattern constructor references are resolved at parse time. This was a mistake but isn’t a slam dunk to fix.

anovstrup:
This is ready for review, but I'm not _that_ confident about the pretty-printing code.

atacratic:
Sorry for not taking a look at this yet - I'll review the pretty-printer changes on Wednesday night.

Great that you've done this! :+1:

unison-core/src/Unison/Term.hs
```diff
@@ -781,6 +782,23 @@ unLamsUntilDelay' t = case unLamsPred' (t, (/=) $ Var.named "()") of
...
+-- Same as unLamsUntilDelay', but only matches if the lambda body is a match
+-- expression, where the scrutinee is also the last argument of the lambda
+unLamsMatch'
+  :: Var v
+  => AnnotatedTerm2 vt at ap v a
+  -> Maybe ([v], [MatchCase ap (AnnotatedTerm2 vt at ap v a)])
+unLamsMatch' t = case unLamsUntilDelay' t of
+    Just (reverse -> (v1:vs), Match' (Var' v1') branches) |
+      (v1 == v1') && not (Set.member v1' (Set.unions $ freeVars <$> branches)) ->
+        Just (reverse vs, branches)
```
pchiusano: These three lines are pretty epic and are doing a lot, but pretty easy to follow.

I might have written as `not (any (Set.member v1') (freeVars <$> branches))` to avoid needing to build up the union'd `Set` (which does a bunch of tree rebalancing stuff) but I think it's totally fine as is.

Issue #1196, opened 2020-02-04
reflog gets multiple entries for certain commands that do independent modifications to the root.


aryairani:
...necessitating multiple `undo`s to undo the command. (Also necessitating remembering how many `undo`s a given command requires!)

examples: 
* `merge` and `pull` get two entries: one for the namespace change, and another for the auto-`patch` that happens afterwards.

* `pr.load` gets three: one for loading "base", one for loading "head" and one for merging them together.

![image](https://user-images.githubusercontent.com/538571/73760383-3e2cfe00-473b-11ea-9419-5650ac42929f.png)

Propose moving reflog logic outside of the main Input handler, so that there's only one update per `Input`.  This may mean undoing some of the initial reflog refactors (including places that the `inputDescription` are threaded through).

aryairani:
Disregard; I later realized that skipping over / squashing entries in the reflog isn't the right solution.  It would be better to not create intermediate entries to begin with.

aryairani:
superceded by #1201

Issue #1194, opened 2020-02-04
hide git output for successful executions


aryairani:
![image](https://user-images.githubusercontent.com/538571/73759124-77fd0500-4739-11ea-96df-4416a78b176a.png)

Yuck, right?

aryairani:
Maybe some sort of activity spinner using `\r` would be good instead?

aryairani:
@pchiusano What do you think about the current "easy-win" behavior?  It prints the ⚙️commands but no longer prints the commands' output.

aryairani:
In the screenshot above, it would just be the lines that start with "⚙️".

aryairani:
Closed via 8687e1d

Issue #966, opened 2019-11-14
implement command to generate pull request text


aryairani:
Generates some text to be shared with prospective reviewers
```
pr.create <base-url> <head-url>
```
or eventually:
```
pr.create-push <base-url> <head-url> <head-path>
-- pushes <head-path> to <head-url> and diffs vs <base-url>
```

aryairani:
```
.> pr.create https://github.com/unisonweb/base https://github.com/puffnfresh/base 

  + Adds / updates:
  
    Either.swap Tuple.swap Weighted.eithers Weighted.tuples

.> pr.load https://github.com/unisonweb/base https://github.com/puffnfresh/base 

	checked out https://github.com/unisonweb/base to base
	checked out https://github.com/puffnfresh/base to head

	merged namespace is in merged
	use `diff.namespace base merged` for details

	merged has 999 todos

.> diff.namespace base merged
```

undecided on argument order for each

aryairani:
```
  The changes summarized below are available for you to review, using the following command:
  
    `pr.load https://github.com/unisonweb/base https://github.com/puffnfresh/base`
  
  Added definitions:
  
    1. Either.swap      : Either a b -> Either b a
    2. Tuple.swap       : Tuple a b -> Tuple b a
    3. Weighted.eithers : Weighted a -> Weighted b -> Weighted (Either a b)
    4. Weighted.tuples  : Weighted a -> Weighted b -> Weighted (Tuple a b)
```

aryairani:
One issue here is that the numbered args don't refer to anything that exists locally, so you'll get errors if you try to interact with them.
```
.> view 1

  ⚠️
  
  The following names were not found in the codebase. Check your spelling.
    .Either.swap#8hr15i0otmkr4npjlva0c06lgvif4qagktr9tl4cls6och8uellgtoef614rivqj9d3s27ajilc6i0b2fpu4j3ol2n4hutc0ok1727g
```

aryairani:
captured the bug as #1203

PR #1197, opened 2020-02-04
implement pr.create / pr.load

aryairani:
This implements the `pr.create` and `pr.load` commands described in #966.

Apart from #1194 and #1196, the result seems decent.

image

image

Not sure where that extra prompt line came from at the bottom; seems like a fswatch glitch.


atacratic:
Awesome!! Nice one! 🤓👍 Really excited to try this out.

aryairani:
Funny that this has a green checkmark even while it "has conflicts that must be resolved".

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -329,6 +331,12 @@ loop = do
           LoadI{} -> wat
           PreviewAddI{} -> wat
           PreviewUpdateI{} -> wat
+          CreatePullRequestI{} -> wat
+          LoadPullRequestI base head ->
+            "pr.load "
+              <> (uncurry3 printNamespace) base
```
mitchellwrosen: unnecessary parens here and below

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -500,12 +508,49 @@ loop = do
...
+            (ppe, diff) <- diffHelper (Branch.head baseBranch) (Branch.head merged)
+            respondNumbered $ ShowDiffAfterCreatePR baseRepo headRepo ppe diff
+        where
+        gitError = eval . Notify . GitError input
+
+      LoadPullRequestI baseRepo headRepo ->
+        if Branch.isEmpty0 currentBranch0 then do
+          let textToRelative = 
+                Path.Relative . Path.Path . Seq.singleton . NameSegment.NameSegment
+              base = textToRelative "base"
```
mitchellwrosen: I wonder how builtin strings like this will interact with the pluggable syntax; I guess there'd just be a partial function pulled out of the sky to parse a `Text` as a `Maybe NameSegment`, and if `"base"` is an invalid name segment... then you can't load a pull request 😅

aryairani: Hmm, I guess I'd think "base" here would be the literal underlying name, and that any pluggable syntax should have some way to represent it (e.g. syntax-specific escaping).

Like, we should probably allow a namespace called `.where`; we currently don't but it's because the ucm command parser is inappropriately (IMO) sharing the basic unison syntax parser for wordyIds.

mitchellwrosen: Ah, I see... so then there would be global rules for what a name segment _is_, i.e. maybe it's just defined as a non-empty string of unicode code points

aryairani: Yeah, I think it's that.

mitchellwrosen: In that case I think it's safe/smart to put an `IsString` instance for relative paths (which creates a singleton path of course)

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -324,6 +326,9 @@ cons' n (Path' e) = case e of
 consAbsolute :: NameSegment -> Absolute -> Absolute
 consAbsolute n a = Absolute . cons n $ unabsolute a
 
+snocRelative :: Relative -> NameSegment -> Relative
```
mitchellwrosen: Could write a [Snoc](http://hackage.haskell.org/package/lens-4.19/docs/Control-Lens-Cons.html#t:Snoc) instance for all these similar cons/snoc things

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: I worry about accidentally trying to create a non-singleton path from a string constant and ending up with an unexpected singleton path; WDYT?  NameSegment currently has an `IsString` instance though.  Maybe a constructor like `Relative.singleton` though?

parser-typechecker/src/Unison/Codebase/Path.hs
aryairani: Interesting...

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
mitchellwrosen: Yeah; `Relative.singleton` is a good idea. I do think you're right about the confusion since the codebase today does have strings like `"Int.+"` that are turned into a multi-segment name. But when the concept of pluggable syntax matures, I think it will become more be intuitive/obvious that `.` is not special in any way, so a string literal really cannot be multi-segment.

aryairani:
> lgtm 👍 ! This patch leaves me wondering about unifying the concept of remote and local namespaces. What if I want to maybe peek at a PR to a remote base with a local head?

Yeah, that seems like it should be a thing.  We could do more to unify the concept of remote and local namespaces; just need to avoid like doing an external git pull every time someone saves a .u file.  We'll need a more efficient git client; to that end, I'd like to understand better how the command-line git client gets data from github (without involving API limits).

parser-typechecker/src/Unison/Codebase/Path.hs
aryairani: I created #1200 to track this.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: Also I forgot that there was an `instance IsString NameSegment`.  I got rid of the call to `textToRelative` and added a call to `relativeSingleton :: NameSegment -> Relative` using the `IsString`.  Thanks for pointing it out!

PR #1199, opened 2020-02-04
`Path.consAbsolute` function existed when it shouldn't

aryairani:
It was being used to implement `builtins.merge` incorrectly.

```
.a.b.c> builtins.merge
```
is supposed to create `.a.b.c.builtins`, but instead created `.builtins.a.b.c`; whoops!


aryairani:
Wait, I'll add a transcript.

PR #1208, opened 2020-02-05
Allow `link` and `unlink` to take numbered range arguments

### LINKED ISSUE #1127: link command that accepts 1 or more source definitions
Currently, `link` command just takes one source definition, and one target, so `link thing .licenses.MIT`. Instead, let's make it work for any number of source definitions:

```
.mylib> find
.mylib> link 1-148 .authors.pchiusano
.mylib> link 1-148 .licenses.MIT
.mylib> link def1 def2 .authors.runarorama
```

It has the same effect as running `link` on each definition in the namespace. Primary use case is probably exactly this - assigning authorship and license info in bulk. We'll want to do this for base and others will want to do this for preexisting Unison packages they've written.

Separate issues:

* base library should be licensed MIT and can get an author of myself, Runar, and Arya. As new definitions get added to base they'll be authored by whoever writes them!
* we might want to have a config setting for default author and license used for new definitions

runarorama:
Fixes #1127


aryairani:
@runarorama Can you look over and commit the changed transcript outputs (currently causing CI to fail)

aryairani:
Also could you add some comments explaining `manageLinks` and `OneOrNumbers`, when should you use that instead of `OnePlus`?  I assumed the numbers ended up counting as normal arguments.

pchiusano:
@runarorama took a look, nice that you knocked out unlink as well! Can you:

1. Remove accidentally added file 🙃
2. Add a transcript or add some test coverage of this to an existing transcript. (Like a ‘find’ followed by a bulk link of all the found definitions, then a ‘links’ to verify the metadata is now there)
3. As part of the transcript, can you show the history command after doing one of these bulk link/unlinks? Will let us eyeball that we aren’t creating needless nodes in the causal graph. (Like if 100 definitions in List package get new metadata, that should be node in the history)

runarorama:
@aryairani There's a comment on the `OneOrNumbers` constructor. Does it need more explanation? `OnePlus` doesn't do it since that assumes the arguments are at the end whereas `link` assumes the metadata is at the end. Maybe we should swap the order?

aryairani:
>There's a comment on the `OneOrNumbers` constructor. Does it need more explanation? `OnePlus` doesn't do it since that assumes the arguments are at the end whereas `link` assumes the metadata is at the end. Maybe we should swap the order?

Hm yeah, I saw the comment but still wasn't sure how it worked.  Partly needed to page in the context, which is always challenging for PRs touching less recent code.  Also, I think I should've documented this code better to start with than I did.  

Unless we beef up `argType` to a regular-expression sort of thing, or more sophisticated parser to figure out which arguments of which type have been supplied at a given point, I think variable-length arg specifiers have to come at the end, not in the middle; it can't guess whether you're done with an early variable-length series of args or not.  It doesn't actually matter in the case of `links` because all the arguments have the same type:
```haskell
  [(OneOrNumbers, exactDefinitionQueryArg), (Required, exactDefinitionQueryArg)]
```
so it could equivalently be expressed as `[(Required, ...), (OnePlus...)]`, but WDYT?

`minArgs` and `maxArgs` I guess aren't being used anywhere; I think they were supposed to be helpers to eyeball the number of args and short-circuit the cases where the user supplied an impossible number, but they should probably get a sanity check too if we add another `IsOptional` case.  They look like they might currently be broken anyway though, so not sure which way to go with them.

> `manageLinks` is literally just a "promote to function" refactoring of the two cases, `LinkI` and `UnlinkI`, which for some reason were exactly identical.

Maybe just a note that would help someone know whether and how to use that function without having to read the full implementation.  It's already better as-is than it was before the PR (it was already undocumented AND duplicated), but when we're forced to look at this stuff (for code review) is when it's easiest to notice spots that maybe already needed improvement, and now has the benefit of someone having understood it recently enough to explain it :)

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -932,17 +932,25 @@ viewPatch = InputPattern "view.patch" [] [(Required, patchArg)]
...
+  "link"
+  []
+  [(Required, exactDefinitionQueryArg), (OnePlus, exactDefinitionQueryArg)]
+  (fromString $ concat
+    [ "`link dest src` creates a link to `dest` from `src`. "
+    , "Use `links src` or `links src <type>` to view outgoing links, "
+    , "and `unlink dest src` to remove a link. The `src` can be either the "
+    , "name of a term or type, multiple such names, or a range like `1-4` "
+    , "for a range of definitions listed by a prior `find` command."
+    ]
```
aryairani: I get confused (especially with the earlier wording) about what `dest` and `src` are.  What do you think about `defn` and `metadata...`?

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -689,55 +731,11 @@ loop = do
...
-            mdValuel = toList (getHQ'Terms mdValue)
-        case (srcle, srclt, mdValuel) of
-          (srcle, srclt, [Referent.Ref mdValue])
-            | length srcle < 2 && length srclt < 2 -> do
-              mdType <- eval $ LoadTypeOfTerm mdValue
-              case mdType of
-                Nothing -> respond $ LinkFailure input
-                Just ty -> do
-                  stepAt (parent, step (Type.toReference ty))
-                  success
```
aryairani: Was it intentional that the new definition produces some output that the old one doesn't?  Seems fine, just double-checking.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -369,6 +372,45 @@ loop = do
...
+          case (srcle, srclt, mdValuel) of
+            (srcle, srclt, [Referent.Ref mdValue]) -> do
+              mdType <- eval $ LoadTypeOfTerm mdValue
+              case mdType of
+                Nothing -> respond $ LinkFailure input
+                Just ty -> do
+                  let steps =
+                        second (const . step $ Type.toReference ty)
+                        . first (Path.unabsolute .
+                                   Path.toAbsolutePath currentPath') <$> srcs
```
aryairani: The conflicting edit was just to replace `Path.toAbsolutePath currentPath'` with `resolveToAbsolute`.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -46,6 +46,12 @@ isRoot' = either isRoot (const False) . unPath'
 isRoot :: Absolute -> Bool
 isRoot = Seq.null . toSeq . unabsolute
 
+root :: Absolute
+root = Absolute (Path mempty)
```
aryairani: Could use the existing `Path.absoluteEmpty` instead of introducing new name.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -46,6 +46,12 @@ isRoot' = either isRoot (const False) . unPath'
...
+
+root' :: Path'
+root' = Path' $ Left root
```
aryairani: Looks like this might be unused.

parser-typechecker/src/Unison/CommandLine/InputPattern.hs
```diff
@@ -55,14 +55,14 @@ argType ip i = go (i, args ip) where
...
-  -- If requesting a later parameter, decrement and drop one.
-  go (n, (Required, _) : args) = go (n - 1, args)
   -- Vararg parameters should appear at the end of the arg list, and work for
   -- any later argument number.
   go (_, [(ZeroPlus, t)]) = Just t
   go (_, [(OnePlus, t)]) = Just t
   -- Optional parameters only work at position 0, under this countdown scheme.
   go (_, [(Optional, _)]) = Nothing
+  -- If requesting a later parameter, decrement and drop one.
+  go (n, (Required, _) : args) = go (n - 1, args)
```
aryairani: I don't _think_ this move changes the result; so did it just read better below?

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -963,17 +971,18 @@ links = InputPattern
...
 unlink :: InputPattern
-unlink = InputPattern "unlink" ["delete.link"]
-  [(Required, exactDefinitionQueryArg),
-   (Required, exactDefinitionQueryArg) ]
-  "`unlink src dest` removes a link from `src` to `dest`."
+unlink = InputPattern
+  "unlink"
+  ["delete.link"]
+  [(Required, exactDefinitionQueryArg), (OnePlus, exactDefinitionQueryArg)]
+  "`unlink dest src` removes a link to `dest` from `src`."
```
aryairani: Should update the help here (like `link`) to explain the multiple metadata option.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -316,8 +317,10 @@ loop = do
           PropagatePatchI p scope -> "patch " <> ps' p <> " " <> p' scope
           UndoI{} -> "undo"
           ExecuteI s -> "execute " <> Text.pack s
-          LinkI from to -> "link " <> hqs' from <> " " <> hqs' to
-          UnlinkI from to -> "unlink " <> hqs' from <> " " <> hqs' to
+          LinkI froms to ->
+            "link " <> Text.pack (show $ hqs' <$> froms) <> " " <> hqs' to
+          UnlinkI froms to ->
+            "unlink " <> Text.pack (show $ hqs' <$> froms) <> " " <> hqs' to
```
aryairani: I think the arg order needs to be swapped here too

PR #1209, opened 2020-02-06
Refactor pull / merge / pr.load

### LINKED ISSUE #1200: create and use Snoc instances for Path.
[Snoc](http://hackage.haskell.org/package/lens-4.19/docs/Control-Lens-Cons.html#t:Snoc) candidates include:
- [x] `Path -> NameSegment -> Path`
- [x] `Path' -> NameSegment -> Path'`
- [x] `Relative -> NameSegment -> Relative`
- [x] `Absolute -> NameSegment -> Absolute`
- [x] `Name -> NameSegment -> Name`
Note that `Cons` is generally wrong for `Absolute` — we had a `consAbsolute`, but it was only used in one place, to implement a bug (#1199).  😅

### LINKED ISSUE #1201: refactor `LoadPullRequestI` to only produce one root update
`pullRemoteBranchAt` is called once in the implementation of `PullRemoteBranchI` and twice in the implementation of `LoadPullRequestI`.  `LoadPullRequestI` also calls `updateAtM` to complete a merge.

These three should be consolidated into a single update at the root, refactoring `pullRemoteBranchAt` if appropriate.

### LINKED ISSUE #1202: refactor `MergeLocalBranchI` and `PullRemoteBranchI` to produce only one root update
Currently these each update the root twice, creating a messy history & potentially messing up a live demo :P

Let's refactor them to do at most one update (refactoring `pullRemoteBranchAt` and `propagatePatch` as appropriate); possibly by delaying the actions until after whatever conditionals have passed.

Related to #1201.

aryairani:
Welcome to my PR of spiraling scope.

### What:
* We decided not to combine merge and propagate steps after all for pull/merge (closes #1202).

  @pchiusano and I decided it's good to have these steps distinct in the reflog, and although we might like to see them combined when a user does an `undo`, that's best addressed by changing how `undo` works.

* Combine the pull/pull/merge steps of `pr.load` into one Causal update (closes #1201).  The propagate step remains separate.

* Remove automatic `todo` output after a `merge`.  We may revisit this later and print a quantitative summary, rather than the standard `todo` output, which duplicates some of the post-merge diff information.

* Added optional local staging dest arg to `pr.load` eg
  ```
  pr.load https://github.com/unisonweb/base https://github.com/puffnfresh/base .prs.puffnfresh
  ```

* Made a few helpers that do the work of pulling, merging, diffing, propagating.

### Notable changed behavior
* A merge can result in a conflicted patch, and when ucm tries to auto-propagate a conflicted patch after the merge, it will generate a `PatchNeedsToBeConflictFree` message, which must go in a `ucm:error` block for transcripts.  I think this was the expected behavior before, and although I don't see the change that changed it, it evidently wan't happening before.

### Notable codebase changes
* Added some versions of HandleInput.step* functions that return `Bool` instead of `()`

  They return `True` if they resulted in a change / distinct step; `False` otherwise.  This is needed if we want to have some output conditional on whether a command had an effect.

* Generalized some of the Branch.step functions:
  e.g.
  ```haskell
  stepDistinctM :: (Applicative m, Eq e, Hashable e)
    => (e -> m e) -> Causal m h e -> m (Causal m h e)
  ```
  became
  ```haskell
  stepDistinctM  :: (Applicative m, Functor n, Eq e, Hashable e)
    => (e -> n e) -> Causal m h e -> n (Causal m h e)
  ```
  allowing a different monad in the step function than in the `Causal`.  I didn't end up using these though, so I could back them out if desired.
* add some Lens.Cons / Lens.Snoc instances (closes #1200) to consolidate several `Path` helper functions into `Lens.cons` / `Lens.snoc` calls:

  in `module Unison.Codebase.Path`:
  ```haskell
   -- extract the top parent from a path, or nest a path one level deeper.
  instance Cons Path Path NameSegment NameSegment

  -- extract the leaf segment from a path, or descend one level deeper into one.
  -- (could probably simplify `parsePath'`, `parsePath'Impl`, `parseSplit'`, etc. with these?)
  instance Snoc Relative Relative NameSegment NameSegment
  instance Snoc Absolute Absolute NameSegment NameSegment
  instance Snoc Path Path NameSegment NameSegment
  instance Snoc Path' Path' NameSegment NameSegment
  ```
  in `module Unison.Codebase.NameSegment`:
  ```haskell
  instance Snoc Name Name NameSegment NameSegment
  ```

* add `class Path.Resolve` to consolidate several `Path` helper functions:
  ```haskell 
  where
    resolve :: l -> r -> l
  ```
  resolves one path against another, e.g.:
  ```haskell
  resolve .foo .bar == .bar
  resolve .foo  bar == .foo.bar
  resolve  foo .bar == .bar
  resolve  foo  bar ==  foo.bar
  ```
  in `module Unison.Codebase.Path`:
  ```haskell
  instance Resolve Path Path
  instance Resolve Relative Relative
  instance Resolve Absolute Relative
  instance Resolve Path' Path'
  instance Resolve Path' Split'
  instance Resolve Absolute Path'
  ```
### Misc
* removed `{-# OPTIONS_GHC -Wno-unused-top-binds #-}` and some unused top binds from `HandleInput.hs` (towards #1111).
* Added `type InputDescription = Text` and used it in places where we're passing input description for the reflog.  Could do a `newtype`.
* Eliminated `Output.NothingToDo Input` in favor of explicit cases `MergeAlreadyUpToDate` and `PreviewMergeAlreadyUpToDate` (toward #1141).

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -37,6 +37,7 @@ module Unison.Codebase.Branch
   , stepEverywhere
   , uncons
   , merge
+  , adjustHeadMN
 
     -- * Branch children
     -- ** Children lenses
@@ -211,6 +212,13 @@ makeLenses ''Raw
 toNames0 :: Branch0 m -> Names0
 toNames0 b = Names (R.swap . deepTerms $ b)
                    (R.swap . deepTypes $ b)
+
+adjustHeadMN :: (Monad m, Monad n)
+  => (Branch0 m -> n (Branch0 m))
+  -> (forall a. m a -> n a)
+  -> Branch m
+  -> n (Maybe (Branch m))
+adjustHeadMN f g = (fmap . fmap) Branch . Causal.adjustHeadMN f g . _history
 
 -- This stops searching for a given ShortHash once it encounters
 -- any term or type in any Branch0 that satisfies that ShortHash.
@@ -562,7 +570,7 @@ isEmpty = (== empty)
 step :: Applicative m => (Branch0 m -> Branch0 m) -> Branch m -> Branch m
 step f = over history (Causal.stepDistinct f)
 
-stepM :: Monad m => (Branch0 m -> m (Branch0 m)) -> Branch m -> m (Branch m)
+stepM :: (Monad m, Monad n) => (Branch0 m -> n (Branch0 m)) -> Branch m -> n (Branch m)
 stepM f = mapMOf history (Causal.stepDistinctM f)
 
 cons :: Applicative m => Branch0 m -> Branch m -> Branch m
@@ -600,8 +608,8 @@ stepAtM p f = modifyAtM p g where
     b0' <- f (Causal.head b)
     pure $ Branch . Causal.consDistinct b0' $ b
 
-stepManyAtM :: (Monad m, Foldable f)
-            => f (Path, Branch0 m -> m (Branch0 m)) -> Branch m -> m (Branch m)
+stepManyAtM :: (Monad m, Monad n, Foldable f)
+            => f (Path, Branch0 m -> n (Branch0 m)) -> Branch m -> n (Branch m)
 stepManyAtM actions = stepM (stepManyAt0M actions)
 
 -- starting at the leaves, apply `f` to every level of the branch.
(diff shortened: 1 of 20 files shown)


pchiusano:
LGTM! I like the snoc typeclass.

PR #1211, opened 2020-02-07
Issue 1015: find by type is affected by the var names chosen.

pete-ts:
1. my editor seems to have picked up a number of trailing whitespaces and removed them :man_facepalming: 
2. `unison-core/src/Unison/Type.hs` has a local definition of `freeVars` which is calling ABT.freeVars, so I removed the `ABT.` part
3. I have added a transcript that shows that `find` does indeed work as expected, if `generalizeLowercase` respects the ordering of the Term that contains the `freeVars` set.

However I doubt my fix is actually sound: ideally `freeVars` should be an `Ordered.Set` so that it respects the order of free variable in the Arrow in this instance. :thinking:

### DIFF
--- a/unison-core/src/Unison/ABT.hs
+++ b/unison-core/src/Unison/ABT.hs
@@ -279,6 +279,13 @@ allVars t = case out t of
   Abs v body -> v : allVars body
   Tm v -> Foldable.toList v >>= allVars
 
+orderedFreeVars :: (Foldable f, Ord v, Show v) => Term f v a -> [v]
+orderedFreeVars t =
+    filter (`elem` unordered) ordered
+  where
+    ordered = nub $ allVars t
+    unordered = Set.toList $ freeVars t
+
 freshes :: Var v => Term f v a -> [v] -> [v]
 freshes = freshes' . freeVars
 
@@ -559,16 +566,16 @@ components :: Var v => [(v, Term f v a)] -> [[(v, Term f v a)]]
 components = Components.components freeVars
 
 -- Converts to strongly connected components while preserving the
--- order of definitions. Satisfies `join (orderedComponents bs) == bs`. 
+-- order of definitions. Satisfies `join (orderedComponents bs) == bs`.
 orderedComponents' :: Var v => [(v, Term f v a)] -> [[(v, Term f v a)]]
-orderedComponents' tms = go [] Set.empty tms 
+orderedComponents' tms = go [] Set.empty tms
   where
   go [] _ [] = []
   go [] deps (hd:rem) = go [hd] (deps <> freeVars (snd hd)) rem
   go cur deps rem = case findIndex isDep rem of
     Nothing -> reverse cur : let (hd,tl) = splitAt 1 rem
                              in go hd (depsFor hd) tl
-    Just i  -> go (reverse newMembers ++ cur) deps' (drop (i+1) rem) 
+    Just i  -> go (reverse newMembers ++ cur) deps' (drop (i+1) rem)
                where deps' = deps <> depsFor newMembers
                      newMembers = take (i+1) rem
     where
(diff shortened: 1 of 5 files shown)


aryairani:
Hey, thanks for taking a stab at this.  I'm not quite sure what we'll do yet though, because @pchiusano had raised the concern that any fix could affect the hashes of existing definitions in the codebase, and as such it should wait until our next codebase format migration.  On the other hand, the universally quantified variables are stored in the codebase even when they not specified by hand.  So suppose someone manually created this function: `const : forall a b. b -> a -> b`.  Shouldn't it still show up when searching for `b -> a -> b`, which would expand to `forall b a. b -> a -> b` or when searching for `a -> b -> a`, which would expand to `forall a b. a -> b -> a`; neither of which match `const`. ?

> 1. my editor seems to have picked up a number of trailing whitespaces and removed them 🤦‍♂

This is okay.  It's better if you can catch and revert them in your git client before committing, but really we just need to stop all the work, close all the PRs, and remove all the trailing whitespaces from all files, and resynchronize.  I bet I'm responsible for many of them due to a bug in my editor.  😓 

> 2. `unison-core/src/Unison/Type.hs` has a local definition of `freeVars` which is calling ABT.freeVars, so I removed the `ABT.` part

Side note, as much as possible, it is helpful to resist fixing unrelated things in a single PR; better to make a note and do those as a separate PR, reducing the review load for each.  Same for other little things like fixing lint errors like eta reduction or redundant `$`.

> 3. I have added a transcript that shows that `find` does indeed work as expected, if `generalizeLowercase` respects the ordering of the Term that contains the `freeVars` set.
>
> However I doubt my fix is actually sound: ideally `freeVars` should be an `Ordered.Set` so that it respects the order of free variable in the Arrow in this instance. 🤔

If we do end up going with this sort of approach, why not use an ordered set, like you mentioned?

pete-ts:
> If we do end up going with this sort of approach, why not use an ordered set, like you mentioned?

Lack of stack-fu, to get the `Data.Set.Ordered` dependency into the project. Might just happen given some more time to look into it.

pete-ts:
> Hey, thanks for taking a stab at this. I'm not quite sure what we'll do yet though, because @pchiusano had raised the concern that any fix could affect the hashes of existing definitions in the codebase, and as such it should wait until our next codebase format migration. On the other hand, the universally quantified variables are stored in the codebase even when they not specified by hand. So suppose someone manually created this function: `const : forall a b. b -> a -> b`. Shouldn't it still show up when searching for `b -> a -> b`, which would expand to `forall b a. b -> a -> b` or when searching for `a -> b -> a`, which would expand to `forall a b. a -> b -> a`; neither of which match `const`. ?
> 

Valid point, indeed currently `const` with manually specified `forall a b.` would be found when searched without the `forall`.

pchiusano:
@pete-ts I can think of a couple ways of fixing:

1. Do the normalization just before insertion into the search index, and also just before doing a lookup into the search index. No repoformat change. Unfortunately, this only fixes for new definitions going forward. Old definitions might still be in the index at the wrong key and wouldn't be located as expected. If/when that happens, re-adding the definition should trigger the search index update so it could be fixed up incrementally.
  * We could also have a "rebuild index" command which rebuilds the search index in its entirety. Slightly annoying but if it's done now, not a big deal.
2. Just produce all permutations of quantifier order, and look for each one and union their results. (Possibly bail if number of quantifiers is bigger than say 6)

I think my vote is for option 1. The rebuild index command be separately added.

We could also wait to address this until next repoformat change. That doesn't seem great though, I'd like it if this worked properly.

aryairani:
So #1 would strip off and replace any `forall` in the type?

pete-ts:
I meant to ask for some additional information as well. I have a basic grasp of what the moving parts in handling the "find" command is but not 100% clear on `before insertion to the search index` or `before doing a lookup in the search index`, some pointing bits out would be of help.

PR #1214, opened 2020-02-08
Replace case-of with match-with

anovstrup:
This PR will replace `case .. of` syntax with `match .. with`.

Changes:
- [x] Update parser/lexer
- [x] Update pretty-printer
- [x] Update tests 
- [x] Update transcripts
- [x] Update editor support

When merging:
- [ ] Update unisonweb-org docs to use the new syntax


parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -696,23 +704,36 @@ keywords = Set.fromList [
...
+  case findNearest l (Set.fromList ["handle", "match"]) of
+    Just "match"  -> "match-with"
+    Just "handle" -> "handle-with"
+    _ -> "with"
+layoutCloseAndOpenKeywordMap kw _ = kw
+
+openingKeyword :: BlockName -> String
+openingKeyword        "then" = "if"
+openingKeyword        "else" = "then"
+openingKeyword        "with" = "match or handle" -- hack!!
```
anovstrup: This is used only when `with` appears without a matching `match` or `handle` first, which is an error.  The "match or handle" string winds up in the error message, colored red.

pchiusano: Epic hack! It's fine though. In general, I'd put this info in a comment in the code rather than the PR notes.

Issue #843, opened 2019-10-04
Compute minimum unique hash prefix length for branches


aryairani:
(related to #842)

Add a member like:
```
shortBranchHashLength :: Int
```
which should return the smallest number of base32hex characters necessary to uniquely identify any known branch hash by prefix, in some way that doesn't do O(n) work on every call.

atacratic:
Is this definitely worthwhile? Why not just always use, say, 8, like you see with GitHub hashes?

And why make the length decision globally instead of on a per hash basis at the point you're printing out each hash?

aryairani:
@atacratic The minimum size needed grows with the size of the repo; Github works the same way.  Didn’t understand the second question.

atacratic:
Are you sure github works that way?  I see only seven characters of hash both [here](https://github.com/torvalds/linux/commits/master) and [here](https://github.com/seagreen/unison-code-explorer/commits/master).  7 chars of hex is a space of size 256 million.  

I guess I'm not clear what this will be used for and whether the truncated hashes need to be globally collision-free across the repo.

I think it would be good to get a clear statement of benefit before people start contributing code for this.

aryairani:
Sorry, I misspoke when I said Github works that way; it doesn’t necessarily need to, because its short hashes are hyperlinked to full hashes.  So even if you had two colliding hashes on the same HTML page, they can still refer to two distinct things.  But if you are entering the hash freeform, it needs to uniquely identify one.

Command-line git works that way (7+ characters in a short hash), so I would bet that if you asked it for short hashes, it would give different lengths on those two repos (am on mobile though and can’t check).

I think 7 hex chars is only 4^7=16k definitions.

atacratic:
Ah, interesting about the command line.

I'm just thinking that in the edge case where the user sees a truncated hash come out of UCM, then inputs it again in some command or code, only to see UCM complain about an ambiguity, they can work around it by getting the full hash and trying again (is that find.verbose?)  And maybe 7 chars is enough that this is already vanishingly rare?

It's not the full birthday problem, because you don't care if there is *some* pair of colliding truncated names in the repo - just if the current one you're talking about has a collision.

(Also I think it's 16^7 == (2^4)^7 == 2^(4*7) == 2^28, right?)

aryairani:
> (Also I think it's 16^7 == (2^4)^7 == 2^(4*7) == 2^28, right?)

Yeah you’re right.

[...]

PR #1224, opened 2020-02-10
Start using tries to calculate minimum hash unique prefix length

anchpop:
See this issue - https://github.com/unisonweb/unison/issues/843

There's a lot of code involving constructing Tries here, a lot of which comes from https://github.com/viercc/trie-simple/blob/master/src/Data/Trie/Map/Hidden.hs. I added the code in directly instead of adding it as a dependency but I would be happy to change that (I didn't add it as a dependency to begin with because there are some optimizations that could be made that would require changing the trie implementation, but I didn't implement them yet).

Also, it only works for types/terms, not namespaces. I'm planning on fixing the namespaces tonight, just submitting the pull request in case anyone wants to suggest anything in the meantime.


aryairani:
I guess that until some concrete implementation changes are planned, I would just use it as a dependency if it exists in stackage.

Also — sorry we didn’t talk about this, but listing all the files in the directory is too expensive to do on each call to the function.  We need to brainstorm some disk-based trie or something so that we don’t have to do all that work often.  (Let me know if I misunderstood the current implementation.)

anchpop:
Yes, I agree that it's too expensive. I'd like to actually start working on a disk based trie, and I have an optimization in mind that will make it cheap to get the minimum unique prefix length even on disk, but I'm not sure where to get started. If you could point out some source files I should look at or give a general rundown of what changes you had in mind I could try it today.

aryairani:
Haven't got anything figured out in mind yet, but whatever format we consider doing, we wanna think about:
* can it be automatically introduced into existing codebases?
* does it need to be synced to github (a permanent part of the codebase), or could it be a locally computed / cached thing, that ucm can update it smoothly, independently after a `pull` or other external change?

Issue #915, opened 2019-10-30
Transcript runner should pick GUIDs for `unique` types deterministically


pchiusano:
Some discussion in #889 basically we'd like to avoid having the output always change on every run.

I think fix is a relatively straightforward refactoring, [this function](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Parser.hs#L63-L65) should accept an `IO Rng`, then fix all the compile errors. :) 

At all call sites other than the transcript runner, can pass the `Crypto.Random.getSystemDRG` that is currently hardcoded. For the transcript runner, use a deterministic RNG with seed set to some constant.

Looking at [`Crypto.Random`](https://www.stackage.org/haddock/lts-14.25/cryptonite-0.25/Crypto-Random.html) seems like signature becomes:

```Haskell
uniqueBase58Namegen :: DRG gen => IO gen -> IO UniqueName
```

Other stuff to possibly include as part of this PR:

* change `uniqueBase58Namegen` to `uniqueBase32Namegen`... it's not using base58 anymore :)
* Maybe bump [number of characters up to 64](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Parser.hs#L71), since 32 base32 characters is "only" 160 bits, which is kind of low for a GUID. 320 bits is well into "totally overkill" territory. :)

aryairani:
Thanks for opening this ticket.

I imagine everyone will want it deterministic for generating docs that don’t keep changing.

So I suggest we do that initially, and if that’s a problem, we add a flag (`-deterministic-guid` or `-guid-seed <n>`) or something.

pchiusano:
Okay, yeah @aryairani I am sold. :)

aryairani:
Let's revert https://github.com/unisonweb/unison/pull/889/commits/c7722061a01102469d968552a1cc6dcb31de9b2f when this is done.

aryairani:
Probably just remove all unique type guids from the transcripts when this is done.

aryairani:
Let's also revert dc26ff0 when this is done.

anchpop:
In [this code](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs#L118), do we know if we're in "transcript mode" or "codebase mode"? I'm trying to figure out exactly how I have to modify this code.

anovstrup:
[...]

PR #1226, opened 2020-02-12
PR to verify master now good

pchiusano:
Broken build was due to #915


pchiusano:
Okay, Travis succeeded.

aryairani:
Hmm, I wonder why the PR is only showing 2 changed files; that seems like a Github glitch

PR #1227, opened 2020-02-12
replace auto-derived Ord HQ instance which had ordered `z` before `a#x`

aryairani:
This PR replaces the auto-derived `Ord (HashQualified n)` instance which had ordered `z` before `a#x`. 

Possibly controversial: The new Ord HQ instance does order `#x` before `a#a`, though, because it had a simpler implementation and wasn't obviously worse than `#x` after `a#a`.

This also deletes a crazy `HQ.sort` function which claimed to sort by name, but actually ignored names and other constructor parameters.


aryairani:
I don't get why this still has shuffled `link.output.md` results.

aryairani:
Oh.  The `push` build is broken, because it inherited the un-fixed `link.md`; the `pr` build is okay, so this should be safe to merge.

Issue #1128, opened 2020-01-06
base library metadata updates


pchiusano:
- [ ] Update base library to have metadata types for authorship and licensing (an `Author` type and `License` type)
  - [ ] `.licenses.{MIT,BSD3,BSD2,Apache2...}`
  - [ ] `.authors.{pchiusano,runarorama,aryairani}`, and whoever else contributes to base
- [ ] Assign authorship and license to all definitions in base

atacratic:
Will there be a convention for what namespace the author's handle is drawn from?  e.g. 'use a github ID'.  Or will the full text of a name ("Joe Bloggs") be possible?  Some people might be unconvinced by asserting authorship through a handle like this, especially if it's a handle you can make up on the spot, and I think attribution is important to get right for licensing reasons.  

Maybe not a worry for now.

aryairani:
Git uses email and full name.  We could do the same.  Something else with signing could come later.

pchiusano:
Design session with @runarorama and @dolio to discuss `Author` and `License` representation.

Here's some basic metadata types we're proposing adding to `base` (and possibly `builtins` if needed): 

```
unique type Author = { guid : GUID, name : Text }

unique type CopyrightHolder = { guid : GUID, name : Text}

unique type License = 
  { copyrightHolders : [CopyrightHolder]
  , years : [Year], 
  , licenseType : LicenseType }

-- Use `Doc` here to get nice text-wrapping when viewing
-- and to avoid needing to stick hard line breaks in the license
unique type LicenseType = LicenseType Doc

unique type GUID = GUID Bytes

-- Years since 1970, could be negative
unique type Year = Year Int
```

#### Some notes on this

* Expected we'll have `.licenseTypes.{MIT,BSD2,BSD3,Apache2..} : LicenseType` in `base`
* It's expected that `Author` and `License` info will be set automatically based on user config. [This ticket tracks that](https://github.com/unisonweb/unison/projects/2#card-31311925).
  * `unique type CreationDate = CreationDate Int` - seconds (ms? µs?) since 1970, could be negative
  * Setting `CreationDate` automatically would be nice as well. I don't think we want to create a separate named piece of metadata for every creation date, but I think UCM could create anonymous `CreationDate` values and link them to definitions as they are added.
  * If we want to do this, it's gotta be added to `builtins` 
[...]

Issue #1056, opened 2019-12-12
Don't populate `.builtin` namespace; instead prompt user to fetch base library on empty codebase


aryairani:
Also delete pure unison functions from IOSource.hs.

Having them in `.builtin` (and thinking about how to upgrade them) adds a maintenance burden, and doesn't provide any particular benefit IMO.

Let's have `.builtin` be just be a source of names for references that are hard-coded into `ucm` and the runtime; `.base` can be where we maintain user-facing stuff, using normal `ucm` refactor/upgrade/update processes once they're usable.

We could also:

* not add `.builtin` to the initial namespace; instead, adding it only on demand with `builtins.merge` or whatever.  The thought being that `.builtin` is generally going to be out of date (populated from whatever version of ucm you had when you first initialized your codebase), so what's the point?

  * So when you start `ucm`, download a library that names the builtins (e.g. unisonweb/base that we maintain; or some other library that someone else maintains), or run `builtins.merge` and use it to bootstrap your own library.

```
Hello, you have an empty codebase. Please choose:

 1. Download the Unison standard library [recommended]
 2. I don't need that, but thanks! [insane expert mode]

> 1
```

* let `.builtin` be dynamic, according to whatever is in your version of `ucm`?  (maybe later, with some design)

PR #1228, opened 2020-02-12
Add Author and License metadata types to builtins

pchiusano:
See discussion [here](https://github.com/unisonweb/unison/issues/1128#issuecomment-584732483). 

Notes:

* Year is `Year Nat`, common era year. If someone discovers ancient Babylonian tablets containing Unison code, we may need to revisit.
* This deletes `builtins` from the initial codebase namespace - it starts out empty, see #1056 and there's a transcript, see `emptyCodebase.md` in the transcripts directory.
* As a result of deleting `builtins`, I had to add a `ucm` block at the start of most transcripts to do a `builtins.merge` to avoid transcript failures or changes to their output. Some of the transcript output is changed just due to there being an extra entry in the history (history starts out empty, then the `builtins.merge` adds a bunch of names into the namespace).

Next up, we will try using the PR process to add these to base!!!


pchiusano:
@runarorama - you can run a transcript via ucm transcript.fork to start the transcript from some existing codebase. Like the transcripts used to generate some of the docs on the website start from a codebase that just has base.

But if you are making a totally self contained transcript that starts from an empty codebase, then the transcript will generally begin by fetching its dependencies - pulling base or doing builtins.merge.

aryairani:
I guess we could have a command-line flag that says whether `builtin.` gets added or not?

pchiusano:
Yeah, maybe ucm transcript.builtins or something that runs a transcript starting from just the builtins. I think that’s my vote - the transcript runner should by default start from the same state as a blank codebase.

Or maybe ucm transcript behaves as before (Can access builtins) and we add ucm transcript.fork0 or something to start a transcript without access to even builtins?

runarorama:
I think we can address these concerns in a separate PR. My vote is to either put the builtins behind a flag, or put the blank codebase behind a flag. Either way is fine with me.

aryairani:
Sounds good; I vote we put the builtins behind a flag.

Side note @pchiusano if we had `unique` terms, maybe we could avoid needing a visible GUID in the `Author` object.  (Is there a different ticket this comment should go into?)

PR #1236, opened 2020-02-14
implement `alias.many`

### LINKED ISSUE #1221: `alias.many <relative1> [relative2...] <namespace>`

aryairani:
Check out [the transcript](https://github.com/unisonweb/unison/blob/1009e48f868a278109477e513cd3e9aef683e424/unison-src/transcripts/alias-many.output.md) for usage info.

I could use feedback on the help text, and how to format it better with `Pretty` library.

Closes #1221


aryairani:
Transcript worked before merging master, not sure offhand why it would fail now — maybe case/of vs match/with discrepancy.

PR #1237, opened 2020-02-14
Pick GUIDs deterministically when running transcripts

anchpop:
address #915

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
@@ -26,7 +26,7 @@ import qualified Data.Text                     as Text
 import           System.Directory               ( getXdgDirectory
                                                 , XdgDirectory(..)
                                                 )
-import           System.FilePath                ( (</>) )           
+import           System.FilePath                ( (</>) )
 
 import           Unison.Codebase                ( Codebase )
 import qualified Unison.Codebase               as Codebase
@@ -98,14 +98,14 @@ commandLine
   -> (NumberedOutput v -> IO NumberedArgs)
   -> (SourceName -> IO LoadSourceResult)
   -> Codebase IO v Ann
-  -> gen
+  -> (Int -> IO gen)
   -> Free (Command IO i v) a
   -> IO a
-commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSource codebase rng =
- Free.fold go
+commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSource codebase rngGen =
+ Free.foldWithIndex go
  where
-  go :: forall x . Command IO i v x -> IO x
-  go = \case
+  go :: forall x . Int -> Command IO i v x -> IO x
+  go i x = case x of 
     -- Wait until we get either user input or a unison file update
     Eval m        -> m
     Input         -> awaitInput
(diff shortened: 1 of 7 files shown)


pchiusano:
@anchpop cool! However, should we continue to use the randomness source from cryptonite package? I could be wrong, but from what I understand, System.Random may just be initialized using like the system time, which is not enough entropy for generating GUIDs.

anchpop:
Ah, you're right, we should. System.Random just gets the current time, that definitely won't work. The only reason I changed it was because I found the cryptonite api kind of hard to use but I'll switch it back.

parser-typechecker/src/Unison/Parser.hs
```diff
@@ -60,13 +60,19 @@ instance Monoid UniqueName where
...
 
-uniqueBase58Namegen :: IO UniqueName
-uniqueBase58Namegen = do
-  rng <- Random.getSystemDRG
-  pure . UniqueName $ \pos lenInBase32Hex -> go pos lenInBase32Hex rng
+-- Should be used when we actually do want the result to be different every time
+-- (most cases)
+-- But we actually want it to be deterministic when running transcripts, 
+-- in which case we need to be able to control the random seed. If that is the case you should use 
+-- `uniqueBase32Namegen`
```
aryairani: I didn’t quite follow this comment, does it have a mistake, or could you help me understand?

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -25,7 +26,7 @@ import qualified Data.Text                     as Text
 import           System.Directory               ( getXdgDirectory
                                                 , XdgDirectory(..)
                                                 )
-import           System.FilePath                ( (</>) )
+import           System.FilePath                ( (</>) )
```
aryairani: white space alert :)

aryairani:
Also, should we revert those transcript changesets mentioned in #915 as part of this PR (and rerun the transcripts) or do that separately?

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -115,8 +117,8 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
     Typecheck ambient names sourceName source -> do
       -- todo: if guids are being shown to users,
       -- not ideal to generate new guid every time
-      namegen <- Parser.uniqueBase58Namegen
-      let env = Parser.ParsingEnv namegen names
+      let namegen = Parser.uniqueBase32Namegen rng
+          env = Parser.ParsingEnv namegen names
```
pchiusano: I just spotted something else that I misread last night but is quite important. rng should be an ‘IO gen’ here and you will need to do a bind to obtain the rng. We want a fresh random number generator each time we typecheck. Otherwise the generator will produce the same values - notice it’s a pure interface.

For main you’ll pass in the IO action that grabs the system rng. It’ll be different every time since it’s seeded from /dev/urandom or something.

For the transcript runner, it’s not quite as simple as ‘pure constRng’, since you want to update the state of the rng after each use. I think something like - keep an IORef with an rng in it, and produce an IO action that reads that rng, and returns it but also sets the IORef with an updated version of the rng.

Lmk if that makes any sense.

anchpop: This should be fixed now, we now take a function that takes an integer and returns an `IO Gen`. We pass it a different integer every time (incrementing by 1). It's ignored when called from main and it just returns a new std rng. We use the number as the rng seed when calling from the transcript runner.

pchiusano:
@anchpop - thanks for addressing that. I might have just baked the seed update into the IO gen but I think the way you did it is fine too. 

Could you add yourself to CONTRIBUTORS file? Planning to merge after that if no objections @aryairani

anchpop:
Ah yeah, I did but I forgot to push that change. There you are!

pchiusano:
Branch has conflicts somehow?

anchpop:
@pchiusano Just rebased, how about now?

pchiusano:
Thanks, merging!

PR #1254, opened 2020-02-18
allow `add`/`update` to create aliases

### LINKED ISSUE #990: Types are sometimes both "OK to update" and "require alias"
```
 I found and typechecked these definitions in ~/work/unison/scratch.u. If you do
  an `add` or `update`, here's how your codebase would change:
    ⍟ These new definitions will replace existing ones of the same name and are ok
      to `update`:
      unique type io.Mode  (existing name: .builtin.io.Mode)
    x These definitions would fail on `add` or `update`:
      Reason
      needs alias   unique type io.Mode   (existing name: .builtin.io.Mode)
  Ctrl+C cancels.
```

This is with a codebase where `.builtin.io.Mode` is the exact same `unique` type (same seed) as the definition in my scratch file.

### LINKED ISSUE #1004: Change "needs alias" so that it applies only to current namespace.
The current logic, that a definition is conflicted with a "needs alias" message if it exists _anywhere_ in the codebase, is probably too strict.

We probably do want that message when the definition already exists in the current namespace, but only then. If it exist somewhere else, we should just create the alias and notify the user that we've done so.

### LINKED ISSUE #1253: allow `add`/`update` to add aliases with just a notification

### LINKED ISSUE #1267: The message on `update` is confusing.
The message says

> These new definitions will replace existing ones of the same name and are ok to `update`:

There are a few issues here:

1. If the definition already exists in the codebase, but with a different name, it's not a "new" definition.
2. It's probably better to talk about redefining/updating names than replacing definitions.
3. The existing definition under the name might actually have several other names, and those names will be updated as well.
4. The new definition might also have several other names, and if those are in the namespace already, we should report that.

Then when you `update`, it says

> I've updated to these definitions:

But no definitions are shown, only names. What UCM has really done is _redefine those names to match the scratch file_, and we should say that (preferably saying which file, too). It should also tell you a couple of more things:

1. What other names got redefined? Updating one hash actually redefines all of its names in the current namespace (points those names to the new definition).
2. What other names exist for the new definition?

aryairani:
### Overview

This PR allows `add` and `update` in UCM to create aliases instead of annoyingly failing with "needs alias" and asking the user to create the aliases manually.

After this PR is merged, both of these commands simply create the aliases you need, and tell you they've done so.

Before:

image

After:

image

Another example (couldn't get this far in `master`):
image

### Implementation notes

Most of the meat of this change is in `SlurpResult.hs`, the code that handles the result of saving a scratch file when working in UCM. Some changes are also in `HandleInput` where we call into `SlurpResult`.

### Controversial and interesting decisions

We made the decision that the user only really wants to get notified about aliases in the current namespace (and subnamespaces). If `foo` exists somewhere as `.my.cool.branch.2020.nov4.foo` or whatever, we thought it would just be unnecessary noise to tell you about it if you're currently in some totally unrelated location in the codebase.

### Loose ends

Tangentially related issues created in the course of this PR: #1275, #1273, #1265, #1264

- [x] Filter out aliases that are for absolute names outside the current namespace.
- [x] Do add all the (local) aliases for a name inside the current namespace. Currently only shows one of the names.
- [x] Don't count duplicates (defs already in codebase) as updates.
- [x] Don't show both `x` and `currentNamespace.x` in names.
- [x] Only show the first n aliases, where n is some small number.
- [x] Close #990? — let's try reproducing the issue in master and then checking against the PR.
- [x] Close #1004
- [x] Close #1253 
- [x] Close #1267

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -27,7 +27,7 @@ import Unison.Codebase.Editor.Input
 import Unison.Codebase.Editor.Output
 import Unison.Codebase.Editor.DisplayThing
 import qualified Unison.Codebase.Editor.Output as Output
-import Unison.Codebase.Editor.SlurpResult (SlurpResult(..))
+import Unison.Codebase.Editor.SlurpResult (SlurpResult(..), NameExists(..))
 import qualified Unison.Codebase.Editor.SlurpResult as Slurp
 import Unison.Codebase.Editor.SlurpComponent (SlurpComponent(..))
 import qualified Unison.Codebase.Editor.SlurpComponent as SC
@@ -755,7 +755,7 @@ loop = do
         where
         -- a list of missing sources (if any) and the actions that do the work
         go :: ([Path.HQSplit], [(Path, Branch0 m -> Branch0 m)])
-           -> Path.HQSplit 
+           -> Path.HQSplit
            -> ([Path.HQSplit], [(Path, Branch0 m -> Branch0 m)])
         go (missingSrcs, actions) hqsrc =
           let
@@ -2115,9 +2115,9 @@ toSlurpResult currentPath uf existingNames =
       ]
 
   -- alias (n, r) if (n' /= n, r) exists in names0
-  termAliases :: Map v (Set Name)
+  termAliases :: Map v (NameExists, Set Name)
   termAliases = Map.fromList
-    [ (var n, aliases)
+    [ (var n, (isExistingName, aliases))
     | (n, r@Referent.Ref{}) <- R.toList $ Names.terms fileNames0
     , aliases               <-
       [ Set.map (Path.unprefixName currentPath) . Set.delete n $ R.lookupRan
@@ -2126,12 +2126,14 @@ toSlurpResult currentPath uf existingNames =
       ]
     , not (null aliases)
     , let v = var n
+    , let isExistingName =
+            NameExists . R.memberDom n $ Names.terms existingNames
     , Set.notMember v (SC.terms dups)
     ]
 
-  typeAliases :: Map v (Set Name)
+  typeAliases :: Map v (NameExists, Set Name)
   typeAliases = Map.fromList
-    [ (v, aliases)
+    [ (v, (isExistingName, aliases))
     | (n, r) <- R.toList $ Names.types fileNames0
     , aliases <-
       [ Set.map (Path.unprefixName currentPath) . Set.delete n $ R.lookupRan
@@ -2140,6 +2142,8 @@ toSlurpResult currentPath uf existingNames =
       ]
     , not (null aliases)
     , let v = var n
+    , let isExistingName =
+            NameExists . R.memberDom n $ Names.types existingNames
     , Set.notMember v (SC.types dups)
     ]
 
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -27,7 +27,7 @@ import Unison.Codebase.Editor.Input
 import Unison.Codebase.Editor.Output
 import Unison.Codebase.Editor.DisplayThing
 import qualified Unison.Codebase.Editor.Output as Output
-import Unison.Codebase.Editor.SlurpResult (SlurpResult(..), NameExists(..))
+import Unison.Codebase.Editor.SlurpResult (SlurpResult(..))
 import qualified Unison.Codebase.Editor.SlurpResult as Slurp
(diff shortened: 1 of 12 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2103,21 +2103,19 @@ toSlurpResult currentPath uf existingNames =
...
+      | (n, Referent.Ref{}) <- R.toList (Names.terms fileNames0)
+      , [Referent.Ref{}]     <- [toList $ Names.termsNamed existingNames n]
       ]
     types =
       [ var n
-      | (n, r') <- R.toList (Names.types fileNames0)
-      , [r]     <- [toList $ Names.typesNamed existingNames n]
-      , r' /= r
+      | (n, _) <- R.toList (Names.types fileNames0)
+      , [_]    <- [toList $ Names.typesNamed existingNames n]
```
aryairani: I think we want to revert this change.  if `(n, r') == (n, r)`, then it's an no-op; it belongs in `dups` (and I assume it's already there).

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2103,21 +2103,19 @@ toSlurpResult currentPath uf existingNames =
...
       ]
 
   -- alias (n, r) if (n' /= n, r) exists in names0
-  termAliases :: Map v (Set Name)
+  termAliases :: Map v (NameExists, Set Name)
   termAliases = Map.fromList
-    [ (var n, aliases)
+    [ (var n, (isExistingName, aliases))
```
aryairani: Would `IsUpdate` be clearer than `NameExists`?

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2148,14 +2150,10 @@ toSlurpResult currentPath uf existingNames =
...
-      go (n, r@Referent.Ref{}) = (not . R.memberDom n) existingNames
-                              && (not . R.memberRan r) existingNames
+      go (n, Referent.Ref{}) = (not . R.memberDom n) existingNames
       go _ = False
     addTypes existingNames = R.filter go where
-      go (n, r) = (not . R.memberDom n) existingNames
-               && (not . R.memberRan r) existingNames
-
-
+      go (n, _) = (not . R.memberDom n) existingNames
```
aryairani: This change is the main one that stops excluding new aliases from being added.

parser-typechecker/src/Unison/Codebase/Editor/SlurpResult.hs
```diff
@@ -154,116 +165,158 @@ pretty
...
+        (P.green
+        $ "These new definitions will replace existing ones of the same name and "
+        <> "are ok to `update`:"
+        )
+        (updates sr)
+      , notOks
+        (P.red "These definitions failed:")
+        (P.wrap $ P.red "These definitions would fail on `add` or `update`:")
+        sr
+      ]
```
aryairani: This chunk ended up being massive due to brittnification. 😓 
I think we only changed `okType`, `okTerm`, and `aliases`, but it is hard to know for sure.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: Reminder to fix up the comment on `adds`, which no longer corresponds to the implementation.

parser-typechecker/src/Unison/Codebase/Editor/SlurpResult.hs
aryairani: ☝️ Github truncated the chunk displayed here to 100 lines, so you don't really get the full effect.

parser-typechecker/src/Unison/Codebase/Editor/SlurpResult.hs
```diff
@@ -154,116 +166,167 @@ pretty
...
+        aliases = case Map.lookup v (termAlias sr) of
+          Nothing -> []
+          Just (_, splitAt 5 . toList -> (shown, rest)) ->
+            [ ( P.indentN 2
+                $  P.hiBlack "(also named "
+                <> P.oxfordCommas
+                     ((P.shown <$> shown) ++ map
+                       (const (P.shown (length rest) <> " more"))
+                       (take 1 rest)
+                     )
```
aryairani: Clever way of writing
```haskell
(map P.shown shown) ++
  (if length rest > 0 then [P.shown (length rest) <> " more"] else mempty)
```
or
```haskell
-- import Unison.Util.Monoid (whenM)
(P.shown <$> shown) ++
  (whenM (length rest > 0) [P.shown (length rest) <> " more"])
```

PR #1255, opened 2020-02-19
reproduces the new-lines-after update bug

pete-ts:


pete-ts:
An `update` on an entity, when unit tests exist that use the entity produces a number of new lines after the heavy check mark. The number of new lines added increases as the number of unit tests affected by the update increases.

In interactive mode, this manifests as new lines in the terminal.

PR #1263, opened 2020-02-20
Fix #1250

### LINKED ISSUE #1250: Issue with 3-way merge
See https://github.com/unisonweb/base2/issues/3 - the merged result appears to be computed incorrectly. We originally thought it was something with `pr.load` but if you do a pull manually and do the merge, you get the same weird results. So it appears to be a problem with the merge function itself.

This could use a more minimal failing example, but the issue linked above does show the behavior.

/cc @runarorama

pchiusano:
Fix for #1250 

There were two related issues:

* The `BranchDiff` type didn't have anything to say about additions or removals of children, which meant that for non-fast-forward namespace merges, adds or removals of _subnamespaces_ were just straight up dropped!
* After fixing that, we noticed a second bug in the `Causal.lca` function. Has to do with treatment when one side or the other runs out of history - you still need to keep searching the other side at that point rather than just saying "there's no LCA".

These are both fixed, and we have [a transcript with docs](https://github.com/unisonweb/unison/blob/9b5c7eb9a3142be36a4d99e1d8066f4ce93a3a55/unison-src/transcripts/merge.output.md) that describes the new algorithm and exercises all the relevant cases.

I'm not sure if I understand why a couple other transcript outputs changed, but that is worth understanding before merging this.


parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -341,8 +341,26 @@ deepEdits' b = go id b where
 
 merge :: forall m . Monad m => Branch m -> Branch m -> m (Branch m)
 merge (Branch x) (Branch y) =
-  Branch <$> Causal.threeWayMerge merge0 diff0 apply x y
+  Branch <$> Causal.threeWayMerge combine x y
  where
+  combine :: Maybe (Branch0 m) -> Branch0 m -> Branch0 m -> m (Branch0 m)
```
pchiusano: The new `threeWayMerge` is parameterized on a function, `combine`, which takes the LCA if it exists, and the two descendants, and produces the merged result.

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -135,69 +135,49 @@ lca a b =
...
+          -- arguments, and the order shouldn't really matter.
+          else go seenRight (Set.insert (currentHash left) seenLeft) remainingRight (as <> children left)
+  search seen remaining =
+    case Seq.viewl remaining of
+      Seq.EmptyL -> pure Nothing
+      a :< as -> do
+        current <- a
+        if Set.member (currentHash current) seen
+          then pure $ Just current
+          else search seen (as <> children current)
```
pchiusano: This is worth a close review! /cc @runarorama

parser-typechecker/tests/Unison/Test/Codebase/Causal.hs
```diff
@@ -146,19 +146,33 @@ oneCausal = Causal.one (Set.fromList [1])
 -- generateRandomCausals :: Causal Identity Hash (Set Int64)
 -- generateRandomCausals = undefined
 
+easyCombine
+  :: (Monad m, Semigroup d)
+  => (e -> e -> m e)
+  -> (e -> e -> m d)
+  -> (e -> d -> m e)
+  -> (Maybe e -> e -> e -> m e)
+easyCombine comb _    _    Nothing l r = comb l r
```
pchiusano: This shows that the new signature of `threeWayMerge` is strictly more powerful than the old - given a `combine`, `diff` (with associated semigroup), and `apply` function, you can construct the argument to `Causal.threeWayMerge`. We considered putting this in `Causal`, but it wouldn't be used anywhere in the codebase, so left it here for now.

parser-typechecker/src/Unison/Codebase/Causal.hs
aryairani: Pro tip: There are `Cons` and `Snoc` instances for `Seq`, meaning you can destructure using `Control.Lens.((<|), (|>))`

aryairani: Just curious, is the `viewl` / `EmptyL` stuff for performance reasons, or because it's dicey to find the destructuring patterns in Data.Sequence?  (They are `Empty`, `:|>` and `:<|`; can also use `<|` and `|>` from `Control.Lens`.

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -135,69 +135,49 @@ lca a b =
   go Set.empty Set.empty (Seq.singleton $ pure a) . Seq.singleton $ pure b
```
aryairani: Not part of the PR, but I'd say the use of `.` here is probably counterproductive.

pchiusano:
Note to self - 

The hash becomes different after the initial builtins.merge. Confirmed that the namespace looks the same after this operation, so there’s something not visible that’s affecting the hash.

I suspect different hashes might be due to the new merge algo leaving (or eliminating) empty subnamespaces, in a way that is different than the old impl. I think we want to just uniformly ensure no empty subnamespaces.

parser-typechecker/src/Unison/Codebase/Causal.hs
pchiusano: I think we just left the existing pattern matching structure that was there. Agree it’d be a little nicer using the special patterns. Don’t think performance is an issue here.

dolio:
Okay, I think I finally figured out why the hashes are different.

Note that the old `threeWayMerge` called `mergeWithM` in the case of no lca. `mergeWithM` computes the hash of the merged `Causal` using just the hashes of the predecessor `Causal`s, while the new code includes the resulting `Branch0` in the hash.

This causes the hash of merging with the builtins to be different, and everything downstream of that ends up different because of this.

The hashing could be changed to match `mergeWithM`, but was that what you actually wanted?

pchiusano:
@dolio - nice sleuthing!

New behavior is correct IMO. Old behavior seems wrong... might have been okay (but still dubious) if you were willing to assume that the merge algorithm would never change.

dolio:
I think either behavior is technically okay from a hashing perspective. The information in the merged branch has already influenced the hash of the children, which will determine the hash of the merge. However, it's kind of weird to do it in two different ways depending on whether there's a common ancestor. So, if you want to just keep the current behavior, that seems fine to me.

I have one additional change to get in before this merges, though. I do think you're correct that we don't delete the empty namespace from the map in some situations. That wasn't the cause of the hashing discrepancy, but I think it's still not what is desired.

aryairani:
We should definitely do it in the same way regardless of whether there's a common ancestor.  

Also, I think we shouldn't expect that the `Branch0` contents of the merge node is a deterministic function of the parents unless we have some documentation / proof / enforcement of that invariant.  In particular, I am fuzzy on how patch-propagation-on-merge factors into the idea that merges are associative and commutative, which I think that proof depends on.

So that means I'm thinking we should incorporate the `Branch0` into even the hash of a merge node (this is how git works too, since you can resolve merge conflicts in various ways).

dolio:
Okay, makes sense to me. Right now I think Unison isn't in that situation, because the merge node is _just_ the result of mechanically merging the branches. There aren't any additional fix-up changes in the branch (that would have to go in another `Cons` on top of the merge). But it's safer to include the branch contents, even if it's redundant.

No changes are necessary to follow this policy, I think, so this can be merged at this point from my perspective.

unison-src/transcripts/merge.md
```diff
@@ -0,0 +1,99 @@
...
+.P0> view foo.w
+```
+
+```ucm:error
+.P0> view baz.x
+```
+
+```ucm:error
+.P0> view quux.x
+```
```
aryairani: What are we trying to show with these?

dolio: These show that the things deleted in a branch are still deleted in the merge.

dolio:
Property tests for LCA might be a good idea, since it was broken. I can try to write some of those if desired.

I don't know how complicated it would be to write good property tests for the full merge algorithm. The transcript was trying to test the cases of that.

unison-src/transcripts/merge.md
runarorama: Would be good to put that intent in the transcript.

aryairani: I was experimenting with something like this, WDYT?

    To summarize:
    ```ucm
    .> diff.namespace P0 P1
    .> diff namespace P0 P2
    ```

    Then let's look at the merged diff:
    ```ucm:hide
    .> fork P1 P3
    .> merge P2 P3
    ```
    ```ucm
    .> diff.namespace P0 P3
    ```

w/ specifics for what we should be looking for, WDYT?

dolio:
I added some property tests. However, I'm not certain how your testing library works. Will what I added run multiple tests? Or just a single test? Since it's random, it's probably better to run more than one per session.

unison-src/transcripts/merge.md
dolio: I added the intent to the transcript, but if you want to change the approach, I guess that's fine, too. I think Paul wanted to encode the specific expected behavior via commands that will succeed in those cases (thus the `ucm:error` parts). But since the CI checks for changes in transcripts, I guess any change in behavior would be caught and have to be looked at anyway. I don't have a strong preference for one or the other.

aryairani: Ah, makes sense about succeed / fail.  Good point about CI checks for changes.  I don't have a strong preference at the moment either.

aryairani:
> I added some property tests. However, I'm not certain how your testing library works. Will what I added run multiple tests? Or just a single test? Since it's random, it's probably better to run more than one per session.

I'm not sure either. @pchiusano ?

pchiusano:
@dolio cool, LGTM. I applied ready-to-merge label, it will merge automatically once CI completes successfully.

PR #1279, opened 2020-02-27
Fix #1276 hang on push

### LINKED ISSUE #1276: Memory leak / hang during `push`
To reproduce:

```
.> pull https://github.com/unisonweb/base2 origin.base2
.origin.base2> push git@github.com:pchiusano/zonk.git
```

Where `git@github.com:pchiusano/zonk.git` is any empty remote git repo. For me and also @dolio this never completes and memory usage grows seemingly indefinitely (I saw it grow from 250MB to like 6GB, @dolio let it run until it reached 32 GB).

runarorama:
Fixes #1276 in two ways:

1. Eliminates hangs when there are loops in the histories by tracking in memory which hashes it has queued for writing, rather than relying on the files having been written to disk.

2. Fixes a bug that caused loops in histories.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Causal.hs
+++ b/parser-typechecker/src/Unison/Codebase/Causal.hs
@@ -71,6 +71,26 @@ data Causal m h e
           , tails :: Map (RawHash h) (m (Causal m h e))
           }
 
+toGraph
+  :: Monad m
+  => Set (RawHash h)
+  -> Causal m h e
+  -> m (Seq (RawHash h, RawHash h))
+toGraph seen c = case c of
+  One _ _           -> pure Seq.empty
+  Cons h1 _ (h2, m) -> if Set.notMember h1 seen
+    then do
+      tail <- m
+      g    <- toGraph (Set.insert h1 seen) tail
+      pure $ (h1, h2) Seq.<| g
+    else pure Seq.empty
+  Merge h _ ts -> if Set.notMember h seen
+    then do
+      tails <- sequence $ Map.elems ts
+      gs    <- Seq.fromList <$> traverse (toGraph (Set.insert h seen)) tails
+      pure $ Seq.fromList ((h, ) <$> Set.toList (Map.keysSet ts)) <> join gs
+    else pure Seq.empty
+
 -- A serializer `Causal m h e`. Nonrecursive -- only responsible for
 -- writing a single node of the causal structure.
 data Raw h e
@@ -111,7 +131,7 @@ sync
   -> Causal m h e
   -> StateT (Set (RawHash h)) m ()
 sync exists serialize c = do
-  b <- lift $ exists (currentHash c)
+  b <- lift . exists $ currentHash c
   unless b $ go c
  where
   go :: Causal m h e -> StateT (Set (RawHash h)) m ()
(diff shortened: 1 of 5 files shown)


runarorama:
It's a little faster with the latest change, but there's still quite a lag as it's writing out 3000 files or so.

runarorama:
I'm not sure I understand why there's a loop in a `Merge` node in the `Causal` in the first place. How can that happen?

aryairani:
@runarorama It isn't supposed to happen; I opened #1280 to track.

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -132,27 +164,38 @@ instance Hashable (RawHash h) where
...
-      a :< as -> do
-        current <- a
-        if Set.member (currentHash current) seen
-          then pure $ Just current
-          else search seen (as <> children current)
+          else go seenRight
+                  (Set.insert (currentHash left) seenLeft)
+                  remainingRight
+                  (as <> children left)
+  search seen remaining = case Seq.viewl remaining of
```
aryairani: Is this just formatting changes?

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -170,17 +213,17 @@ threeWayMerge combine = mergeInternal merge0
...
+          n           <- new
+          mayAncestor <- lca' acc (Seq.singleton (pure n))
+          newHead     <- combine (head <$> mayAncestor) e (head n)
+          pure (newHead, pure n Seq.<| acc)
+    in  case Map.elems m of
+          []       -> error "Causal.threeWayMerge empty map"
+          me : mes -> do
+            e            <- me
+            (newHead, _) <- foldM k (head e, Seq.singleton (pure e)) mes
+            pure $ Merge (RawHash (hash (newHead, Map.keys m))) newHead m
```
aryairani: For readability (since I'm reading top to bottom, but the context for understanding `k` isn't there yet), would you move `k` below, into a `where clause`, and/or put a type signature on it?

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -132,27 +164,38 @@ instance Hashable (RawHash h) where
...
+
+-- Find the lowest common ancestor of two causal-forests.
+-- The convention is that the causals in a forest are the ancestors of some
+-- other Causal.
+lca'
+  :: Monad m
+  => Seq (m (Causal m h e))
+  -> Seq (m (Causal m h e))
+  -> m (Maybe (Causal m h e))
+lca' = go Set.empty Set.empty where
```
aryairani: What's the importance of the causal forest? Would you explain here or in the PR comments

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -69,6 +71,26 @@ data Causal m h e
...
+      g    <- toGraph (Set.insert h1 seen) tail
+      pure $ (h1, h2) Seq.<| g
+    else pure Seq.empty
+  Merge h _ ts -> if Set.notMember h seen
+    then do
+      tails <- sequence $ Map.elems ts
+      gs    <- Seq.fromList <$> traverse (toGraph (Set.insert h seen)) tails
+      pure $ Seq.fromList ((h, ) <$> Set.toList (Map.keysSet ts)) <> join gs
+    else pure Seq.empty
+
```
aryairani: Is this still used anywhere, and if not, should we use it or delete it or neither?

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -132,27 +164,38 @@ instance Hashable (RawHash h) where
 -- Find the lowest common ancestor of two causals.
 lca :: Monad m => Causal m h e -> Causal m h e -> m (Maybe (Causal m h e))
 lca a b =
-  go Set.empty Set.empty (Seq.singleton $ pure a) . Seq.singleton $ pure b
- where
+  lca' (Seq.singleton $ pure a) . Seq.singleton $ pure b
```
aryairani: I know this has been in here a while now, but as long as we're modifying it, could we flip it around to make the arguments equal?
```haskell
lca' (Seq.singleton $ pure a) (Seq.singleton $ pure b)
```

aryairani:
Does anyone have a sense of whether this bug means there are broken histories out in the wild, and if so, can we do anything to repair / ignore them?  (Maybe this patch already successfully ignores them?)

parser-typechecker/src/Unison/Codebase/Causal.hs
runarorama: This is a handy function for debugging Causals. What's the best thing to do, comment it out?

aryairani: No; maybe just a note that says it's currently kept around in case of debugging Causals — or you could wire it up to a ucm command, like `debug.history`, which I guess is similar but different.

runarorama:
There are probably broken histories out in the wild. This fix does successfully ignore them, but I don't think we can repair them.

parser-typechecker/src/Unison/Codebase/Causal.hs
runarorama: Yes

aryairani: Hah, ok. 👍

unison-src/transcripts/mergeloop.md
```diff
@@ -0,0 +1,51 @@
+# Merge loop test
+
+This tests for regressions of https://github.com/unisonweb/unison/issues/1276 where trivial merges cause loops in the history.
+
```
aryairani: A trivial merge is one where the head values are the same?

runarorama: Yes.

PR #1294, opened 2020-03-02
implements the `back` command using a stack of currentNamespace

### LINKED ISSUE #635: `cd ..` doesn't work
I've been enjoying the `cd` alias for `namespace`, moving around my namespaces like a boss.

But then I try and go back up the tree using `cd ..`, and I see

```
.temp> cd ..

  InvalidWordyId ""
```
I'd prefer it if this worked, since it's going to be the first thing I type every time regardless...  I don't want to have to retrain any neurons if I can help it :-)

aryairani:
## Overview

Implements the `back` command shown below:

image

and closes #635 unless we remove this line before merging.

## Implementation notes

See self-review below.

## Test coverage

See `cd-back.output.md`.

## Note re. transcripts, `cd`, and `back`

Using `back` in your transcript can make it sensitive to mistakes.  It's important to remember that the transcript runner injects `cd` commands into its execution when your virtual markdown prompt doesn't match the transcript runner's internal state. This only matters if you're trying to demonstrate something specific with `back`.  😅

See if you can spot the issue with this transcript, which I expected to result in an error message about using `back` too many times:

    ```ucm:error
    .> cd arya.base
    .arya.base> cd arya.myproject
    .arya.myproject> back
    .arya.base> back
    .> back
    ```

Because I mistakenly used `cd arya.myproject` instead `cd .arya.myproject`, the markdown and internal states are out of sync.  The TR injects extra `cd` commands to sync up again, which makes the path stack not-what-I-expected, and `back` to behave unexpectedly:

```
.> cd arya.base
.arya.base> cd arya.myproject
.arya.base.arya.myproject> cd .arya.myproject -- TR correction
.arya.myproject> back
.arya.base.arya.myproject> cd .arya.base -- TR correction
.arya.base> back
.arya.base.arya.myproject> cd . -- TR correction
.> back
.arya.base.arya.myproject>
```
and no error. 🙁 

Anyway, be careful :-)

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -169,10 +169,10 @@ type InputDescription = Text
 makeLenses ''LoopState
 
 -- replacing the old read/write scalar lens with peek/push lens
-peekPushNel :: Lens (NonEmpty Path.Absolute) (NonEmpty Path.Absolute) Path.Absolute Path.Absolute
-peekPushNel = lens Nel.head (flip Nel.cons)
 currentPath :: Lens (LoopState m v) (LoopState m v) Path.Absolute Path.Absolute
-currentPath = currentPathStack . peekPushNel
+currentPath = currentPathStack . peekPushNel where
+  peekPushNel :: Lens (NonEmpty a) (NonEmpty a) a a
+  peekPushNel = lens Nel.head (flip Nel.cons)
 
 loopState0 :: Branch m -> Path.Absolute -> LoopState m v
 loopState0 b p = LoopState b (pure p) Nothing Nothing Nothing []
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -16,7 +16,7 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE EmptyCase #-}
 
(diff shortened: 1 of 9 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -166,8 +168,12 @@ type InputDescription = Text
 
 makeLenses ''LoopState
 
+-- replacing the old read/write scalar Lens with "peek" Getter for the NonEmpty
+currentPath :: Getter (LoopState m v) Path.Absolute
+currentPath = currentPathStack . to Nel.head
```
aryairani: In place of the old read/write `currentPath` lens, I added a getter that peeks at the head of `currentPathStack`, so the usage sites don't have to change.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -678,10 +685,14 @@ loop = do
...
       SwitchBranchI path' -> do
         let path = resolveToAbsolute path'
-        currentPath .= path
+        currentPathStack %= Nel.cons path
         branch' <- getAt path
         when (Branch.isEmpty branch') (respond $ CreatedNewBranch path)
 
+      PopBranchI -> use (currentPathStack . to Nel.uncons) >>= \case
+        (_, Nothing) -> respond StartOfCurrentPathHistory
+        (_, Just t) -> currentPathStack .= t
```
aryairani: push and pop the stack

pchiusano:
@aryairani ff to merge after adding that transcript

PR #1307, opened 2020-03-06
Remove term names from evaluation in docs

### LINKED ISSUE #1305: Evaluation in Docs should not show the name of the evaluated definition
Currently, given `foo = 2 + 2` the Doc `[: @[evaluate] foo :]` renders as `foo = 4`. We should change this so it just results in `4`. The old behaviour can be accomplished with `[: foo = @[evaluate] foo :]` instead.

This change should be made in `DisplayValues.hs`.

runarorama:
Fixes #1305

### DIFF
--- a/unison-src/transcripts/docs.output.md
+++ b/unison-src/transcripts/docs.output.md
@@ -97,12 +97,12 @@ docs.List.take = [:
 
 @[source] List.take.ex1
 🔽
-@[evaluate] List.take.ex1
+@List.take.ex1 = @[evaluate] List.take.ex1
 
 
 @[source] List.take.ex2
 🔽
-@[evaluate] List.take.ex2
+@List.take.ex2 = @[evaluate] List.take.ex2
 :]
 ```
 
@@ -158,13 +158,11 @@ Now that documentation is linked to the definition. We can view it if we like:
   ## Examples:
   
   List.take.ex1 = builtin.List.take 0 [1, 2, 3, 4, 5]
-  🔽
-  List.take.ex1 = []
+  🔽 ex1 = []
   
   
   List.take.ex2 = builtin.List.take 2 [1, 2, 3, 4, 5]
-  🔽
-  List.take.ex2 = [1, 2]
+  🔽 ex2 = [1, 2]
   
 
 ```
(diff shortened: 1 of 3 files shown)


unison-src/transcripts/docs.output.md
```diff
@@ -158,13 +158,11 @@ Now that documentation is linked to the definition. We can view it if we like:
...
   List.take.ex1 = builtin.List.take 0 [1, 2, 3, 4, 5]
-  🔽
-  List.take.ex1 = []
+  🔽 ex1 = []
   
   
   List.take.ex2 = builtin.List.take 2 [1, 2, 3, 4, 5]
-  🔽
-  List.take.ex2 = [1, 2]
+  🔽 ex2 = [1, 2]
```
pchiusano: @runarorama weird what happened to the line breaks here? Is that related to your change? Also unfortunate that the link expands to the suffixed name, not the fully qualified name, I think it makes this idiom of `@blah = @[evaluate] blah` less useful. Maybe just:

```
@[source] blah
⧩
@[evaluate] blah
```

Is fine, but it should preserve those line breaks.

runarorama: It has always been the case that

```
doc = [:
  foo
  bar 
  baz
:]
```

displays as

```
foo bar baz
```

Is that incorrect?

pchiusano: @runarorama can you explain why the output has changed? Previously it used to preserve the line break after the 🔽.

I'm not sure it's correct... I think it depends what we want. Current behavior is like markdown where single line breaks are turned into spaces. We'd need an explicit way to insert single line breaks if we keep this behavior.

@atacratic Would it work to just preserve all line breaks? The main thing we wanted to be able to do is dynamically rewrap the document on display to fit the terminal width.

runarorama: It didn't preserve a line break after the 🔽, but inserted one before the `evaluate` output.

pchiusano: K. Let's merge this then. Can rethink the doc display / rewrapping separately.

PR #1325, opened 2020-03-08
Fix/1320

### LINKED ISSUE #1320: unfortunately-punctuated message when creating new codebase in `.`
```
$ ucm -codebase . init
  ⚠️
  
  Initializing a new codebase in: ..
```
Recommend: 
- removing the period
- maybe removing the colon (bikeshed away!)
- removing the ⚠️ and extra lines
- maybe removing the indent

pete-ts:
## Overview

This change aims to fix #1320 

When interacting with the `-codebase` flag a user might select `.` or `..` as a target directory both for an existing code base and for initialising a codebase.

## Implementation notes

* Removes warning triangle and extra lines.
* Removes the punctuation
* Removes indentation
* Translates `.` and `..`

* I had previously added `Pretty.endSentence` to achieve punctuation after a path. No longer needed, removed.
* `PrettyTerminal.putPrettyLn'` is only used for these messages in `FileCodebase`, removing the "indentN 2" to prevent indentation.
* When `.` or `..` is used as the target directory only the error message translates it to the full path. Since the user requested `-codebase .` the  suggestion after the error is to `run -codebase . init`  

## Test coverage + Loose Ends

None of this is covered by automation as there are no tests that currently track changes to `ucm` output.

This can be achieved by :
* adding a new project (`stack exec output-tests`, along the lines of `stack exec transcripts`)
* creating a test structure parallel to the transcript tester using `easytest` and tests consisting of shell commands  to keep track of changes.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
@@ -51,12 +51,11 @@ import           UnliftIO.Directory             ( createDirectoryIfMissing
 import           System.FilePath                ( FilePath
                                                 , takeBaseName
                                                 , takeFileName
-                                                , takeDirectory
                                                 , (</>)
                                                 )
 import           System.Directory               ( copyFile
                                                 , getHomeDirectory
-                                                , makeAbsolute
+                                                , canonicalizePath
                                                 )
 import           System.Path                    ( replaceRoot
                                                 , createDir
@@ -128,10 +127,8 @@ initCodebaseAndExit mdir = do
   exitSuccess
 
 absDir :: FilePath -> IO FilePath
-absDir dir 
-  | dir == "." = makeAbsolute dir
-  | dir == ".." = takeDirectory <$> makeAbsolute "."
-  | otherwise = pure dir
+absDir dir =
+  if dir == "." || dir == ".." then canonicalizePath dir else pure dir
 
 initCodebase :: FilePath -> IO (Codebase IO Symbol Ann)
 initCodebase dir = do
--- a/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
@@ -51,9 +51,13 @@ import           UnliftIO.Directory             ( createDirectoryIfMissing
 import           System.FilePath                ( FilePath
                                                 , takeBaseName
                                                 , takeFileName
+                                                , takeDirectory
                                                 , (</>)
                                                 )
-import           System.Directory               ( copyFile, getHomeDirectory )
+import           System.Directory               ( copyFile
+                                                , getHomeDirectory
+                                                , makeAbsolute
+                                                )
 import           System.Path                    ( replaceRoot
                                                 , createDir
                                                 , subDirs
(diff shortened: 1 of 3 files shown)


pete-ts:
Output of
*  `-codebase ./../not dots` (codebase does not exist, error)
* `-codebase ./../not dots init`(works)
* `-codebase ./../not dits init` (issued again, error)

```
$ ./ucm -codebase .
No codebase exists in /home/x/temp/x/y/z
Run `ucm -codebase . init` to create one, then try again!
$ ./ucm -codebase ..
No codebase exists in /home/x/temp/x/y
Run `ucm -codebase .. init` to create one, then try again!
$ ./ucm -codebase zz
No codebase exists in zz
Run `ucm -codebase zz init` to create one, then try again!
$ ./ucm -codebase . init
Initializing a new codebase in: /home/x/temp/x/y/z
$ ./ucm -codebase . init
It looks like there's already a codebase in: /home/x/temp/x/y/z
$ ./ucm -codebase .. init
Initializing a new codebase in: /home/x/temp/x/y
$ ./ucm -codebase .. init
It looks like there's already a codebase in: /home/x/temp/x/y
$ ./ucm -codebase zz init
Initializing a new codebase in: zz
$ ./ucm -codebase zz init
It looks like there's already a codebase in: zz
```

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -123,20 +127,26 @@ initCodebaseAndExit mdir = do
   _ <- initCodebase dir
   exitSuccess
 
+absDir :: FilePath -> IO FilePath
+absDir dir 
+  | dir == "." = makeAbsolute dir
+  | dir == ".." = takeDirectory <$> makeAbsolute "."
```
aryairani: WDYT about using `canonicalizePath` instead of `takeDirectory` if that’s what you want to do; in fact I guess `absDir` is really just trying to implement `canonicalizePath`, right?

aryairani:
I kind of like the idea of output-tests for documenting and tracking other output changes; though if it can be done just as elegantly with a shell script than with another haskell executable that adds to the build time, I’d personally rather have the shell script.

aryairani:
Btw thanks for this PR!

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
pete-ts: Good catch; I am not as familiar as I would like to be with Haskell libraries, so in this instance I read the bold note and  shied away from using `canonicalizePath`. (...and ended up reimplementing it naively of course.) 

Will update this :+1: .

aryairani: It’s okay, we’re all learning.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -128,10 +127,8 @@ initCodebaseAndExit mdir = do
   exitSuccess
 
 absDir :: FilePath -> IO FilePath
-absDir dir 
-  | dir == "." = makeAbsolute dir
-  | dir == ".." = takeDirectory <$> makeAbsolute "."
-  | otherwise = pure dir
+absDir dir =
+  if dir == "." || dir == ".." then canonicalizePath dir else pure dir
```
aryairani: Oh, I meant... `absDir` could just be replaced with `canonicalizaPath`, right?

pete-ts: The behaviour is slightly different (for the better) in that now when the user supplies a path like "blah" and gets an error, the message shows the full path, which is the behaviour for `.`/`..`. 
So uniform and matching errors for all instances.

aryairani: Sorry, I didn't quite understand.  Is there something special that you wanted to do when displaying a path that _ends_ in `.` or `..`, but don't want to do if there's a `/./` or `/../` in the middle?

Maybe a few examples of what you had in mind?

pete-ts: I updated the examples of usage in my previous comment, but omitted to mention it explicitly. (https://github.com/unisonweb/unison/pull/1325#issuecomment-596184930)

When a user supplies a path and as long as `canonicalyzePath` can translate it, the error produced ("run init, cannot init exists") will contain the translated path exactly the way unison/haskell/os see it internally.

I have attached examples showing how the errors and output work, I am not trying to do permutation testing with `/.`/ or `/../` as ultimately that would be testing the implementation of `canonicalyzePath`.

aryairani:
Thanks @pete-ts!

PR #1328, opened 2020-03-09
Unabbreviate pr.{create,load}

pchiusano:
This just changes the command name `pr.create` to `pull-request.create` and `pr.load` to `pull-request.load`. The `pr.create` and `pr.load` names are kept as aliases.

Screen Shot 2020-03-08 at 10 04 07 PM


pchiusano:
Also just want to note that this PR does not prevent us from more bikeshedding about whether to call a "pull request" something else (though I'm out of stamina on that), it's just un-abbreviating a command name that might have been confusing.

aryairani:
It would be amazing if there were a PR that _could_ prevent us from bike-shedding!

PR #1331, opened 2020-03-09
correct annotation of match cases to always include the case body

### LINKED ISSUE #1329: match terms are sometimes annotated incorrectly
When match terms are annotated during parsing, they are _sometimes_ annotated incorrectly.

`match x with 1 -> 0` is annotated correctly:

image

`match x with x -> x` is annotated incorrectly:

image

Related to #1297.

anovstrup:
## Overview

Fixes #1329.

## Implementation notes

Explicitly sets the annotation for the case body of each match case, rather than relying on `ABT.absChain'` (which was annotating the case body term with the annotation for a bound variable from the case pattern).

## Test coverage

The change can be tested by adding the following to a scratch file and typechecking with/without the patch:
```
foo : Nat -> Text
foo x = match x with x -> x
```
Before the PR, the `-> x` at the end will not be highlighted in the typechecking output. With the PR, it will be highlighted.

Test coverage could be improved by adding tests in TermParser that explicitly check the annotations. This PR does not do that.

### DIFF
--- a/parser-typechecker/src/Unison/TermParser.hs
+++ b/parser-typechecker/src/Unison/TermParser.hs
@@ -137,9 +137,11 @@ match = do
 matchCase :: Var v => P v (Term.MatchCase Ann (Term v Ann))
 matchCase = do
   (p, boundVars) <- parsePattern
+  let boundVars' = snd <$> boundVars
   guard <- optional $ reserved "|" *> infixAppOrBooleanOp
   t <- block "->"
-  pure . Term.MatchCase p (fmap (ABT.absChain' boundVars) guard) . ABT.annotate (ann t) $ ABT.absChain' boundVars t
+  let absChain vs t = foldr (\v t -> ABT.abs' (ann t) v t) t vs
+  pure . Term.MatchCase p (fmap (absChain boundVars') guard) $ absChain boundVars' t
 
 parsePattern :: forall v. Var v => P v (Pattern Ann, [(Ann, v)])
 parsePattern =
--- a/parser-typechecker/src/Unison/TermParser.hs
+++ b/parser-typechecker/src/Unison/TermParser.hs
@@ -139,7 +139,7 @@ matchCase = do
   (p, boundVars) <- parsePattern
   guard <- optional $ reserved "|" *> infixAppOrBooleanOp
   t <- block "->"
-  pure . Term.MatchCase p (fmap (ABT.absChain' boundVars) guard) $ ABT.absChain' boundVars t
+  pure . Term.MatchCase p (fmap (ABT.absChain' boundVars) guard) . ABT.annotate (ann t) $ ABT.absChain' boundVars t
 
 parsePattern :: forall v. Var v => P v (Pattern Ann, [(Ann, v)])
 parsePattern =


aryairani:
@anovstrup Do you have a sense of whether there might be a "more correct" version of `ABT.absChain'` that would just do the right thing (and end up being the same as `(ann t)`?)

anovstrup:
@aryairani Depends on how it's being used I guess? Something like `annotate (annotation t) $ foldr (\(a,v) t -> abs' a v t) t vs` would work for this case, I think, but not sure that would be right in general. Depends on whether one cares about the annotations for the bound variables—in this case, we don't.

anovstrup:
It _does_ seem wrong that it's completely ignoring the annotation of `t`, though...

anovstrup:
@aryairani maybe
```
absChain' :: (Ord v) => [(a, v)] -> Term f v a -> Term f v a
absChain' vs t = foldr (\(_,v) t -> abs' a v t) t vs
  where a = annotation t
```

This seems more correct for the way it's being used from TermParser, where a case body is being turned into a chain of lambdas -- the annotation of the whole chain (and each lambda in the chain) should include only the case body itself and not the annotations of the bindings from the match pattern. `absChain'` is also used in ANF.hs, though, and I'm not sure what the ramifications would be of changing the annotations there (if any).

pchiusano:
@anovstrup I'm not sure `absChain'` should have any opinion on what the annotations should be, that's why it takes the `[(a,v)]` currently. However why don't we introduce another function:

```Haskell
-- open to better name ideas!
absChain'' :: Ord v => [v] -> Term f v a -> Term f v a
absChain'' vs t = foldr (\v t -> abs' (annotation t) v t) t vs
```

And then just have the term parser use that instead.

anovstrup:
@pchiusano FWIW, `absChain'` is used in exactly three places, and two of them are on the same line in `TermParser.matchCase` (the third is in ANF.hs). Should we just introduce the new function locally there for now?

aryairani:
@anovstrup @pchiusano Could the `a`s being passed to `absChain'` be wrong?

anovstrup:
> @anovstrup @pchiusano Could the `a`s being passed to `absChain'` be wrong?

No, I think they're right: they're the annotations for the variables bound in the match pattern. They just aren't the correct annotations for the "new" variables (of the same names) for the match case body, when it gets turned into a lambda chain.

aryairani:
＠_＠ I guess I'd need a diagram 🥴

anovstrup:
e.g., a case like 
```
x +: xs -> (show x) ++ (show xs)
```
gets turned (notionally) into 
```
MatchCase (Pattern x +: xs) None (x -> xs -> (show x) ++ (show xs)
```

The annotations for the "x" and "xs" variables in the lambda abstractions _should_ be the same as the ones for the ones in the pattern (I _think_?), but the annotations for lambda chain as a whole should be that of the case body (because the `x -> xs ->` part was just conjured up and doesn't appear in the source anywhere).

pchiusano:
@anovstrup sure you can just inline it if it's just used in that one spot

PR #1333, opened 2020-03-10
Fix #1225

### LINKED ISSUE #1225: Too many entries in history for sub-namespaces
E.g.

```unison
myLibrary.f x = x + 9
myLibrary.g x = x + 10
myLibrary.h x = x + 3
```

```ucm
.> add
.> history
.> cd myLibrary
.myLibrary> history
```

The first `history` command (from the root) has one entry for all three functions, but the second command (issued from the `myLibrary` namespace) shows three entries; one for each function.

pchiusano:
This is all set. To be merged after #1337.

## Overview

Fixes #1225, where a separate history entry was being created for each definition being added to a namespace. Now uses a post-order traversal of the branch as suggested by @aryairani, rebuilding from the bottom up and visiting each node exactly once. This avoids creating excessive history nodes. See the changed output for `link.md` transcript.

## Implementation notes

Change was very surgical, to `Branch.stepManyAt0` and `Branch.stepManyAt0M`.

None that I can think of. There was an interesting change to the `delete.output.md`, but I stared at it a while and convinced myself the old output was wrong (annotated below). I don't totally understand why this change has fixed that bug though.

Also important note: this affects hashes for other transcripts just due to there being fewer steps in the history. All the transcripts still behave correctly though.

## Test coverage

Existing transcripts provide pretty good test coverage I think. I'd implemented it incorrectly at first and got lots of failures. :) 

I'll see if I can think of some more specific tests though.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -727,18 +727,27 @@ stepManyAt0 :: forall f m . (Applicative m, Foldable f)
 stepManyAt0 actions b = go (first Path.toList <$> toList actions) b where
   go :: [([NameSegment], Branch0 m -> Branch0 m)] -> Branch0 m -> Branch0 m
   go actions b = let 
+    -- combines the functions that apply to this level of the tree
     currentAction b = foldl' (\b f -> f b) b [ f | ([], f) <- actions ]
 
+    -- groups the actions based on the child they apply to
     childActions :: Map NameSegment [([NameSegment], Branch0 m -> Branch0 m)]
     childActions = List.multimap [ (seg, (rest,f)) | ((seg:rest), f) <- actions ]
 
+    -- alters the children of `b` based on the `childActions` map
     stepChildren :: Map NameSegment (Branch m) -> Map NameSegment (Branch m)
     stepChildren m0 = foldl' g Map.empty (Map.keysSet m0 <> Map.keysSet childActions)
       where
+      -- this skips adding any empty children, which could happen depending
+      -- on the action functions provided (think: `b -> Branch.empty0`)
       insert seg child m = if isEmpty child then m else Map.insert seg child m 
       g m seg = insert seg child m where
         child = case Map.lookup seg childActions of
-          Just actions -> step (go actions) (Map.findWithDefault empty seg m0) 
+          Just actions -> 
+            -- this `findWithDefault` is important, allows the stepManyAt
+            -- to create new children at deeper paths than exist in the
+            -- original Branch
+            step (go actions) (Map.findWithDefault empty seg m0) 
           Nothing -> Map.findWithDefault empty seg m0 
     in currentAction $ over children stepChildren b 
 
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -725,14 +725,14 @@ stepManyAt0 :: forall f m . (Applicative m, Foldable f)
            -> Branch0 m -> Branch0 m
 stepManyAt0 actions b = go (toList actions) b where
   go :: [(Path, Branch0 m -> Branch0 m)] -> Branch0 m -> Branch0 m
-  go actions b = let 
+  go actions b = let
     -- combines the functions that apply to this level of the tree
     currentAction b = foldl' (\b f -> f b) b [ f | (Path.Empty, f) <- actions ]
 
     -- groups the actions based on the child they apply to
     childActions :: Map NameSegment [(Path, Branch0 m -> Branch0 m)]
-    childActions = 
-      List.multimap [ (seg, (rest,f)) | (Path.Cons seg rest, f) <- actions ]
+    childActions =
+      List.multimap [ (seg, (rest,f)) | (seg :< rest, f) <- actions ]
 
     -- alters the children of `b` based on the `childActions` map
     stepChildren :: Map NameSegment (Branch m) -> Map NameSegment (Branch m)
@@ -740,37 +740,37 @@ stepManyAt0 actions b = go (toList actions) b where
       where
       -- this skips adding any empty children, which could happen depending
       -- on the action functions provided (think: `b -> Branch.empty0`)
-      insert seg child m = if isEmpty child then m else Map.insert seg child m 
+      insert seg child m = if isEmpty child then m else Map.insert seg child m
       g m seg = insert seg child m where
         child = case Map.lookup seg childActions of
-          Just actions -> 
+          Just actions ->
             -- this `findWithDefault` is important, allows the stepManyAt
             -- to create new children at deeper paths than exist in the
             -- original Branch
-            step (go actions) (Map.findWithDefault empty seg m0) 
-          Nothing -> Map.findWithDefault empty seg m0 
-    in currentAction $ over children stepChildren b 
+            step (go actions) (Map.findWithDefault empty seg m0)
+          Nothing -> Map.findWithDefault empty seg m0
+    in currentAction $ over children stepChildren b
 
 stepManyAt0M :: forall m n f . (Monad m, Monad n, Foldable f)
              => f (Path, Branch0 m -> n (Branch0 m))
              -> Branch0 m -> n (Branch0 m)
 stepManyAt0M actions b = go (toList actions) b where
   go :: [(Path, Branch0 m -> n (Branch0 m))] -> Branch0 m -> n (Branch0 m)
-  go actions b = let 
+  go actions b = let
     currentAction b = foldM (\b f -> f b) b [ f | (Path.Empty, f) <- actions ]
 
     childActions :: Map NameSegment [(Path, Branch0 m -> n (Branch0 m))]
-    childActions = 
-      List.multimap [ (seg, (rest,f)) | (Path.Cons seg rest, f) <- actions ]
+    childActions =
+      List.multimap [ (seg, (rest,f)) | (seg :< rest, f) <- actions ]
 
     stepChildren :: Map NameSegment (Branch m) -> n (Map NameSegment (Branch m))
     stepChildren m0 = foldM g Map.empty (Map.keysSet m0 <> Map.keysSet childActions)
       where
-      insert seg child m = if isEmpty child then m else Map.insert seg child m 
+      insert seg child m = if isEmpty child then m else Map.insert seg child m
       g m seg = do
         child <- case Map.lookup seg childActions of
-          Just actions -> stepM (go actions) (Map.findWithDefault empty seg m0) 
-          Nothing -> pure $ Map.findWithDefault empty seg m0 
+          Just actions -> stepM (go actions) (Map.findWithDefault empty seg m0)
+          Nothing -> pure $ Map.findWithDefault empty seg m0
         pure $ insert seg child m
     in do
       c2 <- stepChildren (view children b)
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -735,7 +735,8 @@ stepManyAt0 actions b = go (first Path.toList <$> toList actions) b where
     stepChildren :: Map NameSegment (Branch m) -> Map NameSegment (Branch m)
     stepChildren m0 = foldl' g Map.empty (Map.keysSet m0 <> Map.keysSet childActions)
       where
-      g m seg = Map.insert seg child m where
+      insert seg child m = if isEmpty child then m else Map.insert seg child m 
+      g m seg = insert seg child m where
         child = case Map.lookup seg childActions of
           Just actions -> step (go actions) (Map.findWithDefault empty seg m0) 
           Nothing -> Map.findWithDefault empty seg m0 
@@ -755,11 +756,12 @@ stepManyAt0M actions b = go (first Path.toList <$> toList actions) b where
     stepChildren :: Map NameSegment (Branch m) -> n (Map NameSegment (Branch m))
     stepChildren m0 = foldM g Map.empty (Map.keysSet m0 <> Map.keysSet childActions)
       where
+      insert seg child m = if isEmpty child then m else Map.insert seg child m 
       g m seg = do
         child <- case Map.lookup seg childActions of
           Just actions -> stepM (go actions) (Map.findWithDefault empty seg m0) 
           Nothing -> pure $ Map.findWithDefault empty seg m0 
-        pure $ Map.insert seg child m
+        pure $ insert seg child m
     in do
       c2 <- stepChildren (view children b)
       currentAction (set children c2 b)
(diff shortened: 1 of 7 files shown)


unison-src/transcripts/link.output.md
```diff
@@ -196,27 +196,15 @@ myLibrary.h x = x + 3
...
-    + Adds / updates:
-    
-      g
-  
-  ⊙ #seb6uuoocl
-  
-    + Adds / updates:
-    
-      f
+      f g h
```
pchiusano: Here's where we see the new behavior: `f`, `g`, and `h` are all added in one go, rather than 3 separate steps. This then impacts later namespace hashes.

unison-src/transcripts/reflog.output.md
```diff
@@ -65,16 +65,16 @@ y = 2
...
+    `reset-root #d7reeo8lc4`  to reset the root namespace and
                               its history to that of the
                               specified namespace.
   
-  1. #bqd4d6mrnk : add
-  2. #onc5fpvvb7 : add
-  3. #upd8bubg9e : builtins.merge
+  1. #up853v4413 : add
+  2. #d7reeo8lc4 : add
+  3. #i5f8gcfb14 : builtins.merge
```
pchiusano: I'm not seeing an obvious reason why this hash should be different so going to investigate. `builtins.merge` doesn't use `stepManyAt` that I can see.

unison-src/transcripts/delete.output.md
```diff
@@ -182,10 +188,6 @@ type Foo = Foo Boolean
     5. ┌ Foo.Foo#d97e0jhkmd#0 : Nat -> Foo
     6. └ Foo.Foo#gq9inhvg9h#0 : Boolean -> b.Foo
   
-  Added definitions:
-  
-    7. foo : Nat
-
```
pchiusano: This old output was incorrect: `foo` should have been deleted in the `a` namespace, so it shouldn't be resurfacing here.

unison-src/transcripts/reflog.output.md
pchiusano: Okay, it does actually call `stepManyAt` (indirectly via `BranchUtils.addFromNames0`). So mystery solved.

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -719,32 +720,60 @@ modifyAtM path f b = case Path.uncons path of
...
-
-stepManyAt0M :: (Monad m, Monad n, Foldable f)
+stepManyAt0 actions b = go (first Path.toList <$> toList actions) b where
+  go :: [([NameSegment], Branch0 m -> Branch0 m)] -> Branch0 m -> Branch0 m
+  go actions b = let 
+    -- combines the functions that apply to this level of the tree
+    currentAction b = foldl' (\b f -> f b) b [ f | ([], f) <- actions ]
+
+    -- groups the actions based on the child they apply to
+    childActions :: Map NameSegment [([NameSegment], Branch0 m -> Branch0 m)]
```
aryairani: What's the `[NameSegment]`?

unison-src/transcripts/delete.output.md
aryairani: Seems like a merge error.  We had `.a.foo = 1` and `.b.foo = 2`, then we merged `.b` into `.a`, and deleted `.a.foo`.  So we have no `.a.foo` but we still have `.b.foo`.  Then we merged `.b` into `.a` again, and had previously ended up  with `.a.foo = 2` (presumably) again.

But why did it change now?

Since the version of `.b` that contained `foo` was already part of the history of `.a`, it seems like it shouldn't have been added again.  I wouldn't have been surprised if this changed with some `lca` or `merge` thing, but I still can't really imagine the issue and don't get what it has to do with this PR.

pchiusano: I think I get it: old implementation would have a chain of history for each definition added, and moreover the definitions in the chain would be in an arbitrary order (or maybe alphabetical). This then affects LCA calculation and the merge result. I’ll see if I can confirm this hypothesis.

parser-typechecker/src/Unison/Codebase/Branch.hs
pchiusano: It’s just the Path - notice the initial call to go which converts the paths to lists of NameSegment.

aryairani: Oh, I see.  What do you think about leaving it as `Path` for self-documentation purposes?  Maybe the cons (`<|`) and snoc (`|>`) patterns will help?

unison-src/transcripts/delete.output.md
aryairani: Interesting, okay.

pchiusano: Okay, got to the bottom of this and it's not what I thought. The difference in behavior is because the new code leaves around an empty sub-branch which remembers the causal history that led to that sub-branch being empty. The old code was over-eager about deleting sub-branches that became empty. Here's [transcript output, showing the history](https://gist.github.com/pchiusano/9edb8e546149bce49ef1574cdf195577)

The key is that in the `a` namespace in this PR, history after the delete looks like what you'd expect:

```
.a> history

        message.
  
  ⊙ #gcq0ml6diu
  
    - Deletes:
    
      foo#0ja foo#jk1
  
  This segment of history starts with a merge. Use
  `history #som3n4m3space` to view history starting from a given
  namespace hash.
  
  #biv51qeect
  #u8lsnrulgl
  ⑂
  ⊙ #l24km9hmve
```

This reflects that a merge occurred and then `foo` (both versions) were then deleted. Whereas in `master`, this same portion of the transcript looks like:

```
.a> history

  ☝️  The namespace .a is empty.
```

As a result, when doing the subsequent merge of `b` into `a`, `master` has "forgotten" that at an earlier point in `a` history, `foo` was deleted. So it treats the `foo` as a new definition and adds it.

So this branch has the correct behavior, however I was curious what made the old code so eager to delete empty branches. Old code was defined in terms of:

https://github.com/unisonweb/unison/blob/b5cae909fc15639d8f6e86fc86b83141700a2675/parser-typechecker/src/Unison/Codebase/Branch.hs#L689-L696

Which as you can see just deletes the key if you ever go to insert something whose `head` is `Branch.empty0`.

So basically, this PR just happens to work due better in this instance due to a coincidence of the implementation. I'd like to understand better when it is safe to delete empty sub-branches. This seems to be a case where being overeager to delete them leads to worse merge results. I suspect you should just never delete empty sub-branches, unless the user has literally typed `delete.namespace foo.bar`. You shouldn't just delete all the history just because the namespace is currently empty... that would be a little like deleting a directory automatically once you delete the last file.

aryairani: Doh, good work.  Not sold on the relevance of the directory analogy, but agree we shouldn't be deleting info needed to merge correctly :)

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -312,6 +312,9 @@ relativeToAncestor (Path a) (Path b) = case (a, b) of
   _ -> (empty, Path a, Path b)
 
 pattern Parent h t = Path (NameSegment h :<| t)
+pattern Cons h t <- Path (h :<| (Path -> t)) where
```
aryairani: Instructions unclear, sorry — I suggest using the cons pattern `(<|)` that comes with Control.Lens, for which we have instances below; though if you think it’s better to have an explicit one, that’s fine by me too.

aryairani: Struggling to understand what `Parent` is there too, relative to Cons and Snoc, though I acknowledge it’s not part of this PR.

pchiusano: Yeah, not sure where `Parent` is used, seems a bit weird to me. I'll try out the `<|` pattern, still want the `Empty` pattern though.

aryairani: Just wanted to point out that those instances existed, but I'm good with your choice at that point.

aryairani:
and switch back to `updateChildren` if you want.

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -725,52 +725,52 @@ stepManyAt0 :: forall f m . (Applicative m, Foldable f)
...
+            step (go actions) (Map.findWithDefault empty seg m0)
+          Nothing -> Map.findWithDefault empty seg m0
+    in currentAction $ over children stepChildren b
 
 stepManyAt0M :: forall m n f . (Monad m, Monad n, Foldable f)
              => f (Path, Branch0 m -> n (Branch0 m))
              -> Branch0 m -> n (Branch0 m)
 stepManyAt0M actions b = go (toList actions) b where
   go :: [(Path, Branch0 m -> n (Branch0 m))] -> Branch0 m -> n (Branch0 m)
-  go actions b = let
```
pchiusano: kill trailing whitespace

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -725,52 +725,52 @@ stepManyAt0 :: forall f m . (Applicative m, Foldable f)
...
              => f (Path, Branch0 m -> n (Branch0 m))
              -> Branch0 m -> n (Branch0 m)
 stepManyAt0M actions b = go (toList actions) b where
   go :: [(Path, Branch0 m -> n (Branch0 m))] -> Branch0 m -> n (Branch0 m)
-  go actions b = let 
+  go actions b = let
     currentAction b = foldM (\b f -> f b) b [ f | (Path.Empty, f) <- actions ]
 
     childActions :: Map NameSegment [(Path, Branch0 m -> n (Branch0 m))]
-    childActions =
```
pchiusano: kill trailing whitespace

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -719,32 +719,62 @@ modifyAtM path f b = case Path.uncons path of
...
+    currentAction b = foldl' (\b f -> f b) b [ f | (Path.Empty, f) <- actions ]
+
+    -- groups the actions based on the child they apply to
+    childActions :: Map NameSegment [(Path, Branch0 m -> Branch0 m)]
+    childActions =
+      List.multimap [ (seg, (rest,f)) | (seg :< rest, f) <- actions ]
+
+    -- alters the children of `b` based on the `childActions` map
+    stepChildren :: Map NameSegment (Branch m) -> Map NameSegment (Branch m)
+    stepChildren m0 = foldl' g Map.empty (Map.keysSet m0 <> Map.keysSet childActions)
```
aryairani: I might rename `m0` to `children`, since that's whats passed in.

Then... we're folding over the names of the existing children + the names that have actions associated (may correspond to existing children or to children that don't yet exist).

aryairani: `actions` gets partitioned between `currentAction` (and preemptively composed together), and `childActions`.

Why do we fold over `keysSet m0`?

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -719,32 +719,62 @@ modifyAtM path f b = case Path.uncons path of
...
+      -- this skips adding any empty children, which could happen depending
+      -- on the action functions provided (think: `b -> Branch.empty0`)
+      insert seg child m = if isEmpty child then m else Map.insert seg child m
+      g m seg = insert seg child m where
+        child = case Map.lookup seg childActions of
+          Just actions ->
+            -- this `findWithDefault` is important, allows the stepManyAt
+            -- to create new children at deeper paths than exist in the
+            -- original Branch
+            step (go actions) (Map.findWithDefault empty seg m0)
```
aryairani: `(go actions)` aggregates all of the actions at this level and below into a single action. :+1:

parser-typechecker/src/Unison/Codebase/Branch.hs
pchiusano: Notice the `foldl'` starts with `Map.empty`.

If we just folded over `keysSet childActions`, the `foldl'` would delete all sub-branches that the `childActions` didn't touch... likewise, if we just folded over `keysSet m0`, then we wouldn't be able to have actions that added new child namespaces (I made this mistake in an earlier implementation and lots of tests failed).

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -719,32 +719,62 @@ modifyAtM path f b = case Path.uncons path of
...
+      where
+      insert seg child m = if isEmpty child then m else Map.insert seg child m
+      g m seg = do
+        child <- case Map.lookup seg childActions of
+          Just actions -> stepM (go actions) (Map.findWithDefault empty seg m0)
+          Nothing -> pure $ Map.findWithDefault empty seg m0
+        pure $ insert seg child m
+    in do
+      c2 <- stepChildren (view children b)
+      currentAction (set children c2 b)
```
aryairani: Is there a tidy, correct way to share an implementation between `stepManyAt0` and `stepManyAt0M`

pchiusano: Yeah, I think it'll work to define both in terms of the monadic one, I'll try that out in separate PR.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Oh, hm... <stares more>

aryairani: What about `fold' g m0 (Map.keysSet childActions)`? instead of starting with `Map.empty`?

aryairani: So like there's Map.lookup on the childActions, below.  If it's `Just`, then we call step.  If it's `Nothing`, then that means it's a `seg` that's in `m0` aka `existingChildren`, but not in `childActions`.

When it's `Nothing`, we `Map.findWithDefault empty seg m0` which — it will never use the default, because `seg` is definitely in `m0`, as we know it isn't in `childActions`, and it had to have come from somewhere.

So I think maybe that fold in my previous message will do the trick; leaving such children unchanged from m0, which we're doing anyway, "with extra steps", as Rick and Morty would say.

aryairani: Moreover, we could get rid of that `Map.lookup` I think, since everything we need is in `childActions`; correct me if I'm wrong.

pchiusano: Yeah these are good improvements... agree with all of them and PR coming shortly. Maybe next time do this as part of the initial review though. :)

aryairani: Yeah, my bad. 😓 

I think a sort of mental or definitional shift was needed on my end to decide/realize what code review should be.  I think it's this. 😅

pchiusano: Np :) It is more work to do a detailed review of implementation details and suggest improvements... so there's some cost-benefit there and I wouldn't say it always needs to be done. Not sure what general guidance is on when to do it.

PR #1334, opened 2020-03-10
supporting hashes in `alias`; and names and builtins in `replace`

### LINKED ISSUE #780: the source arg for `alias.term / alias.type` should allow hash-only.
```
.> alias.type ##Sequence base.List

  HQSplit' doesn't have a hash-only option.

.> 
```

### LINKED ISSUE #1047: `replace.term/type` should accept `NameOnly` arguments if unambiguous
```
.> ls

  1.  Fruit                 (type)
  2.  Fruit/                (2 definitions)
  3.  alice/                (283 definitions)
  4.  bob/                  (566 definitions)
  5.  builtin/              (278 definitions)
  6.  fruitStand#g15ka70592 ([Fruit])
  7.  fruitStand#hn233uk80o ([Fruit])
  8.  mergedFruitStand      ([Fruit])
  9.  numFruit#3ujaaara0q   (Nat)
  10. numFruit#a6bifjv41m   (Nat)

.> replace.term 6 mergedFruitStand

  mergedFruitStand is not a valid hash.

.> 
```
😞

aryairani:
## Overview

Previously, the `alias.term` and `alias.type` would fail if the source argument was hash-only, and there was no way to create an alias for a definition that didn't already have a name.  Also, the `replace.term` and `replace.type` _only_ worked on hashes, and they had to be _full_ hashes.

With this PR, the source of an alias can be a short hash (even of a definition that doesn't currently have a name in the namespace) along with a name or hash-qualified name from the current namespace as usual, and the arguments to `replace.term` and `replace.type` can be a short hash, a name, or a hash-qualified name.

Closes #780 
Closes #1047

## Implementation notes

* Exploded Codebase.referencesByPrefix into three functions:
  ```
  - referencesByPrefix :: Text -> m (Set Reference.Id)
  + termReferencesByPrefix :: ShortHash -> m (Set Reference.Id)
  + typeReferencesByPrefix :: ShortHash -> m (Set Reference.Id)
  + termReferentsByPrefix :: ShortHash -> m (Set (Referent' Reference.Id))
  ```
  for top-level terms, top-level types, and for Ref/Con mix.

* Exploded Command.ReferencesByShortHash into three functions
  ```
  - ReferencesByShortHash :: ShortHash -> Command m i v (Set Reference.Id)
  + TypeReferencesByShortHash :: ShortHash -> Command m i v (Set Reference)
  + TermReferencesByShortHash :: ShortHash -> Command m i v (Set Reference)
  + TermReferentsByShortHash :: ShortHash -> Command m i v (Set Referent)
  ```

*  **Note:** As it stands, the types in the Codebase API can only returns `Derived` references; `Builtin` references get added by the Command API.

* By design, the current implementation doesn't let you abbreviate `Builtin` references, but you can abbreviate hash references.

- I added a type arg `r` to Referent, to have a reusable representation for Referents based on `Reference.Id` (i.e. those which are not builtins), and added type alias and patterns for backwards compatibility.
  ```
  data Referent' r = Ref' r | Con' r Int ConstructorType
    deriving (Show, Ord, Eq, Functor)

  type Referent = Referent' Reference
  pattern Ref :: Reference -> Referent
  pattern Ref r = Ref' r
  pattern Con :: Reference -> Int -> ConstructorType -> Referent
  pattern Con r i t = Con' r i t
  {-# COMPLETE Ref, Con #-}
  ```

## Test coverage

Transcript output is [here](https://github.com/unisonweb/unison/blob/6b5d60cb20f9dd1eccbe12fff1b4fa23892f49f3/unison-src/transcripts/fix1334.output.md).

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -212,29 +212,24 @@ loop = do
       getHQ'TermsIncludingHistorical p =
         getTermsIncludingHistorical (resolveSplit' p) root0
 
-      -- HHQS' -> Term References (for `replace`)
       getHQ'TermReferences :: Path.HQSplit' -> Set Reference
       getHQ'TermReferences p =
         Set.fromList [ r | Referent.Ref r <- toList (getHQ'Terms p) ]
-      -- HHQS' -> Term Referents (for `alias`)
       getHQ'Terms :: Path.HQSplit' -> Set Referent
       getHQ'Terms p = BranchUtil.getTerm (resolveSplit' p) root0
-      -- HHQS' -> Type References (for `replace` & `alias`)
       getHQ'Types :: Path.HQSplit' -> Set Reference
       getHQ'Types p = BranchUtil.getType (resolveSplit' p) root0
       resolveHHQS'Types :: HashOrHQSplit' -> Action' m v (Set Reference)
       resolveHHQS'Types = either 
         (eval . TypeReferencesByShortHash)
-        (pure . getHQ'Types)        
+        (pure . getHQ'Types)
+      -- Term Refs only
       resolveHHQS'Terms = either
         (eval . TermReferencesByShortHash)
         (pure . getHQ'TermReferences)
+      -- Term Refs and Cons
       resolveHHQS'Referents = either
-        (\h ->
-          (<>) <$> eval (TermReferentsByShortHash h)
-               <*> fmap (Set.map Referent.Ref)
-                        (eval (TermReferencesByShortHash h))
-               )
+        (eval . TermReferentsByShortHash)
         (pure . getHQ'Terms)
       getTypes :: Path.Split' -> Set Reference
       getTypes = getHQ'Types . fmap HQ'.NameOnly
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -211,8 +211,31 @@ loop = do
         eval . Eval $ Branch.getMaybePatch seg (Branch.head b)
       getHQ'TermsIncludingHistorical p =
         getTermsIncludingHistorical (resolveSplit' p) root0
+
+      -- HHQS' -> Term References (for `replace`)
+      getHQ'TermReferences :: Path.HQSplit' -> Set Reference
+      getHQ'TermReferences p =
+        Set.fromList [ r | Referent.Ref r <- toList (getHQ'Terms p) ]
+      -- HHQS' -> Term Referents (for `alias`)
+      getHQ'Terms :: Path.HQSplit' -> Set Referent
       getHQ'Terms p = BranchUtil.getTerm (resolveSplit' p) root0
+      -- HHQS' -> Type References (for `replace` & `alias`)
+      getHQ'Types :: Path.HQSplit' -> Set Reference
       getHQ'Types p = BranchUtil.getType (resolveSplit' p) root0
+      resolveHHQS'Types :: HashOrHQSplit' -> Action' m v (Set Reference)
+      resolveHHQS'Types = either 
+        (eval . TypeReferencesByShortHash)
+        (pure . getHQ'Types)        
+      resolveHHQS'Terms = either
+        (eval . TermReferencesByShortHash)
+        (pure . getHQ'TermReferences)
+      resolveHHQS'Referents = either
+        (\h ->
+          (<>) <$> eval (TermReferentsByShortHash h)
+               <*> fmap (Set.map Referent.Ref)
+                        (eval (TermReferencesByShortHash h))
+               )
+        (pure . getHQ'Terms)
       getTypes :: Path.Split' -> Set Reference
       getTypes = getHQ'Types . fmap HQ'.NameOnly
       getTerms :: Path.Split' -> Set Referent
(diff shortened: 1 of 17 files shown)


aryairani:
Bleh, it's supposed to be a draft.

aryairani:
```
.> ls

  nothing to show

.> alias.term ##Nat.+ Nat.+

  Done.

.> find

  1. Nat.+ : ##Nat -> ##Nat -> ##Nat
  

.> ls Nat

  1. + (##Nat -> ##Nat -> ##Nat)

.> alias.type ##Nat Nat

  Done.

.> find 

  1. builtin type Nat
  2. Nat.+ :  ->  -> 
  
```
ಠ_ಠ

aryairani:
Got some clues by tracing, the empty strings appear after a call to `suffixify`.  The input looks okay to me though?
```
.> find

names:
  Names {
    currentNames = Names {
      terms = [(Nat.+,Ref' ##Nat.+)], 
      types = [(Nat,##Nat)]
    }, 
    oldNames = Names {
      terms = [(Nat.+,Ref' ##Nat.+)], 
      types = [(Nat,##Nat)]
    }
  }

let suffixified = Names3.suffixify names

suffixified:
  Names {
    currentNames = Names {
      terms = [(,Ref' ##Nat.+),(+,Ref' ##Nat.+),(Nat.+,Ref' ##Nat.+)], 
      types = [(,##Nat),(Nat,##Nat)]
    },
    oldNames = Names {
      terms = [(Nat.+,Ref' ##Nat.+)], 
      types = [(Nat,##Nat)]
    }
  }

  1. builtin type Nat
  2. Nat.+ :  ->  -> 
  
.> 
```

aryairani:
Hypothesis: a bug relating to suffixifying term/type names that appear at the root of the namespace?

```
.> find

  1. builtin type Nat
  2. Nat.+ :  ->  -> 
  

.> move.type Nat x.Nat

  Done.

.> find

  1. Nat.+ :  ->  -> 
  2. builtin type x.Nat
  

.> 
```
No, I guess not.

aryairani:
```
λ> Unison.Name.suffixes "Nat.+"
[Nat.+,+,]
```
👀

aryairani:
Hypothesis: `suffixify`'s call to `Names.suffixes` generates the suffix `""` for every term and type; however, `suffixify` also removes suffixes that are not unique among terms or not unique among types.  It fails in my simple codebase because there is only one type, thus the `""` suffix is unique to `##Nat`.

pchiusano:
Nice sleuthing. So is this still “do not merge yet” or is it ready for review?

aryairani:
@pchiusano I’d say it’s ready for review and I’ll open a new ticket for unsuffixify on singleton namespaces.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -211,8 +211,31 @@ loop = do
...
+      resolveHHQS'Terms = either
+        (eval . TermReferencesByShortHash)
+        (pure . getHQ'TermReferences)
+      resolveHHQS'Referents = either
+        (\h ->
+          (<>) <$> eval (TermReferentsByShortHash h)
+               <*> fmap (Set.map Referent.Ref)
+                        (eval (TermReferencesByShortHash h))
+               )
+        (pure . getHQ'Terms)
```
pchiusano: Just going to assume these are all correct and that they do something useful.

aryairani:
TBH I would like to see a transcript that demonstrates all the modes of the updated commands; both as a tutorial and as a regression test, but am pretty fried and could use a buddy (or a break).

unison-core/src/Unison/Referent.hs
```diff
@@ -19,8 +19,15 @@ import qualified Unison.ConstructorType as CT
...
 -- Slightly odd naming. This is the "referent of term name in the codebase",
 -- rather than the target of a Reference.
-data Referent = Ref Reference | Con Reference Int ConstructorType
-  deriving (Show, Ord, Eq)
+type Referent = Referent' Reference
+pattern Ref :: Reference -> Referent
+pattern Ref r = Ref' r
+pattern Con :: Reference -> Int -> ConstructorType -> Referent
+pattern Con r i t = Con' r i t
+{-# COMPLETE Ref, Con #-}
```
pchiusano: Did you actually manage to get `COMPLETE` pragma to work? I feel like we tried to use it elsewhere without success.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: `HashOrHQSplit'` is an `Either`, and these just handle both sides of it using existing helpers.

`resolveHHQS'Referents` I think no longer needs to be a `<>` though.

unison-core/src/Unison/Referent.hs
aryairani: In this case, a clean build got it work for me.  I don't remember which version of ghc we were on when we couldn't get it to work elsewhere compared to now; I could also believe "sometimes it doesn't work".

The issue I remember running into with it in the past was: adding one `COMPLETE` pragma broke another seemingly unrelated one.  (???)  Was it a ghc issue?  a stack issue?

aryairani:
@pchiusano 
> Question: is the behavior that `alias.term` with a hash only argument doesn't copy any metadata, but the name only or hash qualified versions do copy the metadata? I'd say that is correct, but that can also be something that is tested in the transcript.

Right.

unison-core/src/Unison/Referent.hs
aryairani: @pchiusano I could also have let the actual constructors be called `Ref` and `Con` like they were before instead of `Ref'` and `Con`', but I chose add these patterns so that I could add type annotations to improve whatever inference situations might arise in the future.

PR #1337, opened 2020-03-11
Fix Branch.updateChildren to not elide empty branches with history

pchiusano:
My thinking is merge this after it passes CI, then merge #1333 

## Overview

Currently, `Branch.updateChildren` deletes branches that have no definitions in them but which do have history. Use of this function caused previously deleted definitions to resurface during a merge. See the discussion [in this PR](https://github.com/unisonweb/unison/pull/1333#discussion_r390628109). 

I think the correct behavior is to only delete a child if it is being replaced with the empty `Branch` (with no history), as is done in a `delete.namespace`. That's what this PR does.

## Test coverage

The existing transcripts provide pretty good coverage along with the implementation that's in `master` which still uses `updateChildren` for various operations on `Branch`. Assuming those transcripts still pass in this PR, this is safe to merge I'd say.

### DIFF
--- a/unison-src/transcripts/delete.output.md
+++ b/unison-src/transcripts/delete.output.md
@@ -182,10 +182,6 @@ type Foo = Foo Boolean
     5. ┌ Foo.Foo#d97e0jhkmd#0 : Nat -> Foo
     6. └ Foo.Foo#gq9inhvg9h#0 : Boolean -> b.Foo
   
-  Added definitions:
-  
-    7. foo : Nat
-  
   Tip: You can use `todo` to see if this generated any work to
        do in this namespace and `test` to run the tests. Or you
        can use `undo` or `reflog` to undo the results of this
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -691,7 +691,7 @@ updateChildren ::NameSegment
                -> Map NameSegment (Branch m)
                -> Map NameSegment (Branch m)
 updateChildren seg updatedChild =
-  if isEmpty0 (head updatedChild)
+  if isEmpty updatedChild
   then Map.delete seg
   else Map.insert seg updatedChild
 


unison-src/transcripts/delete.output.md
```diff
@@ -182,10 +182,6 @@ type Foo = Foo Boolean
     5. ┌ Foo.Foo#d97e0jhkmd#0 : Nat -> Foo
     6. └ Foo.Foo#gq9inhvg9h#0 : Boolean -> b.Foo
   
-  Added definitions:
-  
-    7. foo : Nat
-
```
pchiusano: This is the same erroneous undeletion that #1333 also resolves (by avoiding use of the `updateChildren` function in the first place). Nice to see that it also is fixed here by changing how `updateChildren` works.

pchiusano:
Just noting that this currently passes CI with the same diff to delete.output.md as in #1333.

unison-src/transcripts/delete.output.md
aryairani: So after this patch you would switch back to updateChildren?

aryairani:
I think we need to doublecheck the `ls` output and hide empty children if that’s what we want to have happen (and I think that it is), like @pchiusano proposed at the time I’m guessing I introduced this bug.

unison-src/transcripts/delete.output.md
pchiusano: Might be a good idea.

pchiusano:
@aryairani I'm not at all sold anymore on just hiding empty namespaces in `ls`. Knowing that there's a currently empty namespace is useful information IMO - it is telling you that there was once some definitions in there and they were deleted. You might ask for `history` of that now-empty namespace, etc.

We could introduce a flag or option to `ls` later that does the filtering, maybe `ls.nonEmpty` or something. Doesn't seem super important to me though.

aryairani:
Hm, it came across to me as looking like a bug to have 0 definitions, but I'm not sure.

PR #1341, opened 2020-03-11
Some nice cleanup of `stepManyAt0` and friends

pchiusano:
@aryairani suggested these improvements, see discussion [here](https://github.com/unisonweb/unison/pull/1333#pullrequestreview-372819650). 

It worked out very nicely, except one downside is various functions picked up a `Monad m` constraint where before they could get away with `Applicative m`. I don't think this was a big deal at all as `m` is always a `Monad` anyway. Here's the new implementation:

https://github.com/unisonweb/unison/blob/de5261a81c67e37a4322ff77ba50ab7234fbb60c/parser-typechecker/src/Unison/Codebase/Branch.hs#L729-L755

## Test coverage

Existing tests cover this quite well and it's just an implementation change.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -625,7 +625,7 @@ stepAt p f = modifyAt p g where
   g :: Branch m -> Branch m
   g (Branch b) = Branch . Causal.consDistinct (f (Causal.head b)) $ b
 
-stepManyAt :: (Applicative m, Foldable f)
+stepManyAt :: (Monad m, Foldable f)
            => f (Path, Branch0 m -> Branch0 m) -> Branch m -> Branch m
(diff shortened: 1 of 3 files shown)


aryairani:
Probably stupid question (as I've gone through and tried to eliminate or downgrade these before, but don't remember the outcomes), but does `stepManyAt0M` need all of its existing constraints?  Eyeballing it, I don't see it.

aryairani:
I'll take a quick stab at it

pchiusano:
@aryairani Yeah the constraints are needed (the `Monad n` due to the `foldM`, the `Foldable f` due to the `toList`, and the `Monad m` due to the `stepM` call).

aryairani:
> @aryairani Yeah the constraints are needed (the `Monad n` due to the `foldM`, the `Foldable f` due to the `toList`, and the `Monad m` due to the `stepM` call).

Okay, great.

Issue #1295, opened 2020-03-02
`ucm:hide` blocks are hidden from the transcript failure output


aryairani:
For example:

    Dangerous scary words!
    
    ```ucm:hide
    .> move.namespace foo bar
    ```

Yields:
```
⚙️   Processing stanza 2 of 2.Dangerous scary words!



🛑

Transcript failed due to the message above.
```

which is confusing, because the error is hidden.

* `unison:hide` blocks are okay-ish, as they still print the offending source.
* `unison:hide:all` blocks again hide all of the problem.

pete-ts:
@aryairani I think I might have some time to look into this, after a spell of being tied up. Do you have a suggested target/fixed behaviour in mind?

pete-ts:
(As long as "help-wanted" of course!)

aryairani:
@pete-ts 
![image](https://user-images.githubusercontent.com/538571/75915264-77e64700-5e56-11ea-9da9-06771006a70f.png)
😁

I was thinking the behavior should be:  If a transcript fails as a result of a hidden stanza (either due to encountering an error, or failing to encounter an expected error), then the stanza should still be printed to the console before "The transcript failed due to an error encountered in the stanza above."

"The transcript failed to fail as expected in the stanza above." also seems reasonable, if you are willing to set up a separate message for the two cases.

PR #1345, opened 2020-03-12
Failing transcripts should always show the failing stanza

pete-ts:
## Overview

Currently the `stack exec transcripts` command runs the content of `Transcripts.hs`.

1. find all `.md` files in the `transcripts folder`
2. build a "ucm transcript command for each"
3. run each command as a test and expect it to return exit code 0.
4. if changes to `ucm` have affected the output to one or more transcripts, you now need to commit these, or these are caught by the build system and a build failure is produced.

The above behaviour is maintained and not altered at all.

This PR aims to add a new section of tests in the `stack exec transcripts` project that will allow failing transcripts to run to completion while capturing the error messages produced and tracking them as output transcripts.

1. find all `.md` files in the `errors folder`
2. build a "ucm transcript command for each"
3. run each transcript as a test and only capture the standard error.
4. if changes to `ucm` have affected the stderr of the transcripts, commit as necessary, or get a build failure. 

Following that, I have managed to address aspects of #1295 as seen in the output files.
1. when a ucm or unison block fails the failing stanza is always displayed
2. the error message for unexpected success has been updated
3. the error message for a failing stanza has been updated

## Implementation notes

`Transcripts.hs`
[shellmet](https://hackage.haskell.org/package/shellmet-0.0.3.0/docs/Shellmet.html) does not allow for running a failing command and capturing the error, but achieving this is quite straightforward (as `shellmet` is a wrapper around `System.Process` functions).

`TranscriptParser`
1. the stanza being worked on is stored in an IORef
2. this makes it easy to append it during an error

* This can probably be achieved by `bash` but I feel that the approach suggested fits in with the existing mechanism for testing transcripts. There is no need to maintain an additional testing structure in parallel, as the "error" scenarios are testing the transcript parser, under the existing `stack exec transcripts`

* The local `die` function had to be renamed for clarity as the name clashes with `System.Exit (die)`

## Test coverage

This PR adds a way to enable additional testing, as well as examples of what these tests would look like.

Finally the PR tests the fix for #1295 

## Loose ends
* Once these two are sorted, a change so that `Transcripts.runAndCaptureError` only succeeds if it gets a non zero exit code.


parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
```diff
@@ -68,7 +69,7 @@ instance Show Stanza where
   show s = case s of
     Ucm _ _ cmds -> unlines [
       "```ucm",
-      show cmds,
+      foldl (\x y -> x ++ show y) "" cmds,
```
pete-ts: This means that when showing a complete failing stanza, there are no angle braces around each command

parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
```diff
@@ -150,10 +152,12 @@ run dir configFile stanzas codebase = do
       awaitInput = do
         cmd <- atomically (Q.tryDequeue cmdQueue)
         case cmd of
+          -- end of ucm block
           Just Nothing -> do
-            output "\n```\n" -- this ends the ucm block
-            writeIORef hidden Shown
+            output "\n```\n"
+            dieUnexpectedSuccess
```
pete-ts: At the end of the `ucm` block check if there was an unexpected success
There is no need to update the `hidden`, as the next stanza will do that.

parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
```diff
@@ -244,20 +242,45 @@ run dir configFile stanzas codebase = do
...
-      die = do
+      -- Looks at the current stanza and decides if it is contained in the
+      -- output so far. Appends it if not.
+      appendFailingStanza :: IO ()
+      appendFailingStanza = do
+        stanzaOpt <- readIORef mStanza
+        currentOut <- readIORef out
+        let stnz = maybe "" show (fmap fst stanzaOpt :: Maybe Stanza)
+        unless (stnz `isSubsequenceOf` concat currentOut) $
+          modifyIORef' out (\acc -> acc <> pure stnz)
```
pete-ts: This function is required to allow us to overcome the fact that some blocks are hidden, after the fact. When either of `dieWithMsg` or `dieUnexpectedSuccess` gets called, the `awaitInput` method has broken down a stanza, and is working on individual `ucm` commands in the stanza. 

The output so far is read, and if the stanza being worked on and containing an error is already printed (by an `outputEcho` earlier) it is not appended a second time.

unison-src/transcripts/errors/ucm-hide.output.md
```diff
@@ -0,0 +1,17 @@
...
+and surface a helpful message.
+
+```ucm
+.> move.namespace foo bar
+```
+
+
+🛑
+
+The transcript failed due to an error encountered in the stanza above.
```
pete-ts: I believe this the expected behaviour for #1295 @aryairani

unison-src/transcripts/errors/unison-hide.output.md
```diff
@@ -0,0 +1,16 @@
...
+
+```unison
+g 3
+```
+
+
+
+🛑
+
+The transcript failed due to an error encountered in the stanza above.
```
pete-ts: unison:hide shows the stanza

unison-src/transcripts/errors/unison-hide-error.output.md
```diff
@@ -0,0 +1,16 @@
...
+
+```unison
+myVal = 3
+```
+
+
+
+🛑
+
+The transcript failed to fail as unexpected in the stanza above.
```
pete-ts: unison hide error shows the stanza and the new error message

unison-src/transcripts/errors/unison-hide-all-error.output.md
```diff
@@ -0,0 +1,16 @@
...
+
+```unison
+myVal = 3
+```
+
+
+
+🛑
+
+The transcript failed to fail as unexpected in the stanza above.
```
pete-ts: unison:hide:all:error shows the stanza and updated message

unison-src/transcripts/errors/ucm-hide-all-error.md
```diff
@@ -8,5 +8,5 @@ then the transcript parser should print the stanza
 and surface a helpful message.
 
 ```ucm:hide:all:error
-.> ls
+.> history
```
pete-ts: @aryairani came back to this with a clear head, and managed to debug it. `ls` was returning an error, so both of these were actually interpreted as "an error occurs in a block where an error is expected."

aryairani:
@pete-ts Sorry for the delay on this.  I will dig into it "soon".  😬

PR #1360, opened 2020-03-14
Allow deleting type and term replacements in patches

runarorama:
## Overview

This change adds `delete.termReplacement` and `delete.typeReplacement` commands to UCM which allows deleting entries from a patch for term edits and type edits, respectively.

## Implementation notes

This implements the `RemoveTermReplacementI` and `RemoveTypeReplacementI` input types that have been sitting unimplemented for a while. Though, with a few changes.

I decided to only take one name/hash as input, in addition to the (optional) patch. This is the type or term that _got replaced_. I found that this makes it easier to use, and you can't have the same term or type replaced by multiple things in the same patch anyway.

## Test coverage

I added a transcript that adds term/type replacements and then removes them.


runarorama:
You can use the hash instead of a historical name. That's fine.

pchiusano:
@runarorama Cool! Comments -

Can we go with the names unreplace.term and unreplace.type instead? (Like we’ve got link and unlink)

I feel like most common workflow for this will be a view.patch, followed by issuing these commands. Could we add a tip at the end of the view.patch output on how to use? (“If you’d like to remove any of these entries, do...”) Tip can use specific examples chosen from the patch. I think like the reflog output does this sort of thing.

 (Separately from this PR, would be cool if view.patch output were numbered so you could refer to it by number in subsequent commands rather than copy/pasting like a caveman... so like view.patch and then ‘unreplace.term 4’)

I was sort of unclear on the transcript - can you just pass the hash qualified name mentioned in the patch? If so, can we just simplify the transcript? Might mean hard-coding the hash in the transcript but that should be stable.

pchiusano:
Also what happens if the patch is conflicted and a definition has two replacements? Can we see what happens in the transcript? 

A good experience: “Not sure which of these you mean to remove from the patch” and then show a subset of the patch (as a numbered list ideally).

runarorama:
What happens currently is that it deletes all replacement for that hash.

aryairani:
I dunno if further logic changes belong in this PR vs a different one, but since we're on the subject:

@pchiusano  `replace.term` and `replace.type` propagate changes to dependents throughout the namespace.  An `unreplace.term` and `unreplace.type` sound like they might undo this, but presumably wouldn't.

@runarorama Along the same lines, there might be a place for `add.termReplacement` and `add.typeReplacement` that only manipulate the patch and don't propagate; I dunno.  (I don't recall having encountered a use for it, but it's a possible thing.)

pchiusano:
@aryairani good point re: naming. Am fine with naming in this PR, but can we be consistent about camel case vs dashes in UCM command names? Like it's currently `pull-request.create` not `pullRequest.create`. Suggest sticking with that here or changing the other one to camel case.

@runarorama These are only other two holdups for me on this PR, which I mentioned above:

* Could we add a tip at the end of the view.patch output on how to use? (“If you’d like to remove any of these entries, do...”) Tip can use specific examples chosen from the patch. I think like the reflog output does this sort of thing. If you really want, create a separate issue to track this, but tI think it'd be easy to bust out.
* Can you just pass the hash qualified name mentioned in the patch? If so, can we just simplify the transcript? Might mean hard-coding the hash in the transcript but that should be stable.

runarorama:
Actually, using a hash-qualified name there doesn't seem to work and I'm working out how to make it work.

runarorama:
@aryairani I believe the latest changes resolve your concerns. LMK if not.

PR #1361, opened 2020-03-14
Fix #1338: adds the missing "List.empty" to the builtins Map

### LINKED ISSUE #1338: runtime doesn't implement `##List.empty`
When you place this line in a scratch file

&gt; base.List.empty

You'll receive the following error:

https://gist.github.com/drewctaylor/5d8cf40b808d1985c3de82589c35c482

    The program being compiled referenced this definition ##List.empty

`ucm --version` returns `release/M1h`.

pete-ts:
## Overview

This PR aims to fix #1338 by adding `List.empty` to the `builtins` compilation environment

Before:
Attempting to use List.empty directly crashes `ucm` as per the bug.

After:
Using `List.empty` typechecks as `[a]`.

## Implementation notes

By adding the entry to the `CompilationEnv` the call `toIR env r` now finds the builtin, instead of failing and erroring out.

## Test coverage

Transcript attached.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/IR.hs
+++ b/parser-typechecker/src/Unison/Runtime/IR.hs
@@ -872,6 +872,7 @@ builtins = Map.fromList $ arity0 <> arityN
   var = Var.named "x"
   arity0 = [ (R.Builtin name, val value) | (name, value) <-
         [ ("Text.empty", T "")
+        , ("List.empty", Sequence mempty)
         , ("Sequence.empty", Sequence mempty)
         , ("Bytes.empty", Bs mempty)
         ] ]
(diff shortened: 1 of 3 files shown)


pete-ts:
I am of the opinion that adding a transcript that pulls base is actually a bad idea, as this does add a quite the overhead. If there is a better suggestion for how to test this without the pull to `base` please let me know, otherwise the transcripts can always be removed in the interest of speed?

aryairani:
You can `alias.term ##List.empty List.empty` now instead of pulling base.  For the time being, we won’t merge automated tests that pull base.

pete-ts:
Closing; a different approach will be followed for the fix.

Issue #1346, opened 2020-03-12
RFC: design for efficient lookup of old names


pchiusano:
In the repo format, add the following directory:

```
names/<hash1-2>/<hash3-4>/.../<hash511-512/
  namehash1.name
  namehash2.name
  ...
```

Where `<hash1-2>` are the first two chars of the hash, `<hash3-4>` are the 3rd and 4th characters of the hash, etc. `namehash1.name` is a hash of the name and the file contents are a UTF-8 encoded `Name`.

So it's an append-only, on-disk `Trie HashChar (Set Name)`. Whenever definitions are added to the codebase, they are inserted into this trie. Supports efficient lookup of any partial hash. Merges via file system operations efficiently without copying any data into Haskell memory.

The way this would be used:

* Before pretty-printing anything, any references not covered by current namespace would get looked up in the names index. These would get used as the second`Names3` and get hash qualified via the existing mechanism. So this replaces the historical search through the branch that is sometimes super slow.
* Suggest that these get synced via `push` / `pull`, with the behavior that on `pull`, if a name already exists for a reference, names from the source are ignored, and if pushing, only fill in a name if none exists for a hash.
* Later, we could potentially expose UCM commands for maintenance of this names index, like updating the indexed name / picking a winner.

aryairani:
How do we add this to codebases that are currently missing it?  

Option 1: Build from scratch if it's missing, and then once it exists, assume it is complete?  
Option 2: Include a list of namespace hashes that are represented in the index.  Start indexing from the head namespace.  To index a namespace `ns`:  If `hash ns` is already in the index, then you're done.  Otherwise, push `ns` onto a stack, index `predecessor ns`, add the names from `ns` (optionally delete older names for the same hash), add `hash ns`, pop the stack and continue.

aryairani:
Superseded by #2128

PR #1368, opened 2020-03-18
Make `link`, `unlink`, and `diff.namespace` work with unnamed metadata (fix for #1356)

### LINKED ISSUE #1356: `link` allows re-linking and breaks ucm on `unlink`
The usage of the link command to reproduce this issue might not technically be correct, but `ucm` allows it and thus ends in a bad state (I think)

When linked documentation gets updated, the user can `re-link` the documentation. This leads to a situation where multiple documentation links are shown, and an attempt to `unlink` crashes `ucm`

https://gist.github.com/pete-ts/ddb8dea0fc4ffc9f21e73c7820781844

pchiusano:
Thanks to @pete-ts for reporting original issue.

## Overview

Fixes #1356, but more generally, `link`, `unlink`, and `diff.namespace` commands all now work fine even if metadata values are unnamed in the current namespace. See [the transcript](https://github.com/unisonweb/unison/blob/083ed75f3e3b5ffe56e1da7c24a4141628419c72/unison-src/transcripts/fix1356.output.md), which previously bombed with a runtime error but now has been beefed up and works great.

`link #somehash List.map` and  `unlink #somehash my.stuff` now work, previously they were rejected by the parser.

## Implementation notes

See self review.

When you do `unlink foo#xyzpq mydef`, the name `foo` is ignored - just the hash is used to resolve the metadata. @aryairani and I decided this was better than the alternative of doing a historical names search.

## Test coverage

It's pretty reasonable, see [the transcript](https://github.com/unisonweb/unison/blob/083ed75f3e3b5ffe56e1da7c24a4141628419c72/unison-src/transcripts/fix1356.output.md).

## Loose ends

It's kind of a bummer that the namespace diff isn't getting access to historical names, so old metadata that's been updated is just shown as its hash. (Previously it would straight up crash)

When we do #1346 it'll be nice to get those hash qualified using whatever old names exist for the hash.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -42,7 +42,6 @@ import           Data.Configurator              ()
 import qualified Data.List                      as List
 import           Data.List                      ( partition, sortOn )
 import           Data.List.Extra                ( nubOrd, sort )
-import           Data.Maybe                     ( fromJust )
 import qualified Data.Map                      as Map
 import qualified Data.Text                     as Text
 import qualified Text.Megaparsec               as P
@@ -220,7 +219,7 @@ loop = do
       getHQ'Types :: Path.HQSplit' -> Set Reference
       getHQ'Types p = BranchUtil.getType (resolveSplit' p) root0
       resolveHHQS'Types :: HashOrHQSplit' -> Action' m v (Set Reference)
-      resolveHHQS'Types = either 
+      resolveHHQS'Types = either
         (eval . TypeReferencesByShortHash)
         (pure . getHQ'Types)
       -- Term Refs only
@@ -1327,7 +1326,8 @@ loop = do
 
           allTypes :: Map Reference (Type v Ann) <-
             fmap Map.fromList . for (toList neededTypes) $ \r ->
-              (r,) . fromJust <$> (eval . LoadTypeOfTerm) r
+              (r,) . fromMaybe (Type.builtin External "unknown type")
+              <$> (eval . LoadTypeOfTerm) r
 
           let typing r1 r2 = case (Map.lookup r1 allTypes, Map.lookup r2 hashTerms) of
                 (Just t1, Just t2)
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -435,18 +435,14 @@ loop = do
                         ->  Branch.Star r NameSegment)
                     -> MaybeT (StateT (LoopState m v) (F m (Either Event Input) v)) ()
         manageLinks srcs mdValue2 op = do
-          traceM "In manage links, before srcle"
           let !srcle = toList . getHQ'Terms =<< srcs
               !srclt = toList . getHQ'Types =<< srcs
-          traceM "In manage links, AFTER srcle"
           mdValuel <- toList <$> getHQTerms mdValue2
-          traceM "Got mdValuel"
           names0 <- basicPrettyPrintNames0
           ppe <- PPE.suffixifiedPPE <$> prettyPrintEnvDecl (Names names0 mempty)
           case (srcle, srclt, mdValuel) of
             (srcle, srclt, [r@(Referent.Ref mdValue)]) -> do
               mdType <- eval $ LoadTypeOfTerm mdValue
-              traceM $ "Loaded type for " <> show mdValue
               case mdType of
                 Nothing -> respond $ MetadataMissingType ppe r
                 Just ty -> do
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -360,9 +360,9 @@ loop = do
           PropagatePatchI p scope -> "patch " <> ps' p <> " " <> p' scope
           UndoI{} -> "undo"
           ExecuteI s -> "execute " <> Text.pack s
-          LinkI defs md ->
+          LinkI md defs ->
             "link " <> HQ.toText md <> " " <> intercalateMap " " hqs' defs
-          UnlinkI defs md ->
+          UnlinkI md defs ->
             "unlink " <> HQ.toText md <> " " <> intercalateMap " " hqs' defs
           UpdateBuiltinsI -> "builtins.update"
           MergeBuiltinsI -> "builtins.merge"
@@ -902,10 +902,10 @@ loop = do
 --                      | r <- toList $ Names.typesNamed ns name ]
 --              in (terms, types)
 
-      LinkI srcs mdValue ->
+      LinkI mdValue srcs ->
         manageLinks srcs mdValue Metadata.insert
 
-      UnlinkI srcs mdValue ->
+      UnlinkI mdValue srcs ->
         manageLinks srcs mdValue Metadata.delete
 
       -- > links List.map (.Docs .English)
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -218,6 +218,19 @@ loop = do
       getHQ'Terms p = BranchUtil.getTerm (resolveSplit' p) root0
       getHQ'Types :: Path.HQSplit' -> Set Reference
       getHQ'Types p = BranchUtil.getType (resolveSplit' p) root0
+      getHQTerms :: HQ.HashQualified -> Action' m v (Set Referent) 
+      getHQTerms hq = case hq of
+        HQ.NameOnly n -> let
+          -- absolute-ify the name, then lookup in deepTerms of root
+          path :: Path.Path' 
+          path = Path.fromName' n
+          Path.Absolute absPath = resolveToAbsolute path
+          in pure $ R.lookupRan (Path.toName absPath) (Branch.deepTerms root0) 
+        HQ.HashOnly sh -> hashOnly sh 
+        HQ.HashQualified _ sh -> hashOnly sh
+        where
+        hashOnly sh = eval $ TermReferentsByShortHash sh
+
       resolveHHQS'Types :: HashOrHQSplit' -> Action' m v (Set Reference)
       resolveHHQS'Types = either
         (eval . TypeReferencesByShortHash)
@@ -347,10 +360,10 @@ loop = do
           PropagatePatchI p scope -> "patch " <> ps' p <> " " <> p' scope
           UndoI{} -> "undo"
           ExecuteI s -> "execute " <> Text.pack s
-          LinkI froms to ->
-            "link " <> hqs' to <> " " <> intercalateMap " " hqs' froms
-          UnlinkI froms to ->
-            "unlink " <> hqs' to <> " " <> intercalateMap " " hqs' froms
+          LinkI defs md ->
+            "link " <> HQ.toText md <> " " <> intercalateMap " " hqs' defs
+          UnlinkI defs md ->
+            "unlink " <> HQ.toText md <> " " <> intercalateMap " " hqs' defs
           UpdateBuiltinsI -> "builtins.update"
           MergeBuiltinsI -> "builtins.merge"
           PullRemoteBranchI orepo dest ->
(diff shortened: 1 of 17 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -219,8 +218,21 @@ loop = do
       getHQ'Terms p = BranchUtil.getTerm (resolveSplit' p) root0
       getHQ'Types :: Path.HQSplit' -> Set Reference
       getHQ'Types p = BranchUtil.getType (resolveSplit' p) root0
+      getHQTerms :: HQ.HashQualified -> Action' m v (Set Referent)
```
pchiusano: Handy utility function which we thought should exist, but didn't yet. Converts a hash qualified to a `Set Referent`, taking into account current path and branch.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -415,23 +427,24 @@ loop = do
...
+          mdValuel <- toList <$> getHQTerms mdValue2
+          names0 <- basicPrettyPrintNames0
+          ppe <- PPE.suffixifiedPPE <$> prettyPrintEnvDecl (Names names0 mempty)
           case (srcle, srclt, mdValuel) of
-            (srcle, srclt, [Referent.Ref mdValue]) -> do
+            (srcle, srclt, [r@(Referent.Ref mdValue)]) -> do
               mdType <- eval $ LoadTypeOfTerm mdValue
               case mdType of
-                Nothing -> respond $ LinkFailure input
+                Nothing -> respond $ MetadataMissingType ppe r
```
pchiusano: We previously just had `LinkFailure`, which I split into `MetadataMissingType` and `MetadataAmbiguous` messages.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1327,7 +1341,8 @@ loop = do
 
           allTypes :: Map Reference (Type v Ann) <-
             fmap Map.fromList . for (toList neededTypes) $ \r ->
-              (r,) . fromJust <$> (eval . LoadTypeOfTerm) r
+              (r,) . fromMaybe (Type.builtin External "unknown type")
+              <$> (eval . LoadTypeOfTerm) r
```
pchiusano: I removed all `fromJust` calls from the codebase as part of this PR (original but was a fromJust runtime crash). 

This treatment here is a little dubious (it'll get a type of "unknown type"), but should only get hit if codebase is corrupted somehow.

parser-typechecker/src/Unison/Codebase/Editor/Input.hs
```diff
@@ -109,10 +109,10 @@ data Input
...
   | TestI Bool Bool -- TestI showSuccesses showFailures
   -- metadata
-  -- link from to
-  | LinkI [Path.HQSplit'] Path.HQSplit'
-  -- unlink from to
-  | UnlinkI [Path.HQSplit'] Path.HQSplit'
+  -- `link metadata definitions` (adds metadata to all of `definitions`)
+  | LinkI HQ.HashQualified [Path.HQSplit'] 
+  -- `unlink metadata definitions` (removes metadata from all of `definitions`)
+  | UnlinkI HQ.HashQualified [Path.HQSplit']
```
pchiusano: Some cleanup: switched `LinkI` and `UnlinkI` to take args in the same order as their input pattern, it's `link somedoc def1 def2`.

parser-typechecker/src/Unison/Codebase/Editor/Output/BranchDiff.hs
```diff
@@ -91,7 +91,7 @@ type SimpleTypeDisplay v a = (HashQualified, Reference, Maybe (DeclOrBuiltin v a
 type UpdateTermDisplay v a = (Maybe [SimpleTermDisplay v a], [TermDisplay v a])
 type UpdateTypeDisplay v a = (Maybe [SimpleTypeDisplay v a], [TypeDisplay v a])
 
-type MetadataDisplay v a = SimpleTermDisplay v a
+type MetadataDisplay v a = (HQ.HashQualified, Referent, Maybe (Type v  a))
```
pchiusano: Metadata can be unnamed, so should be `HashQualified`,  not `HashQualified'`.

parser-typechecker/src/Unison/Codebase/Editor/Output/BranchDiff.hs
```diff
@@ -329,7 +329,9 @@ toOutput typeOf declOrBuiltin hqLen names1 names2 ppe
   where
   fillMetadata :: Traversable t => PPE.PrettyPrintEnv -> t Metadata.Value -> m (t (MetadataDisplay v a))
   fillMetadata ppe = traverse $ -- metadata values are all terms
-    \(Referent.Ref -> mdRef) -> (HQ'.unsafeFromHQ $ PPE.termName ppe mdRef, mdRef, ) <$> typeOf mdRef
```
pchiusano: This `unsafeFromHQ` was the line that was bombing in the original. report.

parser-typechecker/src/Unison/Codebase/Metadata.hs
```diff
@@ -43,7 +44,15 @@ insert :: (Ord a, Ord n) => (a, Type, Value) -> Star a n -> Star a n
...
-delete (a, ty, v) = Star3.deleteD23 (a, ty, (ty,v))
+delete (a, ty, v) s = let
+  s' = Star3.deleteD3 (a, (ty,v)) s
+  -- if (ty,v) is the last metadata of type ty
+  -- we also delete (a, ty) from the d2 index
+  metadataByType = List.multimap (toList (R.lookupDom a (Star3.d3 s)))
+  in 
+  case Map.lookup ty metadataByType of
+    Just vs | all (== v) vs -> Star3.deleteD2 (a, ty) s' 
+    _ -> s'
```
pchiusano: Found and fixed this bug as part of this PR. Previous implementation was just WRONG and would delete ALL the metadata of the same type as the argument passed to delete, like if you deleted one `Doc` you deleted all of them. This implementation fixes that. 

This fix is exercised by the transcript (I got it to fail in the transcript first, then made this fix).

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -1033,10 +1033,12 @@ link = InputPattern
     ]
   )
   (\case
-    dest : srcs -> first fromString $ do
-      srcs <- traverse Path.parseHQSplit' srcs
-      dest <- Path.parseHQSplit' dest
-      Right $ Input.LinkI srcs dest
+    md : defs -> first fromString $ do
```
pchiusano: Better nomenclature here, moving away from "source" and "target" to "metadata" and "definitions", as in, you "link/associate this metadata with these definitions".

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -323,7 +324,25 @@ notifyUser dir o = case o of
     else putPretty' "  🚫  "
     pure mempty
 
-  LinkFailure input -> pure $ P.warnCallout . P.shown $ input
```
pchiusano: Previous garbage error message.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -323,7 +324,25 @@ notifyUser dir o = case o of
...
+  MetadataMissingType ppe ref -> pure . P.fatalCallout . P.lines $ [
+    P.wrap $ "The metadata value " <> P.red (prettyTermName ppe ref)
+          <> "is missing a type signature in the codebase.",
+    "",
+    P.wrap $ "This might be due to pulling an incomplete"
+          <> "or invalid codebase, or because files inside the codebase"
+          <> "are being deleted external to UCM."
+    ]
```
pchiusano: New, pretty error message!

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -323,7 +324,25 @@ notifyUser dir o = case o of
...
+  MetadataAmbiguous ppe [] -> pure . P.warnCallout .
+    P.wrap $ "Nothing to do. I couldn't find any matching metadata."
+  MetadataAmbiguous ppe refs -> pure . P.warnCallout . P.lines $ [
+    P.wrap $ "I'm not sure which metadata value you're referring to"
+          <> "since there are multiple matches:",
+    "",
+    P.indentN 2 $ P.spaced (P.blue . prettyTermName ppe <$> refs),
+    "",
+    tip "Try again and supply one of the above definitions explicitly."
+    ]
```
pchiusano: Thought about doing numbering for these, but didn't bother with that yet and wasn't sure it was a good idea. The `docs` and `links` commands already give you a numbered list.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -891,11 +889,14 @@ calcImports im tm = (im', render $ getUses result)
     --   - is > 1, or
     --   - is 1, and S is an infix operator.
     -- Also drop names with an empty prefix.
+    lookupOrDie s m = fromMaybe msg (Map.lookup s m) where
+      msg = error $ "TermPrinter.enoughUsages " <> show (s, m)
+
     enoughUsages :: Map Suffix (Prefix, Int) -> Map Suffix (Prefix, Int)
-    enoughUsages m = (Map.keys m) |> filter (\s -> let (p, i) = fromJust $ Map.lookup s m
+    enoughUsages m = (Map.keys m) |> filter (\s -> let (p, i) = lookupOrDie s m
```
pchiusano: Getting rid of `fromJust` in favor of actual error message.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -904,13 +905,17 @@ calcImports im tm = (im', render $ getUses result)
...
                           k2s = k1s |> Map.fromSet (\k1' -> Map.keysSet m
                                                               |> Set.filter (\(k1, _) -> k1 == k1')
                                                               |> Set.map snd)
                           maxk2s = Map.map maximum k2s
-                      in Map.mapWithKey (\k1 k2 -> fromJust $ Map.lookup (k1, k2) m) maxk2s
+                          err k1 k2 = error $ 
+                            "TermPrinter.longestPrefix not found " 
+                            <> show (k1,k2) 
+                            <> " in " <> show maxk2s
+                      in Map.mapWithKey (\k1 k2 -> fromMaybe (err k1 k2) $ Map.lookup (k1, k2) m) maxk2s
```
pchiusano: Getting rid of `fromJust` in favor of meaningful error message.

parser-typechecker/src/Unison/Typechecker/Components.hs
```diff
@@ -61,7 +60,8 @@ minimize (Term.LetRecNamedAnnotatedTop' isTop ann bs e) =
               -- loop).
               cs = sortOn (\(_,e) -> Term.arity e == 0) <$> cs0
               varAnnotations = Map.fromList ((\((a, v), _) -> (v, a)) <$> bs)
-              annotationFor v = fromJust $ Map.lookup v varAnnotations
+              msg v = error $ "Components.minimize " <> show (v, Map.keys varAnnotations)
```
pchiusano: `fromJust` cleanup

parser-typechecker/src/Unison/Util/Star3.hs
```diff
@@ -160,13 +160,19 @@ insertD23 (f, x, y) s = Star3 fact' (d1 s) d2' d3' where
...
+          -> Star3 fact d1 d2 d3
+deleteD3 (f, x) s = Star3 (fact s) (d1 s) (d2 s) d3' where
+  d3' = R.delete f x (d3 s)
+
+deleteD2 :: (Ord fact, Ord d1, Ord d2, Ord d3)
+          => (fact, d2)
           -> Star3 fact d1 d2 d3
           -> Star3 fact d1 d2 d3
-deleteD23 (f, x, y) s = Star3 (fact s) (d1 s) d2' d3' where
+deleteD2 (f, x) s = Star3 (fact s) (d1 s) d2' (d3 s) where
```
pchiusano: GitHub diff is misleading - I deleted the unused `deleteD23` and introduced `deleteD2` and `deleteD3`.

unison-core/src/Unison/HashQualified'.hs
```diff
@@ -33,9 +32,6 @@ fromHQ = \case
   HQ.HashQualified n sh -> Just $ HashQualified n sh
   HQ.HashOnly{} -> Nothing
 
-unsafeFromHQ :: HQ.HashQualified' n -> HashQualified' n
-unsafeFromHQ = fromJust . fromHQ
```
pchiusano: This was only being used in one spot, where it was crashing! Deleted it!

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: Better than crashing 👍

parser-typechecker/src/Unison/TermPrinter.hs
aryairani: better than crashing with no message :)

pchiusano:
>     Do we want to try to do the slow old names search in the mean time? We could delay constructing the PPE?

Considered this but basically trying to avoid the history search wherever possible, unless the old names are particularly meaningful. Metadata doesn't often have meaningful name anyway. We could revisit this later.

>    Do we want to update links as part of update propagation? (I may have missed this in related conversation.)

Maybe not in all cases and I'd say it needs more thought. Considered out of scope for this PR. :)

pchiusano:
Bouncing to retrigger travis.

Issue #1298, opened 2020-03-02
rethink absolute/relative/suffixified/non-suffixified/hash-qualified


aryairani:
I'm wondering, can we get things more correct by doing transformations (like suffixifying, relativizing) later than we do; passing in the `currentPath` at the last minute?

Or maybe something different.

This could improve or close a bunch of output issues:
- #1284 diff after delete shows external aliases
- #1273 update uses non-suffixified ppe
- #1272 pattern matching requires full qualification, construction does not
- #1264 update shows unnecessarily long names for types
- #1241 / #1071 add suffixed-based name resolution for `view`/`display`
- #957 suggested command in Tip about name conflict after `pull` misses necessary namespace prefix
- #702 undo shows incorrectly relativized names
- #978 rethink PPE to simplify its usage when dealing with suffixified / unsuffixified names

Design may be informed by:
- https://github.com/unisonweb/unison/issues/1080#issuecomment-567071035 incorporate local names from file into the suffixification calculation
- #1193 / #977 bias names in output according to input
- #1130 show both names when viewing multiple aliases for the same definition
- #1161 namespace.diff ppe missing types needed to render the deletion of those types
- #1140 add a syntax to hash-qualify patches
- #553 let `add`/`update` support new definitions with absolute names

PR #1371, opened 2020-03-19
Allow `view` and `display` commands to resolve any unique suffix

pchiusano:
## Overview

`view List.map` now works, regardless of where in the tree you are. Likewise for `display myMagicConstant`. This isn't a comprehensive rethink as suggested in #1298 but was a simple change and a nice quality of life improvement IMO.

Here's `view |>` in action - 

Screen Shot 2020-03-19 at 3 04 49 PM

## Implementation notes

This was very straightforward.

I left autocomplete alone but switched to fuzzy matching for `exactDefinitionQueryArg`. This has the effect that `List.zip` will autocomplete to `base.List.zip` - that is, the completions still show the full names, which seemed better to me.

This approach of leaving autocomplete alone means existing commands aren't required to accept suffixes, which gives me more confidence this won't break anything. :)

I discovered that you have to prepend `.` to get autocomplete to look outside the current namespace. I didn't change that behavior. So if you do `.zoink> view List.<tab>`, it won't complete all the `List` functions, but if you do `.zoink> view .List.<tab>` then it will. This seemed OK to me but we might have more of an opinion on this after spending some time with it.

## Test coverage

I added a couple tests to the suffixes.md transcript, exercising `view` and `display`. Existing transcripts still pass. There was one transcript that was relying on `view` NOT looking outside the namespace (to detect a deletion), which I adjusted.

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
+++ b/parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
@@ -258,19 +258,19 @@ patch = InputPattern
   )
 
 view :: InputPattern
-view = InputPattern "view" [] [(OnePlus, exactDefinitionQueryArg)]
+view = InputPattern "view" [] [(OnePlus, definitionQueryArg)]
       "`view foo` prints the definition of `foo`."
       (pure . Input.ShowDefinitionI Input.ConsoleLocation)
 
 display :: InputPattern
-display = InputPattern "display" [] [(Required, exactDefinitionQueryArg)]
+display = InputPattern "display" [] [(Required, definitionQueryArg)]
       "`display foo` prints a rendered version of the term `foo`."
       (\case
         [s] -> pure (Input.DisplayI Input.ConsoleLocation s)
         _ -> Left (I.help display))
 
 displayTo :: InputPattern
-displayTo = InputPattern "display.to" [] [(Required, noCompletions), (Required, exactDefinitionQueryArg)]
+displayTo = InputPattern "display.to" [] [(Required, noCompletions), (Required, definitionQueryArg)]
       (P.wrap $ makeExample displayTo ["<filename>", "foo"]
              <> "prints a rendered version of the term `foo` to the given file.")
       (\case
@@ -278,7 +278,7 @@ displayTo = InputPattern "display.to" [] [(Required, noCompletions), (Required, 
         _ -> Left (I.help displayTo))
 
 docs :: InputPattern
-docs = InputPattern "docs" [] [(Required, exactDefinitionQueryArg)]
+docs = InputPattern "docs" [] [(Required, definitionQueryArg)]
       ("`docs foo` shows documentation for the definition `foo`.")
       (\case
         [s] -> first fromString $ Input.DocsI <$> Path.parseHQSplit' s
@@ -291,7 +291,7 @@ undo = InputPattern "undo" [] []
 
 viewByPrefix :: InputPattern
 viewByPrefix
-  = InputPattern "view.recursive" [] [(OnePlus, exactDefinitionQueryArg)]
+  = InputPattern "view.recursive" [] [(OnePlus, definitionQueryArg)]
     "`view.recursive Foo` prints the definitions of `Foo` and `Foo.blah`."
     (pure . Input.ShowDefinitionByPrefixI Input.ConsoleLocation)
 
@@ -381,7 +381,7 @@ renameType = InputPattern "move.type" ["rename.type"]
 
 delete :: InputPattern
 delete = InputPattern "delete" []
-    [(OnePlus, exactDefinitionQueryArg)]
+    [(OnePlus, definitionQueryArg)]
     "`delete foo` removes the term or type name `foo` from the namespace."
     (\case
       [query] -> first fromString $ do
@@ -443,7 +443,7 @@ aliasType = InputPattern "alias.type" []
 
 aliasMany :: InputPattern
 aliasMany = InputPattern "alias.many" ["copy"]
-  [(Required, exactDefinitionQueryArg), (OnePlus, exactDefinitionOrPathArg)]
+  [(Required, definitionQueryArg), (OnePlus, exactDefinitionOrPathArg)]
   (P.group . P.lines $
     [ P.wrap $ P.group (makeExample aliasMany ["<relative1>", "[relative2...]", "<namespace>"])
       <> "creates aliases `relative1`, `relative2`, ... in the namespace `namespace`."
@@ -804,8 +804,8 @@ replaceEdit f s = self
   self = InputPattern
     ("replace." <> s)
     []
-    [ (Required, exactDefinitionQueryArg)
-    , (Required, exactDefinitionQueryArg)
+    [ (Required, definitionQueryArg)
+    , (Required, definitionQueryArg)
     , (Optional, patchArg)
     ]
     (P.wrapColumn2
@@ -855,7 +855,7 @@ edit :: InputPattern
 edit = InputPattern
   "edit"
   []
-  [(OnePlus, exactDefinitionQueryArg)]
+  [(OnePlus, definitionQueryArg)]
   (  "`edit foo` prepends the definition of `foo` to the top of the most "
   <> "recently saved file."
   )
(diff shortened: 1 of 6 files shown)


parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -1212,8 +1212,8 @@ exactDefinitionOrPathArg =
 exactDefinitionQueryArg :: ArgumentType
 exactDefinitionQueryArg =
   ArgumentType "definition query" $
-    bothCompletors (termCompletor exactComplete)
-                   (typeCompletor exactComplete)
+    bothCompletors (termCompletor fuzzyComplete)
+                   (typeCompletor fuzzyComplete)
```
aryairani: Changing `exactDefinitionQueryArg` from `exact` to `fuzzy` seems suspect :)

maybe a `suffixComplete` and a `exactSuffixQueryArg`?

unison-src/transcripts/merge.md
```diff
@@ -89,13 +89,13 @@ and `quux` namespaces.
...
 ```
 
 ```ucm:error
-.P0> view baz.x
+.> view P0.baz.x
 ```
 
 ```ucm:error
-.P0> view quux.x
+.> view P0.quux.x
```
aryairani: Why?

unison-src/transcripts/suffixes.output.md
```diff
@@ -13,31 +13,35 @@ optional.isNone = cases
...
   
-    ⍟ These new definitions are ok to `add`:
-    
-      foo.bar.a       : Int
-      optional.isNone : Optional a -> Boolean
-   
-  Now evaluating any watch expressions (lines starting with
-  `>`)... Ctrl+C cancels.
+    foo.bar.a       : Int
+    optional.isNone : Optional a -> Boolean
```
aryairani: Note to self: This is just the typechecking message being replaced with the `add` output.

unison-src/transcripts/merge.md
pchiusano: To make the suffixes unique and refer to the P0 namespace only. (There’s a foo.w in different namespace at this point in transcript). Mentioned in the writeup but might not have been clear.

aryairani: Sorry, was in a daze and forgot to read it.

aryairani: I have substantially read it now :)

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
aryairani: For context: we already have a `fuzzyDefinitionQueryArg`, which is distinguished from `exactDefinitionQueryArg` by being `fuzzyComplete` instead of `exactComplete`; it had some [forgotten?] undesirable qualities though, which is why we added `exact`.  Something that dealt in suffixes would still be good though!

Not sure if existing suffixify code would work here?

pchiusano: I’m thinking just rename this to ‘definitionQueryArg’. Whether this uses fuzzy or exact or whatever completions more of implementation detail IMO.

FYI fuzzy matching is just substring matching (might be case insensitive). It used to be some bonkers edit distance thing that gave terrible results but we axed that. 😀 

No need for suffix completion imo since a suffix match is also a substring match, and I haven’t yet felt like the substring match is returning too many results. One interesting example - List.map finds base.List.map.doc which seems nice to me. 😀

pchiusano: Suspect the separation was because the fuzzy matching was overly fuzzy at one time, so we didn’t want to use it everywhere. 😀

It’s safe to have the completion provider do something sloppy as long as the thing it completes to is a subset of what the command can accept. It’s more that you don’t want to be shown a gajillion completions that aren’t very good. I think the substring matching is pretty good balance.

aryairani: Agree we don't want completions that aren't very good; we do want auto-complete to let the user make progress to their desired identifier as quickly as possible.  I'm thinking it's going to do the former though.

Could you do an audit of what was using fuzzy match (if anything) and what was using exact match, and make an inventory (here I guess) of what's going to be affected by the change?  Or make a smaller change and hold off on the comprehensive analysis for now.

aryairani: But eventually (or now) we should make an taxonomy of the types and purposes of arguments, and have a consistent way of decoding and auto-completing them.  Within the implementation for each identified argument type, we can have changeable implementation details! /cc @runarorama

pchiusano: Okie dokie, did that renaming to `definitionQueryArg` and you can peep [this diff](https://github.com/unisonweb/unison/pull/1371/commits/cda53566b51366a931203e983d5401e51f84e18d) to see the commands whose completions will be affected by this PR. I think these are all very reasonable spots to do fuzzier completion.

> It’s safe to have the completion provider do something sloppy as long as the thing it completes to is a subset of what the command can accept.

This is important to understand, can you ack? Basically, "be liberal in what you autocomplete, but strict in what you accept" Example is if `names List.map<tab>` provides completions of `base.List.map` and `base.List.map.doc` - the user still ultimately has to pick which one they want, they are just shown more options than the exact (really prefix) matching.

I basically don't think prefix-only completion is super useful, for this reason (exception is maybe `ls` and `cd` commands). Something we can consider for later is doing a sort / cutoff if there's too many results, and then we can develop some interesting criteria for doing the sorting. Like if we really think prefix matches should come before others, we could do that.

aryairani: > It’s safe to have the completion provider do something sloppy as long as the thing it completes to is a subset of what the command can accept.

and
 
> Basically, "be liberal in what you autocomplete, but strict in what you accept"

I suppose there's a spectrum for what this could mean, but at face value, I disagree — too many sloppy completions can make an auto-complete feature worse than no auto-complete feature. e.g. You can lose time looking through completions that aren't what you want, before giving up; all cost, no benefit. So I think it's important to be thoughtful about how to give a good set of completions.

I'm not implying that you haven't already been thoughtful and have reached a great solution (and I myself definitely haven't), just that it's not clear from the PR description and diff alone!  Agree that the completions you suggested in the `List.map` example sound good.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -381,7 +381,7 @@ renameType = InputPattern "move.type" ["rename.type"]
 
 delete :: InputPattern
 delete = InputPattern "delete" []
-    [(OnePlus, exactDefinitionQueryArg)]
+    [(OnePlus, definitionQueryArg)]
```
aryairani: Could you confirm this one?

pchiusano: Yes confirmed.

aryairani:
Or is it that you would see the completions, and then you'd backspace-out `List.map` and type `base.List.map`?

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
pchiusano: Yeah I think we agree that if autocomplete shows mostly nonsense that’s worse. 😀 Basically I think we are on the same page. 

Autocomplete provider is easy to futz with later, too so we can easily revisit the matching later once we have more experience. IMO it’s a big improvement.

Are you ok with closing this out for now?

aryairani: Do you have a sense of whether this will create a feeling of inconsistency relative to whatever arg. types are still using `exactComplete`?  Or are those ones that should stay as-is?

Anyway, we can merge this and see what happens.

pchiusano: Created #1377 to track taking a look at other command args.

pchiusano:
@aryairani and I discussed on phone, but for anyone lurking - 

If there’s a unique completion, it’ll complete it, even if the input is not a prefix of the completion (example is that `List.map<tab>` will become `base.List.map`, assuming that's the only match).  If the completion isn’t unique then you gotta disambiguate by more typing.

Haskeline is really missing a way to cycle through the available completions and it would be great to add that. Maybe `<tab>` once the completion options are up cycles forward, `<shift>+<tab>` cycles back or something.

PR #1372, opened 2020-03-19
Removes empty string at the end of Name.suffixes output and unittests

### LINKED ISSUE #1339: fix Name.suffixes and add a unit test
eg `suffixes “foo.Bar.baz”` should include `[“foo.Bar.baz”, “Bar.baz”, “baz”]` and not include `””`.

pete-ts:
## Overview

Updates Name.suffixes and adds relevant unit tests

* The before behaviour can be seen in the first commit: `suffixes` adds an empty string at the end of the list.
* The after behaviour can be seen in the unit tests: `suffixes` no longer adds the empty string.

fix #1339

## Implementation notes

* Input `""` is unexpected and should return an empty list of names
* Otherwise split, get all tails, remove empty string, then put back together with `.` in the middle

There were no unit tests for `unison-core/`. I have added the tests for `Name` as part of the `tests` binary, but under a `Unison.Core.` namespace, on purpose so future `unison-core` unit tests can be grouped together.

## Test coverage

Covered.

## Loose ends

Not that I am aware of.

### DIFF
--- a/parser-typechecker/tests/Suite.hs
+++ b/parser-typechecker/tests/Suite.hs
@@ -29,6 +29,7 @@ import qualified Unison.Test.Codebase as Codebase
 import qualified Unison.Test.Codebase.FileCodebase as FileCodebase
 import qualified Unison.Test.UriParser as UriParser
 import qualified Unison.Test.Git as Git
+import qualified Unison.Core.Test.Name as Name
 
 test :: Test ()
 test = tests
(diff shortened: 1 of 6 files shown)


aryairani:
Thanks @pete-ts!  To really put a bow on this, could you update `unison-src/transcripts/fix1334.md`?  The current version creates two types, just to get around the `Names.suffixes` bug.  Could you remove `Twoolean` and `==` to show that even with only one type named in the codebase (`alias.type ##Nat Cat`) that [this](https://github.com/unisonweb/unison/pull/1334#issuecomment-597338520) won't happen anymore?

pete-ts:
Will have a look at updating the transcript. :+1:

PR #1378, opened 2020-03-20
Let more places use Reference.Id

aryairani:
- `Codebase` functions are now parameterized by `Reference.Id` or `Referent.Id`
  where appropriate, to
  enforce at the type level that Codebases can't contain builtins, and
  to eliminate jank that was enforcing or assuming this at runtime.  
  (A `Referent.Id` is a `Referent` that doesn't refer to a built-in.)

  Ditto for `CodeLookup`.

  Ditto for `UnisonFile`, `UnisonFile.Env`, and `TypecheckedUnisonFile`; with some backwards-
  compatible functions and pattern synonyms to support the old types,
  and some new functions for the new types.

- Off-topic: `Codebase.getRootBranch` and `.getBranchForHash` now return a sum type instead
  of a magic value (`Branch.empty`) on error. `BranchLoadMode` goes away.

  `getRootBranch`'s result distinguishes between "I don't know what the
  head is supposed to be" vs "I do know what it is supposed to be but
  couldn't find it".

- Even more off-topic (oops), I moved the actual
  builtin `Decl`s from `unison-core/Unison.DataDeclaration` to
  `unison-parser-typechecker/Unison.Builtin.Decls`.

  I let this new module be qualified as `DD` to minimize diffs, although it may
  make it harder to know which module a `DD.` definition is coming from.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -4,7 +4,6 @@ module Unison.Codebase where
 
 import Unison.Prelude
 
-import           Control.Category               ( (>>>) )
 import           Control.Lens                   ( _1, _2, (%=) )
 import           Control.Monad.State            ( State, evalState, get )
 import           Data.Bifunctor                 ( bimap )
@@ -21,7 +20,6 @@ import qualified Unison.Names2                 as Names
 import           Unison.Reference               ( Reference )
 import qualified Unison.Reference              as Reference
 import qualified Unison.Referent as Referent
-import Unison.Referent (Referent, Referent')
 import qualified Unison.Term                   as Term
 import qualified Unison.Type                   as Type
 import           Unison.Typechecker.TypeLookup  (TypeLookup(TypeLookup))
@@ -29,6 +27,7 @@ import qualified Unison.Typechecker.TypeLookup as TL
 import qualified Unison.Parser                 as Parser
 import qualified Unison.UnisonFile             as UF
 import qualified Unison.Util.Relation          as Rel
+import qualified Unison.Util.Set               as Set
 import qualified Unison.Var                    as Var
 import           Unison.Var                     ( Var )
 import qualified Unison.Runtime.IOSource       as IOSource
@@ -38,8 +37,6 @@ import Unison.Term (Term)
 import Unison.Type (Type)
 import Unison.Codebase.ShortBranchHash (ShortBranchHash)
 import Unison.ShortHash (ShortHash)
-
---import Debug.Trace
 
 type DataDeclaration v a = DD.DataDeclaration' v a
 type EffectDeclaration v a = DD.EffectDeclaration' v a
@@ -52,10 +49,10 @@ data Codebase m v a =
            , putTerm            :: Reference.Id -> Term v a -> Type v a -> m ()
            , putTypeDeclaration :: Reference.Id -> Decl v a -> m ()
 
-           , getRootBranch      :: m (Branch m)
+           , getRootBranch      :: m (Either GetRootBranchError (Branch m))
            , putRootBranch      :: Branch m -> m ()
            , rootBranchUpdates  :: m (m (), m (Set Branch.Hash))
-           , getBranchForHash   :: Branch.Hash -> m (Branch m)
+           , getBranchForHash   :: Branch.Hash -> m (Maybe (Branch m))
 
            , dependentsImpl     :: Reference -> m (Set Reference.Id)
            -- This copies all codebase elements (except _head) from the
@@ -77,18 +74,20 @@ data Codebase m v a =
            , appendReflog       :: Text -> Branch m -> Branch m -> m ()
 
            -- list of terms of the given type
-           , termsOfTypeImpl    :: Reference -> m (Set Referent)
+           , termsOfTypeImpl    :: Reference -> m (Set Referent.Id)
            -- list of terms that mention the given type anywhere in their signature
-           , termsMentioningTypeImpl :: Reference -> m (Set Referent)
+           , termsMentioningTypeImpl :: Reference -> m (Set Referent.Id)
            -- number of base58 characters needed to distinguish any two references in the codebase
            , hashLength         :: m Int
            , termReferencesByPrefix :: ShortHash -> m (Set Reference.Id)
            , typeReferencesByPrefix :: ShortHash -> m (Set Reference.Id)
-           , termReferentsByPrefix :: ShortHash -> m (Set (Referent' Reference.Id))
+           , termReferentsByPrefix :: ShortHash -> m (Set Referent.Id)
 
            , branchHashLength   :: m Int
            , branchHashesByPrefix :: ShortBranchHash -> m (Set Branch.Hash)
            }
+
+data GetRootBranchError = NoRootBranch | CouldntLoadRootBranch Branch.Hash
 
 bootstrapNames :: Names.Names0
 bootstrapNames =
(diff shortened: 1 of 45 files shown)


parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -160,31 +156,26 @@ typeLookupForDependencies codebase = foldM go mempty
...
+        foldM (transitiveDependencies code) seen (getIds $ Term.dependencies t)
+      Nothing -> 
+        CL.getTypeDeclaration code rid >>= \case 
+          Nothing        -> pure seen
+          Just (Left ed) -> foldM (transitiveDependencies code)
+                                  seen
+                                  (getIds $ DD.dependencies (DD.toDataDecl ed))
+          Just (Right dd) -> foldM (transitiveDependencies code)
+                                   seen
+                                   (getIds $ DD.dependencies dd)
```
aryairani: This is mostly whitespace diff, apart from the calls to `getIds`.  Could have folded those into new methods in the `Term` and `DataDeclaration` modules?

parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -198,47 +189,59 @@ makeSelfContained'
   -> UF.UnisonFile v a
   -> m (UF.UnisonFile v a)
 makeSelfContained' code uf = do
-  let UF.UnisonFile ds0 es0 bs0 ws0 = uf
-      deps0 = Term.dependencies . snd <$> (UF.allWatches uf <> bs0)
+  let UF.UnisonFileId ds0 es0 bs0 ws0 = uf
+      deps0 = getIds . Term.dependencies . snd <$> (UF.allWatches uf <> bs0)
+        where getIds = Set.mapMaybe Reference.toId
```
aryairani: Hmm, another case for `Term.dependenciesId` 🤔

parser-typechecker/src/Unison/Codebase/BranchUtil.hs
```diff
@@ -25,6 +25,17 @@ import qualified Unison.Codebase.Metadata as Metadata
...
+
+fromNames0 :: Monad m => Names0 -> Branch m
+fromNames0 names0 = Branch.one $ addFromNames0 names0 Branch.empty0
+
+-- can produce a pure value because there's no history to traverse
+hashesFromNames0 :: Monad m => Names0 -> Map Branch.Hash (Branch m)
+hashesFromNames0 = deepHashes . fromNames0 where 
+  deepHashes :: Branch m -> Map Branch.Hash (Branch m)
+  deepHashes b = Map.singleton (Branch.headHash b) b 
+    <> (foldMap deepHashes . view Branch.children . Branch.head) b
```
aryairani: Complementary to `fromNames0`, `hashesFromNames0` returns a map of the hashes and branches implied by the `Names0`.

Not as universally applicable as `Branch m -> m (Map Branch.Hash (Branch m))`, but I chose it because it could be pure, by taking advantage of the fact that `fromNames0` does not produce history.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -157,18 +158,12 @@ initCodebase dir = do
...
-      let errMsg = P.lines [ "No codebase exists in " <> P.string dir
-                           , "Run `ucm init` to create one, then try again!" ]
-      pure (dir, errMsg)
-
+  dir <- getCodebaseDir mdir
+  prettyDir <- P.string <$> canonicalizePath dir
+  let errMsg = P.lines
+        [ "No codebase exists in " <> prettyDir
+        , "Run `ucm -codebase " <> prettyDir
+          <> " init` to create one, then try again!"]
```
aryairani: Refactor to make use of the existing `getCodebaseDir` which already looks up the user's home directory.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -328,25 +323,28 @@ branchFromFiles loadMode rootDir h@(RawHash h') = do
...
+    (pure $ Left Codebase.NoRootBranch)
  where
+  go' :: [String] -> m (Either Codebase.GetRootBranchError (Branch m))
+  go' = \case
+              []       -> pure $ Left Codebase.NoRootBranch
+              [single] -> runExceptT $ go single
+              conflict -> runExceptT (traverse go conflict) >>= \case
+                Right (x : xs) -> Right <$> foldM Branch.merge x xs
+                Right _ -> error "FileCodebase.getRootBranch.conflict can't be empty."
+                Left e -> Left <$> pure e
```
aryairani: Gosh, why is this so indented.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -451,7 +455,7 @@ syncToDirectory fmtV fmtA codebase localPath branch = do
   b <- (liftIO . exists) localPath
   if b then do
     let code = codebase1 fmtV fmtA localPath
-    remoteRoot <- Codebase.getRootBranch code
+    remoteRoot <- fromMaybe Branch.empty . rightMay <$> Codebase.getRootBranch code
```
aryairani: Instead of `BranchLoadMode` with a failing option, we convert failure to empty at the call site.

parser-typechecker/src/Unison/Codebase/Runtime.hs
```diff
@@ -64,7 +65,8 @@ evaluateWatches
 evaluateWatches code ppe evaluationCache rt uf = do
   -- 1. compute hashes for everything in the file
   let m :: Map v (Reference, Term.Term v a)
-      m = Term.hashComponents (Map.fromList (UF.terms uf <> UF.allWatches uf))
+      m = first Reference.DerivedId <$>
+            Term.hashComponents (Map.fromList (UF.terms uf <> UF.allWatches uf))
```
aryairani: Minimal, superficial changes to this module; just need it to typecheck.

parser-typechecker/src/Unison/Codebase/ShortBranchHash.hs
```diff
@@ -21,6 +21,9 @@ fromHash :: Int -> Branch.Hash -> ShortBranchHash
 fromHash len =
   ShortBranchHash . Text.take len . Hash.base32Hex . Causal.unRawHash
 
+fullFromHash :: Branch.Hash -> ShortBranchHash
+fullFromHash = ShortBranchHash . Hash.base32Hex . Causal.unRawHash
```
aryairani: `fromHash` requires a length; this one doesn't.  Analogous to `Reference.toShortHash`, but `Branch.Hash` doesn't have its own module.

parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
```diff
@@ -222,10 +224,10 @@ run dir configFile stanzas codebase = do
       loadPreviousUnisonBlock name = do
         ufs <- readIORef unisonFiles
         case Map.lookup name ufs of
-          Just uf -> do
-            return $ LoadSuccess uf
+          Just uf ->
+            return (LoadSuccess uf)
           Nothing ->
-            return $ InvalidSourceNameError
+            return InvalidSourceNameError
```
aryairani: Didn't mean to change these, but this is where we are.

parser-typechecker/src/Unison/UnisonFile.hs
```diff
@@ -170,13 +204,17 @@ dependencies' file = let
 -- Returns the dependencies of the `UnisonFile` input. Needed so we can
 -- load information about these dependencies before starting typechecking.
 dependencies :: (Monoid a, Var v) => UnisonFile v a -> Set Reference
-dependencies uf = Term.dependencies (typecheckingTerm uf)
+dependencies (UnisonFile ds es ts ws) =
+  foldMap (DD.dependencies . snd) ds
+  <> foldMap (DD.dependencies . DD.toDataDecl . snd) es
+  <> foldMap (Term.dependencies . snd) ts
+  <> foldMap (foldMap (Term.dependencies . snd)) ws
```
aryairani: Inlined the relevant parts of `typecheckingTerm`, so if it were to go away...

parser-typechecker/src/Unison/UnisonFile.hs
```diff
@@ -244,21 +286,27 @@ bindNames names (UnisonFile d e ts ws) = do
...
+  , effectsId :: Map v (Reference.Id, EffectDeclaration' v a)
   -- Naming environment
   , names   :: Names0
 }
 
+datas :: Env v a -> Map v (Reference, DataDeclaration' v a)
+datas = fmap (first Reference.DerivedId) . datasId
+
+effects :: Env v a -> Map v (Reference, EffectDeclaration' v a)
+effects = fmap (first Reference.DerivedId) . effectsId
```
aryairani: Source-compatible changes to `UnisonFile.Env`

unison-core/src/Unison/DataDeclaration.hs
```diff
@@ -391,216 +395,17 @@ hashDecls0 decls =
...
+      bindTypes dd = dd { constructors' = over _3 (Type.bindExternal varToRef') <$> constructors' dd }
       typeNames0 = Names.names0 mempty
-                 $ Rel.fromList (first Name.fromVar <$> varToRef)
+                 $ Rel.fromList (first Name.fromVar <$> varToRef')
       -- normalize the order of the constructors based on a hash of their types
       sortCtors dd = dd { constructors' = sortOn hash3 $ constructors' dd }
       hash3 (_, _, typ) = ABT.hash typ :: Hash
   decls' <- fmap sortCtors <$> traverse (bindNames mempty typeNames0) decls'
   pure  [ (v, r, dd) | (v, r) <- varToRef, Just dd <- [Map.lookup v decls'] ]
-
```
aryairani: Moved to `Unison.Builtin.Decls` module

unison-core/src/Unison/Term.hs
```diff
@@ -470,6 +470,9 @@ var' = var() . Var.named
 ref :: Ord v => a -> Reference -> Term2 vt at ap v a
 ref a r = ABT.tm' a (Ref r)
 
+refId :: Ord v => a -> Reference.Id -> Term2 vt at ap v a
+refId a = ref a . Reference.DerivedId
+
```
aryairani: Opted to create some new functions instead of modifying the existing ones.

unison-core/src/Unison/DataDeclaration.hs
aryairani: (This comment refers to the hundreds of lines below this one; I didn't move just a single blank line to a different module.)

PR #1380, opened 2020-03-22
remove all trailing whitespace

aryairani:
Get trailing whitespace out of the way in the hopes that it stops showing up in other PRs.


pchiusano:
Everyone check their editor settings

aryairani:
> Everyone check their editor settings

My editor is broken, so I may be part of the problem. 😞 https://youtrack.jetbrains.com/issue/IDEA-228850

We could add it to git commit hooks or CI?

PR #1383, opened 2020-03-26
easytest.pending runs hidden tests and reports accordingly

### LINKED ISSUE #928: Fail if a pending test actually passes
E.g.

```haskell
pending ok
```

would be a failing test.

### Motivation:
1. When a PR is opened that adds a pending test, then a passing CI build will be sufficient evidence that the added test is actually failing.
2. If a previously failing test becomes passing incidentally as a side-effect of some change, it will let the author know that the test is now passing.

pete-ts:
## Overview

This PR aims to change the behaviour of the `pending` test wrapper.
When using `pending` from now on:
* the test being wrapped will actually be run, with logging switched off, and not showing any output
* If the inner test fails, `pending` reports a success
* If the inner test passes, this is unexpected, and `pending` reports a failure

fix #928 

## Implementation notes

I have implemented "switch off logging" and "run hidden in a new scope" as two functions, for the sake of granularity.

I have allowed a commented out `pending` test that produces an `inner success`
When uncommented this will produce the following output:

```
🦄  pending.failure
pending.success FAILURE This pending test should not pass! CallStack (from HasCallStack):
  crash, called at src/EasyTest.hs:377:7 in easytest-0.1-GHUNtMcGbspKNzFvNGkuQU:EasyTest
  pending, called at tests/Suite.hs:18:30 in main:Main
💥  pending.success


  1 passed
    "pending.success"

    EasyTest.rerun 3851235856694737643
    EasyTest.rerunOnly 3851235856694737643 "pending.success"
```

## Test coverage

Yes

## Loose ends
Maybe remove `Pending` from `Status` altogether?

### DIFF
--- a/yaks/easytest/src/EasyTest.hs
+++ b/yaks/easytest/src/EasyTest.hs
@@ -352,9 +352,30 @@ crash msg = do
       msg' = msg ++ " " ++ prettyCallStack trace
   Test (Just <$> putResult Failed) >> noteScoped ("FAILURE " ++ msg') >> Test (pure Nothing)
 
--- skips the test but makes a note of this fact
-pending :: Test a -> Test a
-pending _ = Test (Nothing <$ putResult Pending)
+-- | Overwrites the env so that note_ (the logger) is a no op
+nologging :: HasCallStack => Test a -> Test a
+nologging (Test t) = Test $ do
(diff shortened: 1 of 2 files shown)


pchiusano:
Wow, cool!

PR #1387, opened 2020-03-28
four git ux fixes

aryairani:
## Overview

Check out this screencast, or scrutinize the still image:
[asciicast](https://asciinema.org/a/EMV7jGA1nCfvV0QIN1V76PqsW)
(Too bad it can't play inline from a GH issue!)

**From here, you can probably skip over to my self-review.**

---
### 1. Fix an issue preventing use of local, relative-path git repos.

`push repo1.git bar` _should_ work, but ended up resulting in an error like:
```
fatal: 'repo1.git' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

  I couldn't access a git repository at repo1.git. Make sure the repo exists and that you
  have access to it.
```

The only workaround was to specify an absolute path:
```
.> push /Users/arya/unison7/repo1.git bar
```
which, in addition to being annoying, made it a bit harder to create bug reports that involve git repos — a bug-fixer would have to adjust the absolute paths for their machine.

This was occurring because `git` was resolving the relative path _relative to_ a random temp directory, rather than ucm's current path.

This PR uses `--git-dir` and `--work-tree` to indicate where `git` should work, instead of having it change directories (`git -C`).

### 2. Avoid the output from `git rev-parse --git-dir`.

The `git` docs sort of make it sound like you must pass `--git-dir` if you want `git rev-parse` to exit with an error when it can't find a local `.git` dir, but the primary effect of passing `--git-dir` is that git _prints_ the git dir it finds to stdout.

But it turns out we don't need to pass `--git-dir` to get the exit codes we want, and we don't want to print the git dir anyway, so I removed the flag.

### 3. Eliminate Shellmet output.

With an updated Shellmet, I could turn off the printing of git commands.

### 4. Added some progress messages.

Well, without the git output, it pretty much just looked like ucm was frozen. So I added some status messages to keep you hopeful.

## Test coverage

`push` / `pull` could use some better regression tests in general; this PR doesn't change that.

## Loose ends

At the end of the screencast, you see these annoying warnings from `git` about stuff I know about and don't need to see:
```
.prs.puffnfresh> push repo.git merged
warning: --depth is ignored in local clones; use file:// instead.
warning: You appear to have cloned an empty repository.
warning: --depth is ignored in local clones; use file:// instead.
warning: You appear to have cloned an empty repository.
                                          
  Done.
```

One way to hide these would be to redirect stderr.

Another way would be to not pass `--depth 1` for local repos, but the function that passes `--depth 1` doesn't know how to parse the url to know if it's a local repo.  https://github.com/unisonweb/unison/compare/713a746...topic/more-remoterepo-structure does the first part of bringing this structure forward.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/Git.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Git.hs
@@ -35,6 +35,18 @@ import qualified Unison.Util.Exception         as Ex
 import qualified Unison.Codebase.Branch        as Branch
 import qualified Unison.Names3                 as Names
 import Unison.Codebase.ShortBranchHash (ShortBranchHash)
+import UnliftIO.IO (hFlush, stdout)
+
+withStatus :: MonadIO m => String -> m a -> m a
+withStatus str ma = do
+  flushStr str
+  a <- ma
+  flushStr (const ' ' <$> str)
+  pure a
+  where
+  flushStr str = do
+    liftIO . putStr $ "  " ++ str ++ "\r"
+    hFlush stdout
 
 -- Given a local path, a remote git repo url, and branch/commit hash,
 -- checks for git, the repo path, performs a clone, and verifies resulting repo
@@ -48,10 +60,11 @@ pullBranch localPath uri treeish = do
   case e of
     Left e -> throwError (SomeOtherError (Text.pack (show e)))
     Right _ -> pure ()
-  "git" (["clone", "--quiet"] ++ ["--depth", "1"]
-     ++ maybe [] (\t -> ["--branch", t]) treeish
-     ++ [uri, Text.pack localPath])
-    `onError` throwError (NoRemoteRepoAt uri)
+  withStatus ("Downloading from " ++ Text.unpack uri ++ " ...")
+    ("git" $^ (["clone", "--quiet"] ++ ["--depth", "1"]
+       ++ maybe [] (\t -> ["--branch", t]) treeish
+       ++ [uri, Text.pack localPath]))
+      `onError` throwError (NoRemoteRepoAt uri)
   isGitDir <- liftIO $ checkGitDir localPath
   unless isGitDir . throwError $ NoLocalRepoAt localPath
 
@@ -96,7 +109,8 @@ pullGitBranch localPath codebase url treeish sbh = do
           Just b -> pure b
           Nothing -> throwError $ NoRemoteNamespaceWithHash url treeish sbh
         _ -> throwError $ RemoteNamespaceHashAmbiguous url treeish sbh branchCompletions
-  lift $ Codebase.syncFromDirectory codebase gitCodebasePath
+  withStatus "Importing downloaded files into local codebase..." $
+    lift $ Codebase.syncFromDirectory codebase gitCodebasePath
   pure branch
   where gitCodebasePath = localPath </> codebasePath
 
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Codebase/Editor/Git.hs
```diff
@@ -48,10 +60,11 @@ pullBranch localPath uri treeish = do
...
     Right _ -> pure ()
-  "git" (["clone", "--quiet"] ++ ["--depth", "1"]
-     ++ maybe [] (\t -> ["--branch", t]) treeish
-     ++ [uri, Text.pack localPath])
-    `onError` throwError (NoRemoteRepoAt uri)
+  withStatus ("Downloading from " ++ Text.unpack uri ++ " ...")
+    ("git" $^ (["clone", "--quiet"] ++ ["--depth", "1"]
+       ++ maybe [] (\t -> ["--branch", t]) treeish
+       ++ [uri, Text.pack localPath]))
+      `onError` throwError (NoRemoteRepoAt uri)
```
aryairani: The `withStatus` line was added, and the rest was indented into it.
The `$^` was inserted between "git" and "clone" to suppress printing the command.

parser-typechecker/src/Unison/Codebase/Editor/Git.hs
```diff
@@ -96,7 +109,8 @@ pullGitBranch localPath codebase url treeish sbh = do
           Just b -> pure b
           Nothing -> throwError $ NoRemoteNamespaceWithHash url treeish sbh
         _ -> throwError $ RemoteNamespaceHashAmbiguous url treeish sbh branchCompletions
-  lift $ Codebase.syncFromDirectory codebase gitCodebasePath
+  withStatus "Importing downloaded files into local codebase..." $
+    lift $ Codebase.syncFromDirectory codebase gitCodebasePath
```
aryairani: Wrap & indent.

parser-typechecker/src/Unison/Codebase/Editor/Git.hs
```diff
@@ -107,18 +121,23 @@ checkForGit = do
 
 checkGitDir :: FilePath -> IO Bool
 checkGitDir dir =
-  (const True <$> gitIn dir ["rev-parse", "--git-dir"]) $? pure False
+  (const True <$> gitIn dir ["rev-parse"]) $? pure False
```
aryairani: `--git-dir` didn't seem to change the behavior except to add unwanted output.

parser-typechecker/src/Unison/Codebase/Editor/Git.hs
```diff
@@ -136,7 +155,9 @@ pushGitRootBranch localPath codebase branch url gitbranch = do
   -- Clone and pull the remote repo
   pullBranch localPath url gitbranch
   -- Stick our changes in the checked-out copy
-  merged <- lift $ syncToDirectory codebase (localPath </> codebasePath) branch
+  merged <-
+    withStatus ("Staging files for upload to " ++ Text.unpack url ++ " ...") $
+      lift $ syncToDirectory codebase (localPath </> codebasePath) branch
```
aryairani: Wrap & indent.

parser-typechecker/src/Unison/Codebase/Editor/Git.hs
```diff
@@ -155,5 +176,5 @@ pushGitRootBranch localPath codebase branch url gitbranch = do
         case gitbranch of
           Nothing        -> gitIn localPath ["push", "--quiet", "--all", url]
           Just gitbranch -> gitIn localPath ["push", "--quiet", url, gitbranch]
-  liftIO push `onException` throwError (NoRemoteRepoAt url)
-
+  withStatus ("Uploading to " ++ Text.unpack url ++ " ...") $
+    liftIO push `onException` throwError (NoRemoteRepoAt url)
```
aryairani: Lol @ this diff.  Wrap & indent.

stack.yaml
```diff
@@ -16,7 +16,7 @@ resolver: lts-14.7
 extra-deps:
 - base58-bytestring-0.1.0
 - relation-0.2.1
-- shellmet-0.0.1
+- shellmet-0.0.3.0
```
aryairani: 🎉

Issue #1277, opened 2020-02-26
Possible variable capture in ABT implementation


dolio:
This is the rename function for ABTs:

```
rename :: (Foldable f, Functor f, Var v) => v -> v -> Term f v a -> Term f v a
rename old new t0@(Term _ ann t) = case t of
  Var v -> if v == old then annotatedVar ann new else t0
  Cycle body -> cycle' ann (rename old new body)
  Abs v body -> if v == old then abs' ann v body
                else abs' ann v (rename old new body)
  Tm v -> tm' ann (fmap (rename old new) v)
```
In the `Abs` case, if `v == new`, then the new variable may be captured. Further the place that calls into this (`subst'`) doesn't check that the new variable is disjoint from all _bound_ variables in the term being substituted into.

I'll try to get a test case in to demonstrate before fixing.

PR #1389, opened 2020-03-31
Fix issue with `putSymbol` which resulted in extra numbers getting appended to the name on each round trip

pchiusano:
See self review.

After fixing this, some serialization round trip tests started failing. This ended up being due to the Eq instance for ABT being busted due to its use of rename, which has a subtle variable capture bug that @dolio spotted a while ago https://github.com/unisonweb/unison/issues/1277 #1277 needs a proper fix but in the meantime, I've tweaked the Eq instance to be a bit more efficient and also avoid the #1277 bug in this particular case.

## Test coverage

All the existing typechecker tests get a serialization round trip, so we have some confidence that it hasn't goofed anything up. 

I also added a minimal failing test against current master. It just checks that `putSymbol` and `getSymbol` round trip successfully for symbols nonzero fresh ids.

## Loose ends

Not sure what we do about existing repositories. 

We could do a cleanup pass, or I think people can fix weird type signatures manually by deleting and re-adding the term.

### DIFF
--- a/parser-typechecker/tests/Unison/Test/ABT.hs
+++ b/parser-typechecker/tests/Unison/Test/ABT.hs
@@ -7,6 +7,8 @@ import EasyTest
 import Unison.ABT as ABT
 import Unison.Symbol (Symbol(..))
 import Unison.Var as Var
+import           Unison.Codebase.Serialization    ( getFromBytes, putBytes )
+import qualified Unison.Codebase.Serialization.V1 as V1
 
 test :: Test ()
 test = scope "abt" $ tests [
@@ -20,5 +22,11 @@ test = scope "abt" $ tests [
     in tests
       [ scope "first"  $ expect (not $ Set.member fresh (ABT.freeVars t1))
       , scope "second" $ expect (not $ Set.member fresh (ABT.freeVars t2))
-      ]
+      ],
+  -- confirmation of fix for https://github.com/unisonweb/unison/issues/1388
+  -- where symbols with nonzero freshIds did not round trip
+  scope "putSymbol" $ let
+    v = Symbol 10 (User "hi")
+    v' = getFromBytes V1.getSymbol (putBytes V1.putSymbol v)
+    in expectEqual (Just v) v'
   ]
--- a/parser-typechecker/tests/Unison/Test/ABT.hs
+++ b/parser-typechecker/tests/Unison/Test/ABT.hs
@@ -14,8 +14,6 @@ test :: Test ()
 test = scope "abt" $ tests [
   scope "freshInBoth" $
     let
-      symbol i n = Symbol i (Var.User n)
-      var i n = ABT.var $ symbol i n
       t1 = var 1 "a"
       t2 = var 0 "a"
       fresh = ABT.freshInBoth t1 t2 $ symbol 0 "a"
@@ -23,6 +21,17 @@ test = scope "abt" $ tests [
       [ scope "first"  $ expect (not $ Set.member fresh (ABT.freeVars t1))
       , scope "second" $ expect (not $ Set.member fresh (ABT.freeVars t2))
       ],
+  scope "rename" $ do
+    -- rename x to a in \a  -> [a, x] should yield
+    --                  \a1 -> [a1, a]
+    let t1 = ABT.abs (symbol 0 "a") (ABT.tm [var 0 "a", var 0 "x"])
+        t2 = ABT.rename (symbol 0 "x") (symbol 0 "a") t1
+        fvs = toList . ABT.freeVars $ t2
+    -- make sure the variable wasn't captured
+    expectEqual fvs [symbol 0 "a"]
+    -- make sure the resulting term is alpha equiv to \a1 -> [a1, a]
+    expectEqual t2 (ABT.abs (symbol 0 "b") (ABT.tm [var 0 "b", var 0 "a"])),
+
   -- confirmation of fix for https://github.com/unisonweb/unison/issues/1388
   -- where symbols with nonzero freshIds did not round trip
   scope "putSymbol" $ let
@@ -30,3 +39,6 @@ test = scope "abt" $ tests [
     v' = getFromBytes V1.getSymbol (putBytes V1.putSymbol v)
     in expectEqual (Just v) v'
   ]
+  where
+    symbol i n = Symbol i (Var.User n)
+    var i n = ABT.var $ symbol i n
--- a/unison-core/src/Unison/ABT.hs
+++ b/unison-core/src/Unison/ABT.hs
@@ -244,12 +244,25 @@ into' a abt = case abt of
 
 -- | renames `old` to `new` in the given term, ignoring subtrees that bind `old`
 rename :: (Foldable f, Functor f, Var v) => v -> v -> Term f v a -> Term f v a
-rename old new t0@(Term _ ann t) = case t of
-  Var v -> if v == old then annotatedVar ann new else t0
-  Cycle body -> cycle' ann (rename old new body)
-  Abs v body -> if v == old then abs' ann v body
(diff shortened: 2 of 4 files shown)


unison-core/src/Unison/ABT.hs
```diff
@@ -528,14 +528,13 @@ find' :: (Ord v, Foldable f, Functor f)
...
-  -- alpha equivalence, works by renaming any aligned Abs ctors to use a common fresh variable
+  -- alpha equivalence, works by renaming any aligned Abs ctors to use a common variable
   t1 == t2 = go (out t1) (out t2) where
     go (Var v) (Var v2) | v == v2 = True
     go (Cycle t1) (Cycle t2) = t1 == t2
     go (Abs v1 body1) (Abs v2 body2) =
       if v1 == v2 then body1 == body2
-      else let v3 = freshInBoth body1 body2 v1
-           in rename v1 v3 body1 == rename v2 v3 body2
+      else rename v1 v2 body1 == body2
```
pchiusano, line 537: @dolio I'm not sure why the old code felt the need to rename both variables, this seems fine to just rename one to  match the other, right?

unison-core/src/Unison/Var.hs
```diff
@@ -37,22 +37,25 @@ named n = typed (User n)
...
+  Inference Other -> "𝕩"
+  Inference PatternPureE -> "𝕞"
+  Inference PatternPureV -> "𝕧"
+  Inference PatternBindE -> "𝕞"
+  Inference PatternBindV -> "𝕧"
+  Inference TypeConstructor -> "𝕗"
+  Inference TypeConstructorArg -> "𝕦"
+  MissingResult -> "_"
+  Blank -> "_"
+  UnnamedWatch k guid -> fromString k <> "." <> guid
```
pchiusano, line 55: Factored out the "create a name from a `Var.Type`" from the logic that appends the id.

aryairani:
@pchiusano Do you think it would be visible in existing codebases?  Fwiw, at least all the ones I've seen so far are stripped away before displaying.  Maybe the synthetic variables could be cleaned up anyway, though.

Do user variables ever have nonzero ids?

pchiusano:
@aryairani yes, it could show up if the user has written some code, allowed Unison to infer the type, and there are multiple variables of the same `Var.Type` in the resulting signature (such that some of the ids are nonzero). So, not that common, but it can happen.

aryairani:
@pchiusano So I'm thinking that initially, they are `Inference` variables if they are inferred, right?  Although once they make a round trip through the serializer, they are `User` variables. 😕

pchiusano:
@aryairani okay, I added a nice little unit test confirming the fix, also verified that it fails in master without the fix. I think this is good to merge. 

FYI it was a deliberate decision to discard the variable type when writing, basically, didn't want those variable types to be part of the codebase format, they are more for purposes of generating nicer error messages and may change regularly. That said I think the conversion should be happening elsewhere, going to open a ticket for that.

pchiusano:
@dolio could you just take a quick peek at https://github.com/unisonweb/unison/pull/1389/files#diff-65bed18b30ffc9f3afffcb6762b3f370 and if it seems okay LMK or just go ahead and merge.

unison-core/src/Unison/ABT.hs
dolio: I think it's possible that `body1` has free occurrences of `v2` that will cause erroneous results. Like:

    λ v1 -> f v1 v2
    λ v2 -> f v2 v2

pchiusano: Good catch, glad I got your review. I’ll revert and do a proper fix of rename I think. Did you have a failing test of it somewhere?

Might also do the de Bruijn based comparison, using annotateBound.

parser-typechecker/tests/Unison/Test/ABT.hs
```diff
@@ -7,18 +7,38 @@ import EasyTest
...
+  scope "rename" $ do
+    -- rename x to a in \a  -> [a, x] should yield
+    --                  \a1 -> [a1, a]
+    let t1 = ABT.abs (symbol 0 "a") (ABT.tm [var 0 "a", var 0 "x"])
+        t2 = ABT.rename (symbol 0 "x") (symbol 0 "a") t1
+        fvs = toList . ABT.freeVars $ t2
+    -- make sure the variable wasn't captured
+    expectEqual fvs [symbol 0 "a"]
+    -- make sure the resulting term is alpha equiv to \a1 -> [a1, a]
+    expectEqual t2 (ABT.abs (symbol 0 "b") (ABT.tm [var 0 "b", var 0 "a"])),
```
pchiusano, line 33: @dolio Unit test for that specific variable capture case, previously failed, now passes.

unison-core/src/Unison/ABT.hs
```diff
@@ -244,12 +244,25 @@ into' a abt = case abt of
...
 -- | renames `old` to `new` in the given term, ignoring subtrees that bind `old`
 rename :: (Foldable f, Functor f, Var v) => v -> v -> Term f v a -> Term f v a
-rename old new t0@(Term _ ann t) = case t of
-  Var v -> if v == old then annotatedVar ann new else t0
-  Cycle body -> cycle' ann (rename old new body)
-  Abs v body -> if v == old then abs' ann v body
-                else abs' ann v (rename old new body)
-  Tm v -> tm' ann (fmap (rename old new) v)
+rename old new t0@(Term fvs ann t) =
+  if Set.notMember old fvs then t0
```
pchiusano, line 248: This line is an optimization I didn't spot before - if the variable `old` isn't free in this subtree (either due to it being bound, or it just not appearing), the rename has no effect on this subtree, so skip it.

unison-core/src/Unison/ABT.hs
```diff
@@ -244,12 +244,25 @@ into' a abt = case abt of
...
+    Abs v body ->
+      -- v shadows old, so skip this subtree
+      if v == old then abs' ann v body
+
+      -- the rename would capture new, freshen this Abs
+      -- to make that no longer true, then proceed with
+      -- renaming `old` to `new`
+      else if v == new then
+        let v' = freshIn (Set.fromList [new,old] <> freeVars body) v
+        in abs' ann v' (rename old new (rename v v' body))
```
pchiusano, line 261: @dolio This is the case I added to deal specifically with the case where `v == new`.

Idea for later, I think you mentioned this at some point - would be more efficient to have a rename operation that took a `Map v v`  or some such, rather than nesting the calls like this.

dolio: I implemented simultaneous substitution in the normalized version of ABTs in my new-run branch, so we can port it back to normal ABTs after that gets merged or something.

unison-core/src/Unison/ABT.hs
dolio: Make sense to me. I'm surprised that wasn't in there, because I thought that was one of the main points of storing the free variables.

unison-core/src/Unison/ABT.hs
pchiusano: Great sounds perfect.

pchiusano:
Thanks @dolio and @aryairani merging now

unison-core/src/Unison/ABT.hs
pchiusano: Yeah it is done in the subst implementation but wasn’t being done here for some reason.

aryairani:
@pchiusano @dolio Do we need a ticket to track the backported fix, or is it already done?

pchiusano:
@aryairani existing implementation after my fix is okay, could be a bit more efficient but it isn't wrong. @dolio your call if you want to create ticket to track, or even if you think it's worth doing. Don't think it's a bottleneck ATM.

Issue #1399, opened 2020-04-04
`alias.term` doesn't work for constructor hashes


aryairani:
```
.> find

  1. unique type basics.Author
  2. unique type basics.CopyrightHolder
  3. unique type basics.GUID
  4. builtin type basics.Nat
  
.> view 1

  unique type basics.Author = #5hi1vvs5t1#0 GUID ##Text

.> alias.term #5hi1vvs5t1#0 basics.Author

  I could't find a term with hash #5hi1vvs5t1#0
```

This works for me in the branch where I fixed the `referentsByShortHash` lookup, so I assume it's that.

PR #1400, opened 2020-04-04
implement `create.author` command

### LINKED ISSUE #1392: `create.author` command
Proposed usage:

```
.foo> create.author alicecoder "Alice McGee"
   
  I created:
    
    metadata.authors.alicecoder
    metadata.copyrightHolders.alicecoder 

  Suggestion: Add `License` values for `alicecoder` under `metadata.authors.alicecoder`.

.foo> link metadata.authors.alicecoder def1 def2
...
.foo> link metadata.authors.alicecoder.defaultLicense def3 def4
```

So it creates the `Author` and `CopyrightHolder` values under the current path, `alicecoder` is the unison identifier name used for the definitions. Would typically be issued at the root of a project.

aryairani:
## Overview

image

closes #1392

## Implementation notes

I created a new `Command` to construct the three new terms, as `HandleInput` doesn't know about `Parser.Ann`.  The command calls out to `AuthorInfo.hs`.  `HandleInput` then puts the new terms into the codebase, and adds the names under the current namespace.

* initially I had included two names for the guid, `metadata.authors.<author>.guid` and `metadata.copyrightHolders.<author>.guid`, but they looked kind of ugly linked together in the diff output:
  image
  so I deleted one of the names and left the one under `authors`.

* Currently, the hashes for `GUID`, `Author`, and `CopyrightHolder` are hard-coded because thought they lived in base.  But I guess they actually live in `IOSource.hs`, so I guess later I could pull them out of there.  OTOH, didn't we want to break that up? I forget.

* `create.author foo "Foo Bar"` has the same effect as
  `create.author foo Foo Bar`.
Only `"` is given special treatment here.

## Test coverage

There is [a transcript](https://github.com/unisonweb/unison/blob/5bfe950850a6562a38bf22f2d67387d65525b5de/unison-src/transcripts/create-author.output.md) showing essentially the same thing as the screenshot at the top.

## Loose ends

There's a hash-only in the transcript output, due to #1399.  Once the #1401 is merged, this transcript could be updated to give a term alias to that hash with the following command; but not super urgent.
```
alias.term #aohndsu9bl#0 .basics.CopyrightHolder
```

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
@@ -54,6 +54,7 @@ import           Unison.FileParsers             ( parseAndSynthesizeFile
 import qualified Unison.PrettyPrintEnv         as PPE
 import Unison.Term (Term)
 import Unison.Type (Type)
+import qualified Unison.Codebase.Editor.AuthorInfo as AuthorInfo
 
 typecheck
   :: (Monad m, Var v)
@@ -190,6 +191,7 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
     Execute ppe uf -> void $ evalUnisonFile ppe uf
     AppendToReflog reason old new -> Codebase.appendReflog codebase reason old new
     LoadReflog -> Codebase.getReflog codebase
+    CreateAuthorInfo t -> AuthorInfo.createAuthorInfo Parser.Intrinsic t
 
   eval1 :: PPE.PrettyPrintEnv -> Term v Ann -> _
   eval1 ppe tm = do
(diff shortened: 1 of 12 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -190,6 +191,7 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
     Execute ppe uf -> void $ evalUnisonFile ppe uf
     AppendToReflog reason old new -> Codebase.appendReflog codebase reason old new
     LoadReflog -> Codebase.getReflog codebase
+    CreateAuthorInfo t -> AuthorInfo.createAuthorInfo Parser.Intrinsic t
```
pchiusano, line 194: `Parser.External` here I think: `Intrinsic` is the location used for builtin functions. The definitions being created here are just regular pure Unison definitions that user could in principle create themselves.

aryairani: @pchiusano Cool, I will update that before merging.

aryairani:
The bot removed the tag but didn't merge?

PR #1405, opened 2020-04-07
Fix #1402: Only display watch message if there're watch expressions

### LINKED ISSUE #1402: Remove or change output about watch expressions when there are none
```
.temp.inside> load /Users/arya/unison4/u.u

  I found and typechecked the definitions in ~/unison4/u.u. This file has been previously added to
  the codebase.
   
.temp.inside>
```

Maybe just delete the line? wdyt?  Where's the best place to teach people about watch expressions?

noahhaasis:
Closes #1402  and only displays a message about watch expressions if there are any.

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
+++ b/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
@@ -25,7 +25,7 @@ import           Control.Lens
 import qualified Control.Monad.State.Strict    as State
 import           Data.Bifunctor                (bimap, first)
 import           Data.List                     (sort, sortOn, stripPrefix)
-import           Data.List.Extra               (nubOrdOn, nubOrd)
+import           Data.List.Extra               (nubOrdOn, nubOrd, notNull)
 import qualified Data.Map                      as Map
 import qualified Data.Set                      as Set
 import qualified Data.Sequence                 as Seq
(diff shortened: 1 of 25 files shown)


anovstrup:
Should we say something like "I found no watch expressions (lines starting with `>`) to evaluate." in that case, so that watch expressions are still discoverable within the tool? Or maybe even something like:

<pre>
I found no watch expressions to evaluate.

  TIP: start a line with `> ` to have Unison evaluate an expression 
  when your scratch file is typechecked. For example:

  > foo 1 2
</pre>

pchiusano:
@noahhaasis cool! You probably have to regenerate the transcript output via `stack exec transcripts`

I think let's hold off on the tip @anovstrup - this feature is already pretty discoverable in the sense that it's in like all the docs and examples.

aryairani:
> @noahhaasis cool! You probably have to regenerate the transcript output via `stack exec transcripts`

Yah, re-run `stack build && stack exec transcripts`, sanity-check the changed.output files, and then check them in.

Though actually wait, can you double-check if it's adding stray spaces here (highlighted in red), and remove them if it's easy to do?  I can't tell if the red means they're deleted, or added and bad :)
image

aryairani:
@noahhaasis I'm gonna update the transcripts and get this merged.  Thanks again!

noahhaasis:
@aryairani sorry, saw your comments too late.

aryairani:
@noahhaasis No worries!

PR #1412, opened 2020-04-08
use -O2 by default?  can always use `stack build --fast` for `-O0`

aryairani:
I guess I should time the build each way, and put the result here.


pchiusano:
I am interested to see these results.

aryairani:
It's about 2.5x as slow for me.  I'm gonna close this for now, but noting that there is some other cargo-culted kind of stuff in there that can probably go.

PR #1416, opened 2020-04-10
split `tests` and `transcripts` into nonoverlapping source directories

aryairani:
`ghci` and dependent tooling complain when a given source file can't be attributed to a distinct build artifact.   

This PR splits the testing source into a convenience library, which the `tests` and `transcripts` executables, now each in their own distinct source directory, can both depend on.

No files were changed apart from the .cabal file.

Questionable: I moved `Suite.hs` (the `Main` of `tests`) into a new directory `testSuite`.  "Better" might have been to leave `Suite.hs` in `tests`, and move the 27 supporting source files into `unison-tests` or `test-lib` or something, but I was afraid of angering the Git Merge gods; also we'd have to learn the new path name.  Still could, though.

Feel free to merge if you like it as-is.


aryairani:
Heads up — this PR triggers a regression in stack (commercialhaskell/stack#4148, https://github.com/commercialhaskell/stack/issues/4148) with the result that if you check out the project fresh, or delete `.stack-work`, then `ghcid` or `stack repl` will not work until you have once successfully `stack build`t.

The error message is:
```
<command line>: cannot satisfy -package z-unison-parser-typechecker-z-unison-tests
    (use -v for more information)
```

To work around it, just `stack build` once before you start `ghcid`ing.

If you can't `stack build` because you have some WIP, use: 
```
git stash -u && stack build && git stash pop
```
once before you start `ghcid`ing.

Issue #1246, opened 2020-02-14
`link` command gives funny output when the metadata name is not found


aryairani:
```
.base2> link aryairani Set.difference Set.delete List.find Map.delete Map.delete.tests.prop1 Map.delete.tests.prop1.doc

  ⚠️
  
  LinkI [(Set,difference),(Set,delete),(List,find),(Map,delete),(Map.delete.tests,prop1),(Map.delete.tests.prop1,doc)] (,aryairani)

.base2> 
```

In addition to making a nicer error message, why don't we come up with some nice primitives for dealing in suffix-based names across all the ucm commands, including here?

aryairani:
Related to #1247

Issue #1244, opened 2020-02-14
No explicit error when `link` argument doesn't exist


aryairani:
```
.base2> add

  ⍟ I've added these definitions:
  
    List.find                  : a -> [a] -> Optional Nat
    Map.delete                 : a -> Map a b -> Map a b
    Map.delete.tests.prop1     : [Result]
    Map.delete.tests.prop1.doc : Doc

.base2> link

  `link metadata defn` creates a link to `metadata` from `defn`. Use `links defn` or `links defn <type>` to view outgoing links, and `unlink metadata defn` to remove a link. The `defn` can be either the name of a term or type, multiple such names, or a range like `1-4` for a range of definitions listed by a prior `find` command.

.base2> link Map.delete.tests.prop1.doc Map.delete.test.prop1

  The namespaces are identical.

.base2> 
```
@pchiusano Suggested it silently failed before the diff, possibly due to user error?

aryairani:
@runarorama Any ideas? :)

aryairani:
I guess it failed because the `defn` argument doesn't exist; `.test.` vs `.tests.`

runarorama:
Previously it would just print the `LinkI` data structure. The messages here need to be improved.

aryairani:
Hah, ok.

aryairani:
Appears to be fixed.
```
.> alias.term ##Nat.+ +

  Done.

.> ls

  1. + (##Nat -> ##Nat -> ##Nat)

.> link asdf +

  ⚠️
  
  I couldn't find any metadata matching asdf


  I didn't make any changes.

.> 
```

Issue #1220, opened 2020-02-10
Improve error messages for terms/types not found when linking/unlinking


runarorama:
Try e.g.

```
link blah blah
```

Where `blah` doesn't exist in your codebase. There should be a nice error message here.

Preferably consolidate the code so that any error where a term or type is not found displays the same message, regardless of the command.

PR #1417, opened 2020-04-10
Updates error messages for `link-unlink`

pete-ts:
## Overview

(I had an issue with my rebase on the previous PR - reopened to have clean git history)

When a user attempts `link metadata defns`, and metadata or definitions are not part of the codebase, the error surfaced explicitly points to the fact that the metadata or definitions have not been found.

A before and after can be seen in the link.output.md

Unclear if I should link this against #1220 or #1246 or #1244.

## Implementation notes

Extends `Output` with a new constructor tailored to the "metadata not found" message. 
* If the metadata `Name` cannot be extracted the original message is show.
* If the metadata `Name` can be extracted, the message mentions it explicitly.

Extends `Output` with a new constructor tailored to the "definition(s) not found" message.
* A definition will always come as a relative or absolute path. It will be displayed in the message as such, so that the user sees "the value that they entered"

* When a user attempts to `link existing_medatada existing_def non_existing_def`, the error message says "non_existing_def does not exist" and no link is created.

* I had to add unit tests for `Pretty` to figure out what I was doing wrong. Do we want these, if not I can blow them away?

## Test coverage

All scenarios touched are covered by `link.md`.
Here are the test cases added
- [x] Absolute/Relative metadata does not exist + absolute/relative definition exists
- [x] Metadata `##name` cannot be named and does not exist
- [x] Absolute/relative metadata and definition do not exist
- [x] Absolute/relative metadata exists + all absolute/relative definition(s) do not exist
- [x] Absolute/relative metadata exists + some of absolute/relative definition(s) do not exist

## Loose ends
- `link` an existing link (separate PR, unclear how to proceed or what the intended behaviour should be)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -488,31 +488,39 @@ loop = do
...
+                  step mdType b0 = let
+                    tmUpdates terms = foldl' go terms srcle
+                      where
+                      go terms src = op (src, mdType, mdValue) terms
+                    tyUpdates types = foldl' go types srclt
+                      where
+                      go types src = op (src, mdType, mdValue) types
+                    in over Branch.terms tmUpdates . over Branch.types tyUpdates $ b0
+            (_srcle, _srclt, []) ->
+              respond $ MetadataNotFound $ HQ.toName mdValue2
```
aryairani, line 520: Instead of passing `Maybe Name`, which would be much use if it's `Nothing`, try just passing `mdValue2` directly.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -488,31 +488,39 @@ loop = do
...
-                  tyUpdates types = foldl' go types srclt
-                    where
-                    go types src = op (src, mdType, mdValue) types
-                  in over Branch.terms tmUpdates . over Branch.types tyUpdates $ b0
-            (_srcle, _srclt, mdValues) ->
-              respond $ MetadataAmbiguous ppe mdValues
+             let defnsAndTerms = Map.fromList $ srcs `zip` (toList . getHQ'Terms <$> srcs)
+             let noReferents = Map.keys $ Map.filterWithKey (\_ v -> null $ v) defnsAndTerms
+             let paths = (\(path', segment) -> Path.snoc' path' $ HQ'.toName segment) <$> noReferents
+             when (not . null $ noReferents) $ respond $ DefinitionsNotFound paths
```
aryairani, line 494: Should this reuse `SearchTermsNotFound` instead of introducing a new `Output` type?

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+{-# OPTIONS_GHC -fwarn-incomplete-patterns #-}
```
aryairani, line 2: Uh oh, is this one that we want to keep?

parser-typechecker/tests/Unison/Test/Util/Pretty.hs
```diff
@@ -0,0 +1,44 @@
...
+  [ scope "commas" $
+    let
+      actual = P.render 80 $ P.commas input
+      expected = "1, 2, 3"
+    in expectEqual actual expected
+  , scope "wrap.commas" $
+     let
+       actual = P.render 80 $ P.wrap $ P.commas input
+       expected = "1 , 2 , 3"
+     in expectEqual actual expected
```
aryairani, line 19: Is this actually expected?

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
pete-ts: I stumbled upon the need for this due to the fact that `DefinitionsNotFound` was not being handled, I expected to get a compilation error, instead got an unhandled exception. Can remove it, but it feels like the sum type is too big to manoeuvre without it. I can remove it, not a problem.

parser-typechecker/tests/Unison/Test/Util/Pretty.hs
pete-ts: Was wondering about this (these rather). I realise now that: no, do not use this particular combination as the white-spacing is messed up. Then again there might be value in keeping these "tests for broken usage" for first time contributors, to see that this is the expected behaviour, and not what you should be using. 

Can remove no problem. (I take it your comment means remove the two `wrap.` cases).

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
pete-ts: In which case this becomes a use of `SearchTermsNotFound`, right?

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
pete-ts: Should be doable indeed, I think I went down this path cause I could not find a way to go from `(Path', NameSegment.HQSegment)` or from `Path'` to `HashQualified` which is what `SearchTermsNotFound` needs. :thinking:

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
aryairani: Oh, this adds warnings; I thought it was removing them.  

Correct me if I'm wrong, but this is already included in `-Wall`, which is turned on in the .cabal file.  Did adding the flag again actually do anything?

pete-ts: As per my original comment, if I am adding a new constructor in Output, the "turn warnings on" is needed, as otherwise the code compiles and ends up in a runtime error (unhandled exception). I only added it cause I got the exception in tests.

aryairani: Right, but I'm asking whether you observed a difference after adding the flag?  Or did you added it to be safe, for next time?

(If you observed a difference after adding the flag, then it means something is broken with the build file.  If you didn't observe a difference after adding the flag, then it's safe to delete.)

pete-ts: Sorry, should have clarified. Upon adding the flag, the ghc immediately gave me an "this case is unhandled error".

pete-ts: Should be reproducable on master I would expect, just remove 1-2 cases and try to compile.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
pete-ts: So from what I can tell about HQ and HQ':
* `manageLinks` takes a `[(Path', HQ'.HashQualified' NameSegment)]`, where `HQ'.HashQualified` and NameSegment` is equivalent to `NameSegment.HQSegment`.

* I can change the `paths` val to:
```haskell
let paths :: [HQ.HashQualified' Path'] = HQ'.toHQ <$> Path.unsplitHQ' <$> noReferents
```
meaning I am how in HQ land, but still there appears to be no way to convert from `HQ.HashQualified' Path'` to `HQ.HashQualified' Name`, which is what `SearchTermsNotFound` needs. 

Now at the same time, I am of the impression that an equivalent restriction exists for other error messages:

```haskell
  | PatchNotFound Path.Split'
  | TypeNotFound Path.HQSplit'
  | TermNotFound Path.HQSplit'
  | TypeNotFound' ShortHash
  | TermNotFound' ShortHash
  | SearchTermsNotFound [HQ.HashQualified] -- HQ.HashQualified' Name
```
I think that in this instance since the `Path.HQSplit' or `Path', HQ'.HashQualified' NameSegment` is indeed what the user has requested gets linked, it is what should be serialised in the response.

pete-ts: Updated to reuse `SearchTermsNotFound`. As discussed, with an intermediate cast so that the `Text` is extracted from the `Path'` and used as a `NameOnly Name`

pete-ts:
Based on latest master behaviour I don't think this PR is needed anymore.
It is slightly diverging from current behaviour in that: it broke `ucm` with an error if the user would provide multiple `defns` to link and one or more did not exist (alongside existing ones). 

Closing!

PR #1424, opened 2020-04-13
Couple of fixes to whitespace handling in doc literals

### LINKED ISSUE #1363: Doc value parses incorrectly
This doc value ends up being parsed incorrectly:

```unison
void.doc = [: `@void op` executes an effectful computation `op` for its effects,
  ignoring the return value. `@void op` is equivalent to `@ignore . op`.

  __See also__ @ignore

:]
```

After adding the definition, `view void.doc` prints the following:
```
  void.doc : #v00j3buk6m
  void.doc =
    [: `@voidop` executes an effectful computation `op` for its effects,
    ignoring the return value. `@void op` is equivalent to `@ignore .
    op`.

    __See also__ @ignore

    :]
```
(Note the missing whitespace in "voidop".)

Based on this showing up in the _view_ output, this seems to be a parser issue? The behavior is inconsistent; a similar definition is parsed/rendered as expected:

```
ignore.doc = [: `@ignore a` ignores the argument `a` and returns `()`. :]
```

**_See also_** [this gist](https://gist.github.com/anovstrup/941af998bd503abc2886d94889f4fa4d), and [the gist output](https://gist.github.com/anovstrup/63d2269adce3efe7744390216bdd29eb)

atacratic:
Tagging @pchiusano, although I don't think this really calls for much review beyond the text in this PR.  I'll merge it in a couple of days if I haven't heard.  

## Overview

- Preserve spaces after @ directives in first line when unindenting (fixes #1363)
- Stop eating spaces that immediately follow @[source], @[evaluate] or @[include] directives on the same line, while unindenting.
- Fix docs.md so its example isn't asking for line-wrapping.
- Improve commenting, including a note in TermParser.hs on what we could do if we just wanted to blow away all this whitespace munging code - pasted in below.

## Implementation notes

Only touches TermParser.docNormalize (plus transcripts, plus adding a list utility function.)

Nothing.

## Test coverage

Both the code fixes are covered by the changes to the transcripts.

## Loose ends

Only the doubt about whether we'll want to keep `docNormalize` long term.  I think keep it for now.  

```haskell
-- This normalization allows the pretty-printer and doc display code to do
-- indenting, and to do line-wrap of paragraphs, but without the inserted
-- newlines being then frozen into the text for ever more over subsequent
-- edit/update cycles.
--
-- The alternative would be to stop line-wrapping docs on view/display by adding
-- newlines in the pretty-printer, and instead leave wrapping to the
-- terminal/editor.  Might be worth considering if this code ends up being
-- too buggy and fragile to maintain.  Maybe display could add newlines,
-- and view could refrain from doing so.
```


atacratic:
Ah I see review is required now, OK will wait for you to take a look Paul.

PR #1425, opened 2020-04-13
Swaps the arguments and output sequence of ListNames

### LINKED ISSUE #1394: let `names` command output list types, then terms
Just a quick swap in `OutputMessages.hs` at `ListNames`.

If you spot other sections that list the terms before the types, we could swap those too.

pete-ts:
## Overview

Before: `names` shows terms before types
After: `names` shows types first, then terms

This PR aims to close #1394

## Implementation notes

Swaps the constructor arguments, and swaps the sequence of content being serialised.

Changing the sequence of arguments not strictly needed; but the constructor fits in in a more uniform manner with the rest of the constructors.

## Test coverage

Transcript added, further future changes can thus be tracked.

## Loose ends

N/A

### DIFF
--- a/unison-src/transcripts/names.md
+++ b/unison-src/transcripts/names.md
@@ -1,0 +1,20 @@
+ Example uses of the `names` command and output
+```ucm:hide
+.> builtins.mergeio
+```
+
+```unison:hide
+type IntTriple = IntTriple (Int, Int, Int)
+intTriple = IntTriple(+1, +1, +1)
+```
+
+```ucm:hide
+.> add
+```
+
+```ucm
+.> alias.type IntTriple namespc.another.TripleInt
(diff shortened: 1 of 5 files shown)


unison-src/transcripts/names.md
```diff
@@ -0,0 +1,20 @@
+ Example uses of the `names` command and output
+```ucm:hide
+.> builtins.mergeio
```
aryairani, line 3: Avoid using `builtins.mergeio` where possible, as it just adds unnecessary lag at startup.

In this case, you can just use `builtins.merge`, or even better: just `alias.type ##Int .builtins.Int`

aryairani:
I left one request; otherwise looks great, thanks!

Issue #1157, opened 2020-01-17
Parsing "❤"


Sheyne:
While troubleshooting Unicode output (#1152), I created a scratch file with a unicode heart in it. When @aryairani ran the file, he got: 

``` 
/Users/arya/unison/u.u:1:1:
  unexpected UnknownLexeme
  expecting ability, type, or use
      1 | ❤ : a -> a
```
and wrapping it in parens like a symboly-id doesn't help:
```
  /Users/arya/unison/u.u:1:1:
  unexpected (
  expecting ability, type, or use
      1 | (❤) : a -> a
```

The original file contained: 
```
❤ : a -> a
❤ a = a

> ❤
```

anovstrup:
The error message reported in #1152 (`ucm: <stdout>: commitBuffer: invalid argument (invalid character`) suggests that this lexer/parser bug is distinct from the bug(s) that @Sheyne encountered, which both seem to be lower level character encoding issues (possibly with the same root cause).

aryairani:
I get:
```
  /Users/arya/unison/u.u:1:1:
  unexpected UnknownLexeme
  expecting ability, type, or use
      1 | ❤ : a -> a
```
and wrapping it in parens like a symboly-id doesn't help:
```
  /Users/arya/unison/u.u:1:1:
  unexpected (
  expecting ability, type, or use
      1 | (❤) : a -> a
```

aryairani:
This is kind of weird.:

**Bare heart:**
```
❤ : a -> a
❤ a = a

> ❤
```
First of all, it takes a long long time to parse, and then reports:
```
  I got confused here:
  
      1 | ❤ : a -> a
  
  
  resetting state
```
A  **parenthesized heart** results in the same.


**Backticked heart:**
```
`❤` : a -> a
`❤` a = a

> `❤`
```
```
  I got confused here:
  
      1 | `❤` : a -> a
  
  
  I was surprised to find a ❤ here.
  I was expecting one of these instead:
  
  * end of input
  * identifier (ex: abba1, snake_case, .foo.bar#xyz, .foo.++#xyz, or 🌻)
```

^Inspired by the previous error message:
**Bare flower:**
```
🌻 : a -> a
🌻 a = a

> 🌻
```
(very slow parse again)
```
    ⍟ These new definitions are ok to `add`:
    
      🌻 : a -> a
  
  Now evaluating any watch expressions (lines starting with `>`)... Ctrl+C cancels.


    4 | > 🌻
          ⧩
[...]

PR #1437, opened 2020-04-19
Pretty-printer of non ascii characters in `Text`

### LINKED ISSUE #1429: Pretty-printing of non-ascii characters in Text is broken.
Unison:
```unison
> "Rúnar"
```

UCM:
```ucm
    1 | > "Rúnar"
          ⧩
          "R\250nar"
```

But `\250` is not a valid escape sequence in Unison.

pete-ts:
## Overview

Before: Text output of the pretty printer cannot display non ascii characters for `Text`.
After: non ascii characters displayed in watch outputs of Text

Fix #1429 

## Implementation notes

Using the unicode-show package, changes the call to the default show for `Text` with one `ushow` that allows Unicode output to be displayed correctly.

Some questions asked in the comments for #1429 

## Test coverage

1. The test case mentioned in the issue has been added as a trancript, as well as some more.
2. After @atacratic 's suggestion I have also added the same case in the `TermPrinter` tests, which (I think) shows that the lexer can handle some but not all unicode. ( #1157 , or should another be raised? )

## Loose ends

Not aware of any at this time.


atacratic:
I've reviewed this and I think it's good to go.  I'll leave it to one of the core team to actually give the nod though.

Issue #1423, opened 2020-04-12
ucm should also sync dependencies of `patch` objects


aryairani:
or else how can I be sure I can apply them?

Issue #1342, opened 2020-03-11
idea: record `pull` and `push` urls to a text file, and use them to populate `gitUrlArg` autocomplete


aryairani:
This may (though arguably may not) become irrelevant if we start attaching git remotes as metadata in other ways.

Issue #667, opened 2019-08-06
Ability to see what's new in a repo since last `pull`


runarorama:
It's crucial to any kind of collaboration on a codebase that we can check whether there's any new work that we need to pull.

Also, need to be able to see what's new, instead of blindly pulling everything and having no clue.

pchiusano:
Temporary workaround is you can do a `pull` to some temporary namespace and then a `merge.preview`

```
.> pull giturl temp
.> merge.preview temp mynamespace
```

That'll show you the diff of what would happen if you merged `temp` into `mynamespace`. Assuming that looks okay, you do:

```
.> merge temp mynamespace
.> delete.namespace temp
```

aryairani:
Some kind of pull-dry-run that shows a diff? #2703

PR #1439, opened 2020-04-20
git caching / selective copy on pull

### LINKED ISSUE #940: selectively copy codebase elements on pull
To pull a remote unison repo into your current involves feels safe to assume you need to copy all of the branches, types, terms, watches, indices, which we do by literally copying all the files other than .git and _head.  To pull just a single remote namespace and its dependencies needs generally a tiny fraction of those files.  To avoid significant codebase bloat, we'll eventually want to select files more precisely.

One reason we began with the copy-all approach was that identifying the appropriate codebase elements might be slow, since there's no "dependencies" index; however, there will be a point where copy-all is slower—not per file, but per codebase as they begin to include all the code transitively that anyone has ever referenced.

Also, it would be nice to remove the implicit dependency on the v1 `FileCodebase` format `syncFromDirectory :: FilePath -> m ()` from the `Codebase` interface in favor of some generic codebase data exchange like `importFromBranch :: Codebase m v a -> Branch m -> m ()` that can be used against other kinds of Codebase implementations.  Then instead of `syncToDirectory`, we can just do the create the new merged root branch, and import it in the other direction.

Do you have to do something TBD extra for copying `watches` too?

### LINKED ISSUE #1281: design: `push`/`pull` are sluggish
Figure out how to make them fast(er).

### LINKED ISSUE #1291: issue pushing to as-of-yet nonexistent git branches
Not sure if this is a regression, or never tested before.

### LINKED ISSUE #1296: confirm whether pulling a particular git commit works
(after merging #1281)

e.g.
```
.tmp> pull git@github.com:unisonweb/base:7ec9cf12006d524dba7486af5540d4649ab07c06 base7ec
.tmp> pull git@github.com:unisonweb/base:0181e80a95ca3a7bbe32f40abf037f8bb2e9e77f base018
.tmp> diff.namespace base018 base7ec
 
  The namespaces are identical.
```
Shouldn't they be different? But possibly a bad example.

### LINKED ISSUE #1335: Codebase/Sync refactor proposal
My main goals here are to be able to:
- compose a virtual `Codebase` from less general ones:
  - implement a MemoryCodebase with all the builtins & essentials
    - split out the builtins (`IO`, `Optional`, etc.) from the FileCodebase into a MemoryCodebase, to avoid having to write out a big database for each new codebase (adds up when running many transcripts).
    - have such a thing as a FileCodebase without 1925 files in it already, in case I might want to debug something about FileCodebase construction.  (e.g. #1317 :))
  - operate on definitions from remote codebases (e.g. a `CachingGithubCodebase`) without the requirement to first fully ingest it into the local codebase
- provide a "fast sync" between codebases with compatible internal representations in a type-safe way

Plan:
- Create a first-class representation of `FileCodebase`, e.g.
    ```haskell
    data FileCodebase v a = FileCodebase
      { root :: CodebasePath
      , fmtV :: S.Format v
      , fmtA :: S.Format a
      }
    ```
    (these are the same as the current arguments to `FileCodebase.codebase1`)

- pull out the `sync*` functions from the `Codebase`
- who should own reflog functions?
- Move `IOSource.hs` and the like to an in-memory codebase which gets composed with the `FileCodebase`
- Spawn a thread in certain cases (e.g. interactive ucm only) to sync that in-memory codebase to disk?

wip (I already know this is not what I want, but it's all I can write at the moment, so I'm saving it anyway)

aryairani:
## Overview

This PR implements:
1. selective copy on `pull` (avoids copying extra objects from a cloned remote git repo)
2. caching of git default branch (e.g. `master) of unison repos for `push`/`pull` operations (avoids git-downloading the same data on successive pulls)

Some benchmarks for the caching:
|   | M1i | master | PR | |
| ------------- | ------------- | --- | --- | - |
| 1.  create codebase | .71s | .04s | .04s | |
| 2. pull `unisonweb/base`  | 10.3s | 11.1s | 10.6s | |
| 3. pull `unisonweb/base` again  | 9.4s |  9.6s  | 0.5s | cached |
| 4. push `.base` to a new github repo | 54.0s | 45.0s | 37.4s | |
| 5. push again to the same github repo | 16.1s | 15.0s | 3.5s | cached |
| 6. pull `unisonweb/base` one more time  | 9.6s |  10.1s  | 0.8s | just checking. |
| 7. pr.load unisonweb/base#23 | 21.5s | 24.2s | 9.2s | half cached |
| 8.pr.load unisonweb/base#23 again | 23.7s | 24.2s | 5.0s | both cached |

- Closes #940.
- Closes #1335 for now.
- Closes #1281, although we can brainstorm more ideas if needed; including benchmarking / switching sync algorithms.
- Cancels/closes #1296, cancels/closes #1291 — the existing support for `push`/`pull` with arbitrary git revisions was incomplete, this PR prints an error message if you try to use it for now, which directs wishful users to add feedback to the ticket.  Opened #1436 to track this.
- Relates to #1342, because we can now recover git urls from the git cache and use them as autocompletions.
- Relates to #667, because new `Command.ViewRemoteBranch` would help us implement `pull.preview`.

### Git caching

The new assumption is that we will only be working with the default git branch / remote HEAD

- `Git.pullBranch` _produces_ the repo/codebase/cache dir instead of receiving it as an arg

convenience stuff:
- added `Pretty.backticked'` which takes an extra arg that will be  grouped onto the ending backtick.

boring stuff:
- lifted some IO stuff to MonadIO
- consolidated use of `P.parse UriParser.repoPath` into `IP.parseUri`

### Selective copy on `pull`

I updated the signatures of `Codebase.syncTo/FromDirectory` to be basically the same, and they now share the same implementation as we use for `push`.

These no longer do any special merge checks (do this ahead of time if desired); they just set a provided root branch in the destination codebase, and copy the branch along with missing dependencies to the
destination codebase as well.

- refactored `LoadRemoteRootBranch` and `LoadRemoteShortBranch` commands into `ImportRemoteBranch` and `ViewRemoteBranch` which will/won't copy files into the local codebase, respectively.
- added `Git.viewGitBranch :: RemoteNamespace -> _ (Branch m)`, which doesn't copy stuff to the local codebase.
- `Git.pushGitRootBranch` no longer performs a merge; checks `before` instead and complains if necessary
- replaced `onError` with `withIOError` that passes the error along
- consolidate `Text`+`Maybe Text` arg pairs into `RemoteRepo` in a bunch of relevant places, where
  
  ```haskell
  data RemoteRepo = GitRepo { url :: Text, commit :: Maybe Text }
  ```
- rejiggered GitError.hs somewhat

There's one cache per git url; during self-review I realized we could maybe do better with a monorepo with lots of remotes, if this is still a bottleneck.

## Test coverage

For caching: no specific tests.  There are probably some weird git situations that people could produce outside of `ucm` which haven't been tested.  If we want to test these, I could probably use help coming up with a list.

For selective pull: There is a new `pull` test in `Unison/Test/Git.hs`, which ensures that only necessary files are pulled from the remote repo, including dependencies outside the pulled namespace.

## Loose ends

* #1423 - ucm should also sync dependencies of patch objects (existing bug)
* pr.load is still kind of slow at 5s?  Maybe 2.5s `git fetch` per repo explains it.


parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -136,31 +136,27 @@ data Command m i v a where
...
+    RemoteNamespace -> Command m i v (Either GitError (Branch m))
 
   -- Syncs the Branch to some codebase and updates the head to the head of this causal.
   -- Any definitions in the head of the supplied branch that aren't in the target
   -- codebase are copied there.
   SyncLocalRootBranch :: Branch m -> Command m i v ()
 
+  SyncRemoteRootBranch ::
+    RemoteRepo -> Branch m -> Command m i v (Either GitError ())
+
```
aryairani, line 154: This is just moved up from below.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -689,44 +692,35 @@ loop = do
...
         respondNumbered $ ShowDiffNamespace beforep afterp ppe outputDiff
 
-      CreatePullRequestI baseRepo headRepo -> do
-        baseBranch <- loadRemoteBranch baseRepo
-        headBranch <- loadRemoteBranch headRepo
-        case (baseBranch, headBranch) of
-          (Left e1, Left e2) -> gitError e1 >> gitError e2
-          (Left e, _) -> gitError e
-          (_, Left e) -> gitError e
-          (Right baseBranch, Right headBranch) -> do
```
aryairani, line 699: No change here, other than use of `unlessGitError`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -689,44 +692,35 @@ loop = do
...
       LoadPullRequestI baseRepo headRepo dest0 -> do
         let desta = resolveToAbsolute dest0
         let dest = Path.unabsolute desta
         destb <- getAt desta
-        if Branch.isEmpty0 (Branch.head destb) then do
-          loadRemoteBranch baseRepo >>= \case
-            Left e -> respond $ GitError input e
-            Right baseb -> loadRemoteBranch headRepo >>= \case
-              Left e -> respond $ GitError input e
-              Right headb -> do
```
aryairani, line 715: No change here, other than use of `unlessGitError`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1652,30 +1646,32 @@ loop = do
           makePrintNamesFromLabeled' (Patch.labeledDependencies patch)
         respond $ ListEdits patch ppe
 
-      PullRemoteBranchI mayRepo path -> do
-        let destAbs = resolveToAbsolute path
-        resolveConfiguredGitUrl Pull path mayRepo >>= \case
-          Left e -> eval . Notify $ e
-          Right ns -> pullRemoteBranchAt (Just path) input inputDescription ns destAbs
```
aryairani, line 1659: Rewrite to use `unlessError` and `importRemoteBranch`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1652,30 +1646,32 @@ loop = do
...
-          Left e -> eval . Notify $ e
-          Right (repo, Nothing, remotePath) -> do
-              -- push from srcb to repo's remotePath
-              eval (LoadRemoteRootBranch repo) >>= \case
-                Left e -> eval . Notify $ GitError input e
-                Right remoteRoot -> do
-                  newRemoteRoot <- eval . Eval $
-                    Branch.modifyAtM remotePath (Branch.merge srcb) remoteRoot
-                  syncRemoteRootBranch input repo newRemoteRoot
-          Right (_, Just _, _) ->
```
aryairani, line 1675: Rewrite to use `unlessError` and `viewRemoteBranch`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1760,10 +1756,10 @@ loop = do
         :: PushPull
         -> Path'
         -> Maybe RemoteNamespace
-        -> Action' m v (Either (Output v) RemoteNamespace)
+        -> ExceptT (Output v) (Action' m v) RemoteNamespace
       resolveConfiguredGitUrl pushPull destPath' = \case
-        Just ns -> pure $ Right ns
-        Nothing -> do
```
aryairani, line 1766: Refactor to return `ExceptT` for convenience with `unlessError`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2129,35 +2125,14 @@ unlessError ma = runExceptT ma >>= either (eval . Notify) pure
...
-  -> RemoteNamespace
-  -> Path.Absolute
-  -> Action' m v ()
-pullRemoteBranchAt p' input inputDescription ns p = do
-  loadRemoteBranch ns >>= \case
-    Left  e -> eval . Notify $ GitError input e
-    Right b -> do
-      let msg = p' <&> \p' -> PullAlreadyUpToDate ns p'
-      mergeBranchAndPropagateDefaultPatch inputDescription msg b p' p
-
```
aryairani, line 2151: Inlined away.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2129,35 +2125,14 @@ unlessError ma = runExceptT ma >>= either (eval . Notify) pure
...
 -- | supply `dest0` if you want to print diff messages
 --   supply unchangedMessage if you want to display it if merge had no effect
 mergeBranchAndPropagateDefaultPatch :: (Monad m, Var v) =>
   InputDescription -> Maybe (Output v) -> Branch m -> Maybe Path.Path' -> Path.Absolute -> Action' m v ()
 mergeBranchAndPropagateDefaultPatch inputDescription unchangedMessage srcb dest0 dest = do
-  mergeDidChange <- mergeBranch inputDescription srcb dest0 dest
-  if mergeDidChange then
-    loadPropagateDiffDefaultPatch inputDescription dest0 dest
-  else for_ unchangedMessage respond
```
aryairani, line 2160: No change, other than to use `ifM`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2181,19 +2156,6 @@ loadPropagateDiffDefaultPatch inputDescription dest0 dest = do
...
-                        (LoadRemoteShortBranch repo) sbh)
-  pure $ Branch.getAt' remotePath <$> eroot
-
-
-syncRemoteRootBranch
-  :: Var v => Monad m => Input -> RemoteRepo -> Branch m -> Action m i v ()
-syncRemoteRootBranch input repo b = do
-  e <- eval $ SyncRemoteRootBranch repo b
-  either (eval . Notify . GitError input) (const $ respond Success) e
-
```
aryairani, line 2196: Refactored away.

parser-typechecker/src/Unison/Codebase/Execute.hs
```diff
@@ -40,6 +40,8 @@ execute codebase runtime mainName =
         die ("Couldn't identify a root namespace.")
       Left (Codebase.CouldntLoadRootBranch h) ->
         die ("Couldn't load root branch " ++ show h)
+      Left (Codebase.CouldntParseRootBranch h) ->
+        die ("Couldn't parse root branch head " ++ show h)
```
aryairani, line 44: We were just crashing in this case, previously.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -175,8 +173,7 @@ codebase1' syncToDirectory fmtV@(S.Format getV putV) fmtA@(S.Format getA putA) p
           (branchHeadUpdates path)
           (branchFromFiles path)
           dependents
-          -- Just copies all the files from a to-be-supplied path to `path`.
-          (copyFromGit path)
+          (flip (syncToDirectory fmtV fmtA) path)
           (syncToDirectory fmtV fmtA path)
```
aryairani, line 177: `syncToDirectory` and `syncFromDirectory` share an implementation now.

parser-typechecker/src/Unison/Codebase/FileCodebase/Common.hs
```diff
@@ -313,30 +304,29 @@ branchFromFiles rootDir h = do
...
+    conflict -> runExceptT (traverse fileToBranch conflict) >>= \case
+      Right (x : xs) -> Right <$> foldM Branch.merge x xs
+      Right _ -> error "FileCodebase.getRootBranch.conflict can't be empty."
+      Left e -> Left <$> pure e
+
+  fileToBranch :: String -> ExceptT Codebase.GetRootBranchError m (Branch m)
+  fileToBranch single = ExceptT $ case hashFromString single of
+    Nothing -> pure . Left $ Codebase.CouldntParseRootBranch single
+    Just (Branch.Hash -> h) -> branchFromFiles root h <&>
+                                maybeToEither (Codebase.CouldntLoadRootBranch h)
```
aryairani, line 327: Just renamed these functions called `go`.  I think it's a slight improvement, but not huge.

parser-typechecker/src/Unison/Codebase/FileCodebase/Common.hs
```diff
@@ -407,30 +397,6 @@ referentIdFromString s = referentFromString s >>= \case
...
-      unless exists . copyFile path $ replaceRoot from to path
-
-copyFromGit :: MonadIO m => CodebasePath -> CodebasePath -> m ()
-copyFromGit to0 from0 = let
-  to = to0 </> codebasePath
-  from = from0 </> codebasePath 
-  in whenM (doesDirectoryExist from) $
-  copyDir (\x -> not ((".git" `isSuffixOf` x) || ("_head" `isSuffixOf` x)))
-          from to
-
```
aryairani, line 433: We're selectively copying rather than bulk copying now.

parser-typechecker/src/Unison/Codebase/FileCodebase/CopyFilterIndex.hs
```diff
@@ -84,22 +84,9 @@ syncToDirectory' :: forall m
...
-          Right existingDestRoot -> Branch.merge branch existingDestRoot
-          -- The destination codebase doesn't advertise a root branch,
-          -- so we'll just use ours.
-          Left Codebase.NoRootBranch -> pure branch
-          Left (Codebase.CouldntLoadRootBranch h) -> fail $
-            "I was trying to merge with the existing root branch at " ++
-            branchPath destPath h ++ ", but the file was missing."
-          )
-        -- else there was no existing codebase structure at `destPath` so whatev
-        (pure branch)
```
aryairani, line 102: Instead of trying to do this error handling and merging during the sync (which we only wanted for `pull` anyway, and not for `push`), we do it before this call. 

The load error handling now happens in `Git.viewRemoteBranch'`, a "fast-forward only" / `Branch.before` check happens in `Git.pushGitRootBranch`, and any merging has to happen in `HandleInput.hs`, not here.

parser-typechecker/src/Unison/Codebase/FileCodebase/CopyRegenerateIndex.hs
```diff
@@ -99,22 +99,9 @@ syncToDirectory' :: forall m v a
...
-          Right existingDestRoot -> Branch.merge branch existingDestRoot
-          -- The destination codebase doesn't advertise a root branch,
-          -- so we'll just use ours.
-          Left Codebase.NoRootBranch -> pure branch
-          Left (Codebase.CouldntLoadRootBranch h) -> fail $
-            "I was trying to merge with the existing root branch at " ++
-            branchPath destPath h ++ ", but the file was missing."
-          )
-        -- else there was no existing codebase structure at `destPath` so whatev
-        (pure branch)
```
aryairani, line 117: Same deal as for the removed code from `CopyFilterIndex.hs`

parser-typechecker/src/Unison/Codebase/FileCodebase/Reserialize.hs
```diff
@@ -71,26 +69,15 @@ syncToDirectory
...
-  -- If there's already a codebase at `destPath` and we can access its root
-  -- branch, then we merge the existing root with the provided `branch`, in
-  -- preparation for writing; otherwise we just plan to write the provided one.
-  branch@(Branch c) <-
-    ifM (codebaseExists destPath)
-      (do
-        remoteRoot <- getRootBranch destPath
-        Branch.merge branch (fromMaybe Branch.empty $ rightMay remoteRoot))
-      (pure branch)
-  -- We build and use the state inside of `serialize`; now we can throw it away.
```
aryairani, line 85: Same deal as for the removed code from `CopyFilterIndex.hs`

parser-typechecker/src/Unison/Codebase/GitError.hs
```diff
@@ -2,19 +2,22 @@ module Unison.Codebase.GitError where
...
-              | NoRemoteNamespaceWithHash Text (Maybe Text) ShortBranchHash
-              | RemoteNamespaceHashAmbiguous Text (Maybe Text) ShortBranchHash (Set Branch.Hash)
-              | Couldn'tLoadRootBranch Text (Maybe Text) (Maybe ShortBranchHash) Branch.Hash
-              | SomeOtherError Text
+              | PushDestinationHasNewStuff RemoteRepo
+              | NoRemoteNamespaceWithHash RemoteRepo ShortBranchHash
+              | RemoteNamespaceHashAmbiguous RemoteRepo ShortBranchHash (Set Branch.Hash)
+              | CouldntLoadRootBranch RemoteRepo Branch.Hash
+              | CouldntParseRootBranch RemoteRepo String
+              | SomeOtherError String
```
aryairani, line 22: Switched out the top 3 for the top 5, and switched `Text (Maybe Text)` to `RemoteRepo` everywhere.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -716,12 +718,10 @@ pull = InputPattern
   (\case
     []    -> Right $ Input.PullRemoteBranchI Nothing Path.relativeEmpty'
     [url] -> do
-      ns <- first (fromString . show)
-              (P.parse UriParser.repoPath "url" (Text.pack url))
```
aryairani, line 720: Consolidated all of these calls to `UriParser.repoPath` to one spot so I could warn not to try using git branches there.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -784,7 +783,7 @@ createPullRequest = InputPattern "pull-request.create" ["pr.create"]
                                                 "https://github.com/me/unison:.libs.pr.base" ]
     ])
   (\case
-    [baseUrl, headUrl] -> first fromString $ do
```
aryairani, line 787: After changing the signature of `parseUri` below, these were no longer needed.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -609,31 +610,37 @@ notifyUser dir o = case o of
 
   TodoOutput names todo -> pure (todoOutput names todo)
   GitError input e -> pure $ case e of
+    CouldntParseRootBranch repo s -> P.wrap $ "I couldn't parse the string"
+      <> P.red (P.string s) <> "into a namespace hash, when opening the repository at"
+      <> P.group (prettyRepoBranch repo <> ".")
```
aryairani, line 615: Started using `prettyRepoRevision` and `prettyRepoBranch` throughout to consolidate the formatting.

parser-typechecker/src/Unison/Util/Pretty.hs
```diff
@@ -786,6 +787,10 @@ blockedCallout = callout "🚫"
 backticked :: IsString s => Pretty s -> Pretty s
 backticked p = group ("`" <> p <> "`")
 
+-- |Attach some punctuation after the closing backtick.
+backticked' :: IsString s => Pretty s -> Pretty s -> Pretty s
+backticked' p end = group ("`" <> p <> "`" <> end)
+
```
aryairani, line 793: Side note, I would love a combinator that took a pretty, applied `P.wrap`, but then grouped a supplied punctuation or something to the last word of the `wrap`.

parser-typechecker/unison-parser-typechecker.cabal
```diff
@@ -313,6 +313,7 @@ library unison-tests
     mtl,
     raw-strings-qq,
     shellmet,
+    split,
```
aryairani, line 316: for `Data.List.Split.splitOn` in `parser-typechecker/tests/Unison/Test/Git.hs`

pchiusano:
Bouncing for Travis refresh.

parser-typechecker/src/Unison/Codebase/Editor/RemoteRepo.hs
```diff
@@ -3,7 +3,7 @@
 module Unison.Codebase.Editor.RemoteRepo where
 
 import Unison.Prelude
-import Unison.Util.Monoid as Monoid
+import Unison.Util.Monoid as Monoid
```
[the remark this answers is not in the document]
aryairani: My editor only sometimes strips trailing whitespace; I have opened a ticket about it — and will continue trying to catch the ones that slip through.

parser-typechecker/src/Unison/Codebase/Editor/Git.hs
```diff
@@ -48,132 +55,187 @@ withStatus str ma = do
...
+  pure branch
+
+-- | Pull a git branch and view it from the cache, without syncing into the
+-- local codebase.
+viewRemoteBranch :: forall m. MonadIO m
+  => RemoteNamespace -> ExceptT GitError m (Branch m)
+viewRemoteBranch = fmap fst . viewRemoteBranch'
+
+viewRemoteBranch' :: forall m. MonadIO m
+  => RemoteNamespace -> ExceptT GitError m (Branch m, CodebasePath)
```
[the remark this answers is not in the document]
aryairani: Yeah there may be a few places where we can.  I’m a little scared to switch it without pairing due to some tough to resolve inference issues that I struggled to resolve a few days ago.  Maybe I got better at it in the process?

aryairani:
> I do think that for the cache, one repo with multiple remotes could be a faster. A lot of these Unison repos are going to have common namespace history, so this could be the difference between needing to fetch thousands of files the first time you load a pull request from someone new vs needing to fetch like 10 files.
> 
> OTOH, I'm not sure if git's sync protocol will do any better with this, since it still has to sync all the commits - the git histories are in general going to be unrelated.

Thank you for documenting exactly what I was thinking when I made that comment but was too tired to elaborate at the time 👍👍👍

> If the remote has made an unrelated commit of a file with the same contents, and the recipient has those same contents (possibly in another commit), do you know if Git will avoid sending those file contents again?

Yes — files in git are identified by hash, then other files representing a directory tree for a given commit references those first files and maps filenames to them.

So at least the file files, if not the tree files or commit files, would be shared/cached appropriately by git.

> Actually, now that I'm saying this, I'm doubting whether git will even let you add remotes that don't all have a common history.

I’m almost completely sure it can.

> Anyway, seems like maybe long-term plan will be to move away from using git for syncing. But this seems more than good for now. :)

parser-typechecker/src/Unison/Codebase/Editor/RemoteRepo.hs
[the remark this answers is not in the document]
aryairani: https://youtrack.jetbrains.com/issue/IDEA-228850

PR #1449, opened 2020-04-20
`todo` and `pull` tab completion crashes

### LINKED ISSUE #1445: Unison crashes on attempted tab-complete
I excitedly pull a juicy-looking PR:
```
.basememo> pull-request.load https://github.com/unisonweb/base https://github.com/anovstrup/unisonweb-base:.prs._memo
```
I type the following and hit tab
```
.basememo> todo merged.patch m
```
`ucm` crashes with no output :-(

image

Other tab-completes seem to be working for me OK.  

This is master@ae95dc.

pete-ts:
## Overview

Before: When using `todo` and `pull` an attempt to auto complete the second argument crashes `ucm`.
After: `todo` and `pull` work as expected. 

fix #1445 

## Implementation notes

Updates the first argument in both cases to `Requred`.
A quick scan seems to show that these are the last 2 cases where this was happening.

 N/A

## Test coverage

Manual

Before:
Screenshot from 2020-04-20 23-00-38

Screenshot from 2020-04-20 23-07-12

After:
Screenshot from 2020-04-20 23-13-04
Screenshot from 2020-04-20 23-13-09

## Loose ends

N/A

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
+++ b/parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
@@ -95,7 +95,7 @@ todo :: InputPattern
 todo = InputPattern
   "todo"
   []
-  [(Optional, patchArg), (Optional, pathArg)]
+  [(Required, patchArg), (Optional, pathArg)]
   (P.wrapColumn2
     [ ( makeExample' todo
       , "lists the refactor work remaining in the default patch for the current"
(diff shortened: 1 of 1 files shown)


anovstrup:
The help output needs to be updated, too, but I'm not sure this is the right solution. Would it be better instead to change how the command parser handles multiple optional arguments, so that the three forms currently listed by the help command could still be supported?

pete-ts:
Indeed, I think you are correct. I missed the fact that `todo` is a valid command without any arguments!

pete-ts:
Let me close this and reopen after some more investigation. Or please feel free (anyone) to go ahead  otherwise :+1:

PR #1456, opened 2020-04-22
Add more bitwise operations to Int/Nat

stew:
## Overview

This adds bitwise and / or / xor / compliment / leadingZeros / trailingZeros to both Int and Nat

## Test coverage

Added basic "methodical" tests for these new methods


aryairani:
Looks good to me.  Should add the corresponding entries to `base`.

stew:
@aryairani what are the entries that need to be added to base? how do I add them?

Is this where I could be adding docs to these functions?

aryairani:
@stew Hi!  Yes. Given a build that includes `stew:bitwise-operations`, ucm knows how to handle these at runtime, but they don't necessarily have names in the namespace for folks to discover (and/or add docs to).

There are two ways to add them, either would need to be done using a build of ucm that includes your PR.

Way 1: add aliases by hash

```
.> alias.term ##Int.and myfork.Int.and
.> link myfork.Int.and.doc myfork.Int.and
```

Way 2: dump the builtins
```
.tmp> builtins.merge
.tmp> ls
  1. builtin/ (316 definitions)
.tmp> alias.term builtin.Int.and .myfork.Int.and
.tmp> link .myfork.Int.and.doc .myfork.Int.and
.tmp> docs .myfork.Int.and
🌷 
.tmp>
```

Let me know if you run into trouble.

pchiusano:
@stew and then also if you are feeling brave you can open a PR to merge this into `trunk` of base - we are beta testing docs on that, see https://www.unisonweb.org/docs/codebase-organization and specifically this section https://www.unisonweb.org/docs/codebase-organization/#day-to-day-development-creating-and-merging-pull-requests

@aryairani I'm starting to cut releases of `base` for each UCM release, named accordingly. So like `base` repo has `releases._M1j` in it, and when we do M1k release we'll add a release for that too.

aryairani:
@pchiusano Excellent

PR #1466, opened 2020-04-26
fix  Nat.complement and Int.complement builtins

atacratic:
## Overview

The builtin `Nat.complement` was being treated as if it had signature `Nat -> Nat -> Nat` for typechecking purposes, when it should be `Nat -> Nat`.  Similarly `Int.complement`.

## Implementation notes

2 line fix to Builtin.hs.

None.  

## Test coverage

Sometime soon I'll PR adding some stuff to base that will depend on these, so we should be OK.  

## Loose ends

Nothing.

### DIFF
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -238,7 +238,7 @@ builtinsSrc =
   , B "Int.and" $ int --> int --> int
   , B "Int.or" $ int --> int --> int
   , B "Int.xor" $ int --> int --> int
-  , B "Int.complement" $ int --> int --> int
+  , B "Int.complement" $ int --> int
   , B "Int.increment" $ int --> int
   , B "Int.isEven" $ int --> boolean
   , B "Int.isOdd" $ int --> boolean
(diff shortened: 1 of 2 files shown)


pchiusano:
@atacratic good catch. Can you add a type annotation to the tuple that’s returned by the existing tests? Or maybe return a list rather than a tuple? (I can’t remember if the test harness can handle this) That would have caught this error in the first place.

aryairani:
> Can you add a type annotation to the tuple that’s returned by the existing tests? Or maybe return a list rather than a tuple? (I can’t remember if the test harness can handle this) That would have caught this error in the first place.

I think adding type annotations too the tuple sounds too painful — can we instead check at the end of evaluating a watch expression, typecheck the output (which should be valid unison syntax, right?) and verify that it has the same type as the input watch expression; otherwise throw a bug message implicating the runtime.

aryairani:
By "we", I don't mean in this PR.  @pchiusano wdyt

pchiusano:
I like idea of decompiling the result and retypechecking it to make sure it matches but yeah that’s a bit bigger task. I think changing the results to a list instead of a tuple might be easy assuming that works. 

@atacratic if you don’t think you’ll get to this today or tomorrow we will merge as is. Thanks again for spotting.

aryairani:
Is the idea: use list instead of a tuple for testing the outputs of the subset of `Int` builtins that return `Int`? (i.e. leave out the ones that return `Boolean`, `Nat`, and `Float`)?  🤔

atacratic:
yeah sorry I think I've timed out for a few days now!

pchiusano:
@atacratic No worries!

@aryairani Well, I guess put all the ones expected to return a value of the same type together in a list, so the result type will be a tuple of lists.

PR #1493, opened 2020-04-30
Fix slowness on viewing docs

### LINKED ISSUE #1481: Viewing a `Doc` that references a data constructor hangs UCM
Unison

```unison
doc = [: @None :]
```

UCM:
```ucm
.> add
.> view doc
```

UCM is pegged at 100% CPU and is unresponsive.

runarorama:
Fixes #1481 

## Implementation notes

1. `generalizedDependencies` was just doing the wrong thing for `TermLink`, so foldHistoryUntil had no hope of finding some of the terms coming out of that. Fixed.
2. Made `foldHistoryUntil` skip nodes it doesn't need to load (or already has loaded). Avoids loading the causals from disk that have hashes it has already seen.

## Test coverage

This is hard to test unless you have a codebase with a lot of history. And even then, the test "fails" by just not terminating. Would love to hear ideas about how to test this kind of thing in general.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Causal.hs
+++ b/parser-typechecker/src/Unison/Codebase/Causal.hs
@@ -339,25 +339,39 @@ unsafeMapHashPreserving f c = case c of
   Cons h e (ht, tl) -> Cons h (f e) (ht, unsafeMapHashPreserving f <$> tl)
   Merge h e tls -> Merge h (f e) $ Map.map (fmap $ unsafeMapHashPreserving f) tls
 
+data FoldHistoryResult a = Satisfied a | Unsatisfied a deriving (Eq,Ord,Show)
+
 -- foldHistoryUntil some condition on the accumulator is met,
 -- attempting to work backwards fairly through merge nodes
 -- (rather than following one back all the way to its root before working
 -- through others).  Returns Unsatisfied if the condition was never satisfied,
 -- otherwise Satisfied.
-data FoldHistoryResult a = Satisfied a | Unsatisfied a deriving (Eq,Ord,Show)
-foldHistoryUntil :: forall m h e a. (Monad m) => --(Show a, Show e) =>
-  (a -> e -> (a, Bool)) -> a -> Causal m h e -> m (FoldHistoryResult a)
+foldHistoryUntil
+  :: forall m h e a
+   . (Monad m)
+  => (a -> e -> (a, Bool))
+  -> a
+  -> Causal m h e
+  -> m (FoldHistoryResult a)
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -339,25 +339,39 @@ unsafeMapHashPreserving f c = case c of
...
   Merge h e tls -> Merge h (f e) $ Map.map (fmap $ unsafeMapHashPreserving f) tls
 
+data FoldHistoryResult a = Satisfied a | Unsatisfied a deriving (Eq,Ord,Show)
+
 -- foldHistoryUntil some condition on the accumulator is met,
 -- attempting to work backwards fairly through merge nodes
 -- (rather than following one back all the way to its root before working
 -- through others).  Returns Unsatisfied if the condition was never satisfied,
 -- otherwise Satisfied.
-data FoldHistoryResult a = Satisfied a | Unsatisfied a deriving (Eq,Ord,Show)
```
aryairani, line 347: moved above the comment

runarorama: The comment is on the function

unison-core/src/Unison/Term.hs
```diff
@@ -851,7 +851,12 @@ generalizedDependencies
 generalizedDependencies termRef typeRef literalType dataConstructor dataType effectConstructor effectType
   = Set.fromList . Writer.execWriter . ABT.visit' f where
   f t@(Ref r) = Writer.tell [termRef r] $> t
-  f t@(TermLink r) = Writer.tell [termRef $ Referent.toReference r] $> t
+  f t@(TermLink (Referent.Ref r)) =
+    Writer.tell [termRef r] $> t
+  f t@(TermLink (Referent.Con r id CT.Data))
+    = Writer.tell [dataConstructor r id] $> t
+  f t@(TermLink (Referent.Con r id CT.Effect))
+    = Writer.tell [effectConstructor r id] $> t
```
aryairani, line 859: Maybe more readable if we  just have one equation and then `case r of` to get the RHSs onto the RHS, so to speak.

parser-typechecker/src/Unison/Codebase/Causal.hs
aryairani: 👍

aryairani: Yeah, not a critique, just when viewing the diff it looked like the moved version appeared from nowhere, because this had originally been buried in with the function.

pchiusano:
Nice!!

Issue #1484, opened 2020-04-29
a messed up codebase?


aryairani:
a codebase to look for anything important in, once we have a codebase history browser.

Current status:  Where's my code?  Because I thought there had been some.

```
15:44 $ ucm

   _____     _             
  |  |  |___|_|___ ___ ___ 
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|
  
  Welcome to Unison!
  
  I'm currently watching for changes to .u files under ~/Desktop/tsv2md
  
  Type help to get help. 😎

.> ls

  1.  ()/              (0 definitions)
  2.  .                ((b ->{𝕖} c) -> (a ->{𝕖} b) -> a ->{𝕖} c)
  3.  <|               ((a ->{𝕖} b) -> a ->{𝕖} b)
  4.  Abort            (type)
  5.  Abort/           (6 definitions)
  6.  Ask              (type)
  7.  Ask/             (4 definitions)
  8.  Author           (type)
  9.  Author/          (8 definitions)
  10. Boolean          (builtin type)
  11. Boolean/         (1 definition)
  12. Bytes            (builtin type)
  13. Bytes/           (9 definitions)
  14. Char             (builtin type)
  15. Char/            (2 definitions)
  16. CopyrightHolder  (type)
  17. CopyrightHolder/ (7 definitions)
  18. Debug/           (1 definition)
  19. Doc              (type)
  20. Doc/             (7 definitions)
  21. Either           (type)
  22. Either/          (3 definitions)
  23. Exception        (type)
  24. Exception/       (6 definitions)
  25. Float            (builtin type)
  26. Float/           (36 definitions)
  27. Function/        (2 definitions)
  28. GUID             (type)
  29. GUID/            (1 definition)
  30. Heap             (type)
  31. Heap/            (12 definitions)
  32. Int              (builtin type)
  33. Int/             (34 definitions)
  34. IsPropagated     (type)
  35. IsPropagated/    (1 definition)
  36. IsTest           (type)
  37. IsTest/          (1 definition)
  38. License          (type)
  39. License/         (10 definitions)
  40. LicenseType      (type)
  41. LicenseType/     (1 definition)
  42. Link             (type)

  1.  ()/              (0 definitions)
  2.  .                ((b ->{𝕖} c) -> (a ->{𝕖} b) -> a ->{𝕖} c)
  3.  <|               ((a ->{𝕖} b) -> a ->{𝕖} b)
[...]

PR #1501, opened 2020-05-01
don't overwrite root on sync

### LINKED ISSUE #1203: pr.create command lists useless numbers
```
  The changes summarized below are available for you to review, using the following command:
  
    `pr.load https://github.com/unisonweb/base https://github.com/puffnfresh/base`
  
  Added definitions:
  
    1. Either.swap      : Either a b -> Either b a
    2. Tuple.swap       : Tuple a b -> Tuple b a
    3. Weighted.eithers : Weighted a -> Weighted b -> Weighted (Either a b)
    4. Weighted.tuples  : Weighted a -> Weighted b -> Weighted (Tuple a b)
```

The numbers here refer to an in-memory branch which is gone by the time they are printed, so you'll get errors if you try to interact with them.
```
.> view 1

  ⚠️
  
    .Either.swap#8hr15i0otmkr4npjlva0c06lgvif4qagktr9tl4cls6och8uellgtoef614rivqj9d3s27ajilc6i0b2fpu4j3ol2n4hutc0ok1727g
```

Suggest passing a flag to the output to suppress the numbering?

aryairani:
## Overview

For background: when we do a `push`, we have been doing fetch from git, loading its root `Branch`, and doing an `updateAtM` on it to update it at the target path, which also produces a new root `Branch`, which we pass to `syncToDirectory`, which writes the branch's dependencies to the staging dir, and then does `updateCausalHead` on the target, to actually set it as its new root branch.

When I reused this same code for `syncFromDirectory`, I forgot to remove the call to `updateCausalHead`, as that is not something we want to do when ingesting a remote repo; that potentially deeply-nested Branch we are ingesting is not likely to be what we want for the new root of our local codebase.

This PR shuffles that call around so that it is not called during sync, but rather after the call to `syncToDirectory` when doing a `push`.

Although the call to `importRemoteBranch` will mess up our local causal head as described above, the final call to `mergeBranchAndPropagateDefaultPatch` was fixing it again, if it was the sort of merge that would cause us to update our root.  As @runarorama suspected, when the merge is a no-op, then this coincidentally corrective step is skipped, and the paths _head is left in the wrong state.

### How to fix your messed up codebase if you encounter this bug

Although when I tried to fix my messed up codebase after #1484, my reflog was mysteriously empty, and although @runarorama had seemed to recall that his reflog was also empty, I have not been able to replicate the "missing reflog" symptom, which leads me to believe that this bug does not impact the reflog, and the reflog can therefore be used to restore your previous head.  

image

You can then re-issue the `pull` with this patch — or maybe even without this patch? (see next section).

## Implementation notes

Just don't call `updateCausalHead` in `syncToDir`:

image
image

## Test coverage

I could try to figure out what the root namespace is supposed to be after a `push` in the git test, and then check for the file, but haven't.

## Loose ends

- [ ] try to fully understand and/or improve the code that does detection of external changes? (`CommandLine.watchBranchUpdates`, `Input.IncomingRootBranch`)
- [ ] update tests to check _head?

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/Git.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Git.hs
@@ -35,7 +35,7 @@ import qualified Unison.Util.Exception         as Ex
 import qualified Unison.Codebase.Branch        as Branch
 import UnliftIO.IO (hFlush, stdout)
 import UnliftIO.Directory (getXdgDirectory, XdgDirectory(XdgCache), doesDirectoryExist, findExecutable, removeDirectoryRecursive)
-import Unison.Codebase.FileCodebase.Common (encodeFileName)
+import Unison.Codebase.FileCodebase.Common (encodeFileName, updateCausalHead, branchHeadDir)
 
 tempGitDir :: MonadIO m => Text -> m FilePath
 tempGitDir url =
@@ -213,6 +213,7 @@ pushGitRootBranch codebase branch repo = do
     let repoString = Text.unpack $ printRepo repo
     withStatus ("Staging files for upload to " ++ repoString ++ " ...") $
       lift (Codebase.syncToDirectory codebase remotePath branch)
+    updateCausalHead (branchHeadDir remotePath) (Branch._history branch)
     -- push staging area to remote
     withStatus ("Uploading to " ++ repoString ++ " ...") $
       unlessM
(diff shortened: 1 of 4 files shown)


aryairani:
lol.  certain kinds of push can close a github PR, and _disable_ the "Reopen" button.

aryairani:
@pchiusano @runarorama Let me know what you guys think about whether to pursue those 3 loose ends.  I think they're worth pursuing, but don't feel at all like doing them solo, so if I'm the only one, then we can just merge as-is.

aryairani:
Great.  Well, I pursued and resolved the last one about ... well I already deleted it.  I think it was about figuring out why this only happened sometimes.

aryairani:
> I think the external branch events stuff will come up again if it turns out the warning is still getting triggered so we can “lazily evaluate” staring at that code until later.

Hehe we didn't get 24 hours in before #1509

Issue #1488, opened 2020-04-30
Pulling from base overwrites the root namespace


runarorama:
```
.> ls

  1. _crs/    (207 definitions)
  2. base/    (749 definitions)
  3. lib/     (22 definitions)
  4. mybase/  (1535 definitions)
  5. runar/   (10 definitions)
  6. scratch/ (1 definition)
  7. work/    (48 definitions)

.> pull https://github.com/unisonweb/base:.trunk base

  Nothing changed as a result of the merge.


  😶

  base was already up-to-date with https://github.com/unisonweb/base:.trunk.

.>
.> ls

  1. _crs/    (207 definitions)
  2. base/    (749 definitions)
  3. lib/     (22 definitions)
  4. mybase/  (1535 definitions)
  5. runar/   (10 definitions)
  6. scratch/ (1 definition)
  7. work/    (48 definitions)

.> exit
~/w/unison (master|✚1) ❯❯❯ stack exec unison

   _____     _
  |  |  |___|_|___ ___ ___
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|

  Welcome to Unison!

  I'm currently watching for changes to .u files under ~/work/unison

  Type help to get help. 😎

.> ls

  1.  ()/              (0 definitions)
  2.  .                ((b ->{𝕖} c) -> (a ->{𝕖} b) -> a ->{𝕖} c)
  3.  <|               ((a ->{𝕖} b) -> a ->{𝕖} b)
  4.  Abort            (type)
  5.  Abort/           (6 definitions)
  6.  Ask              (type)
  7.  Ask/             (4 definitions)
  8.  Author           (type)
  9.  Author/          (8 definitions)
  10. Boolean          (builtin type)
```

It says "nothing changed as a result", but actually my whole codebase is gone after restarting ucm.

aryairani:
damn, that happened to me too.  ref #1484

aryairani:
@runarorama Thanks for this clue; you should be able to restore your codebase using the reflog

runarorama:
I can reproduce reliably, fyi

aryairani:
@runarorama Could you link a transcript as a gist?  That was gonna be my next step, but I haven't done it yet.  It's just for fix development purposes, so it can be slow; do pulls or whatever.

runarorama:
Transcript is just:

```ucm
.> pull https://github.com/unisonweb/base:.trunk base
```

runarorama:
[...]

PR #1503, opened 2020-05-01
uncomment and improve the WarnIncomingRootBranch message

aryairani:
## Overview

Uncommented this message to give the user a little more warning and info to be able to deal with issues like #1484 / #1488.  Anecdotally it seems to work, but I'm not confident that it avoids false positives.  I don't remember why we/I had commented it out to begin with.

image

## Test coverage

A transcript to trigger this would require leaving `ucm` open while simultaneously changing some files.  I guess this could be done using pure Unison, the `{IO}` ability.  I suppose we could write a bunch of trickier tests that way.

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
+++ b/parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
@@ -264,7 +264,7 @@ notifyUser dir o = case o of
       , ""
       , P.wrap $ "but I'm not sure what to do about it."
           <> "If you're feeling lucky, you can try deleting one of the heads"
-          <> "from `.unison/v1/branches/head/`, but please make a backup first."
+          <> "from `.unison/v1/paths/head/`, but please make a backup first."
           <> "There will be a better way of handling this in the future. 😅"
       , ""
       , P.wrap $ "For what it's worth, the hash of the root namespace that's"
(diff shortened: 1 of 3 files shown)


aryairani:
@pchiusano @runarorama  Do you guys have any concerns about potential false positives or lack of testing that should be addressed before merging this?

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -250,25 +250,33 @@ prettyRemoteNamespace =
...
+      [ P.wrap $ (if length hashes == 1 then "A" else "Some")
+         <> "codebase" <> P.plural hashes "root" <> "appeared unexpectedly"
+         <> "with" <> P.group (P.plural hashes "hash" <> ":")
+      , ""
+      , (P.indentN 2 . P.oxfordCommas)
+                (map prettySBH $ toList hashes)
+      , ""
+      , P.wrap $ "but I'm not sure what to do about it."
+          <> "If you're feeling lucky, you can try deleting one of the heads"
+          <> "from `.unison/v1/branches/head/`, but please make a backup first."
```
pchiusano, line 267: paths not branches here

aryairani: I fixed this and then ended up rewriting it completely:

```
.> 

  A codebase root appeared unexpectedly with hash:
  
    #7asfbtqmoj
  
  and I'm not sure what to do about it. The last root namespace hash that I knew about was:
  
    #l35l361e6c
  
  Now might be a good time to make a backup of your codebase. 😬
  
  After that, you might try using the `fork` command to inspect the namespaces listed above, and
  decide which one you want as your root. You can also use `reflog` to see the last few root
  namespace hashes on record.
  
  Once you find one you like, you can use the `reset-root` command to set it.
```

PR #1528, opened 2020-05-05
Greatly speed up `pull` command by short-circuiting merge

pchiusano:
Before this change, this would take about a minute, even if all the code was in the git cache and in the codebase:

```ucm
.> pull git@github.com:runarorama/mybase .runarbase
```

Now it takes a couple seconds, and I think most of that might be printing out the diff.

Pretty simple change, using the fact that `Branch.merge empty b == b`. The old code didn't have this. As a result, for the very common case of pulling to an empty namespace (which is done, for instance, by `pr.load` and `pr.create`), it would be computing the LCA of a namespace, `b`, and `Branch.empty`, which will involve a full scan of `b` history, only to determine that `Branch.empty` is the LCA. :)

I speculate that this might also fix other sources of slowness in `push` / `pull`. Basically, there are a lot of times where you can end up doing a merge of an empty `Branch` and a non-empty one. I can sort of imagine there being some quadratic behavior, where for each point in history, you look for the LCA with the empty namespace (requiring a search all the way back in history).

## Interesting stuff

A couple of the transcript outputs changed. Since we no longer create pointless merge nodes when merging with the empty namespace, this affects namespace hashes.

## Test coverage

No new tests, though any of the tests that do `pull` will exercise.

## Loose ends

None.

### DIFF
--- a/unison-src/transcripts/merges.output.md
+++ b/unison-src/transcripts/merges.output.md
@@ -112,13 +112,13 @@ We can also delete the fork if we're done with it. (Don't worry, it's still in t
   Note: The most recent namespace hash is immediately below this
         message.
   
-  ⊙ #ll42hbp60e
+  ⊙ #1pdgtsicq9
   
     - Deletes:
     
       feature1.y
   
-  ⊙ #h72ma1qlnk
+  ⊙ #9e8gvv7r3e
   
     + Adds / updates:
     
@@ -129,26 +129,26 @@ We can also delete the fork if we're done with it. (Don't worry, it's still in t
       Original name New name(s)
       feature1.y    master.y
   
-  ⊙ #km35ctvltc
+  ⊙ #6sqsddkikv
   
     + Adds / updates:
     
       feature1.y
   
-  ⊙ #nbiibp60kj
+  ⊙ #loe9apjdis
   
     > Moves:
     
       Original name New name
       x             master.x
   
-  ⊙ #d3oe0kr57t
+  ⊙ #p90pbjip7l
   
     + Adds / updates:
     
       x
   
-  ⊙ #gectfnld64
+  ⊙ #givahf3f6f
   
     + Adds / updates:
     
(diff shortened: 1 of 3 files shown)


runarorama:
Do want

unison-src/transcripts/merges.output.md
```diff
@@ -112,13 +112,13 @@ We can also delete the fork if we're done with it. (Don't worry, it's still in t
   Note: The most recent namespace hash is immediately below this
         message.
   
-  ⊙ #ll42hbp60e
+  ⊙ #1pdgtsicq9
```
runarorama, line 115: Why do these hashes change?

aryairani:
This looks great, but I hope it won't prevent us from figuring out why our LCA search is hugely slower than git's.

The change in hashes also surprises me, if you `merge a b` when `before  a b` (and isn't `before empty b`?), I would have expected no merge node to begin with.

aryairani:
@pchiusano any idea on the merge node thing?

pchiusano:
Yes see the “interesting stuff” section of the writeup. 😀

pchiusano:
Oh sorry I read your replies out of order. That is a good question re: merge nodes. 

One factor is that we used to have all the builtins in the root of the history, impossible to remove. That wouldn’t affect the tests though.

pchiusano:
I think this code is worth a close look, might be wrong: https://github.com/unisonweb/unison/blob/86d817a3f3f1c975dabe68f379be708b15647678/parser-typechecker/src/Unison/Codebase/Causal.hs

Note that even if that code is wrong, this PR is still a good idea: avoids a scan of history to determine that empty branch is before everything else.

aryairani:
> I think this code is worth a close look, might be wrong: https://github.com/unisonweb/unison/blob/86d817a3f3f1c975dabe68f379be708b15647678/parser-typechecker/src/Unison/Codebase/Causal.hs

Did you mean to refer to certain lines, or just the whole file?

> Note that even if that code is wrong, this PR is still a good idea: avoids a scan of history to determine that empty branch is before everything else.

Noted.

pchiusano:
Ooop, meant this link: https://github.com/unisonweb/unison/blob/86d817a3f3f1c975dabe68f379be708b15647678/parser-typechecker/src/Unison/Codebase/Causal.hs#L267-L285

PR #1543, opened 2020-05-09
Make dependents and dependencies commands produce numbered args.

### LINKED ISSUE #1408: let `dependents` and `dependencies` command produce numbered args
and update transcript to exercise
see #1401

pt2121:
## Overview

close #1408

Screen Shot 2020-05-10 at 3 24 35 PM

~~I assume the number is a reference to the name not the hash and am not sure how we want to handle `missing`, `(no name available)`.~~

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -1687,7 +1687,9 @@ loop = do
               tm (Referent.Con r _i _ct) = eval $ GetDependents r
               in LD.fold tp tm ld
             (missing, names0) <- eval . Eval $ Branch.findHistoricalRefs' dependents root'
-            respond $ ListDependents hqLength ld names0 missing
+            let names = getTypesAndTerms names0
+            numberedArgs .= fmap (Name.toString . fst) names
+            respond $ ListDependents hqLength ld names missing
       ListDependenciesI hq -> do -- todo: add flag to handle transitive efficiently
         resolveHQToLabeledDependencies hq >>= \lds ->
           if null lds
@@ -1709,7 +1711,9 @@ loop = do
               tm _ = pure mempty
               in LD.fold tp tm ld
             (missing, names0) <- eval . Eval $ Branch.findHistoricalRefs' dependencies root'
-            respond $ ListDependencies hqLength ld names0 missing
+            let names = getTypesAndTerms names0
+            numberedArgs .= fmap (Name.toString . fst) names
+            respond $ ListDependencies hqLength ld names missing
       DebugNumberedArgsI -> use numberedArgs >>= respond . DumpNumberedArgs
       DebugBranchHistoryI ->
         eval . Notify . DumpBitBooster (Branch.headHash currentBranch') =<<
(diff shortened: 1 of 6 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1688,7 +1688,7 @@ loop = do
               in LD.fold tp tm ld
             (missing, names0) <- eval . Eval $ Branch.findHistoricalRefs' dependents root'
             let names = getTypesAndTerms names0
-            numberedArgs .= fmap (Name.toString . fst) names
+            numberedArgs .= fmap (show . Reference.showShort hqLength) ((fmap snd names) <> toList missing)
```
aryairani: For `numberedArgs`, since they're hidden, I'd go ahead and use the full hashes:
```haskell
Text.unpack . Reference.toText
```

pt2121:
Sounds good. I'll update.

pt2121:
Noob question: is there an easy way to get into the missing name situation?

aryairani:
Add one definition that depends on another, and then delete one of the
names.  (Actually I'm not sure if it will let you?)

pt2121:
Looks like I can't delete a name that has dependencies.

pt2121:
@aryairani when you have a moment, can you take a quick look at this in case you have any thoughts?

aryairani:
Hi @pt2121 sorry for the delay and the semi-impossible request.  

One way you can end up with unnamed dependencies is by pushing a namespace that depends on stuff that's outside of it, and then pulling just what was pushed.  That will include the dependencies, but as of now, does not include their names; thus: unnamed dependencies.

Because this involves doing a push, it can't be performed within a single transcript, but see `parser-typechecker/tests/Unison/Test/Git.hs` for an overly-complicated example a related pair of transcripts that rely on the creation and cleanup of a temporary git repo.

But I logged this gist on your branch, and it seems okay to me.  Could you take a look and let me know if anything unusual jumps out at you? https://gist.github.com/aryairani/92dbd30ca258b487bd2e9eed7c3a45cf

pt2121:
Oh, nice. That looks good to me. Thanks!

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1745,6 +1749,12 @@ loop = do
       notImplemented = eval $ Notify NotImplemented
       success = respond Success
 
+      getTypesAndTerms :: Names0 -> [(Name, Reference)]
+      getTypesAndTerms names0 = types <> terms
+        where
+          types = R.toList $ Names3.types0 names0
+          terms = fmap (second Referent.toReference) $ R.toList $ Names.terms names0
+
```
aryairani, line 1757: Could you move this up into `ListDependenciesI`, just because it's a little specific — also note that `Referent.toReference` doesn't always produce term references (specifically: a `Referent.Con` yields a type `Reference`), so in general this function isn't doing the right thing.  

However, it's not a bug in this case, because the `Reference`s that come out `dependencies` and then go into `names0` are always `Referent.Ref`, so which does produce a term `Reference` using `Referent.toReference`.  The `findHistoricalRefs'` functions don't make this obvious, so I apologize.

But in any case, let's make `getTypesAndTerms` local to `ListDependenciesI` for now? And maybe add a note to this effect ^

aryairani: oops, mentioned the wrong username initially above. @pt2121

pt2121:
Thanks for the review!

PR #1545, opened 2020-05-11
Improved three-way merge

runarorama:
This improves `threeWayMerge` on namespaces, as well as the `before` function on namespace histories. Namely, the new `threeWayMerge` doesn't waste time calling `before`, and the new `before` is simply implemented in terms of `lca` (lowest common ancestor).

These implementations are significantly simpler and should be somewhat faster.

## Test coverage

No new tests are added, but a number of tests on `threeWayMerge` are passing.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Causal.hs
+++ b/parser-typechecker/src/Unison/Codebase/Causal.hs
@@ -9,11 +9,8 @@ import           Prelude                 hiding ( head
                                                 , tail
                                                 , read
                                                 )
-import           Control.Lens                   ( (<&>) )
-import           Control.Monad.Loops            ( anyM )
 import qualified Control.Monad.State           as State
 import           Control.Monad.State            ( StateT )
-import           Data.List                      ( foldl1' )
 import           Data.Sequence                  ( ViewL(..) )
 import qualified Data.Sequence                 as Seq
 import           Unison.Hash                    ( Hash )
@@ -21,7 +18,6 @@ import qualified Unison.Hashable               as Hashable
 import           Unison.Hashable                ( Hashable )
 import qualified Data.Map                      as Map
 import qualified Data.Set                      as Set
-import           Util                           ( bind2 )
 
 {-
 `Causal a` has 5 operations, specified algebraically here:
@@ -34,7 +30,7 @@ import           Util                           ( bind2 )
 * `cons : a -> Causal a -> Causal a`, satisfying `head (cons hd tl) == hd` and
           also `before tl (cons hd tl)`.
 * `merge : CommutativeSemigroup a => Causal a -> Causal a -> Causal a`, which is
-           associative and commutative and satisfies:
+           commutative (but not associative) and satisfies:
   * `before c1 (merge c1 c2)`
   * `before c2 (merge c1 c2)`
 * `sequence : Causal a -> Causal a -> Causal a`, which is defined as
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -34,7 +30,7 @@ import           Util                           ( bind2 )
 * `cons : a -> Causal a -> Causal a`, satisfying `head (cons hd tl) == hd` and
           also `before tl (cons hd tl)`.
 * `merge : CommutativeSemigroup a => Causal a -> Causal a -> Causal a`, which is
-           associative and commutative and satisfies:
+           commutative (but not associative) and satisfies:
```
aryairani, line 33: Can we make it associative, passing a list of `Causal` to to `threeWayMerge` and `lca`?  (Doesn't have to be this PR.)

unison-src/transcripts/mergeloop.output.md
```diff
@@ -136,9 +136,8 @@ b = 2
   namespace hash.
   
   #0lf1cvdccp
-  #2quan5n72t
-  #gjfovomom2
+  #ofcsecdak0
   ⑂
-  ⊙ #bpfcuip0ru
+  ⊙ #0ucrusr0bl
```
aryairani, line 141: Why did it change from `Merge [#olf1, #2qua, #gjfo]` to `Merge [#olf1, #bpfc]?

runarorama: Realizing that merges are not associative, all merges are now three-way rather than n-way

parser-typechecker/src/Unison/Codebase/Causal.hs
runarorama: We cannot, as the three-way merge operation is not associative and never was.

aryairani: I see; ok.

aryairani: Would you explain it for me though, and for posterity?  i.e. is there an n-way merge operation we could use that is associative?  I'm guessing no, but not sure why.

runarorama: The issue I was seeing was that merges were actually incorrect with the old algorithm. The reason is that when merging two n-way merge nodes, you don’t want to consider the LCA of all the nodes together. That might give you a very distant ancestor. I was seeing where merging two such nodes resulted in objects being “added back” after one branch removed them, since it appeared to the merge algorithm that they were new in one node and simply absent in the other.

pchiusano: If you picked a single LCA for all the nodes being merged and then implemented an N-way LCA based merge then that would probably be associative. That’s not what you want though since the common LCA among all N nodes is worse than the LCA for the actual pairwise merges the user did. It’s also quadratic I think - every time a new thing gets merged into a chain of N merges you gotta repeat the work.

Fundamentally, LCA selection is dependent on grouping and affects results of a merge, so that kills associativity in general. Though in lots of specific instances, merge will be associative.

runarorama: Consider this case:

1. The root of the tree is `r`, an empty codebase.
2. The node `n` adds a definition `x`.
3. `n` is the LCA of `a` and `b`, where `a` adds `y` and `b` removes `x`.
4. `c` is a direct descendant of `r` and adds `z`.

Now: 
`a` has `{x,y}`
`b` has `{y}`.
`c` has `{z}`

A correct merge of `{a,c}` is `ac = {x,y,z}`.
A correct merge of `{b,c}`, is `bc = {y,z}`.
A correct merge of `{a,b}`, is `ab = {y}`.

Merge of `{a,bc}` is `{x,y,z}`
Merge of `{ab,c}` is `abbc = {y,z}`

Ergo, three-way merge is not associative.

Issue #1547, opened 2020-05-12
What to do about missing definitions in existing codebases?


aryairani:
e.g. the codebase in #1512 is missing 6:

```
MissingTerm #dk56i -- old doc for Optional.isSome
MissingTerm #fm4rr
MissingTerm #gvvm0 -- old doc for Optional.isNone.
MissingTerm #h83dg
MissingTerm #kccnr
MissingTerm #ucsc8
```

We can debug the others with the `topic/bfs-sync-debug-missing-base-deps` branch.

Issue #1535, opened 2020-05-07
Slowness with push / pull / merge and related commands


pchiusano:
As the base library has gotten larger, we've started to notice serious slowness with `push`, `pull`, `merge`, and related commands like `pr.create` and `pr.load` that are defined in terms of them. These commands should take in the 5-20s range, not minutes. This ticket tracks work we are doing to track down what's slow and come up with fixes.

Related tickets: #1276

* Phases of a pull:
  * __Git fetch:__ Git operation to fetch the latest commit of the remote
  * __Sync:__ Copy definitions that are present in the source namespace but not present in the destination namespace.
  * __Merge:__ Merge of the source namespace and the destination namespace (this calls `Causal.before`)
  * Anything else??
* Phases of a push:
  * __Git fetch:__ Git operation to fetch the latest commit of the remote (note: this is the same as a `pull`)
  * __Merge:__ Merge the source namespace and destination namespace. (this calls `Causal.before`)
  * __Git fetch:__ Apart from a round-trip over the network to the remote git server, this should be a no-op if there are no new commits on the remote since the previous __Git fetch__.
  * __Verify fast-forward:__ Reject unless it's a fast forward (calls `Causal.before` to check).
  * __Sync:__ Copy definitions that are present in the source namespace but not present in the destination namespace.
  * __Git push:__ Git operation to add a local commit and push it to the remote repo.

__Scenarios:__

1. `.> pull git@github.com:unisonweb/base:.trunk base` on an empty repo
2. `.> pull git@github.com:unisonweb/base:.trunk base2` on the same repo as 1
3. Some stuff 

__Results:__

### `.> pull <base> .base` in empty repo

__base git commit 245f053, 3/11/2020__

|base git commit       |245f053|CFI|3c928b7|CFI|
|--------              | ----- |---|------ |---|
|**date**              |**11-Mar**           ||**06-May**   ||
|branchFromFiles       | 0.03s / 0.03s       |0.03s|1.3s / 1.45s ||
|syncFromDirectory     | 4.2s / 4.34s        |0.15s|343s / 421s  ||
| --- LoadDependentsDir |                    |0.02s|343s / 421s  ||
| --- CopyTransitiveDeps|                    |0.00s|343s / 421s  ||
| --- Copy Dependents Index|                 |0.00s|343s / 421s  ||
| --- Copy Type Index   |                    |0.03s|343s / 421s  ||
| --- Copy Type Mentions|                    |0.09s|343s / 421s  ||
[...]

PR #1548, opened 2020-05-12
queue-copy sync algorithm

### LINKED ISSUE #1423: ucm should also sync dependencies of `patch` objects

### LINKED ISSUE #1512: Slow push with oldish codebase
Attempting to push the codebase I've attached to this issue seems to cause a memory leak. Issue solved by reinitializing codebase.
[codebase.tar.gz](https://github.com/unisonweb/unison/files/4569456/codebase.tar.gz)

### LINKED ISSUE #1546: pull base.trunk into _base hangs
Is pulling the latest `trunk` of `base` working for others? I have tried this on two separate computers (both Macbooks on Catalina, but this worked on both of them last week), and have even completely removed the `.unison` directory first in case I had some bad state.

```
❯ rm -rf ~/.unison

~
❯ ucm version
ucm version: release/M1l

~
❯ ucm
No codebase exists in /Users/cody.allen
Run `ucm -codebase /Users/cody.allen init` to create one, then try again!

~
❯ ucm -codebase /Users/cody.allen init
Initializing a new codebase in: /Users/cody.allen

~
❯ ucm

   _____     _             
  |  |  |___|_|___ ___ ___ 
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|
  
  Welcome to Unison!
  
  I'm currently watching for changes to .u files under ~
  
  Type help to get help. 😎
  
  Enter pull https://github.com/unisonweb/base:.releases._M1l .base to set up the default base
  library. 🏗

.> pull git@github.com:unisonweb/base:.trunk _base
  Importing downloaded files into local codebase...
```

Pulling the git repo works (fairly quickly), but then it seems to indefinitely stay on `Importing downloaded files into local codebase...`. I've let it go for about 10 minutes, and I don't see any signs of progress. During this time it's using pretty significant CPU and memory resources (though not fully pegged).

aryairani:
## Overview

This replaces the `Causal.sync`-based algorithm that had underlaid ucm's git syncing.

It is way faster and uses way less memory.

The algorithm has three phases:
1. `processBranches`: Work through a queue of branches (which may be in-memory, or on-disk), loading them, aggregating referenced terms (including metadata values), decls, and patches.  Any ancestor branches or child branches encountered are pushed back onto this queue.
2. `processDependencies` Work through through a queue of patches, terms, and decls, loading each, adding its dependencies onto this queue, and building `dependent-index`, `type-index`, and `type-mentions-index` relations in memory.
3. Write the three index relations from memory to the destination codebase.

Some nice timing data at https://github.com/unisonweb/unison/issues/1535#issuecomment-628061034

Closes #1546 "`pull base:.trunk _base` hangs"
Helps #1535 "slowness with push / pull / merge"
Closes #1512 "slow push of oldish codebase"
Closes #1423 "should sync dependents of patches"

## Test coverage

I beefed up the git tests a little, but we could use a regression test to ensure that metadata is synced — although we have seen in #1547 that such metadata is noticed when missing, so it probably syncs fine; just no regression test yet.

## Loose ends

- [ ] various cleanup: docs, organization maybe?
- [x] bombs on some codebases that had been allowed to synced previously despite missing dependencies

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
+++ b/parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
@@ -14,11 +14,12 @@ module Unison.Codebase.FileCodebase.SlimCopyRegenerateIndex (syncToDirectory) wh
 
 import Unison.Prelude
 
+import qualified Data.Set                      as Set
+import           Control.Lens
 import           Control.Monad.State            ( MonadState, evalStateT )
 import           Control.Monad.Writer           ( MonadWriter, execWriterT )
 import qualified Control.Monad.Writer          as Writer
 import           Control.Monad.Except           ( MonadError, runExceptT, throwError )
-import           Control.Lens
 
 import           UnliftIO.Directory             ( doesFileExist )
 import           Unison.Codebase                ( CodebasePath )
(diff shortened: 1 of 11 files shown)


parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
```diff
@@ -0,0 +1,305 @@
...
+syncToDirectory' :: forall m v a
+  . MonadIO m
+  => Var v
+  => S.Get v
+  -> S.Get a
+  -> CodebasePath
+  -> CodebasePath
+  -> Branch m
+  -> m ()
+syncToDirectory' getV getA srcPath destPath newRoot =
```
pchiusano, line 98: This can be future work, but I am guessing you could get additional speedups by parallelizing / pipelining the work. So have the main thread putting items to write on a queue, and one or more threads pulling from that queue and just copying the files.

ceedubs:
Great work @aryairani.

Now when I run `.> pull git@github.com:unisonweb/base:.trunk _base` instead of churning for a really long time, I get `unison: user error (MissingTerm #dk56i) codebase...`, which I believe is a known issue. When I run `pull https://github.com/unisonweb/base:#6cpa1csq7g _base`, everything seems to work 🎉.

parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
aryairani: Good idea.

parser-typechecker/src/Unison/Codebase/Serialization/V1.hs
```diff
@@ -183,6 +184,11 @@ getText = do
   bs <- getBytes len
   pure $ decodeUtf8 bs
 
+skipText :: MonadGet m => m ()
+skipText = do
```
pchiusano, line 188: I'd use: https://hackage.haskell.org/package/bytes-0.15.3/docs/Data-Bytes-Get.html#v:skip Seems more likely to have optimized implementation.

Even with the benefit of laziness, I think you'll still be allocating a thunk with the code that's written here.

parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
```diff
@@ -96,20 +95,18 @@ syncToDirectory' :: forall m v a
...
-                           . pure
-                           . Branch.transform (lift . lift . lift)
-                           ) newRoot )]
-      time "Sync Definitions" $ processDependencies (BD.to' deps)
-      time "Write indices" $ lift do
-        lift . writeDependentsIndex   =<< use dependentsIndex
-        lift . writeTypeIndex         =<< use typeIndex
-        lift . writeTypeMentionsIndex =<< use typeMentionsIndex
-    either (fail . show) pure result -- todo: nicer error messages here?
+  let warnMissingEntities = True in
```
aryairani, line 98: In the laptop rebooting chaos, I managed to let this slip in.

parser-typechecker/src/Unison/Codebase/Serialization/V1.hs
```diff
@@ -709,6 +715,47 @@ getRawBranch =
...
+  childHashes <- fmap (RawHash . snd) <$> getList (getPair skipText getHash)
+  editHashes <- Set.fromList . fmap snd <$> getList (getPair skipText getHash)
+  pure ( childHashes `zip` repeat Nothing
+       , BD.Dependencies editHashes (terms1 <> terms2) (types1 <> types2) )
+  where
+  -- returns things, metadata types, metadata values
+  getStarReferences ::
+    (MonadGet m, Ord r) => m r -> m ([r], [Metadata.Value])
+  getStarReferences getR = do
+    void $ getList getR -- throw away the `facts`
```
pchiusano, line 732: You might want a `skipR` just like you have a `skipText`, and `skipList` combinator as well. This line is scanning every byte in the fact dimension only to throw it all away.

PR #1549, opened 2020-05-13
Add caching layer to avoid hitting disk for namespace, term, and type loads from codebase

pchiusano:
## Overview

This PR adds caching to the Unison codebase, so that operations like `merge`, `pull`, and `push` hit the disk much less often, and similarly adds adds caching for terms, type of term, and decls.

It's pretty easy to come up with examples that are wildly faster (like a `merge` is now imperceptible where before it took 7-8s), but I want to call attention to a couple points that aren't necessarily obvious:

* Before this PR, when you do a `pr.load`, you're doing two pulls, followed by a merge. During the merge, you're doing a bunch of I/O to load the branches back to the LCA. Now, the initial pulls will populate and use the cache, and the merge will have everything cached and be super fast.
* A Unison namespace is actually a DAG, not a tree, since the same namespace may appear at multiple locations at a single point in history, and can also appear at multiple points in history. Our old `Branch.read` would read that DAG as a tree, leading to both duplicated I/O work (because the namespace was loaded from disk N times for each place it appears, rather than once) and additional memory usage (because each such I/O produced an unrelated in-memory value rather than sharing). 

There's a new config setting you can add to `.unisonConfig`: 

```
NamespaceCacheSize = 4096
```

4096 is the current default - that's 4096 namespaces in the cache before it starts cleaning up old ones.

## Implementation notes

The `Cache` implementation is in `Unison.Util.Cache`. There's a simple interface:

https://github.com/unisonweb/unison/blob/386bdc754a1e06079114424618fe003a2fb2f7c7/parser-typechecker/src/Unison/Util/Cache.hs#L10-L13

And then an interesting implementation:

https://github.com/unisonweb/unison/blob/386bdc754a1e06079114424618fe003a2fb2f7c7/parser-typechecker/src/Unison/Util/Cache.hs#L29-L59

And some helper functions for populating the cache in various ways:

https://github.com/unisonweb/unison/blob/386bdc754a1e06079114424618fe003a2fb2f7c7/parser-typechecker/src/Unison/Util/Cache.hs#L61-L69

There's a single "branch cache" created on startup and threaded ultimately to `Causal.cachedRead`:

https://github.com/unisonweb/unison/blob/386bdc754a1e06079114424618fe003a2fb2f7c7/parser-typechecker/src/Unison/Codebase/Causal.hs#L116-L140

Also see `Branch.cachedRead`, which calls `Causal.cachedRead`:

https://github.com/unisonweb/unison/blob/386bdc754a1e06079114424618fe003a2fb2f7c7/parser-typechecker/src/Unison/Codebase/Branch.hs#L470-L489

The rest of the changes are basically plumbing to get the single branch cache threaded to those points.

I left the timing logging messages in, but disabled, with a static flag in `Unison.Util.Timing` re-enable.

## Test coverage

The transcripts hit the cache now so getting some decent coverage there.

~~I think the `Cache` module could use unit tests.~~ This is done now.

## Loose ends

Not exactly a loose end, but I would be curious to spend some time looking at cache hit rates for various operations.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -82,6 +82,7 @@ module Unison.Codebase.Branch
 
     -- * Branch serialization
   , read
+  , cachedRead
   , sync
 
     -- * Unused
@@ -139,6 +140,7 @@ import           Unison.Referent                ( Referent )
 import qualified Unison.Referent               as Referent
 import qualified Unison.Reference              as Reference
 
+import qualified Unison.Util.Cache             as Cache
 import qualified Unison.Util.Relation          as R
 import           Unison.Util.Relation            ( Relation )
 import qualified Unison.Util.Relation4         as R4
@@ -454,6 +456,29 @@ data ForkFailure = SrcNotFound | DestExists
 -- could move a Names0 to a read-only field in Branch0 until it gets too big
 numHashChars :: Branch m -> Int
 numHashChars _b = 3
+
+-- todo: can delete old `read` once we are confident in this
+cachedRead :: forall m . Monad m
+           => Cache.Cache m (Causal.RawHash Raw)
+                            (Causal m Raw (Branch0 m))
+           -> Causal.Deserialize m Raw Raw
+           -> (EditHash -> m Patch)
+           -> Hash
+           -> m (Branch m)
+cachedRead cache deserializeRaw deserializeEdits h =
+ Branch <$> Causal.cachedRead cache d h
+ where
+  fromRaw :: Raw -> m (Branch0 m)
+  fromRaw Raw {..} = do
+    children <- traverse go _childrenR
+    edits <- for _editsR $ \hash -> (hash,) . pure <$> deserializeEdits hash
+    pure $ branch0 _termsR _typesR children edits
+  go = cachedRead cache deserializeRaw deserializeEdits
+  d :: Causal.Deserialize m Raw (Branch0 m)
+  d h = deserializeRaw h >>= \case
+    RawOne raw      -> RawOne <$> fromRaw raw
+    RawCons  raw h  -> flip RawCons h <$> fromRaw raw
+    RawMerge raw hs -> flip RawMerge hs <$> fromRaw raw
 
 -- Question: How does Deserialize throw a not-found error?
 -- Question: What is the previous question?
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -519,8 +519,8 @@ sync
   -> Branch m
   -> m ()
 sync exists serializeRaw serializeEdits b = do
-  written <- State.execStateT (sync' exists serializeRaw serializeEdits b) mempty
-  traceM $ "Branch.sync wrote " <> show (Set.size written) <> " namespace files."
+  _written <- State.execStateT (sync' exists serializeRaw serializeEdits b) mempty
+  -- traceM $ "Branch.sync wrote " <> show (Set.size written) <> " namespace files."
   pure ()
 
 -- serialize a `Branch m` indexed by the hash of its corresponding Raw
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -82,7 +82,6 @@ module Unison.Codebase.Branch
   , modifyPatches
 
     -- * Branch serialization
-  , read
   , cachedRead
   , boundedCache
   , Cache
@@ -467,7 +466,7 @@ type Cache m = Cache.Cache m (Causal.RawHash Raw) (Causal m Raw (Branch0 m))
 boundedCache :: MonadIO m => Word -> m (Cache m)
 boundedCache = Cache.semispaceCache
 
--- todo: can delete old `read` once we are confident in this
+-- Can use `Cache.nullCache` to disable caching if needed
 cachedRead :: forall m . Monad m
            => Cache m
            -> Causal.Deserialize m Raw Raw
@@ -483,29 +482,6 @@ cachedRead cache deserializeRaw deserializeEdits h =
     edits <- for _editsR $ \hash -> (hash,) . pure <$> deserializeEdits hash
     pure $ branch0 _termsR _typesR children edits
   go = cachedRead cache deserializeRaw deserializeEdits
-  d :: Causal.Deserialize m Raw (Branch0 m)
-  d h = deserializeRaw h >>= \case
-    RawOne raw      -> RawOne <$> fromRaw raw
-    RawCons  raw h  -> flip RawCons h <$> fromRaw raw
-    RawMerge raw hs -> flip RawMerge hs <$> fromRaw raw
-
--- Question: How does Deserialize throw a not-found error?
--- Question: What is the previous question?
-read
-  :: forall m
-   . Monad m
-  => Causal.Deserialize m Raw Raw
-  -> (EditHash -> m Patch)
-  -> Hash
-  -> m (Branch m)
-read deserializeRaw deserializeEdits h = Branch <$> Causal.read d h
- where
-  fromRaw :: Raw -> m (Branch0 m)
-  fromRaw Raw {..} = do
-    children <- traverse go _childrenR
-    edits <- for _editsR $ \hash -> (hash,) . pure <$> deserializeEdits hash
-    pure $ branch0 _termsR _typesR children edits
-  go = read deserializeRaw deserializeEdits
   d :: Causal.Deserialize m Raw (Branch0 m)
   d h = deserializeRaw h >>= \case
     RawOne raw      -> RawOne <$> fromRaw raw
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -83,6 +83,8 @@ module Unison.Codebase.Branch
     -- * Branch serialization
   , read
   , cachedRead
+  , boundedCache
+  , Cache
   , sync
 
     -- * Unused
@@ -457,10 +459,16 @@ data ForkFailure = SrcNotFound | DestExists
 numHashChars :: Branch m -> Int
 numHashChars _b = 3
 
+-- This type is a little ugly, so we wrap it up with a nice type alias for
+-- use outside this module.
+type Cache m = Cache.Cache m (Causal.RawHash Raw) (Causal m Raw (Branch0 m))
+
+boundedCache :: MonadIO m => Word -> m (Cache m)
+boundedCache = Cache.semispaceCache
+
 -- todo: can delete old `read` once we are confident in this
 cachedRead :: forall m . Monad m
-           => Cache.Cache m (Causal.RawHash Raw)
-                            (Causal m Raw (Branch0 m))
+           => Cache m
            -> Causal.Deserialize m Raw Raw
            -> (EditHash -> m Patch)
            -> Hash
--- a/parser-typechecker/src/Unison/Codebase/Causal.hs
+++ b/parser-typechecker/src/Unison/Codebase/Causal.hs
@@ -19,6 +19,7 @@ import qualified Data.Sequence                 as Seq
 import           Unison.Hash                    ( Hash )
 import qualified Unison.Hashable               as Hashable
 import           Unison.Hashable                ( Hashable )
+import qualified Unison.Util.Cache             as Cache
 import qualified Data.Map                      as Map
 import qualified Data.Set                      as Set
 import           Util                           ( bind2 )
@@ -111,6 +112,24 @@ data Tails h
   | TailsMerge (Set (RawHash h))
 
 type Deserialize m h e = RawHash h -> m (Raw h e)
+
+cachedRead :: Monad m
+           => Cache.Cache m (RawHash h) (Causal m h e)
+           -> Deserialize m h e
+           -> RawHash h -> m (Causal m h e)
+cachedRead cache deserializeRaw h = Cache.lookup cache h >>= \case
+  Nothing -> do
+    raw <- deserializeRaw h
+    causal <- pure $ case raw of
+      RawOne e              -> One h e
+      RawCons e tailHash    -> Cons h e (tailHash, read tailHash)
+      RawMerge e tailHashes -> Merge h e $
+        Map.fromList [(h, read h) | h <- toList tailHashes ]
+    Cache.insert cache h causal
+    pure causal
+  Just causal -> pure causal
+  where
+    read = cachedRead cache deserializeRaw
 
 read :: Functor m => Deserialize m h e -> RawHash h -> m (Causal m h e)
 read d h = go <$> d h where
--- a/parser-typechecker/src/Unison/Codebase/Causal.hs
+++ b/parser-typechecker/src/Unison/Codebase/Causal.hs
@@ -127,14 +127,6 @@ cachedRead cache deserializeRaw h = Cache.lookup cache h >>= \case
   where
     read = cachedRead cache deserializeRaw
 
-read :: Functor m => Deserialize m h e -> RawHash h -> m (Causal m h e)
-read d h = go <$> d h where
-  go = \case
-    RawOne e -> One h e
-    RawCons e tailHash -> Cons h e (tailHash, read d tailHash)
-    RawMerge e tailHashes ->
-      Merge h e (Map.fromList [(h, read d h) | h <- toList tailHashes ])
-
 type Serialize m h e = RawHash h -> Raw h e -> m ()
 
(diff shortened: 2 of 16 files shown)


parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -455,6 +459,35 @@ data ForkFailure = SrcNotFound | DestExists
...
+  fromRaw Raw {..} = do
+    children <- traverse go _childrenR
+    edits <- for _editsR $ \hash -> (hash,) . pure <$> deserializeEdits hash
+    pure $ branch0 _termsR _typesR children edits
+  go = cachedRead cache deserializeRaw deserializeEdits
+  d :: Causal.Deserialize m Raw (Branch0 m)
+  d h = deserializeRaw h >>= \case
+    RawOne raw      -> RawOne <$> fromRaw raw
+    RawCons  raw h  -> flip RawCons h <$> fromRaw raw
+    RawMerge raw hs -> flip RawMerge hs <$> fromRaw raw
```
pchiusano, line 489: This ended up almost identical to existing `Branch.read`, just with `Causal.cachedRead` in place of `Causal.read`.

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -455,6 +459,35 @@ data ForkFailure = SrcNotFound | DestExists
 numHashChars :: Branch m -> Int
 numHashChars _b = 3
 
+-- This type is a little ugly, so we wrap it up with a nice type alias for
+-- use outside this module.
+type Cache m = Cache.Cache m (Causal.RawHash Raw) (Causal m Raw (Branch0 m))
+
+boundedCache :: MonadIO m => Word -> m (Cache m)
+boundedCache = Cache.semispaceCache
```
pchiusano, line 467: Added this type alias and this `boundedCache` helper to reduce API surface area a bit. You create a `Branch.Cache m` using `Branch.boundedCache`, and then pass that to `Branch.cachedRead`. No need to import other modules.

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -485,8 +518,10 @@ sync
   -> (EditHash -> m Patch -> m ())
   -> Branch m
   -> m ()
-sync exists serializeRaw serializeEdits b =
-  State.evalStateT (sync' exists serializeRaw serializeEdits b) mempty
+sync exists serializeRaw serializeEdits b = do
+  _written <- State.execStateT (sync' exists serializeRaw serializeEdits b) mempty
+  -- traceM $ "Branch.sync wrote " <> show (Set.size written) <> " namespace files."
+  pure ()
```
pchiusano, line 524: Added some logging here, then commented it out again. Considered removing it entirely, but seems okay and kind of helpful as a reminder of how to print out what was sync'd.

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -520,8 +555,8 @@ sync' exists serializeRaw serializeEdits b = Causal.sync exists
     writeB0 b0 = do
       for_ (view children b0) $ \c -> do
         queued <- State.get
-        when (Set.notMember (headHash c) queued)
-          $ sync' exists serializeRaw serializeEdits c
+        when (Set.notMember (headHash c) queued) $
+          sync' exists serializeRaw serializeEdits c
```
pchiusano, line 559: Accidental reformatting, prob doesn't belong here... but seriously folks, who puts the dollar sign on the following line, amirite??

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -112,6 +113,24 @@ data Tails h
...
+    causal <- pure $ case raw of
+      RawOne e              -> One h e
+      RawCons e tailHash    -> Cons h e (tailHash, read tailHash)
+      RawMerge e tailHashes -> Merge h e $
+        Map.fromList [(h, read h) | h <- toList tailHashes ]
+    Cache.insert cache h causal
+    pure causal
+  Just causal -> pure causal
+  where
+    read = cachedRead cache deserializeRaw
```
pchiusano, line 132: Here's where the cache gets used, very similar to existing `Causal.read`, but with some calls to `Cache.insert` and `Cache.lookup`.

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -132,26 +151,28 @@ sync
...
+        Cons currentHash head (tailHash, tailm) -> do
+          -- write out the tail first, so what's on disk is always valid
+          b <- lift $ exists tailHash
+          unless b $ go =<< lift tailm
+          serialize currentHash (RawCons head tailHash)
+        Merge currentHash head tails -> do
+          for_ (Map.toList tails) $ \(hash, cm) -> do
+            b <- lift $ exists hash
+            unless b $ go =<< lift cm
+          serialize currentHash (RawMerge head (Map.keysSet tails))
```
pchiusano, line 175: This change is unrelated to this PR, but slipped in.

Previous code didn't insert `One` into the `queued` set, which seemed possibly wrong, even if we couldn't immediately think of a way this could lead to problems. Inserting the `One` values into `queued` is definitely not wrong.

parser-typechecker/src/Unison/Codebase/Editor/Git.hs
```diff
@@ -122,29 +123,30 @@ importRemoteBranch
   :: forall m v a
    . MonadIO m
   => Codebase m v a
+  -> Branch.Cache m
   -> RemoteNamespace
   -> ExceptT GitError m (Branch m)
-importRemoteBranch codebase ns = do
-  (branch, cacheDir) <- viewRemoteBranch' ns
+importRemoteBranch codebase cache ns = do
+  (branch, cacheDir) <- viewRemoteBranch' cache ns
```
pchiusano, line 130: This file had a bunch of places where the cache is being threaded through. Very mechanical.

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -88,9 +88,10 @@ commandLine
   -> (SourceName -> IO LoadSourceResult)
   -> Codebase IO v Ann
   -> (Int -> IO gen)
+  -> Branch.Cache IO
   -> Free (Command IO i v) a
   -> IO a
-commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSource codebase rngGen =
+commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSource codebase rngGen branchCache =
```
pchiusano, line 94: More plumbing of this new parameter.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -56,6 +56,7 @@ import qualified Unison.Referent               as Referent
 import qualified Unison.Util.TQueue            as TQueue
 import           Unison.Var                     ( Var )
 import qualified Unison.UnisonFile             as UF
+import qualified Unison.Util.Cache             as Cache
```
pchiusano, line 59: Lots of plumbing of the cache in this file, but very mechanical.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -162,27 +164,30 @@ codebase1
...
    . MonadUnliftIO m
   => Var v
   => BuiltinAnnotation a
-  => Common.SyncToDir m v a -> S.Format v -> S.Format a -> CodebasePath -> Codebase m v a
-codebase1' syncToDirectory fmtV@(S.Format getV putV) fmtA@(S.Format getA putA) path =
+  => Common.SyncToDir m v a -> Branch.Cache m -> S.Format v -> S.Format a -> CodebasePath -> m (Codebase m v a)
+codebase1' syncToDirectory branchCache fmtV@(S.Format getV putV) fmtA@(S.Format getA putA) path = do
+  termCache <- Cache.semispaceCache 8192
+  typeOfTermCache <- Cache.semispaceCache 8192
+  declCache <- Cache.semispaceCache 1024
```
pchiusano, line 179: These are the caches for terms, term types, and decls. Currently hardcoded these cache sizes since the config not readily available here.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -162,27 +164,30 @@ codebase1
...
   let c =
         Codebase
-          (getTerm getV getA path)
-          (getTypeOfTerm getV getA path)
-          (getDecl getV getA path)
+          (Cache.applyDefined termCache $ getTerm getV getA path)
+          (Cache.applyDefined typeOfTermCache $ getTypeOfTerm getV getA path)
+          (Cache.applyDefined declCache $ getDecl getV getA path)
```
pchiusano, line 184: `Cache.applyDefined` only caches results that are non-empty, which is the right move here - if you lookup a term in the cache and it isn't found now, you don't want to cache that it's missing, since it might be found later (say, after a `pull`).

parser-typechecker/src/Unison/Codebase/FileCodebase/Common.hs
```diff
@@ -280,13 +281,15 @@ codebaseExists root =
...
   fileExists <- doesFileExist (branchPath rootDir h)
   if fileExists then Just <$>
-    Branch.read (deserializeRawBranch rootDir)
-                (deserializeEdits rootDir)
-                h
+    Branch.cachedRead
+      cache
+      (deserializeRawBranch rootDir)
+      (deserializeEdits rootDir)
+      h
```
pchiusano, line 292: And here's where the cache finally gets used!

parser-typechecker/src/Unison/Codebase/FileCodebase/Reserialize.hs
```diff
@@ -43,20 +43,20 @@ makeLenses ''SyncedEntities
 
 -- Copy all dependents of `branch` from `srcPath` into `destPath`.
 --
--- As a refresher, in the normal course of using `ucm` and updating the 
+-- As a refresher, in the normal course of using `ucm` and updating the
```
pchiusano, line 46: Whitespace... my editor is set to strip trailing line whitespace

parser-typechecker/src/Unison/Util/Cache.hs
```diff
@@ -0,0 +1,88 @@
...
+            Nothing -> pure Nothing
+            Just v -> insert k v $> Just v
+        just -> pure just
+    insert k v = atomically $ do
+      modifyTVar' gen0 (Map.insert k v)
+      m0 <- readTVar gen0
+      when (fromIntegral (Map.size m0) >= maxSize) $ do
+        writeTVar gen1 m0
+        writeTVar gen0 Map.empty
+  pure $ Cache lookup insert
```
pchiusano, line 59: Neato cache implementation.

parser-typechecker/src/Unison/Util/Timing.hs
```diff
@@ -9,7 +9,11 @@ import Data.Time.Clock.System (getSystemTime, systemToTAITime)
 import Data.Time.Clock.TAI (diffAbsoluteTime)
 import Data.Time.Clock (picosecondsToDiffTime)
 
+enabled :: Bool
+enabled = False
```
pchiusano, line 13: Flag to turn timing logs on and off.

parser-typechecker/unison/Main.hs
```diff
@@ -1,5 +1,6 @@
 {-# Language OverloadedStrings #-}
 {-# Language PartialTypeSignatures #-}
+{-# Language ScopedTypeVariables #-}
```
pchiusano, line 3: Comment on this file - I don't think anything interesting here. We create the cache and plumb it everywhere it's needed.

parser-typechecker/unison/Main.hs
```diff
@@ -114,56 +121,61 @@ main = do
            _                                 -> (Nothing, args)
   currentDir <- getCurrentDirectory
   configFilePath <- getConfigFilePath mcodepath
+  config@(config_, _cancelConfig) <-
+    catchIOError (watchConfig configFilePath) $ \_ ->
+      Exit.die "Your .unisonConfig could not be loaded. Check that it's correct!"
+  branchCacheSize :: Word <- Config.lookupDefault 4096 config_ "NamespaceCacheSize"
+  branchCache <- Cache.semispaceCache branchCacheSize
```
pchiusano, line 128: We moved config initialization here, since the branchCache needs to be created here, and we wanted a config setting for the branchCache (the `NamespaceCacheSize` setting).

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Do we still use `Branch.read` anywhere?  (Not that it's necessarily time to delete it yet.)

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
aryairani: Might want to initialize these earlier, as the result of a successful read should be the same regardless of which codebase it comes from. (Cache poisoning aside.)

parser-typechecker/src/Unison/Codebase/FileCodebase/Common.hs
aryairani: Excite!!!!

parser-typechecker/src/Unison/Codebase/FileCodebase/Reserialize.hs
```diff
@@ -70,12 +70,13 @@ syncToDirectory
   -> Branch m
   -> m ()
 syncToDirectory fmtV fmtA srcPath destPath branch = do
-  flip evalStateT mempty $
+  _written <- flip execStateT mempty $
```
aryairani, line 73: Why?  Half-logged I guess?

parser-typechecker/src/Unison/Util/Cache.hs
```diff
@@ -0,0 +1,88 @@
...
+        _ -> pure ()
+
+  pure $ Cache lookup insert
+
+-- Create a cache of bounded size. Once the cache
+-- reaches a size of `maxSize`, older unused entries
+-- are evicted from the cache. Unlike LRU caching,
+-- where cache hits require updating LRU info,
+-- cache hits here are read-only and contention free.
+semispaceCache :: (MonadIO m, Ord k) => Word -> m (Cache m k v)
```
aryairani, line 34: Why `Word`? Faster `>=`?

parser-typechecker/tests/Unison/Test/Git.hs
```diff
@@ -35,6 +36,7 @@ test = scope "git" . tests $ [testPull, testPush]
 -- dependencies
 testPull :: Test ()
 testPull = scope "pull" $ do
+  branchCache <- io $ Branch.boundedCache 4096
```
aryairani, line 39: Was there some reasoning behind the number 4096?

parser-typechecker/unison/Main.hs
aryairani: Maybe we should thread a UnisonConfig monad class throughout the app so we can access it in `FileCodebase.hs`, etc.

parser-typechecker/tests/Unison/Test/Git.hs
runarorama: Seemed like a nice round number. Bigger seemed like a waste, smaller seemed like not enough

parser-typechecker/unison/Main.hs
pchiusano: Sounds intriguing... for later.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
pchiusano: We discussed this, and if we want caching for the sync operation, we can just pass these caches as a new argument to this spot:

https://github.com/unisonweb/unison/blob/386bdc754a1e06079114424618fe003a2fb2f7c7/parser-typechecker/src/Unison/Codebase/FileCodebase.hs#L192-L193

parser-typechecker/src/Unison/Codebase/FileCodebase/Reserialize.hs
pchiusano: Yeah... there was some logging in here originally. Didn't bother with cleanup since we're deleting.

parser-typechecker/src/Unison/Util/Cache.hs
pchiusano: Just because a negative cache size doesn't make sense. :)

parser-typechecker/tests/Unison/Test/Git.hs
pchiusano: Also base last I checked had like 7-10k namespace files. I suspect it's not important to have all the namespace files cached, just "recent history" needed to resolve LCAs of merges etc.

PR #1553, opened 2020-05-15
Properly fix spurious incoming root branch warnings 

### LINKED ISSUE #616: figure out how to handle `WarnIncomingRootBranch` long term
Note that it's currently triggered in normal operation by an `update` or `merge` or `pull` with stuff to propagate.  Maybe we could refactor some of the helpers to support threading an updated branch in memory, instead of having them `sync` or `get` every time?

>I think the spurious branch file arrival message is due to the fact that we calling sync twice in the command, but that IORef with the current branch, which controls the file watch events, is updated on different schedule? Like after each command finishes?

pchiusano:
This is ready to go and fixes #616  

See description here: https://github.com/unisonweb/unison/issues/616#issuecomment-628947439

## Implementation notes

At the point where the new namespace files come in, we don't have a full branch, just the hashes based on the file names. So I added a `beforeHash` function which isn't a full LCA calculation, just searches for a hash in the history of a branch. It is bounded (currently to depth of 20, hardcoded constant); after that, it treats the head as new and raises the same warning as now.

I wasn't sure about the constant to cutoff the search at, but figured lower is better... if one of the commands is issuing 20 separate root branch updates, that seems like a problem. In my testing, I wasn't able to get the warning to trigger anymore with this fix.

## Test coverage

DONE ~~I'd like some unit tests for the `hashBefore` function.~~

## Loose ends

DONE ~~It's not great that there's a full history scan if a branch _does_ actually show up. That incidentally will blow away the namespace cache. I think fix might just be to trim the search to depth 200 or something.~~

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine.hs
+++ b/parser-typechecker/src/Unison/CommandLine.hs
@@ -56,18 +56,23 @@ watchFileSystem q dir = do
     atomically . Q.enqueue q $ UnisonFileChanged (Text.pack filePath) text
   pure (cancel >> killThread t)
 
-watchBranchUpdates :: IO Branch.Hash -> TQueue Event -> Codebase IO v a -> IO (IO ())
+watchBranchUpdates :: IO (Branch.Branch IO) -> TQueue Event -> Codebase IO v a -> IO (IO ())
 watchBranchUpdates currentRoot q codebase = do
   (cancelExternalBranchUpdates, externalBranchUpdates) <-
     Codebase.rootBranchUpdates codebase
   thread <- forkIO . forever $ do
     updatedBranches <- externalBranchUpdates
     currentRoot <- currentRoot
-    -- We only issue the event if the branch is different than what's already
-    -- in memory. This skips over file events triggered by saving to disk what's
-    -- already in memory.
-    when (any (/= currentRoot) updatedBranches) $
-      atomically . Q.enqueue q . IncomingRootBranch $ Set.delete currentRoot updatedBranches
+    -- Since there's some lag between when branch files are written and when
+    -- the OS generates a file watch event, we skip branch update events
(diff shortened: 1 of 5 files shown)


parser-typechecker/src/Unison/CommandLine.hs
```diff
@@ -56,20 +61,47 @@ watchFileSystem q dir = do
...
-    when (any (/= currentRoot) updatedBranches) $
-      atomically . Q.enqueue q . IncomingRootBranch $ Set.delete currentRoot updatedBranches
+    -- Since there's some lag between when branch files are written and when
+    -- the OS generates a file watch event, we skip branch update events
+    -- that are causally before the current root.
+    --
+    -- NB: Sadly, since the file watching API doesn't have a way to silence
+    -- the events from a specific individual write, this is ultimately a
+    -- heuristic. If a fairly recent head gets deposited at just the right
+    -- time, it would get ignored by this logic. This seems unavoidable.
```
aryairani, line 78: I may be being dense, but it seems to me that given that events get generated for files written to disk, and we exactly know which ones we are writing (because we are writing them) that we could exactly silence exactly those.

e.g. When we write a root, we add it to a set; when we receive an event, we remove it from the set.  If we receive an event for something that isn't in the set of the things we said we wrote, then we didn't write that.

Right? 😅  And then wouldn't have to wait for a `IO (Branch IO)` to load, and for a Branch.before to run (or truncate).

pchiusano: That relies for correctness on the file watching not missing any writes. Otherwise this set you maintain could accumulate stale hashes in it. I didn't like that approach, just seemed fragile to me, also more complicated to implement and test I thought (though I could be wrong about that).

PR #1554, opened 2020-05-15
Eliminate second prompt and codebase sync between add/update and default metadata

runarorama:
## Overview

This eliminates a second UCM prompt printing to the screen when calling `add` or `update` with default metadata configured. This is done by stopping these commands syncing the root branch until they're really finished.

Also prevents UCM printing "I added some default metadata" when it actually didn't.

## Implementation notes

Created `stepManyAtNoSync` and friends which acts like `stepManyAt`, but doesn't call `updateRoot`. Tracking two separate root states now, `root` and `lastSyncedRoot`. The latter gets updated only when `updateRoot` is called.

## Test coverage

This gets some coverage from the `diff.md` transcript.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -477,6 +477,7 @@ loop = do
         stepManyAtNoSync =
           Unison.Codebase.Editor.HandleInput.stepManyAtNoSync
         updateRoot = flip Unison.Codebase.Editor.HandleInput.updateRoot inputDescription
+        syncRoot = use root >>= updateRoot
         updateAtM = Unison.Codebase.Editor.HandleInput.updateAtM inputDescription
         unlessGitError = unlessError' (Output.GitError input)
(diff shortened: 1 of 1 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -524,17 +526,19 @@ loop = do
...
+          after  <- Branch.head <$> use root
           (ppe, outputDiff) <- diffHelper before after
           unless silent if OBranchDiff.isEmpty outputDiff
             then respond NoOp
             else respondNumbered $ ShowDiffNamespace Path.absoluteEmpty
                                                      Path.absoluteEmpty
                                                      ppe
                                                      outputDiff
+          when (silent && not (OBranchDiff.isEmpty outputDiff)) $
+            respond DefaultMetadataNotification
```
aryairani, line 541: These two conditions could probably be combined to make this more readable.

pchiusano:
Seems okay. I was curious if we could just get rid of any setting of the root in HandleInput and instead just sync the root after each input is processed.

Or factor the syncing logic to just one spot in HandleInput.

PR #1555, opened 2020-05-15
Update pull-request.create help example text

ceedubs:
## Overview

I think that this help text better reflects that `base` now follows the
`trunk` convention and that a `.prs` namespace is now encouraged.

I think that it would be good to use consistent conventions across docs,
because currently some of them are subtly different, which makes it
tougher to get started with the day-to-day development workflow.

## Test coverage

I haven't added any tests for this. I'm not sure if copy/pasting the
example string into a test is useful, but I'm happy to add something if
it would be desired.


pchiusano:
LGTM! can you just add yourself to CONTRIBUTORS.markdown?

PR #1561, opened 2020-05-18
Use `LANGUAGE Strict` and `fmap B.copy . getBytes` in `V1.hs`

### LINKED ISSUE #1560: Initial pull of base uses a lot of memory
Here's a transcript:

    ```ucm
    .> pull git@github.com:unisonweb/base:.trunk base
    ```

Just from watching activity monitor:

* Initial memory usage is about 14-15MB.
* Usage climbs to about 5.5GB during the "Importing downloaded files into local codebase" phase.
* After completion, memory usage stays at that level even after I do further commands to hopefully trigger some GC.
* After restarting on the same codebase, memory usage is much lower (300-500MB, see #1550).
* These results are with caching turned off - I turned off by hardcoding `nullCache` in `Main`, not relying on the config setting.

@aryairani I could have sworn this took very little memory at one point in the development of this new syncing algorithm

If I had to guess, it's some sort of space leak in `SlimCopyRegenerateIndex`. A couple things I tried (which are all in debug/1560 branch):

* Turned on `{-# Language Strict, StrictData -#}` in SlimCopyRegenerateIndex (total guess).
* Did a `ByteString.copy` everywhere in `V1.hs` where `getBytes` is called - by default, there was one in `getHash` and one in `getText`. Reasoning: `ByteString` values are just offsets into a single array of bytes. To allow that array of bytes to be GC'd you need to copy the bytestring to its own storage.

The fact that memory usage doesn't go back down again seems like a clue and is unexpected to me...

aryairani:
Closes #1560 for now.  This PR improves the memory * time usage (area under the curve) for pulling `base:.trunk` by about 65%.  

Almost all of that improvement comes from just using `LANGUAGE Strict` and `fmap B.copy . getBytes` in `V1.hs`.  I had some timing data that is now probably lost in a browser buffer that said that `B.copy` alone made things a little worse, that `LANGUAGE Strict` made them about 60% better, and `B.copy` on top of `LANGUAGE Strict` was a bit better than that.

A couple percent more improvement came by making `SlimCopy..hs` use `Writer.Strict` and `State.Strict`; I guess they need to be completely evaluated and consumed by the end of the operation anyway.  I almost didn't bother to include them, but went ahead since it seemed to help a little bit. https://docs.google.com/spreadsheets/d/1iBYfVORtHGeUU6ioXU3YXSBNDdQt1KtGt2xRLyFhnXE/edit#gid=0 has some timing numbers.

I also tried timing with just a 1-way relation index instead of an unnecessarily 2-way https://github.com/unisonweb/unison/commit/845bd6a344f4f2728982544aab75096ec39f694, but I was doing timing experiments on my laptop on base.List and the numbers were just too all over the place to tell if it helped, so I left it out for now.

#### Further work

We still hang on to a lot of data during the sync:
image

`ARR_WORDS` is `Text` and `ByteArray`s, so we should find a way to share those that isn't locked in a cache forever.  And then I don't really know what the rest are.  I'm hoping that when we update to ghc 8.8 (#1562), that we can use some of the other profiling modes for clues without it segfaulting.  

Honestly I still don't see how we can get 400MB of bytestrings out of a 200MB codebase when the implementation shouldn't be holding more than one entitity (raw branch, patch, term, or decl) at a time. 512 bits per hash * 15723 distinct elements in `base` is less than 1MB.

Compare to before the PR though:
image

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
+++ b/parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
@@ -16,9 +16,9 @@ import Unison.Prelude
 
 import qualified Data.Set                      as Set
 import           Control.Lens
-import           Control.Monad.State            ( MonadState, evalStateT )
-import           Control.Monad.Writer           ( MonadWriter, execWriterT )
-import qualified Control.Monad.Writer          as Writer
+import           Control.Monad.State.Strict     ( MonadState, evalStateT )
+import           Control.Monad.Writer.Strict    ( MonadWriter, execWriterT )
+import qualified Control.Monad.Writer.Strict   as Writer
 import           UnliftIO.Directory             ( doesFileExist )
 import           Unison.Codebase                ( CodebasePath )
 import qualified Unison.Codebase.Causal        as Causal
--- a/parser-typechecker/src/Unison/Codebase/Serialization/V1.hs
+++ b/parser-typechecker/src/Unison/Codebase/Serialization/V1.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Strict #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE FlexibleContexts, RankNTypes #-}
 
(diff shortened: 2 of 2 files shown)


aryairani:
Found the missing, inconclusive benchmarks: https://gist.github.com/aryairani/251ffb6e91c0b121a0f3e1fcfe94a48b

PR #1566, opened 2020-05-18
Use Text.IO functions to remove unnecessary intermediate String types

aycanirican:
# Overview

This removes unnecessary conversion between `Text` and `String` types when doing IO at Runtime. It should speed up string based IO operations (get and put strings) significantly and reduce memory overhead. 

## Implementation notes

We simply use corresponding functions from Text.IO module and remove packing, unpacking stuff.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Rt1IO.hs
+++ b/parser-typechecker/src/Unison/Runtime/Rt1IO.hs
@@ -43,6 +43,7 @@ import           Data.GUID                      ( genText )
 import qualified Data.Map                      as Map
 import qualified Data.Sequence as Seq
 import           Data.Text                     as Text
+import qualified Data.Text.IO                  as TextIO
 import           Data.Time.Clock.POSIX         as Time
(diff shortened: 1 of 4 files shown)


aryairani:
Hi @aycanirican, this is great.  

It highlighted our lack of tests to exercise those functions though; would you feel up to writing one?

I'm imagining something like in `parser-typechecker/tests/Unison/Test/Git.hs`, which runs a transcript (uses Unison to maybe reads and write a file in Unison?), and then tests afterwards from Haskell that it was written as expected, before cleaning up.

What do you think? 😃

aycanirican:
Hi @aryairani, sounds great. Let me add a few test cases to this PR and then we can merge them all together.

aycanirican:
@aryairani this change has test cases already. Do you want me to add more tests? Otherwise I think it's ok to merge them since more IO tests are possible right now. I'll implement the result returning functionality in a different PR if it's ok for you.

PR #1570, opened 2020-05-20
`debug.push-exhaustive`/`pull-exhaustive`

aryairani:
## Overview

This adds the commands `debug.push-exhaustive` and `debug.pull-exhaustive` which sync with an exhaustively traversal of the dependency trees (no short-circuiting even when an entity already exists at the target).  

These are useful for pushing through dependencies that were missed by earlier sync implementations.  The actual recopying of files is still skipped for an entity that already exists (which also means this won't fix corrupted files), but everything is read once, to find dependencies.

## Implementation notes

It just passes a flag through to the sync algorithm, and replaces

```haskell
ifM (doesFileExist (getFilename destPath h)) dontSync (doSync h)
```

with

```haskell
if mode == SyncMode.Complete then doSync h
else ifM (doesFileExist (getFilename destPath h)) dontSync (doSync h)
```

I deleted the old sync algorithms because I didn't want to maintain them anymore; e.g. adding this functionality.

## Test coverage

I added a quick test in Git.hs, please sanity check it.  It demonstrates pushing a namespace that would otherwise have been short-circuited by the presence of an existing parent namespace, and pushing a term dependency that would otherwise have been short-circuited by the presence of its dependent.

It doesn't explicitly test repairing dependents of patches, or types, (or ...?,) but since the actual core of the patch is a single `if`, that the fact that it works at all hopefully means that it works all over.

There are no tests for the new commands themselves, just for the underlying sync function, as that avoided the need to create a local git repo and run multiple transcripts.  (The previously existing `"git.push."` tests work something like this.)

## Loose ends

I think the commands should produce some output when they encounter a missing dependency during the sync. I have some WIP stashed for that but, as it was starting to get hairy, I  wanted to get this basic version out first, running end to end with a test.


parser-typechecker/tests/Unison/Test/Git.hs
```diff
@@ -72,11 +142,9 @@ test> tests.x = [Ok "Great!"]
     Left e -> crash $ show e
     Right b -> pure b
 
-  io $ Codebase.syncToDirectory codebase targetDir
+  io $ Codebase.syncToDirectory codebase targetDir SyncMode.ShortCircuit
         (Branch.getAt' (Path $ pure "tests") b)
 
-  let makeTitle :: String -> String
-      makeTitle = intercalate "/" . map (take 20) . drop 2 . splitOn "/"
```
aryairani, line 79: This was used in a bunch of places, so I pulled it to the top level.

parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
```diff
@@ -304,16 +308,17 @@ syncToDirectory' getV getA srcPath destPath newRoot =
...
+                 -> (CodebasePath -> h -> FilePath) -- done if this filepath exists
+                 -> SimpleLens s (Set h) -- lens to track if `h` is already done
+                 -> (h -> m ()) -- do!
+                 -> m ()        -- don't
+                 -> m ()
+  ifNeedsSyncing h destPath getFilename l doSync dontSync =
+    ifM (use (l . to (Set.member h))) dontSync $ do
+      l %= Set.insert h
+      if mode == SyncMode.Complete then doSync h
+      else ifM (doesFileExist (getFilename destPath h)) dontSync (doSync h)
```
aryairani, line 324: The `if` on line 323 is the only interesting change here.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -417,7 +418,7 @@ loop = do
           UpdateBuiltinsI -> "builtins.update"
           MergeBuiltinsI -> "builtins.merge"
           MergeIOBuiltinsI -> "builtins.mergeio"
-          PullRemoteBranchI orepo dest ->
+          PullRemoteBranchI orepo dest _syncMode ->
```
pchiusano, line 421: Should this key off the `syncMode` to print either `pull` or `debug.pull-exhaustive`?

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -725,14 +726,41 @@ pull = InputPattern
...
+      Right $ Input.PullRemoteBranchI (Just ns) Path.relativeEmpty' SyncMode.ShortCircuit
     [url, path] -> do
       ns <- parseUri "url" url
       p <- first fromString $ Path.parsePath' path
-      pure $ Input.PullRemoteBranchI (Just ns) p
+      Right $ Input.PullRemoteBranchI (Just ns) p SyncMode.ShortCircuit
+    _ -> Left (I.help pull)
+  )
+
+pullRepair :: InputPattern
```
pchiusano, line 741: Minor thing, maybe call this `pullExhaustive`?

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -654,9 +654,9 @@ notifyUser dir o = case o of
       where
       push = P.group . P.backticked $ IP.patternName IP.push
       pull = case input of
-        Input.PushRemoteBranchI Nothing p ->
+        Input.PushRemoteBranchI Nothing p _ ->
```
pchiusano, line 657: Is this correct? Should you be switching off the mode here to produce different output?

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -654,9 +654,9 @@ notifyUser dir o = case o of
...
           P.sep " " [IP.patternName IP.pull, P.shown p ]
-        Input.PushRemoteBranchI (Just (r, _)) p -> P.sepNonEmpty " " [
+        Input.PushRemoteBranchI (Just (r, _)) p _ -> P.sepNonEmpty " " [
```
pchiusano, line 659: Same here, should you be switching off the mode to produce different output?

parser-typechecker/tests/Unison/Test/Git.hs
```diff
@@ -8,37 +8,107 @@ module Unison.Test.Git where
...
+  , syncComplete
+  , syncTestResults
+  ]
+
+traceTranscriptOutput :: Bool
 traceTranscriptOutput = False
 
+-- | make sure that a definition present in the target dir doesn't prevent
+-- syncing of its dependencies
+syncComplete :: Test ()
```
pchiusano, line 48: LGTM. The test is nice to read, with the comments and helper functions `observe`. 👍

If you wanted you could add a test that covers syncing of child namespaces as well, even when the parent is present. And then syncing of historical namespaces, even when the latest namespace is present.

parser-typechecker/src/Unison/Codebase/FileCodebase/SlimCopyRegenerateIndex.hs
pchiusano: Do we need to do something special here to make sure we don't visit the same hash over and over again every time it appears somewhere in the namespace tree or the history? The short-circuiting version relies on the existence of the file to prevent this worst case, which can be exponential in the worst case (it's based on the number of unique paths to a node in a DAG), so do we need to do something extra here to prevent that? Like maybe a `seen : Set Hash` that gets used in the `SyncMode.Complete` case? (Or maybe you are already doing this...)

I have a question case

aryairani: Right, on lines 321 and 322 we first check/update a ~`seen : Set h`.  So we only get to line 323 once.  Before this PR, that `Set` was being used to skip even the `doesFileExist` check.  Does that make sense?

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: That's a good idea.  My thinking was that this doesn't matter, since `pull` and `debug.pull-exhaustive` have the safe effect, but different side effects.  But maybe showing one or the other would be less surprising though, wdyt?

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
aryairani: Oops, good catch.

parser-typechecker/tests/Unison/Test/Git.hs
aryairani: Thanks!

I think this does test syncing a child namespace (`c`), even when the parent is present (`b`), right?

I will add something for historical namespaces.

Issue #1576, opened 2020-05-21
switch from git submodules to stack git `extra-deps`


aryairani:
spec at https://docs.haskellstack.org/en/v1.0.2/yaml_configuration/

aryairani:
Closed by #1581.

PR #1579, opened 2020-05-21
Bump to lts-15.13

### LINKED ISSUE #1562: update to ghc 8.8
The goal is to give us better tool support; maybe more out of the box builds for hie/ghcide, maybe ghc profiling modes `-pr` etc. to discover what's retaining memory; this currently segfaults for me in our current 8.6.5 build.

I started an attempt in `topic/lts-15`, which is still in the middle of upgrading from megaparsec 6.5 to 8.0.

mitchellwrosen:
- [x] Blocked on #1576 

This PR bumps the resolver to `lts-15.13`, leaving megaparsec where it's at ;)

Fixes #1562


parser-typechecker/src/Unison/Codebase/FileCodebase/Common.hs
```diff
@@ -238,7 +234,8 @@ encodeFileName = let
                  = "$x" <> encodeHex [c] <> "$" <> go rem
                | otherwise = c : go rem
   go [] = []
-  encodeHex = Text.unpack . decodeUtf8 . Hex.hex . encodeUtf8 . Text.pack
+  encodeHex :: String -> String
+  encodeHex = Text.unpack . Text.encodeBase16 . Text.pack
```
aryairani, line 238: Within `FileCodebase`, and `Serialization`, we have to be careful even changing existing behavior at all, even to fix bugs, as it could cause some existing user data to become unreadable.  I think this change is safe though, because we don't have any builtin names that use extended characters, and because git caches can be missed safely.

mitchellwrosen: Good point. In this case, there actually was a change: `hex` uppercase letters `A-F`, but `base16` outputs lowercase letters `a-f`. Is that a problem?

mitchellwrosen: @aryairani I think whether the lowercase/uppercase switch is a problem is the only remaining issue. Tests pass, so I'm not sure :)

aryairani: Hm this morning I’m only “pretty sure” it’s okay too, so I think we should revert this to be safe, until the next codebase format.

aryairani: (In both encodeFileName and decodeFileName)

aryairani: @mitchellwrosen Sorry, I hadn't spotted your question until now. 

I'm not sure; let's throw in a `toUpper` to match the old behavior (to be safe), and then if we have an analogous function in the V2 codebase, we should use lower case letters, because it looks nicer IMO.

aryairani: Btw, good catch on the `A-F` vs `a-f`.

mitchellwrosen: `toUpper` added!

aryairani:
@pchiusano @runarorama @dolio I think we are ready to bump to GHC 8.8.  Any objections to merging?

pchiusano:
No objections..

PR #1584, opened 2020-05-26
Fix nonsense polytype rendered by delayed computations and allow multiple delays in types

### LINKED ISSUE #249: Lexer issues around delay syntax
The parser doesn't handle the following, but should.
1. `a ->'{e} b`  (can't handle lack of space between `->` and `'`; should handle although space is preferred)
2. `a -> '{e} 'b`  (can't handle second `'`)
3. `''a`  (ditto, can't handle second `'`)

> Issue for both is in the lexer I think, we should consider a quote following an arrow to not be part of the arrow lexeme. And if we are in the ‘munch lexeme’ state of the lexer then a single quote should output that quote. --@pchiusano

Came up during implementation of the Type pretty-printer (so will need changes to Test/TypePrinter.hs, which is currently only on the wip/prettyprint.hs branch.)

@atacratic planning to look at this after finishing the Type and Term pretty-printers.

### LINKED ISSUE #689: peculiar universal quantification on inferred type of delayed function
Slurp the following:
```
ability SystemTime where
  systemTime : .base.Nat

tomorrow = '(SystemTime.systemTime + 24 * 60 * 60)
```
You get:
```
  ✅
  
  I found and typechecked these definitions in ~/u1/scratch.u. If you do an
  `add` or `update` , here's how your codebase would change:
  
      ability SystemTime
      tomorrow : ∀ (). () ->{SystemTime} .base.Nat
  
  Ctrl+C cancels.

```

Note the `∀ ()` in the inferred signature for `tomorrow`.  Surely there's only one `()` unit type, we don't need to quantify?  I'd expect it to infer `tomorrow : () ->{SystemTime} .base.Nat`

(M1a)

runarorama:
This adds a special case to the `TypePrinter` to handle types of the form `forall (). () -> x`, which are sometimes generated by the delay syntax `'x`.

Also slightly modifies the type parser to allow types of the form `''x`, which would previously fail the parser.

## Overview

Fixes #689 
Fixes #249

## Implementation notes

Adds a special case to the type printer, doubling down on a horrible hack.

A better implementation of `'x` would be `(_ : () -> x)`, but since the new runtime will likely have a completely different implementation of delayed computations, it seems like overkill especially since it would require a change to the term language (adding type annotations for arguments).

## Test coverage

A transcript is included

## Loose ends

An alternative, much more comprehensive approach is drafted in the branch `/topic/bug689`, which turns all function types of the form `forall a. a -> x`, where `x` does not involve `a`, into `'x`.

### DIFF
--- a/parser-typechecker/src/Unison/TypeParser.hs
+++ b/parser-typechecker/src/Unison/TypeParser.hs
@@ -56,7 +56,7 @@ type2a = delayed <|> type2
 delayed :: Var v => TypeP v
 delayed = do
   q <- reserved "'"
-  t <- effect <|> type2
+  t <- effect <|> type2a
   pure $ Type.arrow (Ann (L.start q) (end $ ann t))
                     (DD.unitType (ann q))
                     t
--- a/parser-typechecker/src/Unison/TypePrinter.hs
+++ b/parser-typechecker/src/Unison/TypePrinter.hs
@@ -89,16 +89,22 @@ prettyRaw n im p tp = go n im p tp
     Effect1' e t ->
       PP.parenthesizeIf (p >= 10) $ go n im 9 e <> " " <> go n im 10 t
     Effects' es         -> effects (Just es)
-    ForallsNamed' vs body -> if (p < 0 && all Var.universallyQuantifyIfFree vs)
-      then go n im p body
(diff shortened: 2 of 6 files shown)


aryairani:
Nice.

Issue #1587, opened 2020-05-28
Consider disallowing name segments with `.` in them.


runarorama:
Since `.` refers to the root of the namespace tree, and is also used as a namespace separator, the fact that this character is allowed in "symboly" identifiers is a source of a lot of bugs. It would be simpler if this character were simply disallowed in mames. We can then revisit later.

aryairani:
Seems reasonable to me, but what do we do about codebases that already have `.` in them?  Make a point of ignoring those names?

aryairani:
Not opposed to disabling this, pending #1589, if that would still help?

aryairani:
This should be the case now.

Issue #1586, opened 2020-05-28
UCM `add` crashes on the name `.`


runarorama:
The default metadata lookup is unable to handle terms or namespaces with `.` in the name, resulting in a crash with:

```
InvalidWordyId ""
```

Transcript:

```unison
(.) f g x = f (g x)
```

```ucm
.> add
```

aryairani:
I kind of forget, is it a consequence of `Name` not having enough structure to distinguish absolute identifiers, or is it just a run-of-the-mill bug in the parser?  (related to https://github.com/unisonweb/unison/issues/1523)

PR #1588, opened 2020-05-28
Properly render calls to functions named (.)

### LINKED ISSUE #1063: Spurious use-clause in roundtrip
The following definition

```
fromDigits = Nat.fromText . join "" . map Nat.toText
```

results in the following when pretty-printed:

```
fromDigits : [Nat] -> Maybe Nat
fromDigits =
  use
   ( Nat.fromText (Text.join "")) (List.map Nat.toText)
```

runarorama:
This changes the `suffixes` of a `Name` such that the shortest suffix can be `.` if the definition has that name (which some definitions do).

## Overview

Before, the definition `noop = not . not` would be rendered as:

```unison
noop =  not not
```

With this change, it gets rendered as defined.

Fixes #1063 

## Test coverage

It's actually impossible to test this in a transcript at the moment, since the name `.` is rejected by `add` because of #1586

## Loose ends

* A fix for #1586 would allow this to be tested.
* A more comprehensive solution to names with `.` in them requires hunting down places where we're segmenting
* Consider disallowing `.` as a name. See #1587.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -9,7 +9,6 @@ module Unison.TermPrinter where
 import Unison.Prelude
 
 import           Data.List
-import           Data.List.Extra                ( dropEnd )
 import qualified Data.Map                      as Map
 import qualified Data.Set                      as Set
 import           Data.Text                      ( splitOn, unpack )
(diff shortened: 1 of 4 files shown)


unison-core/src/Unison/Name.hs
```diff
@@ -114,14 +114,19 @@ parent (Name txt) = case unsnoc (Text.splitOn "." txt) of
   Just ([],_) -> Nothing
   Just (init,_) -> Just $ Name (Text.intercalate "." init)
 
--- suffixes "" -> [] 
+-- suffixes "" -> []
 -- suffixes bar -> [bar]
 -- suffixes foo.bar -> [foo.bar, bar]
 -- suffixes foo.bar.baz -> [foo.bar.baz, bar.baz, baz]
+-- suffixes ".base.." -> [base.., .]
```
aryairani, line 121: Would you add this case to `parser-typechecker/tests/Unison/Core/Test/Name.hs` ?  Optionally add a note here, reminding that this is tested there.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -820,8 +819,10 @@ countName n = let f = \(p, s) -> (s, Map.singleton p 1)
...
 splitName :: Name -> [(Prefix, Suffix)]
-splitName n = let ns = splitOn "." (Name.toText n)
-              in dropEnd 1 ((inits ns) `zip` (map dotConcat $ tails ns))
+splitName n =
+  let ns = splitOn "." (Name.toText n)
+  in  filter (not . Text.null . snd) $ inits ns `zip` map dotConcat (tails ns)
+
 -- > splitName "x" == [([], "x")]
 -- > splitName "A.x" == [(["A"], "x")]
 -- > splitName "A.B.x" == [(["A"], "B.x"), (["A.B"], "x")]
```
aryairani, line 828: Consider adding these (plus any newly relevant test cases) to a unit test?  I couldn't find an existing one for it.

PR #1591, opened 2020-05-29
Improve handling of the name `.`

### LINKED ISSUE #1523: `.` is accepted as namespace name but leads to issues.
A new user may accidentally write `cd ..`, thinking you'll go back one namespace. Instead you will go to the namespace with the name `.` at the root of the codebase. Afterwards, it's impossible to delete the repository other than by deleting all of its contents.

A very otherwise empty codebase is attached below.
[broken-cd-dot-dot-codebase.zip](https://github.com/unisonweb/unison/files/4571476/broken-cd-dot-dot-codebase.zip)

This leads to some other crashes down the line, that possibly don't have anything to do with this particular issue. Such as for example creating a `(base..)` will now crash, but will work on next startup of ucm.
Attempting te delete the namespace will give you a Haskell error.
image

### LINKED ISSUE #1586: UCM `add` crashes on the name `.`

runarorama:
This PR fixes #1586 by way of generally improving the way we are parsing names. Previously there were instances of `Text.splitOn "."` peppered throughout the codebase. I've replaced all of those places with calls to `Name.segments` which handles segments named `.` properly.

## Overview

Fixes #1586 
Fixes #1523 

You can now:

  * `add` or `update` a term called `.`
  * Add and remove namespaces with symboly names including `.`
  * Navigate under such namespaces

Since a lot of things call into `Name.fromText`, those things now handle the name `.` properly.

* Decided to allow `symbolyId`s to be namespace segments in UCM.

## Test coverage

Added a comprehensive test for `Name.segments`.


parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -128,8 +128,8 @@ import           Unison.Codebase.Causal         ( Causal
                                                 )
 import           Unison.Codebase.Path           ( Path(..) )
 import qualified Unison.Codebase.Path          as Path
-import           Unison.Codebase.NameSegment    ( NameSegment )
-import qualified Unison.Codebase.NameSegment   as NameSegment
+import           Unison.NameSegment             ( NameSegment )
+import qualified Unison.NameSegment            as NameSegment
```
runarorama, line 132: I moved `NameSegment` from `Unison.Codebase` to the `Unison` namespace and flipped the dependency between `Name` and `NameSegment` as well as `HashQualified'` and `NameSegment`.

parser-typechecker/src/Unison/Codebase/BranchUtil.hs
```diff
@@ -60,12 +60,16 @@ getTerm (p, hq) b = case hq of
...
+  :: (Path.Path, HQ'.HQSegment) -> Branch0 m -> Metadata.R4 Referent NameSegment
 getTermMetadataHQNamed (path, hqseg) b =
   R4.filter (\(r,n,_t,_v) -> HQ'.matchesNamedReferent n r hqseg) terms
   where terms = Metadata.starToR4 . Branch._terms $ Branch.getAt0 path b
 
-getTypeMetadataHQNamed :: (Path.Path, HQSegment) -> Branch0 m -> Metadata.R4 Reference NameSegment
+getTypeMetadataHQNamed
+  :: (Path.Path, HQ'.HQSegment)
+  -> Branch0 m
+  -> Metadata.R4 Reference NameSegment
```
runarorama, line 72: Formatting change only. Ended up reverting a change to these, but left the formatting change since it's nicer to read.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -523,15 +523,13 @@ loop = do
...
           -> Action m (Either Event Input) v ()
         addDefaultMetadata adds = do
-          let
-            addedVs = Set.toList $ SC.types adds <> SC.terms adds
-            parseResult = Path.parseHQSplit' . Var.nameStr <$> addedVs
-            (errs, addedNames) = partitionEithers parseResult
-          case errs of
-            e : _ ->
+          let addedVs = Set.toList $ SC.types adds <> SC.terms adds
+              addedNs = traverse (Path.hqSplitFromName' . Name.fromVar) addedVs
```
runarorama, line 527: The main change here is the switch from `parseHQSplit'` to `hqSplitFromName'` which doesn't try to parse a `Path` by itself. The name here will never be just a path.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -104,7 +102,7 @@ parsePath' :: String -> Either String Path'
 parsePath' p = case parsePathImpl' p of
   Left e -> Left e
   Right (p, "") -> Right p
-  Right (p, rem) -> case Lexer.wordyId0 rem of
+  Right (p, rem) -> case (Lexer.wordyId0 <> Lexer.symbolyId0) rem of
```
runarorama, line 105: This allows path segments to be symbols.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -195,22 +193,27 @@ parseHQSplit s = case parseHQSplit' s of
...
+      . fmap (\sh -> (p, HQ'.HashQualified seg sh))
+      . SH.fromText
+      $ sh
+ where
   shError s = "couldn't parse shorthash from " <> s
+  parsePath n = do
+    x <- parsePathImpl' $ Text.unpack n
+    pure $ case x of
+      (Path' (Left e), "") | e == absoluteEmpty -> (relativeEmpty', ".")
+      x -> x
```
runarorama, line 216: The main change here is `parsePath` which treats `.` as the _object named `.`_ instead of the absolute namespace root. The thinking here is that `parseHQSplit'` should never be used to parse a path without an object name at the end.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -288,16 +294,15 @@ uncons = Lens.uncons
...
-  _    -> Path' . Right $ Relative path
+fromName' n = case take 1 (Name.toString n) of
+  "." -> Path' . Left . Absolute $ Path seq
+  _   -> Path' . Right $ Relative path
  where
-  path  = fromName n
-  seq   = toSeq path
-  first = Seq.take 1 seq
+  path = fromName n
+  seq  = toSeq path
```
runarorama, line 305: Since `fromName` works differently now, we can't rely on the hack that the first segment is empty to detect the root.

parser-typechecker/tests/Unison/Core/Test/Name.hs
```diff
@@ -4,18 +4,26 @@ module Unison.Core.Test.Name where
...
+      $ expectEqual (suffixes "foo.bar.baz") ["foo.bar.baz", "bar.baz", "baz"]
+    , scope "terms named `.`" $ expectEqual (suffixes "base..") ["base..", "."]
+    ]
+  , scope "segments" $ do
+    numDots <- int' 0 10
+    numSegs <- int' 0 10
+    n       <- int' 0 10
+    segs <- listOf n . pick $ replicate numDots "." ++ replicate numSegs "foo"
+    expectEqual (segments $ Name . pack $ intercalate "." segs)
+                (NameSegment . pack <$> segs)
```
runarorama, line 28: I believe this is a comprehensive test for `segments`

aryairani:
Oooh!

parser-typechecker/src/Unison/Codebase/Path.hs
aryairani: What does this semigroup do?

unison-core/src/Unison/NameSegment.hs
```diff
@@ -0,0 +1,41 @@
...
+
+import qualified Data.Text                     as Text
+import qualified Unison.Hashable               as H
+
+-- Represents the parts of a name between the `.`s
+newtype NameSegment = NameSegment { toText :: Text } deriving (Eq, Ord)
+
+-- Split text into segments. A smarter version of `Text.splitOn` that handles
+-- the name `.` properly.
+segments' :: Text -> [Text]
```
aryairani, line 15: What does the `'` mean here?

aryairani: What kinds of `Text` things should we be passing to `segments'`?  How does it relate to `NameSegment`?

parser-typechecker/tests/Unison/Core/Test/Name.hs
aryairani: What is this test testing?  Are the `segs` things like `"...foofoo"`?

unison-core/src/Unison/NameSegment.hs
aryairani: Would it make sense to be an implementation detail in `Name`?

parser-typechecker/src/Unison/Codebase/Path.hs
runarorama: This is just the `Either` semigroup. It picks the first `Right`, or the last `Left` if they're all `Left`.

parser-typechecker/tests/Unison/Core/Test/Name.hs
runarorama: segs is a list of elements which are either `.` or `foo`. So it generates paths like `foo...foo.foo......`

runarorama: It's testing that a rountrip from a list to a string back to a list via `segments` is the identity.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -116,7 +116,7 @@ todo = InputPattern
   )
   (\case
     patchStr : ws -> mapLeft (warn . fromString) $ do
-      patch  <- Path.parseSplit' Path.wordyNameSegment patchStr
+      patch  <- Path.parseSplit' Path.definitionNameSegment patchStr
```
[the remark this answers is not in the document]
runarorama: It controls how to parse the last segment, so the name of the patch.

unison-core/src/Unison/NameSegment.hs
runarorama: The `'` is here just since `segments` is taken by the `Name` module and these two are usually used in the same context. `NameSegment.segments'` is the internal implementation of `Name.segments`.

`segments'` will accept absolutely any text, just like `Text.splitOn`. It's purely a replacement for `Text.splitOn '.'`, as per the comment.

runarorama: `segments'` has to be called in places where we can't import `Name` because of a dumb circular dependency. It actually is an implementation detail, but one we have to expose. It's just "a better `Text.splitOn '.'`"

parser-typechecker/src/Unison/Codebase/Path.hs
aryairani: Ok, I was missing `instance Semigroup b => Semigroup (a -> b)` altogether, let alone `Semigroup (Either b)` :)

parser-typechecker/tests/Unison/Core/Test/Name.hs
aryairani: Ok, sorry, just learning some Haskell here; it's picking `n` segments out of a distribution with a ratio of `numDots`:`numSegs` dots to "foo"s?

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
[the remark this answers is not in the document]
aryairani: So this will let us have patches named, like `.`, right?  I think that will be confusing, no?  Maybe not.

unison-core/src/Unison/NameSegment.hs
aryairani: Oh, ok. :-\  Would you add a note to the source to that effect?

aryairani: Or can you pull some stuff out to a new module, to avoid the circular dependency?

parser-typechecker/src/Unison/Codebase/Path.hs
aryairani: Not sure I quite got this — is there anything else we need to know about how `fromName` works differently now?

parser-typechecker/tests/Unison/Core/Test/Name.hs
runarorama: That's the one.

parser-typechecker/src/Unison/Codebase/Path.hs
runarorama: The way it works differently is that instead of `...` giving the list `["","","",""]` (the empty strings between the dots), it gives `[".","."]` (the two namespaces named `"."` separated by a dot).

PR #1595, opened 2020-05-31
Fix Delta semigroup instance

### LINKED ISSUE #424: incorrect render width for watch expressions?
```
use Optional Some None
use Universal <

Sequence.foldl : (b -> a -> b) -> b -> [a] -> b
Sequence.foldl f b s =
  go i b = case Sequence.at i s of
    Some a -> go (i + 1) (f b a)
    None -> b
  go 0 b

> foldl (a -> b -> a Nat.+ b)

```

with a terminal width of 71 yielded:

image

whereas it seems like there should be a linebreak after `->` so that the line doesn't overflow?

It doesn't reflow until I reduce the terminal width to 58:

image

mitchellwrosen:
##  Overview

This patch amends `Pretty.Delta` to better model what it's meant to keep track of: the rectangular region that a `Pretty` fits in.

Previously, it kept track of

- Number of newlines
- Width of widest column
- Width of last column

But missing (crucially) was the width of the _first_ column. This is necessary for appending `Delta`s together: the first column of the right hand side is appended to the last column of the left hand side, and this new column may also now be the widest column.

Because this info was missing from `Delta` before, some things ended up overflowing (like #424) when they shouldn't have. This patch does fix #424, though due to an unrelated change in the term printer, the actual example given in that issue description no longer overflows. Still, I cherry-picked this work on top of `67a72c77263c0072cc0c99a190d73b6b9f23f0e9` (a random commit from around that time that exhibited the issue) and saw that it was fixed.

Another problem, also addressed: `Delta`'s `<>` implementation wasn't associative.

## Implementation notes

Before:

```haskell
-- lines, first col, max col
data Delta = Delta Int Int Int
```

Now:

```haskell
data Delta = 
    -- col
    SingleLine Int
    -- first col, last col, max col 
  | MultiLine Int Int Int
```

The switch to a sum type was just motivated by the observation that we don't care how many newlines a `Delta` contains, only whether it's `> 0`.

## Test coverage

The associativity of `Delta` is now under test.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -197,7 +197,7 @@ pretty0
     Request' ref i -> styleHashQualified'' (fmt S.Request) $
       elideFQN im $ PrettyPrintEnv.termName n (Referent.Con ref i CT.Effect)
     Handle' h body -> paren (p >= 2) $
-      if height > 0 then PP.lines [
+      if PP.isMultiLine pb || PP.isMultiLine ph then PP.lines [
         (fmt S.ControlKeyword "handle") `PP.hang` pb,
         (fmt S.ControlKeyword "with") `PP.hang` ph
        ]
@@ -207,7 +207,6 @@ pretty0
           <> (fmt S.ControlKeyword "with") `PP.hang` ph
       ]
       where
-        height = PP.preferredHeight pb `max` PP.preferredHeight ph
         pb = pblock body
         ph = pblock h
         pblock tm = let (im', uses) = calcImports im tm
@@ -224,16 +223,15 @@ pretty0
           <> optSpace <> (fmt S.DelimiterChar $ l "]")
       where optSpace = PP.orElse "" " "
     If' cond t f -> paren (p >= 2) $
-      if height > 0 then PP.lines [
+      if PP.isMultiLine pt || PP.isMultiLine pf then PP.lines [
         (fmt S.ControlKeyword "if ") <> pcond <> (fmt S.ControlKeyword " then") `PP.hang` pt,
         (fmt S.ControlKeyword "else") `PP.hang` pf
        ]
       else PP.spaced [
-        (fmt S.ControlKeyword "if") `PP.hang` pcond <> ((fmt S.ControlKeyword " then") `PP.hang` pt),
+        ((fmt S.ControlKeyword "if") `PP.hang` pcond) <> ((fmt S.ControlKeyword " then") `PP.hang` pt),
         (fmt S.ControlKeyword "else") `PP.hang` pf
        ]
      where
-       height = PP.preferredHeight pt `max` PP.preferredHeight pf
        pcond  = pretty0 n (ac 2 Block im doc) cond
        pt     = branch t
        pf     = branch f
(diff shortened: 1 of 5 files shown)


parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -224,16 +223,15 @@ pretty0
...
       where optSpace = PP.orElse "" " "
     If' cond t f -> paren (p >= 2) $
-      if height > 0 then PP.lines [
+      if PP.isMultiLine pt || PP.isMultiLine pf then PP.lines [
         (fmt S.ControlKeyword "if ") <> pcond <> (fmt S.ControlKeyword " then") `PP.hang` pt,
         (fmt S.ControlKeyword "else") `PP.hang` pf
        ]
       else PP.spaced [
-        (fmt S.ControlKeyword "if") `PP.hang` pcond <> ((fmt S.ControlKeyword " then") `PP.hang` pt),
+        ((fmt S.ControlKeyword "if") `PP.hang` pcond) <> ((fmt S.ControlKeyword " then") `PP.hang` pt),
```
mitchellwrosen, line 231: Oops, I left this in while investigating the relationship between infix `PP.hang` and `<>`. Should I remove it?

parser-typechecker/src/Unison/Util/Pretty.hs
```diff
@@ -132,7 +134,18 @@ instance Functor Pretty where
   fmap f (Pretty d o) = Pretty d (mapLit f $ fmap (fmap f) o)
 
 data F s r
-  = Empty | Group r | Lit s | Wrap (Seq r) | OrElse r r | Append (Seq r)
+  = Empty
+    -- | A group adds a level of breaking. Layout tries not to break a group
```
mitchellwrosen, line 138: I copied a short explanation from Paul on Slack, hope that's ok.

parser-typechecker/src/Unison/TermPrinter.hs
aryairani: I'd keep it.

parser-typechecker/src/Unison/Util/Pretty.hs
```diff
@@ -669,25 +710,49 @@ instance Monoid (Pretty s) where
...
 
 chDelta :: Char -> Delta
-chDelta '\n' = Delta 1 0 0
-chDelta _ = Delta 0 1 1
+chDelta '\n' = MultiLine 0 0 0
+chDelta _ = SingleLine 1
 
 preferredWidth :: Pretty s -> Width
-preferredWidth p = col (delta p)
+preferredWidth p = lastCol (delta p)
```
mitchellwrosen, line 749: This might be a bug. I decided not to change the behavior, but it seems weird to say the "preferred width" of something is however wide its last column is.

Still, I'd like to at least have an idea of what changing this to `maxCol` or something might fix, so I left it as-is for now.

parser-typechecker/unison-parser-typechecker.cabal
```diff
@@ -2,7 +2,7 @@ name:          unison-parser-typechecker
 category:      Compiler
 version:       0.1
 license:       MIT
-cabal-version: >= 1.8
+cabal-version: >= 1.10
```
mitchellwrosen, line 5: This fixes a warning

parser-typechecker/unison-parser-typechecker.cabal
```diff
@@ -223,6 +224,8 @@ library
     TypeApplications
     -- ViewPatterns
 
+  default-language: Haskell2010
```
mitchellwrosen, line 227: These things fix warnings

parser-typechecker/src/Unison/Util/Pretty.hs
aryairani: @pchiusano

parser-typechecker/unison-parser-typechecker.cabal
aryairani: Normally prefer to have this sort of thing in a separate PR, but no need to redo it.

pchiusano:
Cool! Nice work.

PR #1600, opened 2020-06-03
Make haddock-like comments not valid haddocks

mitchellwrosen:
This PR adjusts a few comments to not look like haddocks. `ghcide` (and possibly other tooling) picks them up as parse failures :grimacing:

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Causal.hs
+++ b/parser-typechecker/src/Unison/Codebase/Causal.hs
@@ -236,7 +236,7 @@ squashMerge combine c1 c2 = do
       -- Calling `combine` will recursively call into `squashMerge`
       -- for the children, discarding their history before calling `done`
       -- on the parent.
-      -- | lca == c2 -> pure $ done c1
+      --   | lca == c2 -> pure $ done c1
 
       | otherwise -> done <$> combine (Just $ head lca) (head c1) (head c2)
 
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -621,7 +621,8 @@ ac prec bc im doc = AmbientContext prec bc NonInfix im doc
 fmt :: S.Element -> Pretty S.SyntaxText -> Pretty S.SyntaxText
 fmt = PP.withSyntax
 
-{- # FQN elision
+{-
+   # FQN elision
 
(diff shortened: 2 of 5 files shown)


aryairani:
This is tough.  I ran into this with ghcide but it seems like a “them” problem.

One option is: if we care about haddock (I’m not convinced that we do), confirm that haddock also complains, and then add haddock to CI, or this won’t stick.

Otherwise: can we make ghcide less sensitive?

Can we confirm that ghcide is a fairly useful tool if we’ve jumped through some small hoops to appease it?

mitchellwrosen:
I'll look into these questions but off the top of my head:

> One option is: if we care about haddock (I’m not convinced that we do), confirm that haddock also complains, and then add haddock to CI, or this won’t stick.

Could do that; at work we have a "build haddocks but don't make haddock failures prevent merging" policy, because
- haddocks take a while to build
- it's annoying to fix haddock parse errors when no one's really looking at haddocks regularly
- we also think it's bad to give up on haddocks completely, so we don't want to totally neglect it

I'm not too concerned that this won't stick, so I think not requiring haddocks to pass CI is ok. The occasional comment that sneaks into master occasionally by someone whose workflow _doesnt_ break by this is easily fixed. But I can understand trying to keep everything clean at all times.

> Otherwise: can we make ghcide less sensitive?

I don't think so; I'll look around its issue tracker to see if this has come up before. `ghcide` does want haddock generation to succeed because it can display these docs inline an editor - I just don't know if it's willing to not even try. 

> Can we confirm that ghcide is a fairly useful tool if we’ve jumped through some small hoops to appease it?

I can confirm this, at least for me (vim user) it's the most solid tooling I've used, and has a ton of community support and momentum behind it.

mitchellwrosen:
https://github.com/digital-asset/ghcide/issues/573

According to this thread, you can pass `-haddock` to GHC, which will get these things to fail the build without having to actually generate haddocks.

aryairani:
> [digital-asset/ghcide#573](https://github.com/digital-asset/ghcide/issues/573)
> 
> According to this thread, you can pass `-haddock` to GHC, which will get these things to fail the build without having to actually generate haddocks.

Ah, ok.  Help me remember to do that when we create the .cabal files for the new project structure.

aryairani:
@mitchellwrosen At the end of that same thread, it looks like they've downgraded the haddock errors to warnings in https://github.com/digital-asset/ghcide/pull/608.

PR #1602, opened 2020-06-03
Add snapshot.yaml to speed up CI

mitchellwrosen:
This patch adds a "custom stack snapshot", which you can read about here: https://docs.haskellstack.org/en/stable/pantry/

This doesn't change the build at all, only where stack puts things: all of these dependencies are now cached in `~/.stack`, rather than `.stack-work/`, so CI should be a bit faster.

### DIFF
--- a/snapshot.yaml
+++ b/snapshot.yaml
@@ -1,0 +1,15 @@
+resolver: lts-15.15
+
+packages:
+  - github: unisonweb/configurator
+    commit: e47e9e9fe1f576f8c835183b9def52d73c01327a
+  - github: unisonweb/haskeline
+    commit: 2944b11d19ee034c48276edc991736105c9d6143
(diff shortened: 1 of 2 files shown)


aryairani:
Looking at [lines 4311–4312](https://travis-ci.org/github/unisonweb/unison/builds/694288306#L4311) of the build log, it looks like nothing actually changed — it claims there was nothing new to cache, e.g. from `~/.stack`. 

Is it possible that those stack.yaml extra-deps were already going into `~/.stack` and not `~/.stack-work`?

mitchellwrosen:
Good point...! I just tested this locally, too. Stack does create a snapshot in ~/.stack that includes the extra-deps defined in stack.yaml. :man_shrugging: According to (one person on) the FP slack, this is the difference between extra-deps and a custom snapshot.

Issue #1597, opened 2020-05-31
Create a non-leaky data structure to canonicalize bytestrings and text


aryairani:
## The problem

The codebase is full of 512-bit hashes and `Text`-based names.  When we load these `ByteString` and `Text` values into memory, they are all separate instances of the same data.  Since they are immutable, it would be better to let them all share the same copy in memory.

We could use `Map` or `Unison.Util.Cache`, but they don't really give us a way to release unused memory, at most they let us cap the maximum memory usage of the Map.  I want to be able to garbage-collect the mapping once there are no live references to the data, but the Map contains live references to the data, so every hash will stay around forever.

This would probably be an improvement over the current representation, honestly, but let's just go ahead and build something that will release what isn't needed.

So this leads to: some kind of Map of [`System.Mem.Weak`](http://hackage.haskell.org/package/base-4.14.0.0/docs/System-Mem-Weak.html) pointers.
> A weak pointer expresses a relationship between two objects, the key and the value: if the key is considered to be alive by the garbage collector, then the value is also alive. A reference from the value to the key does not keep the key alive.

```haskell
-- | Establishes a weak pointer to @k@, with value @v@ and a finalizer.
--
-- This is the most general interface for building a weak pointer.
--
mkWeak  :: k                            -- ^ key
        -> v                            -- ^ value
        -> Maybe (IO ())                -- ^ finalizer
        -> IO (Weak v)                  -- ^ returns: a weak pointer object

-- | A specialised version of 'mkWeak', where the key and the value are
-- the same object:
--
-- > mkWeakPtr key finalizer = mkWeak key key finalizer
--
mkWeakPtr :: k -> Maybe (IO ()) -> IO (Weak k)
mkWeakPtr key finalizer = mkWeak key key finalizer

Dereferences a weak pointer.  If the key is still alive, then
@'Just' v@ is returned (where @v@ is the /value/ in the weak pointer), otherwise
'Nothing' is returned.

The return value of 'deRefWeak' depends on when the garbage collector
runs, hence it is in the 'IO' monad.
-}
deRefWeak :: Weak v -> IO (Maybe v)
deRefWeak (Weak w) = IO $ \s ->
   case deRefWeak# w s of
        (# s1, flag, p #) -> case flag of
                                0# -> (# s1, Nothing #)
                                _  -> (# s1, Just p #)
```
## The task

[...]

PR #1605, opened 2020-06-04
Add PinBoard for canonicalizing strings

mitchellwrosen:
## Overview

This PR adds a `PinBoard` type, as described by @aryairani in #1597, which is useful for de-duping values in memory (like hashes).

I'll copy the module doc here:

```
A utility type for saving memory in the presence of many duplicate ByteStrings, etc. If you have data that may be
a redundant duplicate, try pinning it to a pin board, and use the result of that operation instead.

  Without a pin board:

    x ───── "38dce848c8c829c62"
    y ───── "38dce848c8c829c62"
    z ───── "d2518f260535b927b"

  With a pin board:

    x ───── "38dce848c8c829c62" ┄┄┄┄┄┐
    y ────────┘                     board
    z ───── "d2518f260535b927b" ┄┄┄┄┄┘

  ... and after x is garbage collected:

            "38dce848c8c829c62" ┄┄┄┄┄┐
    y ────────┘                     board
    z ───── "d2518f260535b927b" ┄┄┄┄┄┘

  ... and after y is garbage collected:

                                    board
    z ───── "d2518f260535b927b" ┄┄┄┄┄┘
```

and here's the API:

```haskell
new :: IO (PinBoard a)
pin :: (Eq a, Hashable a) => PinBoard a -> a -> IO a
```

When you `pin` something, either you'll get a pointer-equal value back (hasn't been pinned before) or an `Eq`ual-but-not-pointer-equal value back (was pinned already).

## Implementation notes

It's implemented exactly as @aryairani described in #1597, using an `IntMap`.

Data structure: it's currently `IntMap [Weak a]`, but could be `HashMap`, `HashTable`, something else? It wouldn't be hard to change.

Name: I picked `PinBoard` because this abstraction reminded me of a bulletin board, but `BulletinBoard` seemed long.

API: there's no implicit global pin board defined (I didn't know exactly how this thing would be used), but it's simple enough to add. That'd reduce the API to just one function per type, e.g.

```haskell
pinByteString :: ByteString -> IO ByteString
pinText :: Text -> IO Text
```
 
## Test coverage

One test was added that (I believe) covers the entire API.

### DIFF
--- a/parser-typechecker/src/Unison/Util/PinBoard.hs
+++ b/parser-typechecker/src/Unison/Util/PinBoard.hs
@@ -47,7 +47,7 @@ import qualified Data.Text as Text
 import qualified Data.Text.IO as Text
 import Data.Tuple (swap)
 import System.Mem.Weak (Weak, deRefWeak, mkWeakPtr)
-import Unison.Prelude hiding (empty)
+import Unison.Prelude
 
 -- | A "pin board" is a place to pin values; semantically, it's a set, but differs in a few ways:
 --
--- a/parser-typechecker/src/Unison/Util/PinBoard.hs
+++ b/parser-typechecker/src/Unison/Util/PinBoard.hs
@@ -1,0 +1,95 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Unison.Util.PinBoard
+  ( PinBoard,
+    new,
+    pin,
+
+    -- * For debugging
+    debugDump,
+  )
+where
+
+import Control.Concurrent.MVar
+import Data.Foldable (find)
+import Data.Functor.Compose
+import Data.Hashable (Hashable, hash)
+import qualified Data.IntMap as IntMap
+import Data.IntMap.Strict (IntMap)
+import qualified Data.Text as Text
+import qualified Data.Text.IO as Text
+import Data.Tuple (swap)
+import System.Mem.Weak (Weak, deRefWeak, mkWeakPtr)
+import Unison.Prelude hiding (empty)
+
+-- | A "pin board" is a place to pin values, with the following properties:
+--
+--   * Pinned values aren't kept alive by the pin board, they might be garbage collected at any time.
+--   * If you try to pin a value that's already pinned (per its Eq instance), the pinned one will be returned instead.
+newtype PinBoard a
+  = PinBoard (MVar (IntMap (Bucket a)))
+
+new :: IO (PinBoard a)
+new =
+  PinBoard <$> newMVar IntMap.empty
+
+pin :: (Eq a, Hashable a) => PinBoard a -> a -> IO a
+pin (PinBoard boardVar) x =
+  modifyMVar boardVar \board -> pin_ board x
+
+pin_ :: forall a. (Eq a, Hashable a) => IntMap (Bucket a) -> a -> IO (IntMap (Bucket a), a)
+pin_ board x =
+  swap <$> getCompose (IntMap.alterF (fmap Just . Compose . maybe miss hit) (hash x) board)
+  where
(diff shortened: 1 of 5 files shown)


mitchellwrosen:
Thanks :)

> I know it may be hard in an automated test, but were you able to observe that the finalizing / garbage collection works as expected?

Good point! I went back to improve the test and uncovered a bug in the implementation. It was due to the fact that I assumed finalizers are _always_ run during the garbage collection that frees their associated data, when in fact they are not:

> It is not guaranteed that a finalizer will eventually run, and no attempt is made to run outstanding finalizers when the program exits

This was tripping an `assert`, which I removed.

The test now does check that the board doesn't keep the contained values alive, but due to this lack of guarantee, I don't think I can (easily) write a reliable test that ensures the finalizers indeed delete keys from the data structure. But I can see that they do, via playing around the REPL and looking at the output of `PinBoard.debugDump`

pchiusano:
There’s probably little to no contention on this cache, so TVar might be a better fit: https://codeburst.io/the-haskell-concurrency-primitive-shootout-538c21993f1c

mitchellwrosen:
@pchiusano Interesting, I wasn't able to reproduce that finding.

2020-06-04-231634_1016x130_scrot

Code, which compares

```haskell
modifyMVar :: MVar a -> (a -> IO (a, b)) -> IO b
modifyTVarIO :: TVar a -> (a -> IO (a, b)) -> IO b
```

```haskell
{-# LANGUAGE BlockArguments #-}

module Main where

import Control.Concurrent
import Control.Concurrent.STM
import Chronos.Bench
import System.IO.Unsafe
import Control.Exception

main = defaultMain
  [ benchIO "mvar" (modifyMVar mvar (\x -> pure (x, ())))
  , benchIO "tvar" (modifyTVarIO tvar (\x -> pure (x, ())))
  ]

modifyTVarIO :: TVar (Maybe a) -> (a -> IO (a, b)) -> IO b
modifyTVarIO v f =
  mask \restore -> do
    x <- atomically do
      mx <- readTVar v
      case mx of
        Nothing -> retry
        Just x -> do
          writeTVar v Nothing
          pure x
    (x', r) <- restore (f x >>= evaluate) `onException` do
      atomically do
        mx <- readTVar v
        case mx of
          Nothing -> writeTVar v (Just x)
          Just _ -> retry
    atomically do
      mx <- readTVar v
      case mx of
        Nothing -> writeTVar v (Just x)
        Just _ -> retry
    pure r

mvar = unsafePerformIO (newMVar ())
{-# NOINLINE mvar #-}

tvar = unsafePerformIO (newTVarIO (Just ()))
{-# NOINLINE tvar #-}
```

pchiusano:
The benchmark you wrote isn’t really directly comparing MVar vs TVar performance since you’ve got extra stuff in there to allow the modify function to do IO (not needed for canonicalization use case).

You could just compare a simple loop that sums the numbers up to a million by repeatedly adding to a mutable “currentSum” variable which is either an MVar or TVar.

mitchellwrosen:
🤔 I may be missing something. I do need to modify the var in IO, because it involves either creating a weak reference, or dereferencing one that's already there, or both. I will try benchmarking pure updates, too, though.

pchiusano:
@mitchellwrosen you are totally right, forgot that weak refs are in IO. Carry on. 😀

I’m kinda curious on the pure updates case but not really relevant here.

mitchellwrosen:
@pchiusano Worth investigating for sure! At least with this with this quick and dirty command-liney benchmark tool I also observed single-threaded MVar performance to outperform single-threaded TVar performance about 2.2x.

Issue #1603, opened 2020-06-03
bigly rearrange modules


aryairani:
I was thinking something like this:

## transcripts

./parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
./parser-typechecker/transcripts/Transcripts.hs

## term

./unison-core/src/Unison/ABT.hs
./unison-core/src/Unison/Blank.hs
./unison-core/src/Unison/ConstructorType.hs
./unison-core/src/Unison/DataDeclaration.hs
./unison-core/src/Unison/Hash.hs
./unison-core/src/Unison/LabeledDependency.hs
./unison-core/src/Unison/Pattern.hs
./unison-core/src/Unison/PatternP.hs
./unison-core/src/Unison/Reference.hs
./unison-core/src/Unison/Reference/Util.hs
./unison-core/src/Unison/Referent.hs
./unison-core/src/Unison/Symbol.hs
./unison-core/src/Unison/Term.hs
./unison-core/src/Unison/Type.hs
./unison-core/src/Unison/Var.hs

## runtime-v1

./parser-typechecker/src/Unison/Runtime/ANF.hs
./parser-typechecker/src/Unison/Runtime/IR.hs
./parser-typechecker/src/Unison/Runtime/Rt1.hs
./parser-typechecker/src/Unison/Runtime/Rt1IO.hs

## runtime-api

./parser-typechecker/src/Unison/Codebase/CodeLookup.hs
./parser-typechecker/src/Unison/Codebase/Runtime.hs

## runtime-builtin

./parser-typechecker/src/Unison/Builtin.hs
./parser-typechecker/src/Unison/Builtin/Decls.hs
./parser-typechecker/src/Unison/Runtime/IOSource.hs

## runtime-data

./parser-typechecker/src/Unison/Runtime/SparseVector.hs
./parser-typechecker/src/Unison/Runtime/Vector.hs

## codebase

./parser-typechecker/src/Unison/Codebase.hs
./parser-typechecker/src/Unison/Codebase/Branch.hs
./parser-typechecker/src/Unison/Codebase/Branch/Dependencies.hs
./parser-typechecker/src/Unison/Codebase/BranchDiff.hs
./parser-typechecker/src/Unison/Codebase/BranchUtil.hs
./parser-typechecker/src/Unison/Codebase/Causal.hs
./parser-typechecker/src/Unison/Codebase/FileCodebase.hs
./parser-typechecker/src/Unison/Codebase/FileCodebase/Common.hs
[...]

PR #1609, opened 2020-06-08
Add -rtsopts to unison ghc-options

mitchellwrosen:
This PR just adds `-rtsopts` to the unison executable, so a greater number of runtime knobs can be tweaked with command-line options

### DIFF
--- a/parser-typechecker/unison-parser-typechecker.cabal
+++ b/parser-typechecker/unison-parser-typechecker.cabal
@@ -237,7 +237,7 @@ executable unison
   main-is: Main.hs
   hs-source-dirs: unison
   default-language: Haskell2010
-  ghc-options: -Wall -fno-warn-name-shadowing -fno-warn-missing-pattern-synonym-signatures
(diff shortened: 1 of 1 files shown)


aryairani:
We'll have to do this again after #1603.  Also do you know if and how it affects build times?

aryairani:
Also I think `stack build --profile` does this, along with building the libraries with the right settings? or am I mixed up?

mitchellwrosen:
It doesn't affect build times, it's just whether the resulting executable allows to to tweak things (honestly, I'm not really sure why you wouldn't want to set this :shrug:)

About `stack build --profile` - I'm not sure if that sets `-rtsopts` too; at least in GHC they're distinct (`-rtsopts` and `-prof`)

aryairani:
If `-rtsopts` is required to be able to supply `+RTS ...` options to the build executable's commandline, then `--profile` definitely enables at least that;  if it isn't, then I dunno.

mitchellwrosen:
Not all `+RTS` opts are disabled without `-rtsopts`, only "most" according to GHC. `-s` (print stats) and `-p` (profile? or something?) both seem to work, but things like tweaking the nursery size (`-A`) to not/

PR #1611, opened 2020-06-11
Fix #1131 : Add windows Ctrl+C handling

sharno:
Add windows Ctrl+C handling
Fixes #1131 

The test for this maybe adding a windows target in CI?
But there are other issues in #598 and tests that don't pass yet

PR #1615, opened 2020-06-17
Revert "Fix #1131 : Add windows Ctrl+C handling"

### LINKED ISSUE #1131: Add Windows compliant ctrl+c handling
#912 added Unix compliant ctrl c handling in ucm. Discussion there called for a followup to add Windows compatibility. See [ghci](https://gitlab.haskell.org/ghc/ghc/blob/581cbc28e143a4ed8e7f794ed1618161222a5646/libraries/ghci/GHCi/Signals.hs#L35-L46) for an example.

aryairani:
Revert #1611 due to macOS build regression / interaction with https://github.com/haskell/cabal/issues/4739

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -44,4 +44,3 @@ The format for this list: name, GitHub handle, and then optional blurb about wha
 * Dave Nicponski (@virusdave)
 * Cody Allen (@ceedubs)
 * Ludvig Sundström (@lsund)
-* Mohamed Elsharnouby (@sharno)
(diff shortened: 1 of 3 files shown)


sharno:
Instead of reverting, isn't this better handled with `-optP-Wno-nonportable-include-path`?
Seems that it'll succeed if we just silence this error.

example:
https://github.com/PostgREST/postgrest/pull/1332

aryairani:
@sharno Maybe — I'm not sure I want to revert either; what I'd like to do is undo the merge and its effect on the git history, put better CI in place, and then try going forward again from there, including possibly the `-optP-Wno-nonportable-include-path` route or some other.  

Unfortunately, undoing the effect on the git history isn't possible, as others have already pulled the failing `master` branch, so I believe reverting is still the best option for now.  Then we can try to recreate your PR in a way that passes the updated CI.

I'm just waiting to confirm that https://travis-ci.org/github/unisonweb/unison/builds/699385107 fails as expected.

PR #1619, opened 2020-06-18
Part of #598: Fix frozen windows executable

sharno:
Apparently FSNotify requires `-threaded` flag to work properly. And I think it's best to use the flag anyway to use the multi-threaded runtime

Part of #598


aryairani:
Thanks for your help with these, @sharno. 

Here's the info I think you had found:
https://hackage.haskell.org/package/fsnotify
> Requirements
> Windows
> compile with `-fthreaded`

aryairani:
:+1:

Issue #1497, opened 2020-04-30
Updating documentation or other metadata requires tedious unlinking and relinking


pchiusano:
Currently, if you update a `Doc`, you have to unlink the old documentation and then link the new docs. I think that `update` should just replace linked metadata as well. 

So if `mydocs` is linked to a definition, `foo`, and `mydocs -> mydocs'` is in the patch, then `mydocs` will be unlinked from `foo` and `mydocs'` will be linked to it after the patch is applied.

Suggested workaround for now is to just `unlink` the existing docs (say it's called `foo.mydocs`, do the `update` of it), then `link foo.mydocs foo` the now updated version back in.

Originally I was concerned that this could be a bad idea, but now that I understand patch management better, I think it's the right move.

aryairani:
This makes sense to me.  So given `d -> m`, 
updating `m` yields `d -> m'`
updating `d` yields `d' -> m`
updating `d` and `m` yields `d' -> m'`

aryairani:
Same as #1497

pchiusano:
Rúnar commented that it’s not possible to unlink the old metadata after the update, since it no longer has a name. I’m unsure if that’s still and issue but flagging here to check.

ceedubs:
Fixed by #1853

PR #1621, opened 2020-06-18
Update links when propagating changes to metadata

### LINKED ISSUE #1349: Old metadata stays linked after update
1. Add a `Doc` to the codebase.
2. Link it to a definition.
3. Edit the `Doc` from step 1 and `update`.

The link from step 2 stays between the old Doc and the definition. I'd expect it to get removed and replaced with a link to the Doc from step 3.

There's also no way of unlinking the old `Doc`, since it now has a hash-qualified name and `unlink` can't handle those.

runarorama:
## Overview

Fixes #1349 and #1497 

This is a slight change to the `propagate` code so that when an `update` occurs, not only is the name updated to point to the new reference, but also if the old reference participates in any links somewhere under the current namespace, those links are updated to use the new reference.

## Implementation notes

This is a tiny change to the code that updates the `Star3` for the relevant branch during `propagate`.

## Test coverage

A couple of existing transcripts relied on the old behaviour. I've updated them so they now test the new behaviour.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -552,7 +552,7 @@ loop = do
                     -> [(Path', HQ'.HQSegment)]
                     -> [HQ.HashQualified]
                     -> (forall r. Ord r
-                        => (r, Reference, Reference)
+                        => (r, Metadata.Type, Metadata.Value)
                         ->  Branch.Star r NameSegment
                         ->  Branch.Star r NameSegment)
                     -> Action m (Either Event Input) v ()
@@ -2588,6 +2588,8 @@ doSlurpUpdates typeEdits termEdits deprecated b0 =
     Just split -> [ BranchUtil.makeDeleteTermName split (Referent.Ref old)
                   , BranchUtil.makeAddTermName split (Referent.Ref new) oldMd ]
       where
+      -- oldMd is the metadata linked to the old definition
+      -- we relink it to the new definition
       oldMd = BranchUtil.getTermMetadataAt split (Referent.Ref old) b0
   errorEmptyVar = error "encountered an empty var name"
 
--- a/parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
@@ -44,6 +44,7 @@ import           Unison.UnisonFile              ( UnisonFile(..) )
 import qualified Unison.UnisonFile             as UF
 import qualified Unison.Util.Star3             as Star3
 import           Unison.Type                    ( Type )
+import qualified Unison.Type                   as Type
 import qualified Unison.Typechecker            as Typechecker
 import           Unison.ConstructorType         ( ConstructorType )
 import qualified Unison.Runtime.IOSource       as IOSource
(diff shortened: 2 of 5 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -552,7 +552,7 @@ loop = do
                     -> [(Path', HQ'.HQSegment)]
                     -> [HQ.HashQualified]
                     -> (forall r. Ord r
-                        => (r, Reference, Reference)
+                        => (r, Metadata.Type, Metadata.Value)
```
runarorama, line 555: This change just clarifies what these types are.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2588,6 +2588,8 @@ doSlurpUpdates typeEdits termEdits deprecated b0 =
     Just split -> [ BranchUtil.makeDeleteTermName split (Referent.Ref old)
                   , BranchUtil.makeAddTermName split (Referent.Ref new) oldMd ]
       where
+      -- oldMd is the metadata linked to the old definition
+      -- we relink it to the new definition
       oldMd = BranchUtil.getTermMetadataAt split (Referent.Ref old) b0
```
runarorama, line 2593: I needed to know what this line did, so I added a comment.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -465,11 +465,12 @@ applyPropagate patch Edits {..} = do
       (if isPropagated r'
        then Metadata.insert (propagatedMd (Referent.Ref r'))
        else Metadata.delete (propagatedMd (Referent.Ref r'))) .
-      Star3.replaceFact (Referent.Ref r) (Referent.Ref r') $ s
+      Star3.replaceFact (Referent.Ref r) (Referent.Ref r') .
+        Star3.mapD3 (\(tp, v) -> if v == r then (tp, r') else (tp, v)) $ s
```
runarorama, line 469: This is really all there is to this change. Update the range of the `d3` wherever the old reference occurs and point it to the new refefence instead.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: lol

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
aryairani: Is there any chance that `r` and `r'` would not have the same type `tp`?

runarorama: I suppose that's possible, yes. I'll need to look up the type of the new reference.

aryairani: And then in that situation, I guess you'll have to update D2 also :-\

runarorama: That's right. Fixed that now

pchiusano: Nice catch. Code review ftw!

Issue #1594, opened 2020-05-30
overhaul old names behavior


aryairani:
We'd talked yesterday about removing or depth-limiting historical names search.

As I was falling asleep (*disclaimer), I found myself doubting whether a depth-limited search was any good?

My thinking is that the distance to a match is going to be arbitrarily long (at least in the current model of historical names search, which begins at the root), and that there's barely any meaningful concept of recency, that the distance to a match is going to be arbitrarily long. _e.g._, a lot of activity in `.project.x` kills your ability to pick up older work that depends on historical names in `.project.y`.

If a global, depth-limited search isn't practically useful (and I suspect it won't be), we could just remove it altogether, but should make a couple of changes to avoid losing relevant names to begin with, and to deal with already-lost names:

1. Add commands to do explicit historical search: `names.old.all #hash [namespace]` and `names.old.first #hash [namespace]`.  The last argument is the search root, and the current namespace is used if none is supplied.

2. Deleting the last name for a hash that has dependents in the current root namespace requires confirmation.
    * This requires an efficient way to query the dependents of the current root namespace.

3. Updating a definition should not delete the old name unless a) there are no more dependents of the old definition at the end of propagation, or b) there is an alternative name for the old definition outside of the propagation scope.
    * This requires an efficient way to query the dependents of the current root namespace.
  
    So `update foo` would result in a conflicted name for `foo` in some cases.  UCM could warn you in this situation and suggest a rename command, e.g. `rename.term name#oldhash name.old[n].` for some `n` such that `name.old[n]` is unique.

4. `push`ing needs some combination of
    - a) a warning/confirmation that not all of the namespace's dependencies are internally named, and/or 
    - b) an `alias.dependencies <source path> <target path> [query path]` command which copies names of dependencies missing from `query path` out of `source path` into `target path` with relative names.  e.g.  `alias.dependencies .libs.alice.mylib.v94 external.mylib.v94` or
    - c) some algorithm for auto-aliasing external names as needed.

Wdyt @pchiusano @runarorama

aryairani:
Superseded by #2128

PR #1629, opened 2020-07-06
Remove old names

runarorama:
## Overview

This gets rid of all uses of "old names", which is when UCM searches the history for a name for a hash if it can't find one in the namespace root.

The reason is that searching through the history can take a very long time (minutes, if the history is long), rendering UCM unresponsive for the duration. This happened whenever you

1. Viewed or edited any definition in your `todo` set.
2. Viewed any patch.

With this change, UCM simply shows the hashes for definitions that no longer have names.

## Loose ends

See #1594 for a discussion on a more permanent solution.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -32,9 +32,9 @@ module Unison.Codebase.Branch
   , head
   , headHash
   , before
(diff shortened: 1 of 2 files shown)


aryairani:
Howdy, I got the review request but I see the build is failing and the PR is still marked as a draft. Is there something specific I should review?

Issue #1613, opened 2020-06-12
Incorrect zero counting ops on Int


dolio:
The type signatures of `Int.leadingZeros` and `Int.trailingZeros` are `Int -> Nat`. However the checked-in test output has signs, which suggests they are actually returning `Int`.

I've implemented them according to the type signature in the new runtime, but this is causing spurious test failures, and should probably be fixed in the existing runtime.

PR #1631, opened 2020-07-17
New Runtime Implementation

dolio:
This pull request contains the new runtime work. By default, the existing runtime is still used, but the new one can be enabled by passing `--new-runtime` on the command line, or putting `new-runtime: true` in the config file.

The majority of the test suite passes using the new runtime. One failure is an actual incorrect test case (#1613). The other is the recently added IO case, which I haven't looked into just yet. There are also some transcript differences that I haven't resolved yet.

In general, the `IO` implementation is relatively untested, and may have bugs, but I thought this pull request might signal that someone besides me could actually try things out, as I might not find all the issues myself.

The extra levels of code generation down to the executable layer are far from ideal, but even so some simple loops should run faster than the existing runtime. Loops with things like ability usage might be worse or the same, because those are the areas that are noticeably far from optimal with respect to code generation/optimization.

One significant thing I have not done is put the new runtime in a separate subproject, even though I think it should probably be in one. This is because the main program is under `parser-typechecker/`, but that portion of the project also contains a lot of machinery that is prior to the new runtime and code generator layers for it. If the project reorganization is merged prior to this pull request, I can rearrange my new code to be somewhat more self-contained.

Questions/comments welcome.


aryairani:
Awesome!

> The majority of the test suite passes using the new runtime. One failure is an actual incorrect test case (#1613). The other is the recently added IO case, which I haven't looked into just yet. There are also some transcript differences that I haven't resolved yet.

IMO, we should address these before merging, and have CI run the tests on both the old runtime and the new one (if that's not already part of this PR).

> In general, the IO implementation is relatively untested, and may have bugs, but I thought this pull request might signal that someone besides me could actually try things out, as I might not find all the issues myself.

Totally.  I think once existing tests passing under CI, we merge it to master and let people start trying it with the flag.  (Though they are welcome to try it in this branch?) Once an arbitrarily short amount of time passes, we can switch the default to the new runtime; and once the old runtime is known to be strictly worse or at least buggier in a way that we don't want to fix, we can dump it altogether.

atacratic:
Awesome this is landing!

Is it for definite the destination is bin the old runtime, rather than keep both? The old one might act as a simple reference implementation, and (I'm thinking at some point) one with extra instrumentation for analysing the code being run. Instrumenting the new runtime might be more of a challenge since it's more involved, and also the priority will be avoiding compromising speed.

aryairani:
@atacratic I think we can keep both as long as they both work; I was just assuming that over time we'll encounter issues with the old one that nobody wants to fix, and under those circumstances, it'd probably not be worth keeping.

atacratic:
Is there a way to rerun all the tests in a codebase using the new runtime? Without the test cache preventing it?

atacratic:
@dolio I brought the new runtime up in my codebase and ran 'test' and hit the following.  Known issue?  LMK if you want more info.

```
chris@chris-thinkpad:~/unison$ stack exec unison -- --new-runtime

   _____     _             
  |  |  |___|_|___ ___ ___ 
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|
  
  Welcome to Unison!
  
  You are running version: devel/M1l-189-gdafb3c57
  
  I'm currently watching for changes to .u files under ~/unison
  
  Type help to get help. 😎


  I loaded scratch.u and didn't find anything.

.> names base.List.Nonempty.foldr.test

  Term
  Hash:   #07s0393c3l
  Names:  _base.releases._M1m.List.Nonempty.foldr.test
          _base.releases._latest.List.Nonempty.foldr.test
          _base.series._M1m.List.Nonempty.foldr.test
          _base.trunk.List.Nonempty.foldr.test
          base.List.Nonempty.foldr.test

.> cd base.List.Nonempty
.base.List.Nonempty> test

    
    Cached test results (`help testcache` to learn more)
    
    ◉ scanl.test    : Passed 100 tests.
    ◉ scanr.test    : Passed 100 tests.
    
    ✅ 2 test(s) passing
    
    ✅  

14 tests left to run, current test: foldr.testunison: buildMatrix: empty rows
  error, called at src/Unison/Runtime/Pattern.hs:275:18 in unison-parser-typechecker-0.1-AVd5UsYOj5GOFXnV4xcly:Unison.Runtime.Pattern

```

dolio:
No, not a known issue. I've only been running the build time test suite, though, not any tests written inside unison.

pchiusano:
NICE!! Looking forward to this. 

@dolio I would work on getting all the tests in base to pass with the new runtime, and we should turn that into a a test transcript at some point (maybe run separately from the other transcripts since it will fetch base):

```
.> pull https://github.com/unisonweb/base:.releases._latest base
.> cache.clear -- doesn't exist yet, but we should add a command for this
.base> test
```

I'd like to see some high level documentation of the compilation process. Can be pretty short I think. Basically, what are all the types involved, what are the phases of compilation, where does the code for different phases live and what are interesting things along the way. We should do a synchronous review and come up w/ rough outline for this once you've ironed out the bugs you find in getting all the base tests to pass. 

I'd also like to see comments in the code for all types and the constructors if they aren't self-explanatory. Like there's stuff in `MCode` with some abbreviations whose meaning isn't obvious, and probably other places as well.

https://github.com/unisonweb/unison/blob/af12fab03f2638fdd2588d5e4ae3a6bb37bbac5b/parser-typechecker/src/Unison/Runtime/MCode.hs#L360

dolio:
Yeah, I've been neglecting some documentation more than I like as time went on, because I've just been trying to get the implementation done. You can see that there's a big writeup of the runtime ideas from a while ago, but less about newer stuff. I'll try to fill it in. The `MCode` instructions almost all have analogues in the `ANF` code that do say what they are, but it doesn't hurt to say in both places.

dolio:
So, I've been digging more into the IO stuff, and there's a pretty significant problem, I think. The `IOSource` stuff is written in a way that is pretty specific to the old runtime choices. For instance, it thinks file handles are text. So, that carries over to the code it feeds into the new runtime. But, file handles in the new runtime are expected to actually be file handles.

I kind of think that some of these choices don't really make sense anyway. For instance, treating file handles as text keys into a local map is just giving a false sense that they can be stored or transferred to other machines. However, I don't know whether it's worth it to modify the existing runtime at the moment.

pchiusano:
I wouldn't worry about updating the old runtime or trying to have new stuff match what IOSource is currently doing.

dolio:
Well, the point is that this is no longer just swapping out a runtime implementation. There's a big embedded text file that no longer entirely makes sense, and it seems way higher in the pipeline than the runtime. Also, there are definitions in it that are necessary for the new runtime, like `EIther`. The references to those definitions are named as Haskell values, and used in the primops to build the right data and such. I also don't know if it makes sense to swap that text file out from under a repository that builds on the existing one.

parser-typechecker/src/Unison/Builtin/Decls.hs
```diff
@@ -115,6 +129,68 @@ builtinDataDecls = rs1 ++ rs
...
+          (var "a" `arr` Type.apps' (var "Either") [var "a", var "b"])
+      )
+    , ( ()
+      , v "Either.Right"
+      , Type.foralls () [v "a", v "b"]
+          (var "b" `arr` Type.apps' (var "Either") [var "a", var "b"])
+      )
+    ]
+  fmode = DataDeclaration
+    Structural
```
pchiusano: @dolio any of the file / IO enums should be `unique` types. You pick the GUID, via `head -c 1024 /dev/urandom | shasum`.

parser-typechecker/src/Unison/Builtin/Decls.hs
```diff
@@ -115,6 +129,68 @@ builtinDataDecls = rs1 ++ rs
...
+    ()
+    []
+    [ ((), v "FileMode.Read", var "FileMode")
+    , ((), v "FileMode.Write", var "FileMode")
+    , ((), v "FileMode.Append", var "FileMode")
+    , ((), v "FileMode.ReadWrite", var "FileMode")
+    ]
+  bmode = DataDeclaration
+    Structural
```
pchiusano: same deal here, should be `unique`

parser-typechecker/src/Unison/Builtin/Decls.hs
```diff
@@ -115,6 +129,68 @@ builtinDataDecls = rs1 ++ rs
...
+    ()
+    []
+    [ ((), v "BufferMode.NoBuffering", var "BufferMode")
+    , ((), v "BufferMode.LineBuffering", var "BufferMode")
+    , ((), v "BufferMode.BlockBuffering", var "BufferMode")
+    , ((), v "BufferMode.SizedBlockBuffering"
+      , Type.nat () `arr` var "BufferMode")
+    ]
+  ioerr = DataDeclaration
+    Structural
```
pchiusano: `unique` here too

parser-typechecker/src/Unison/Builtin/Decls.hs
```diff
@@ -115,6 +129,68 @@ builtinDataDecls = rs1 ++ rs
...
+    [ ((), v "AlreadyExists", var "IOError")
+    , ((), v "NoSuchThing", var "IOError")
+    , ((), v "ResourceBusy", var "IOError")
+    , ((), v "ResourceExhausted", var "IOError")
+    , ((), v "EOF", var "IOError")
+    , ((), v "IllegalOperation", var "IOError")
+    , ((), v "PermissionDenied", var "IOError")
+    , ((), v "UserError", var "IOError")
+    ]
+  seqview = DataDeclaration
```
pchiusano, line 179: Can you explain how this is used? Just curious.

I think it'd be (in Unison syntax) `SeqView a = Empty | SeqView a (SeqView a)` which is just a cons list, right?

I could also imagine `SeqView a seq = Empty | SeqView a seq` which you might use for uncons-ing or unsnoc-ing "one step at a time".

unison-src/transcripts/alias-many.output.md
```diff
@@ -18,174 +18,251 @@ Let's try it!
...
+  69.  Float.sin : Float -> Float
+  70.  Float.sinh : Float -> Float
+  71.  Float.sqrt : Float -> Float
+  72.  Float.tan : Float -> Float
+  73.  Float.tanh : Float -> Float
+  74.  Float.toText : Float -> Text
+  75.  Float.truncate : Float -> Int
+  76.  IO.clientSocket : Text
+                         -> Text
+                         ->{##IO} Either IOError ##Socket
```
pchiusano: This transcript calls `builtins.merge` at the start, which looks like it includes names for the builtin `IO` functions but not for the `IO` type itself? (since it is showing up as `##IO` here). That's a little odd.

We could use `builtins.mergeio` to give names for both the `IO` type and all these functions. 

Or just have `builtins.merge` include the `IO` type and functions.

@aryairani WDYT?

pchiusano (changes requested):
Hey @dolio awesome that this is finally getting in!! Am excited about all the cool stuff we are going to be able to start adding.

I think the one 'must have' before merging is switching those various enum types to `unique` instead of structural. I left a few other minor comments.

unison-src/transcripts/alias-many.output.md
aryairani: Side note, based solely on transcripts from a while back, it seemed like the IO builtins were being put a level too high.  And yeah we should check for an extra `#`.

parser-typechecker/src/Unison/Builtin/Decls.hs
dolio: Why, though? Is this because a structural definition can only have one set of relevant namings?

parser-typechecker/src/Unison/Builtin/Decls.hs
dolio: Yeah, this was a mistake. It should be more like you say.

unison-src/transcripts/alias-many.output.md
dolio: The old `IO` ability still exists (and is used in the existing runtime), so any ideas on what the names should be?

parser-typechecker/src/Unison/Builtin/Decls.hs
pchiusano: Well, a same structural type can have multiple names, but it's more that without `unique`, the type will be considered the same as like `type Four = One | Two | Three | Four` and they could pass a `Four` where a `FileMode` was expected. Since `FileMode` has a specific meaning that's probably not what we want. (vs say `type List a = Nil | Cons a (List a)` where the structure is really all there is to say about the type) 

Another example is like `unique type Day = Mon | Tue | ...` which is probably not something you want to be confused with "something with 7 nullary constructors" :)

Aside: we are likely going to make it a type error or a very stern warning to have a structural type where two of the constructors have the same type. /cc @aryairani

unison-src/transcripts/alias-many.output.md
aryairani: Ah, yeah if they both have to coexist, I guess we could call it `IO2` and rename it later.  The text would remain `IO2.whatever` even when we give it a new name in the namespace.

pchiusano: The builtin names can just be `IO.whatever`, but I'd just put them all in an `io2` namespace. This could be done with rename directives or in the spot where we `builtins.merge[io]`, wdyt @aryairani ?

The existing `IO` type and functions aren't builtins, so there's no name conflict there.

aryairani: Oh I see, they aren't builtins, right.  So then there's no rename directives (those only applies to Builtin types and terms), but yeah we can something something them in an `io2` namespace in `builtins.mergeio(2?)`).

pchiusano: @aryairani Cool :) Can you help @dolio with making this change?

I was imagining adding like `R "IO" "io2.IO"` to the little DSL in `Builtin.hs` but I think that is a bad idea and better to just change where it is merged into the namespace tree like you suggest.

parser-typechecker/src/Unison/Builtin/Decls.hs
dolio: Here's another question: other builtin unique decls have different length strings. Some look like the length of `sha1sum` and others are probably like `sha256sum` or larger. Does it not matter?

aryairani: It doesn't matter for now; currently they're just strings that are parsed as `wordyId`s. 😅

parser-typechecker/src/Unison/Builtin/Decls.hs
dolio: Okay, I made this basically `type SeqView a b = Empty | Elem a b`. This fixes something else that was bugging me too, because it gets used for both left and right viewing, so either side may be the sequence depending on which you do.

parser-typechecker/src/Unison/Builtin/Decls.hs
dolio: These are now done.

unison-src/transcripts/alias-many.output.md
dolio: Everything related to the new `IO` stuff is now under the `io2` namespace when you merge it, I believe. I made the declared types go there, too, for instance. I presume just giving them that name in the hand-built declaration works, since there's no renaming DSL like there is for the builtin types.

parser-typechecker/src/Unison/Builtin/Decls.hs
pchiusano: Ah, nice! It's always bugged me that there's a separate `ViewL` and `ViewR` type in Haskell for like Data.Sequence.

pchiusano (approved):
Okay, I think this is good to merge! @aryairani can you review how @dolio did the `io2` stuff. I think it should be fine, but I'd like your sign off. You can merge if it looks good and no other blocking issues.

pchiusano:
Also @dolio can you check the CI failures and see what's going on there? I feel like these were all green at one point previously.

dolio:
Yeah. I think I just forgot to redo the transcripts with the new names.

aryairani:
> @aryairani can you review how @dolio did the `io2` stuff.

Will do.

Issue #1070, opened 2019-12-15
`delete` suggestion


mitchellwrosen:
# Current behavior:

- If the name is unambiguous, delete it.
- If the name is ambiguous, complain that you need to be more specific.
  - But if the command is issued a second time, it's like sudo/-f, just do it

# Proposed behavior:

Just delete all terms/types with the given name, even if there are many.

# Why now?

Recent patches improve the story here: delete shows a diff (#1068) and prettified diff is more accurate/informative (#1069).

aryairani:
SGTM

PR #1639, opened 2020-07-27
Modify the behavior of "delete" to remove all terms/types associated with a name.

jsthomas:
## Overview

When a user runs the `delete` command for a name associated with multiple terms/types, `ucm` will display a warning and ask them to enter the command a second time if they want to delete everything associated
with the name.

In Issue #1070, @mitchellwrosen proposed changing the behavior of `delete` so that all terms/types associated with a name are deleted. This change implements the behavior requested in that issue.

## Implementation Notes / Test Coverage

Mostly this appears to be an exercise in making some slight adjustments to existing logic and bringing the existing transcript tests (which verify the current warning behavior) up to date.

## Other Notes

This is my first unison PR, so apologies in advance if I missed something.

It looks like this issue hasn't been discussed since December of 2019 and I'm not sure if the consensus about the right behavior for `delete` might have changed since then. For my part, working on this PR was mostly an opportunity to learn more about the code-base, tests, etc.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -610,12 +610,7 @@ loop = do
           let matchingTypes = toList (getHQ'Types hq)
           case (matchingTerms, matchingTypes) of
             ([], []) -> respond (NameNotFound hq)
-            -- delete one term
-            ([r], []) -> goMany (Set.singleton r) Set.empty
-            -- delete one type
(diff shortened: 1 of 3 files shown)


aryairani (approved):
Looks good to me, thanks @jsthomas!

pchiusano:
Nice! For your next PR can you also add yourself to CONTRIBUTORS.markdown? 😀

aryairani:
Whoops, good catch.

jsthomas:
Will do.

PR #1642, opened 2020-07-29
Fix #1640 caused by incorrect parsing of nested patterns that mention nullary constructors

### LINKED ISSUE #1640: Spurious pattern arity mismatch error when nullary patterns are involved
```haskell
unique type Color = Red | Black
unique type RBTree a = Leaf | Tree Color (RBTree a) a (RBTree a)

RBTree.isRed = cases
  RBTree.Tree Color.Red _ _ _ -> true
  _ -> false
```

Yields this error:

```
  Sorry, you hit an error we didn't make a nice message for yet.
  
  Here is a summary of the Note:
  
    simple cause:
      PatternArityMismatch:
    loc=line 5, columns 3-30
    typ=Color -> RBTree a -> a -> RBTree a -> RBTree a
    args=1
  
    path:
      InMatch firstBody=line 5, columns 34-38
      InSynthesize e=match l13qtrrgf3 wit...
      InCheck e=match l13qtrrgf3 wit..., t=𝕣7
      InSynthesize e=cases
    RBTree.Tree ...
      InSynthesize e=let
    RBTree.isRed =...
```

### Hacky workaround

If we switch `Red` and `Black` to be unary constructors then this typechecks fine:

```haskell
unique type Color = Red () | Black ()
unique type RBTree a = Leaf | Tree Color (RBTree a) a (RBTree a)

RBTree.isRed = cases
  RBTree.Tree (Color.Red _) _ _ _ -> true
  _ -> false
```

So I'd likely stare at the pattern typechecking code and think through what happens when the pattern is nullary.

pchiusano:
The issue was actually in the parser, not the typechecker. A pattern like `Foo Bar a b c` was actually being parsed as `Foo (Bar a b c)`. Whoops. The pattern parser just needed to be tweaked a bit to handle this correctly. I also fixed an unreported pattern parse bug and added a test for that too.

Fixes #1640 

/cc @chiroptical

### DIFF
--- a/parser-typechecker/src/Unison/TermParser.hs
+++ b/parser-typechecker/src/Unison/TermParser.hs
@@ -144,10 +144,10 @@ matchCase = do
   pure . Term.MatchCase p (fmap (absChain boundVars') guard) $ absChain boundVars' t
 
 parsePattern :: forall v. Var v => P v (Pattern Ann, [(Ann, v)])
-parsePattern =
-  chainl1 patternCandidates patternInfixApp
+parsePattern = root
   where
-  patternCandidates = constructor <|> seqLiteral <|> leaf
+  root = chainl1 patternCandidates patternInfixApp
+  patternCandidates = constructor <|> leaf
   patternInfixApp :: P v ((Pattern Ann, [(Ann, v)])
                   -> (Pattern Ann, [(Ann, v)])
                   -> (Pattern Ann, [(Ann, v)]))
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -225,15 +229,21 @@ parsePattern =
...
     tok <- ctor UnknownDataConstructor
     let (ref,cid) = L.payload tok
         f patterns vs =
           let loc = foldl (<>) (ann tok) $ map ann patterns
           in (Pattern.Constructor loc ref cid patterns, vs)
-    unzipPatterns f <$> many patternCandidates
+    unzipPatterns f <$> many leaf
 
-  seqLiteral = Parser.seq f leaf
+  seqLiteral = Parser.seq f root
```
pchiusano, line 246: There was no reason to restrict the patterns that form the elements of a list pattern to be leaf patterns. Like `[Foo a b]` is a perfectly fine pattern but `Foo a b` is not a `leaf` pattern.

dolio (approved):

PR #1644, opened 2020-07-30
Standardize default extensions

seagreen:
Standardize all components on the same `default-extensions`.

Use ones from `parser-typechecker:lib` because they're a superset of the `default-extensions` in `unison-core`.

(I don't think the particular extensions are too important, the goal was just to get each component of the repo using the same ones. Which ones those are can always be changed later.)

Use cabal common stanzas (https://cabal.readthedocs.io/en/latest/cabal-package.html#common-stanzas) so that the extensions only have to be listed once per cabal file. Add a note to each common stanza suggesting to keep them in sync.

Bump the `cabal-version` within the cabal files to `2.2` in order to get access to common stanzas.

Then remove the now redundant extensions from the haskell files.


yaks/easytest/easytest.cabal
```diff
@@ -1,8 +1,8 @@
+cabal-version: 2.2
```
aryairani, line 1: Should this be` >= 2.2` in each case?  Or what happens if you have a version that's too new?

aryairani (approved):

yaks/easytest/easytest.cabal
seagreen: So there's a story here!

To get common stanzas we have to specify at least 2.2. But changed this field in two ways:

1. It has to be at the top of the cabal file (easy enough)

2. It has to be an exact version, not a range.

You can find more documentation on it here: https://cabal.readthedocs.io/en/latest/cabal-package.html#pkg-field-cabal-version

My understanding is that later versions of cabal-the-library will still work.

All that the `cabal-version` field is identifying is which version of the spec this particular cabal file complies with, but later versions of cabal-the-library can still read older versions of cabal-the-spec-for-cabal-files.

I checked, and I think that's borne out by my current setup:
```sh
$ stack exec -- cabal --version
cabal-install version 3.0.0.0
compiled using version 3.0.0.0 of the Cabal library
```

aryairani: I see, great!

aryairani:
Now that https://fgaz.me/posts/2019-11-14-cabal-multiple-libraries/, maybe it can all be in one cabal file, with one `default-extensions:` section?

seagreen:
Perhaps. Support for multiple libraries in a package is janky in stack (eg https://github.com/commercialhaskell/stack/issues/5318). But it's worth considering. I'll bring it up on: https://github.com/unisonweb/unison/issues/779

PR #1647, opened 2020-08-10
untangled PatternP vs Pattern a little

aryairani:
Before this PR, `Pattern.hs` exported `PatternP`, whereas `PatternP.hs` exported `Pattern` 🤪  and some function forwarders.

This PR switches the names around, renames `PatternP.hs` to `PatternCompat.hs` (used only by the runtime).  `PatternCompat` fixes the annotation type to `()` and exports pattern aliases that assume `()`.

The PR also switches `Int` out for `ConstructorId` in relevant places in `Term.hs`.  `ConstructorId` was being reexported from `Pattern.hs` by `Term.hs`, since it's used by both, and at least one other module was importing it from `Term.hs`.  I guess it lives in `Pattern.hs` due to Haskell circular module dependency restrictions.

## Loose ends

`PatternCompat.hs` could be moved to the same package as `IR.hs`, or possibly merged into it, or possibly done away with completely by inlining `()`s into the match cases in `IR.hs`.


pchiusano (approved):
👍👍

aryairani:
I will create a PR which merges this into wip/rt2

PR #1651, opened 2020-08-12
Disable historic name search and allow lookup by hash

runarorama:
## Overview

This disables "historic names", which was the process by which UCM would search the history of the codebase for a name to display for a hash that no longer has a name.

It's a good idea, but the implementation was such that it could take UCM *several minutes* to display patches or render the source code of an updated definition. The problem was that many hashes don't have any names in the history, so the process would traverse all of it.

This PR changes the user experience in the following ways:

* For any hash that doesn't have a name in the namespace root *or the immediate historic ancestor of the root*, UCM just displays the hash. That is, it will look one entry backwards in history for a historic name, but no further.
* `view.patch` completes immediately, whereas before it would take several minutes.
* `view` for definitions that have been edited (e.g. during a PR review) completes immediately instead of taking several minutes. 

## Implementation notes

* This mostly disables the `findInHistory` function. It's still there, and it still searches history, but only a single step.

* We could bikeshed about how many steps back in history to look for names. I settled on *one step*.
* When e.g. using `view` with a hash, the search returns a hash qualified name with an empty name portion. This is instead of modifying the search result type to allow `ShortHash`es.

## Test coverage

No tests were added or removed.


aryairani:
I restarted the build that had failed in CI.

> We could bikeshed about how many steps back in history to look for names. I settled on one step.

Makes sense to me.  Because it's so easy to quickly accumulate intermediate steps, there's no number that's going to be right.  I'm not sure if _one_ buys us any more than _zero_ (do you know if it does?), but I'm pretty confident that anything >1 is not going to be much help.

>When e.g. using view with a hash, the search returns a hash qualified name with an empty name portion. This is instead of modifying the search result type to allow ShortHashes.

What do you think is the right fix long term?  I'm a little nervous about having these degenerate values floating around.

aryairani (approved):

runarorama:
The underlying issue is that some things use `HQ` and other things use `HQ'` and switching from one to the other is expensive in terms of refactoring. Long-term I'd probably just delete the one that has the `HashOnly` option and use a HQN with an empty name portion to represent that in the UI. We can always have a partial function from that to a `ShortHash` when going from UI to backend code.

aryairani:
What do you do in the cases that require a name (like the source and destination of a `move`)?

runarorama:
Do you mean `NameOnly`? Both types support that.

aryairani:
Sorry, I mean cases that require a name that isn't empty, like `move.term foo#xyz bar`; I'm thinking that every place that uses `Name` expects it to be non-empty, so if we start passing around `mempty`s, it's going to bite us.

PR #1659, opened 2020-08-22
Fix bug in lexing hash-qualified symboly id

mitchellwrosen:
## Overview

This patch fixes a bug where `+#foo` was not being lex'd as a hash-qualified "symboly" id.

### DIFF
--- a/parser-typechecker/tests/Unison/Test/Lexer.hs
+++ b/parser-typechecker/tests/Unison/Test/Lexer.hs
@@ -1,7 +1,10 @@
+{-# LANGUAGE OverloadedStrings #-}
+
 module Unison.Test.Lexer where
 
 import           EasyTest
 import           Unison.Lexer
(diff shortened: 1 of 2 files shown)


parser-typechecker/tests/Unison/Test/Lexer.hs
```diff
@@ -78,6 +81,10 @@ test =
         ]
       , t ".Foo.Bar.+" [simpleSymbolyId ".Foo.Bar.+"]
 
+      -- idents with hashes
+      , t "foo#bar" [WordyId "foo" (Just (ShortHash.unsafeFromText "#bar"))]
```
mitchellwrosen, line 85: The HQ wordy id test was added only because it wasn't covered yet, however it was passing before and after this patch. The symboly test was failing before the fix, and passing after.

aryairani (approved):
Great!

PR #1661, opened 2020-08-23
Disable idle GC, to avoid high CPU usage when ucm is idle (and has a large codebase loaded)

### LINKED ISSUE #1660: ucm uses CPU while idle - most time being spent in GHC RTS in scheduleDetectDeadlock
Symptom:  I start ucm, and I see one of my CPU cores go to 40-60% and stay there.

If I pass `+RTS -I0` when I start ucm, then it goes away, so I think something's probably triggering/aggravating the issue described [here](https://gitlab.haskell.org/ghc/ghc/-/issues/13164): the GHC runtime system apparently has a tradeoff where it does GCs even when idle, as a form of deadlock detection.

Why it should be happening a bunch more now I don't know, so I guess that's the next question - read on below for clues related to repeating background work.

The rest of this summary was written before I found the link above and the -I0 flag.

I'm going to down tools on this now!

-----

I recompiled with debug symbols and used 'perf' and gdb (I'm on linux).  ~70% of time is spent in the following stack (most recent function call at the bottom.)
```
__GI___clone (inlined)
start_thread
workerStart
scheduleWorker
schedule
scheduleDetectDeadlock (inlined)
scheduleDoGC
GarbageCollect
scavenge_until_all_done
scavenge_loop1
scavenge_find_work (inlined)
scavenge_block1
```
I use gdb and I find we are moving around in schedule(), part of the GHC runtime system, the main scheduler work loop. (Latest GHC code for that function at the time of writing is [here](https://gitlab.haskell.org/ghc/ghc/-/blob/3f50154591ada9064351ccec4adfe6df53ca2439/rts/Schedule.c#L193) - although I haven't dug in to work out exactly what code revision we're currently running with.)

I can see us executing scheduleDetectDeadlock and scheduleYield.  After much stepping around in gdb I still haven't proven the exact code flow, but I suspect we're going through the `continue` statement [here](https://gitlab.haskell.org/ghc/ghc/-/blob/3f50154591ada9064351ccec4adfe6df53ca2439/rts/Schedule.c#L317) and therefore repeatedly doing the scheduleDetectDeadlock and the GC inside it in a tight loop (same disclaimer that I haven't checked we're running the exact code version at that link - in fact I know from line number mismatches that we're not). 

Note this interesting comment in the RTS code (added 8 years ago) about GC livelock: [here](https://gitlab.haskell.org/ghc/ghc/-/blob/3f50154591ada9064351ccec4adfe6df53ca2439/rts/Schedule.c#L665). 

It's possible that the gdb-based part of this analysis could be being confounded by the gdb attach I'm doing, looking like an interrupt and causing different behavior after the scheduleDetectDeadlock - so would need to hack the RTS sources to add some debug trace to be sure what's going on.   

Not much luck from googling around, but I haven't tried thoroughly.  Recent history of the RTS scheduler code is [here](https://gitlab.haskell.org/ghc/ghc/-/commits/master/rts/Schedule.c), but again no obvious hot leads.  

[This](https://gitlab.haskell.org/ghc/ghc/-/issues/4322#note_44382) might be a clue: are we doing some repeated background work maybe on a repeating timer that is then causing a GC each time?  Maybe there's some GHC RTS logging that will tell us what 'idle GCs' are happening?  And some config to allow them to be less frequent?

Best next steps are probably 
- search recent unison history for stuff to do with repeating background work
- more research around idle GCs
- binary chop to see what unison commit triggered it (wild guesses would be any GHC upgrades or stuff to do with either GC config, multi-threading, or the file watcher or other I/O handing.)

Notes to self on the debugging steps (only good on linux):
 - building ucm with [dwarf debug symbols](https://gitlab.haskell.org/ghc/ghc/-/wikis/dwarf):  `stack build --ghc-options="-g" --no-strip`, and then find the resultant binary and run it directly, as `stack exec` seems funny/[broken](https://github.com/commercialhaskell/stack/issues/4863) about which binary it uses as soon as you start to customize the build; (use `nm` to check if a given binary has debug symbols)
- `perf record -F 99 -p PID --call-graph dwarf sleep 10` as per [this](http://www.brendangregg.com/perf.html), then `perf report` or `perf report --stdio -n -g folded` to decode
- gdb attach directly to the LWP of the offending thread, not just the parent process (easy to see which is which in htop, and the child thread will have a higher PID)

unison commit trunk@519cbe

atacratic:
## Overview

Fixes #1660 - using 40-60% of one core when ucm is idle, due to a garbage collection running every second.

The GHC runtime system's 'idle GC' mechanism (described [here](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/runtime_control.html#rts-flag--I%20%E2%9F%A8seconds%E2%9F%A9)) tries to get ahead with garbage collection while the program is idle, to reduce GC work at peak time.  It defaults to doing this by running a GC after 0.3 seconds of idleness.

But since recently, ucm has done work on a repeating 1 second timer (configurator checking for config file updates), so this ends up meaning a GC every second.  That can burn a lot of CPU depending on the amount of state in memory.  See [this note](https://gitlab.haskell.org/ghc/ghc/-/issues/4322#note_44361) from the GHC team from 9 years ago, which describes the problem.

There's no point bumping the idle timer value up from 0.3 to say 2 seconds, because then idle GC will never happen.  

So this PR switches off idle GC, so ucm uses more like zero CPU when not being actively used.  It may make it slightly less responsive, as more GC work will be done during the first few fractions of a second of each user interaction.  *I have not tried to test or measure this.*

[This note](https://gitlab.haskell.org/ghc/ghc/-/issues/3408#note_35438) suggests there is a drawback to do with this approach making the RTS unable to catch deadlocks (or run finalizers?) - I don't know if that means common-or-garden thread deadlocks in ucm.  I suggest we don't worry.  

## Implementation notes

Pass `-with-rtsopts=-I0` in the ghc-options for the ucm build.

An alternative would be to use `-with-rtsopts=-Iw60` to keep doing idle GCs, but at most one a minute.  But I still don't much like an idle program consuming CPU; it could still generate future confusion/questions/problems; and it massively dilutes any responsiveness benefit anyway.  

## Test coverage

None.  And I haven't even done manual testing of command responsiveness.  Would need finding some command that uses, say, half a second ish of CPU, and seeing how much slower it goes when run repeatedly at intervals of say a couple of seconds.  But the main issue with ucm responsiveness is clamping down on commands that use many seconds of CPU, not saving a small constant increment on commands that use significant but sub-second CPU.  So I propose not to bother.  

## Loose ends

Didn't actually pin it on configurator being the trigger for definite, but it does seem likely and anyway this PR does fix the problem.

### DIFF
--- a/parser-typechecker/unison-parser-typechecker.cabal
+++ b/parser-typechecker/unison-parser-typechecker.cabal
@@ -238,7 +238,7 @@ executable unison
   import: unison-common
   main-is: Main.hs
   hs-source-dirs: unison
-  ghc-options: -Wall -threaded -fno-warn-name-shadowing -fno-warn-missing-pattern-synonym-signatures -rtsopts -optP-Wno-nonportable-include-path
+  ghc-options: -Wall -threaded -fno-warn-name-shadowing -fno-warn-missing-pattern-synonym-signatures -rtsopts -with-rtsopts=-I0 -optP-Wno-nonportable-include-path
   other-modules:
     System.Path
     Version


aryairani (approved):

aryairani:
per https://github.com/unisonweb/unison/issues/1660#issuecomment-678836576, idle-gc deadlock detection kicked on when we enabled `-threaded` in #1619.

Issue #1663, opened 2020-08-25
unknown data reference


aryairani:
In 67bf4aed, evaluating this expression

```
> handle 1 with cases { a } -> a
```

gives
```
unison: unknown data reference: PR {matches = [EffectPure Var], guard = Nothing, body = Var User "a"}
  error, called at src/Unison/Runtime/Pattern.hs:550:18
```

/cc @dolio

dolio:
I see. This actually makes me realize that there is a second way some assumptions I made can go wrong. I think if you do:

    cases
      { a } -> ...
      { A1.e -> k } -> ...
      { A2.e -> k } -> ...

you will only get `A1` matches, because I look for a single reference for the type of a pattern. However, ability patterns match against 0 or more referenced types.

I'll work out a fix for this, and it'll probably cause a compilation issue further down the line due to only having a pure case, but I think that should be easy to check for and just turn into `let a = 1 in a` or something.

aryairani:
works now

```
    1 | > handle 1 with cases { a } -> a
          ⧩
          1
```

PR #1666, opened 2020-08-27
Improve ability pattern compilation for new runtime

dolio:
Previously the ability pattern compilation was treating it almost identically to data types, where you match on exactly one ability signature in a pattern column. However, you can actually match on 0 or more abilities simultaneously. This fixes that issue in the pattern compiler, and tweaks subsequent phases to handle things properly. Should fix issue #1663.


pchiusano:
@dolio can you just add a regression test for this? If we're not already doing this, we should be running the tests with both the old and the new runtime. 

Other than that looks good.

dolio:
I beefed up the ability test case to check for stuff related to this fix, and it actually works on the new runtime now.

pchiusano:
CI isn't currently running the tests with new runtime but #1673 is tracking this. @dolio has manually run the test with new runtime locally and it worked fine.

pchiusano (approved):

PR #1667, opened 2020-08-29
(WIP) Lexing identifiers

mitchellwrosen:

## Overview

## Implementation notes

## Test coverage

## Loose ends


mitchellwrosen:
Looking over the commits here, I think the idea was (more or less) worth pursuing: there are _so_ many variants of an "identifier" (non-empty list of segments, possibly empty list of segments, possibly hash-qualified thing, no hash allowed, etc) it may be preferable to have a generic opaleye-style record (i.e. one type var per field), which is generic enough to model them all.

Here's where this PR ended up, which would no doubt need further tweaking as not much of the codebase was tackled (just kinda starting from the lexer, and bubbling up to the path parsing stuff).

```haskell
data Ident f g = Ident
  { segments :: f NameSegment,
    hash :: g ShortHash
  }
  deriving stock (Generic)
```

But I've long stopped working on this; there's really just too much breakage, and it was just an exploration anyway, initially broken off of some work to implement a syntax that allows spaces in an identifier.

I think when/if the time comes to implement a truly pluggable syntax, some of this could be worth revisiting!

PR #1671, opened 2020-09-03
Add `MVar` and some associated operations, with a document on adding builtins

dolio:
This pull request adds `MVar` as a builtin type, and various operations for using `MVar` in the new builtin `IO` ability. It also adds a document specifying how this is done, so that it can hopefully be accomplished by other people.

The new runtime 'foreign function' machinery that is used for I/O stuff was reworked a fair bit as part of this, because the existing implementation was insufficient for working directly with uninspected closures like `MVar` should. This of course isn't in the document, since it was just a design change motivated by the example.

Not all GHC `MVar` functions have been wrapped, but the relatively simple ones have, and it should allow for some actual non-trivial threading to be used.

### DIFF
--- a/docs/adding-builtins.markdown
+++ b/docs/adding-builtins.markdown
@@ -14,8 +14,10 @@ referred to by `Reference`. Builtin, opaque data types use the
 use are listed in the `Unison.Type` module, so we'll add a definition
 there:
 
-> mvarRef :: Reference
-> mvarRef = Reference.Builtin "MVar"
+```haskell
+mvarRef :: Reference
+mvarRef = Reference.Builtin "MVar"
+```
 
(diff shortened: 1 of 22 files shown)


pchiusano (approved):
LGTM. The guide looks great. Minor formatting thing - I'd do fenced code blocks for the snippets, otherwise it renders all on one line - see https://github.com/unisonweb/unison/blob/2c55c2b6796dd206b0f8076b9505864c2c98b398/docs/adding-builtins.markdown

I think the best way to test out the docs is to have someone who isn't @dolio add a new builtin.

dolio:
Added the code blocks. Look good now?

pchiusano:
Great! Merging.

PR #1677, opened 2020-09-14
Add delay and kill threading operations for new runtime

dolio:
The instructions for these were implemented previously, but the wrapping to make them available was missing.


pchiusano:
The CI failure looks totally unrelated... something with name.segment.

pchiusano (approved):

PR #1685, opened 2020-09-22
New runtime builtin fixes/tweaks

dolio:
This is a pull request to rework some of the new builtin IO related stuff.

- Fork was broken in multiple ways, but it was difficult to test when it was written. Now it should work.
- An enumeration type `StdHandle` has been added to serve as the argument of `stdHandle`, so that it can't be called with incorrect arguments, and doesn't need to return an optional result.

Those are the user facing changes. Some additional implementation details are as follows:
- One of the ways that fork was broken was that it was trying to manipulate the stack in a complicated way, with the assumption that an IO handler was at the bottom, and that it might be running arbitrary code that accessed the current 'frame'. This has been completely eliminated, and now it just creates new stacks to run a closure in, which is much simpler.
- I've also removed a bunch of mask/unmask stuff from the runtime. At the time it was written, I thought maybe it was appropriate to only unmask at points where exceptions could be caught and explicitly returned, but I think we actually only mediate synchronous exceptions into unison anyway. So it was just adding complication and dubious behavior.

It might be good not to merge this immediately, since maybe Stew will have some additional feedback or find additional bugs now that fork can actually be used. Or that could just be a separate merge. Either way is fine with me.

### DIFF
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -444,7 +444,7 @@ ioBuiltins =
   , ("IO.socketSend", socket --> bytes --> ioe unit)
   , ("IO.socketReceive", socket --> nat --> ioe bytes)
   , ("IO.forkComp"
-    , forall1 "a" $ \a -> (unit --> ioe a) --> ioe threadId)
+    , forall1 "a" $ \a -> (unit --> ioe a) --> io threadId)
   , ("IO.stdHandle", nat --> optionalt handle)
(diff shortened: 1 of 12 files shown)


pchiusano (approved):

pchiusano:
Travis error looks spurious, it failed before even fetching Stack.

dolio:
Added some additional fixes I noticed. `openFile` and `seek` had incorrect types, and the latter didn't even have the necessary builtin type declared.

PR #1690, opened 2020-09-22
Handle character pattern matching in ANF translation

dolio:
Character patterns were handled in the pattern compiler step, but overlooked during ANF translation.


pchiusano (approved):

pchiusano:
Again spurious failure fetching stack... tried bouncing.

dolio:
FYI, there are restart buttons on travis. I guess that might be harder to get to than close/open, though.

aryairani:
We could also try caching the stack installation, if that's a frequent failure point; though I don't know why the `travis_retry` that's already there wouldn't be enough.  I don't see it retrying ["up to three times"](https://docs.travis-ci.com/user/common-build-problems/#travis_retry); I see it trying once, failing, and giving up.

dolio:
Yeah, I agree. It doesn't seem to be retrying. I restarted a failed job yesterday, and it failed again, and it seems unlikely that an individual build would re-fail 3 times (as opposed to one of the 5 builds we do, say). Unless it's some intermittent problem causing the download to fail for some relatively short time period, in which case adding to the script wouldn't help either.

PR #1691, opened 2020-09-23
(Stalled) Allow deleting multiple terms at once

### LINKED ISSUE #1526: delete.term doesn't let you delete multiple terms in one go
Whereas you can do `view a b c` and `edit a b c`, you can't do do `delete.term a b c` or `delete.type A B C` (or `delete.term 1-3` for that matter).  Would sometimes be convenient if it did accept this.

Maybe more generally a consistency pass over the commands for this would be good - see also #1396, #1397. 

```
.> delete.term a b c

  ⚠️
  
  `delete.term` takes an argument, like `delete.term name`.

```

sullyj3:
## Overview

Currently, it's not possible to delete multiple things in a single command:
```
  I found and typechecked these definitions in scratch.u. If you do an `add` or `update`,
  here's how your codebase would change:

      x : Nat
      y : Nat
      z : Nat -> Nat

.> add

    x : Nat
    y : Nat
    z : Nat -> Nat

.> delete x y z

  ⚠️

  `delete` takes an argument, like `delete name`.
```

This will allow the delete command, as well as delete.term and delete.type, to take multiple arguments.

Closes: #1526 

Calls `delete` from `Editor.HandleInput` once for each listed name, producing an `Output` for each. @aryairani mentioned in slack: 

> I might write a new version of delete that takes multiple HQSplit' in the 3rd argument, but produces only a single Output to the user [..]

I don't understand the implementation well enough to do so myself.

It also seems like this approach results in n entries to the reflog. This seems bad, right? I'm assuming we'd want just one?

```
.> reflog

  Here is a log of the root namespace hashes, starting with the most recent, along with
  the command that got us there. Try:

    `fork 2 .old`
    `fork #dqc1cae8re .old`   to make an old namespace accessible again,

    `reset-root #dqc1cae8re`  to reset the root namespace and its history to that of the
                              specified namespace.

  1. #cqe5netvov : delete ..x ..y ..z
  2. #dqc1cae8re : delete ..x ..y ..z
  3. #p3rtchrcmj : delete ..x ..y ..z
  4. #8jfd8nkas1 : add
  5. #lqs21dco5d : pull https://github.com/unisonweb/base:.releases._latest .base
  6. #7asfbtqmoj : (initial reflogged namespace)

.>
```

## Loose ends

Yes
### todo

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -410,9 +410,9 @@ loop = do
           MoveBranchI src dest -> "move.namespace " <> ops' src <> " " <> ps' dest
           MovePatchI src dest -> "move.patch " <> ps' src <> " " <> ps' dest
           CopyPatchI src dest -> "copy.patch " <> ps' src <> " " <> ps' dest
-          DeleteI thing -> "delete " <> hqs' thing
-          DeleteTermI def -> "delete.term " <> hqs' def
-          DeleteTypeI def -> "delete.type " <> hqs' def
+          DeleteI things -> "delete " <> intercalateMap " " hqs' things
+          DeleteTermI defs -> "delete.term " <> intercalateMap " " hqs' defs
+          DeleteTypeI defs -> "delete.type " <> intercalateMap " " hqs' defs
           DeleteBranchI opath -> "delete.namespace " <> ops' opath
           DeletePatchI path -> "delete.patch " <> ps' path
           ReplaceTermI src target p ->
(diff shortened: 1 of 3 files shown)


sullyj3:
WIP: currently the output is super weird, need to figure it out:

```
  I found and typechecked these definitions in scratch.u. If you do an `add` or `update`,
  here's how your codebase would change:

      x : Nat
      y : Nat
      z : Nat -> Nat

.> add

    x : Nat
    y : Nat
    z : Nat -> Nat

.> delete x y z

  Removed definitions:

    1. x : Nat

  Removed definitions:

    1. x : Nat
    2. y : Nat

  Removed definitions:

    1. x : Nat
    2. y : Nat
    3. z : Nat -> Nat

.>
```

sullyj3:
Presumably has something to do with https://github.com/unisonweb/unison/blob/e77cce9e7ba56fdb52f7af161ea7ffb490410e9f/parser-typechecker/src/Unison/Codebase/Editor/Output/BranchDiff.hs#L46-L63

Since we're calling delete 3 times, each time the diff from since the last command (???) will have an additional removed thing. I'm fuzzy on the details.

Arya mentioned ideally we want a single update and a single output, but I think that's actually the only way to get it to work with the diff based system. I'll leave this to someone who understands and can modify `delete`
https://github.com/unisonweb/unison/blob/df653686004a6fddb89a559ce501371c75d9a099/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs#L621-L625

sullyj3:
Arya mentioned ideally we want a single update and a single output, but I think that's actually the only way to get it to work with the diff based system.

PR #1699, opened 2020-09-28
Rework 'foreign' builtins and update documentation

dolio:
This makes a few changes to the foreign function based builtins.

- Changes the helper type name from `IOOP` to `ForeignOp`, since they're not necessarily I/O related
  * Some other `IO` names have been changed similarly
- Gets rid of `IOp` instruction type, to avoid an explosion of arbitrary abbreviated names when people add more builtins
  * instead word identifiers are made up and referenced in the wrapper code
- Creates a new method of declaring builtins that associates the builtin name, the wrapper code and the implementation, and generates necessary mappings from the declaration block

The builtin adding documentation has been updated with these changes. I've also added some information on decompilation.

### DIFF
--- a/docs/adding-builtins.markdown
+++ b/docs/adding-builtins.markdown
@@ -100,86 +100,86 @@ rename.
 
 What we have done so far only declares the functions and their types.
 There is nothing yet implementing them. This section will proceed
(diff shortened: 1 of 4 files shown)


pchiusano (approved):
LGTM!

Issue #1578, opened 2020-05-21
Suffix-based name resolution uses definitions from the codebase instead of definitions from the current file


pchiusano:
I don't have a good examples for this one, but if you have `Bar.toList` in your codebase and `toList` is a unique suffix, then define `Foo.toList` in your scratch file, `toList` will resolve to `Bar.toList` when I'd expect it to be ambiguous or perhaps prefer the local `Foo.toList`.

I've been bitten by this a few times and always find it very confusing.

aryairani:
The file / not-yet-added distinction is pretty artificial IMO, I want a way to eliminate it.

chiroptical:
I recently came across a motivating example. Trying to implement red-black tree a-la Okasaki leads to a situation where you can't match on a tree's color, a trivial example:

```
unique type Color = Red | Black
unique type RBTree a = Leaf | Tree Color (RBTree a) a (RBTree a)

use Color Red Black
use RBTree Leaf Tree

RBTree.isRed : RBTree a -> Boolean
RBTree.isRed = cases
  Tree Red _ _ _ -> true
  _ -> false
```
leads to:
```
  Sorry, you hit an error we didn't make a nice message for yet.
  
  Here is a summary of the Note:
  
    simple cause:
      PatternArityMismatch:
    loc=line 9, columns 3-17
    typ=Color -> RBTree a -> a -> RBTree a -> RBTree a
    args=1
  
    path:
      InMatch firstBody=line 9, columns 21-25
      InSynthesize e=match cdlsph512n wit...
      InCheck e=match cdlsph512n wit..., t=base.Boolean
      InCheck e=cases
    RBTree.Tree ..., t=RBTree a7 ->{𝕖4} base.Boolean
      InSynthesize e=(cases
    RBTree.Tree...
      InSynthesize e=let
    RBTree.isRed :...
```
If you remove the pattern match on the color, the code compiles. In this case, it is pretty easy to workaround. However, if you want to implement `balance` it becomes quite tedious. I am on Unison master:

```sh
git rev-parse --short HEAD
8b57038d
```

Let me know if I can provide any more details.

pchiusano:
I've filed a separate issue (#1640) for that spurious pattern arity error (marked off topic above), since I just tried the same thing on an empty codebase and it's still happening. It looks like the cause is just some typechecking bug in pattern matching and it's definitely unrelated to suffix-based resolution.

Issue #913, opened 2019-10-30
Coverage checking for pattern matching


pchiusano:
(@augustss if you are seeing this - we are adding pattern coverage checking to Unison and could use a good reference on how to implement for a language without GADTs)

Right now, you can have incomplete pattern matches and it will compile without even a warning. This should be a compile error that reports which patterns aren't covered by the cases. It should also report redundant patterns, like if you have `4 -> "hi"` after a `x -> "bye"` case.

(If you have a set of patterns which you think are complete, you can explicitly crash using a wildcard pattern that calls the new `error` / `bug` function in #881, which we are still doing some bikeshedding on)

I imagine adding a function like this to the typechecker, and then calling it after typechecking a case expression somewhere [here in Context.hs](https://github.com/unisonweb/unison/blob/master/parser-typechecker/src/Unison/Typechecker/Context.hs#L850-L860):

```Haskell
-- for a given type and a list of patterns at some location, verify that the patterns cover that type, are not redundant, and fail otherwise
verifyCoverage :: loc -> [Pattern loc] -> Type v a -> M v loc ()
```

And then add a new kind of type error in `Context.hs` for if pattern matching is incomplete (maybe a separate error if it's redundant).

Guards should ultimately be incorporated into this check as well, though the logic for guards can be really straightforward / conservative. 

Here are some pointers: 

* https://stackoverflow.com/questions/7883023/algorithm-for-type-checking-ml-like-pattern-matching gives a basic overview.
* Page 17 of [this paper](https://arxiv.org/pdf/1601.05106.pdf) also has an algorithm, but that is for a type system with GADTs so I'm guessing that will be more complicated than necessary.

@TomasMikula given that you're getting pretty familiar with the typechecker I thought you might like to have a go at this one. :)

Also, @mrdziuban I know we have talked about this one and if you're still interested in helping out with this let us know!

TomasMikula:
Thanks for the mention @pchiusano! This is definitely something I would like to work on. I have a few other tweaks in the making, so I don't want to call dibs on this one just yet. :)

noahhaasis:
Unfortunately, I currently don't have the time to write a coverage checker but I've found some hopefully helpful resources.
This is a paper about implementing coverage and redundancy checking for pattern matching in ML-like languages:
http://moscova.inria.fr/~maranget/papers/warn/warn.pdf
It's also what Elm uses.
https://github.com/elm/compiler/blob/master/compiler/src/Nitpick/PatternMatches.hs

[...]

PR #1700, opened 2020-09-28
New syntax: destructuring bind

pchiusano:
## Overview

This introduces new syntax for destructuring binds, for instance:

```Haskell
(x, y, z) = ("uno", "dos", 33)
z + 1
```

This desugars to `match ("uno", "dos",33) with (x,y,z) -> z + 1`. See [the transcript](https://github.com/unisonweb/unison/blob/23d06faa1f2facd37fbe82e40abbebe5e6fd2b47/unison-src/transcripts/destructuring-binds.output.md) for more verbiage.

## Implementation notes

This is just syntax sugar, so there's a change to the parser and a corresponding change to the pretty-printer (similar to what was done for other sugar, like `cases` syntax (#1192).

@atacratic I'd like your review on the pretty-printer changes, in particular how this interacts with calculate imports. 

I killed `namespace` blocks, which had an unused, hacky implementation that conflicted with this change. This feature wasn't documented anywhere and we were encouraging people not to use this feature anyway if they discovered it by accident. I have some ideas for how to resurrect it properly.

`IOSource` was using the `namespace` block syntax so needed to be rewritten to not use namespace blocks. Due to I think #1578 this was more annoying than necessary. Also there were some test files still using namespace blocks which I fixed up.

The heuristics for when to have the pretty-printer show the destructuring bind syntax are somewhat arbitrary, but seem reasonable.

One interesting idea I wasn't sold on and decided was out of scope was to have pattern match failures turn into some default expression, so this could be used for things like list comprehensions. For instance:

```Haskell
ex1 elems = 
  default = Each.fail 
  handle
    Some (a, b) | a < b ?= each elems
    a
  with Each.toList.handler
```

Where `Each` is `ability Each where each : [a] -> a` (standard nondeterminism ability).

Anyway, this seemed out of scope and I wasn't very confident in the design or usefulness.

## Test coverage

Updated pretty-printer tests to cover this change and also added a transcript.


aryairani:
Looking at
```
-- example:
--   (x, y)   = foo
--   hd +: tl | hd < 10 = [1,2,3]
--   stuff
--
-- desugars to:
--
--   match foo with
--     (x,y) -> match [1,2,3] with
--       hd +: tl | hd < 10 -> stuff
```
I guess the goal is to have the guards become an error under pattern exhaustiveness checking, rather than a path to multi-way assignments, right? :)
```unison
(x, y)   = foo
hd +: tl | hd < 10 = [1,2,3]
hd +: tl | hd < 5 = [4,5,6]
stuff
```

Wrt default expressions, wdyt about making a pattern match failure its own ability request —
```unison
unique ability Match a where
  failure : a

default d = \cases
  { failure -> k } -> k d
  { a } -> a
```

or just make it an error with exhaustiveness checks.  Or make non-exhaustive pattern cases require the ability and exhaustive ones not, idk.

parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -760,50 +791,10 @@ importp = do
       suffix <- L.payload <$> suffixes
       pure (suffix, Name.joinDot (L.payload prefix) suffix)
 
---module Monoid where
---  -- we replace all the binding names with Monoid.op, and
---  -- if `op` is free in the body of any binding, we replace it with `Monoid.op`
---  op : Monoid a -> (a -> a -> a)
---  op m = case m of Monoid
```
pchiusano, line 767: cleanup... irrelevant / unused comment

parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -760,50 +791,10 @@ importp = do
...
-    vs :: [Maybe v]
-    vs = snd . fst <$> bs
-    prefix :: v -> v
-    prefix v = Var.named (Text.pack name `mappend` "." `mappend` Var.name v)
-    vs' :: [Maybe v]
-    vs' = fmap prefix <$> vs
-    substs = [ (v, Term.var () v') | (Just v, Just v') <- vs `zip` vs' ]
-    sub = ABT.substsInheritAnnotation substs
-    in [ ((a, v'), sub e) | (((a,_),e), v') <- bs `zip` vs' ]
-  in finishBindings (expand =<< b)
```
pchiusano, line 806: got to delete some code that was only relevant to hacky namespace implementation

pchiusano (commented):
Okay, this should be good to review. Left a few self-review comments.

Going through @aryairani remarks - 

> I guess the goal is to have the guards become an error under pattern exhaustiveness checking, rather than a path to multi-way assignments, right? :)

Your example is kind of ambiguous but I'm guessing you mean multi-equation bindings like what Haskell supports? (like `factorial 0 = 1; factorial n = n * factorial (n - 1)`)

It might be nice to multi-equation bindings. Grammatically, that looks a little different than this, it looks like 

```
<varname> <pattern>* = <block>
<pattern> <op-varname> <pattern> = <block>
```

Whereas a destructuring bind is: 

```
<pattern> = <block>
```

The example you gave would be ambiguous if we supported multi-equation bindings - is it introducing a new term, called `+:`, or is it a destructuring bind?

```Haskell
(x, y)   = foo
hd +: tl | hd < 10 = [1,2,3]
hd +: tl | hd < 5 = [4,5,6]
stuff
```

Haskell works around this ambiguity by disallowing overlap between constructor names and regular terms. I don't love this solution.

We could change the symbol for a destructuring bind to `<-` maybe. This is... ok. I don't love it.

Another option would be to NOT support multi-equation bindings (because who wants to type the same name over and over?) and instead have multi-pattern `cases`, for instance:

```Haskell
merge : [a] -> [a] -> [a]
merge = cases
  (a +: as) (b +: bs) -> 
    if a < b then a +: merge as (b +: bs)
    else b +: merge (a +: as) bs
  as [] -> as
  [] bs -> bs
```

> Wrt default expressions, wdyt about making a pattern match failure its own ability request

Yeah, you could just have pattern match failure require `Abort` (I don't think you want `Abort a` since the match can fail at multiple types depending on the subpattern). I'll move this discussion to #913.

unison-src/transcripts/destructuring-binds.md
```diff
@@ -0,0 +1,72 @@
...
+
+```unison
+ex1 : (a,b,(Nat,Nat)) -> Nat
+ex1 tup =
+  (a, b, (c,d)) = tup
+  c + d
+```
+
+```ucm
+.> add
```
atacratic, line 22: I think the new pretty-printing is somehow broken at the moment.  When I pull this branch and add a `.> view` here, I see

```
.> view ex1

  ex1 : (a, b, (Nat, Nat)) -> Nat
  ex1 = cases
    (a, b, (c, d)) ->
      use Nat +
      c + d
```
(how come not picked up by the TermPrinter tests?)

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -999,6 +1020,31 @@ immediateChildBlockTerms = \case
...
+-- BUT
+--   match (y,x) with (x,y) -> body
+-- Has shadowing, is rendered as a regular `match`.
+--   match blah with 42 -> body
+-- Pattern is a literal, rendered as a regular match (rather than `42 = blah; body`)
+isDestructuringBind :: Ord v => ABT.Term f v a -> [MatchCase loc (ABT.Term f v a)] -> Bool
+isDestructuringBind scrutinee [MatchCase pat _ (ABT.AbsN' vs _)]
+  = all (`Set.notMember` ABT.freeVars scrutinee) vs && not (isLiteralPattern pat)
+    where
+    isLiteralPattern p = case p of
```
atacratic: Do you care about the case where a nested sub-pattern is literal?

Looks like currently it will render the following:

```
  (x, 3) = blah
  body
```

I think ideally it would fall back to match/cases for this.

atacratic (approved):
I committed 37bcd95 to your branch to tweak the interaction with calc imports.

But it looks like the new pretty-printing is not kicking in?  See my other comment, and my addition to the transcript.

unison-src/transcripts/destructuring-binds.md
pchiusano: It looks like if the expression could be rendered as `cases` or destructuring bind, it chooses `cases`. I'm not sure what I think about that. Maybe it's fine?

parser-typechecker/src/Unison/TermPrinter.hs
pchiusano: I think I'm sold, though maybe doesn't matter so much - that would be a partial pattern match anyway and you'd likely have a second catch-all case so the syntax wouldn't apply.

pchiusano:
@atacratic thanks for your review and for the imports-related fix :) I'll take a crack at making pretty-printer avoid using the syntax when the LHS pattern contains literals.

unison-src/transcripts/destructuring-binds.md
atacratic: Yeah, the `cases` version looks OK to me, maybe better since it doesn't repeat `tup` as in
```
ex1 : (a, b, (Nat, Nat)) -> Nat
ex1 tup =
  (a, b, (c,d)) = tup
  c + d
```
Also it would be tricky to change because the `cases` stuff is done early in `specialCases` which is already a bit hairy.

atacratic: Probably worth making the transcript use a different example though.

pchiusano: Yeah... I'm going to leave it in there to mention the behavior, but am adding another example that doesn't overlap with `cases`.

PR #1705, opened 2020-10-01
make conversion from Text to Bytes explicit 

### LINKED ISSUE #1697: IO.putBytes, Text.toUtf8, Text.fromUtf8 builtins
Let's add the following: 

```Haskell
io2.putBytes : Handle -> Bytes ->{IO} ()
Text.toUtf8 : Text -> Bytes
-- returns some message on the left if malformed
Text.tryFromUtf8 : Bytes ->Either Text Text

-- not a builtin, but could be written in pure Unison
io2.putUtf8 : Handle -> Text ->{IO} ()
io2.putUtf8 h t = putBytes h (toUtf8 t)
```

### Question: should we just remove `putText` entirely?

`putText` is not really a thing for file handles... what is the encoding? UTF8? UTF16? Something system specific?? "Something system specific" seems like a bad answer: there's no need for writing to a file handle to invoke magic system environment state to do its job.

If default encodings are a useful thing, we could introduce other builtins like: `defaultEncoding : {IO} (Text -> Bytes)` or some such. Personally, I think that default encodings are bugs waiting to happen... you write some data (in random, system dependent encoding) and then elsewhere read it (in coincidentally, the same system dependent encoding), and then 6 months later read it someplace else (where default encoding differs), leading to crazy errors.

For standard out/in, does `putText` make sense? It seems like the console will support some particular encoding, and that's what you want to be targeting. Ideally you could interrogate the default encoding of the console and use that.

stew:
This gets rid of two builtins:

IO.getText : Handle -> {IO} Either IOError Text
IO.putText : Handle -> Text -> {IO} Either IOError ()

and add's 4 new builtins:

IO.getBytes : Handle -> Nat -> {IO} Either IOError Bytes
IO.putBytes : Handle -> Bytes -> {IO} Either IOError ()
Text.toUtf8 : Text -> Bytes
Text.fromUtf8 : Bytes -> Either Text Text

This is so that we are explicit about how we convert from Text ->
Bytes, instead of relying on the system locale to decide on a Char

Fixes #1697

### DIFF
--- a/parser-typechecker/unison-parser-typechecker.cabal
+++ b/parser-typechecker/unison-parser-typechecker.cabal
@@ -190,6 +190,7 @@ library
     configurator,
     cryptonite,
     directory,
+    either,
     guid,
     data-memocombinators,
     edit-distance,
--- a/unison-src/new-runtime-transcripts/hashing.md
+++ b/unison-src/new-runtime-transcripts/hashing.md
@@ -26,7 +26,6 @@ List.map f as =
     (h +: t) -> go (acc :+ f h) t
   go [] as
 
--- not very efficient, but okay for testing
 hex : Bytes -> Text
 hex b =
   Bytes.toBase16 b
@@ -34,36 +33,13 @@ hex b =
     |> List.map Char.fromNat
     |> Text.fromCharList
 
-ascii : Text -> Bytes
-ascii t = Text.toCharList t |> List.map Char.toNat |> Bytes.fromList
-
-fromHex : Text -> Bytes
-fromHex txt =
-  match Text.toCharList txt
-          |> List.map Char.toNat
-          |> Bytes.fromList
-          |> Bytes.fromBase16
-  with
-    Left e -> bug e
-    Right bs -> bs
-
 check : Boolean -> [Result]
 check b = if b then [Result.Ok "Passed."]
           else [Result.Fail "Failed."]
-
-test> hex.tests.ex1 = check let
-         s = "3984af9b"
-         hex (fromHex s) == s
 ```
 
 ```ucm:hide
 .scratch> add
-```
-
-The test shows that `hex (fromHex str) == str` as expected.
-
-```ucm
-.scratch> test
 ```
 
 ## API overview
@@ -71,18 +47,18 @@ The test shows that `hex (fromHex str) == str` as expected.
 Here's a few usage examples:
 
 ```unison
-ex1 = fromHex "2947db"
+ex1 = toUtf8 "2947db"
         |> crypto.hashBytes Sha3_512
         |> hex
 
-ex2 = fromHex "02f3ab"
+ex2 = toUtf8 "02f3ab"
         |> crypto.hashBytes Blake2b_256
         |> hex
 
 mysecret : Bytes
-mysecret = fromHex "237be2"
+mysecret = toUtf8 "237be2"
 
-ex3 = fromHex "50d3ab"
+ex3 = toUtf8 "50d3ab"
         |> crypto.hmacBytes Sha2_256 mysecret
         |> hex
 
@@ -100,7 +76,7 @@ And here's the full API:
 Note that the universal versions of `hash` and `hmac` are currently unimplemented and will bomb at runtime:
 
 ```
-> crypto.hash Sha3_256 (fromHex "3849238492")
+> crypto.hash Sha3_256 (toUtf8 "3849238492")
 ```
 
 ## Hashing tests
@@ -109,8 +85,8 @@ Here are some test vectors (taken from [here](https://www.di-mgt.com.au/sha_test
 
 ```unison:hide
 ex alg input expected = check let
-  hashBytes alg (ascii input) ==
-  fromHex expected
+  hex (hashBytes alg (toUtf8 input)) ==
+  expected
 
 test> sha3_512.tests.ex1 =
   ex Sha3_512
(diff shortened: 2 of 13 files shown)


aryairani:
@pchiusano @dolio Since this is our first time deleting a built-in (even though it's a built-in that isn't in a milestone release), we should be careful and deliberate.  What do we want to have happen if someone's codebase references a missing builtin?

Sorry I didn't spot it and have the discussion ahead of this PR, but: in https://github.com/unisonweb/unison/pull/1705/files#diff-8117eb68e95042e5ae9d5831b25d93b7R411-R412, I wonder if it's okay that we are bulk creating and renaming? Will we ever want to delete or deprecate any [more] of these?  What's the plan?  Should we be declaring them them `B "io2.<xxx>"` from the start, since these names can't ever be reused for a builtin with different semantics from whatever it was when the first user references one in their codebase.  Or maybe it's too late for that, because people can already have added these names to their codebases?

Do we want to have deprecated implementations of IO.putText for a while?

Is this deprecation system already more complicated than we can handle and we need a different one?

stew:
of course we *could* implement the old functions in terms of the new in pure unison if we want them there for a transition

pchiusano:
I think at this stage, with the definitions not being in any official release, we should simply delete it. I think using the proposed deprecation cycle is overkill for changes to trunk between releases... but it seems reasonable for released code. Basically, no real guarantees for folks using trunk.

Seems we'll have to deal with missing builtins regardless. Anytime we add builtins you might pull some code from a newer codebase that references added builtins you don't have. I think the right thing to do in this situation is like a message on pull. Straw man:

```
You've pulled definitions that use builtin functions you don't have 
in your version of Unison. Your version of Unison is M1m. 

Use `check-builtins` to learn more.

.myproject> check-builtins

Here are a list of builtin definitions referenced in this namespace which 
you don't have in your version of Unison (which is version M1m).

    1. builtin type Hash
    2. Hash.Sha3_512 : Hash
    3. Hash.Sha3_256 : Hash

Use `dependents Hash.Sha3_512` or `dependents 2` to see which 
definitions use a builtin listed here.

If you try running any code that uses builtins you don't have, you'll get an 
error. Do `ucm-upgrade` to download the latest version of Unison.

.myproject> ucm-upgrade

You're at version M1m. The latest version is M2b. Downloading now...

I've downloaded it to /my/current/directory/ucm. 
Your current version of ucm is in `~/.bin/ucm`. 

Here's commands you can use to upgrade:

    mv ~/.bin/ucm /my/current/directory/ucm-M1m # backup the old version
    mv /my/current/directory/ucm ~/.bin/ucm
```

aryairani:
Ok, that's fine with me.

pchiusano:
> Should we be declaring them them B "io2.<xxx>" from the start, since these names can't ever be reused for a builtin with different semantics from whatever it was when the first user references one in their codebase. 

I don't think it matters too much? The next version of `IO.putBytes` can be `Builtin "IO.putBytes.v3"`, and so on. It doesn't really matter whether we call the version @stew just wrote `Builtin "IO.putBytes.v2"` or just `Builtin "IO.putBytes"`. I tend toward adding version numbers only when they are needed later... like `Nat.+` will probably never have another version.

We should have a doc that covers the conventions used. Maybe linked from the adding-builtins.md file.

parser-typechecker/unison-parser-typechecker.cabal
```diff
@@ -191,6 +191,7 @@ library
     configurator,
     cryptonite,
     directory,
+    either,
```
pchiusano, line 194: What’s the license on this? Just checking.

unison-src/new-runtime-transcripts/hashing.md
```diff
@@ -15,7 +15,7 @@ You can skip this section, which is just needed to make the transcript self-cont
 .builtin> ls Bytes
 ```
 
-Notice the `fromBase16` and `toBase16` functions. Here's some (somewhat inefficient) convenience functions for converting `Bytes` to and from base-16 `Text`. These could be replaced by use of `Text.toUtf8` and `Text.tryFromUtf8` once those builtins exist:
+Notice the `fromBase16` and `toBase16` functions. Here's some (somewhat inefficient) convenience functions for converting `Bytes` to and from base-16 `Text`. These could be replaced by use of `Text.toUtf8` and `Text.tryFromUtf8`.
```
pchiusano: Can you simplify this transcript now? toBase16 can just be passed to toUtf8 for encoding to text, and for decoding you just utf8 decode and then pass that to fromBase16. 

You can maybe get rid of the List.map definition if it’s not used elsewhere.

pchiusano (approved):
LGTM. For the name, let’s go with just Text.fromUtf8. (I know I suggested tryUtf8 originally 🙃) I went with Bytes.fromBase16 even though it returns Either. In general I think no need to add type information to the names unless it’s necessary to disambiguate the name. (Like Nat.sum vs Float.sum)

parser-typechecker/unison-parser-typechecker.cabal
stew: we already include it in the -core project, so it's already documented

pchiusano:
@stew you can merge whenever you’re ready, cleanup of the hashing transcript to use the new functions is optional. Looks like there’s just some conflicts with trunk that need fixing.

dolio (approved):
This looks good to me.

pchiusano:
@stew I'll do a PR for the hashing transcript stuff, I'd like to keep the inputs as hex literals for easier comparison with the test vectors rather than using `Bytes.fromList`.

PR #1706, opened 2020-10-01
Pull in some changes from the mobile code development

dolio:
This pull request pulls in some work done in preparation of being able to send code between machines. I'm submitting this because it is useful on its own right now. There is also a bugfix included.

- I've added a flag that changes how UCM will submit code to a runtime. The runtime can say that it requires code to be self-contained, or doesn't care. The old runtime requires self-contained code, while the new one doesn't anymore.
- Even for the new runtime, UCM will put things in a let rec block unless they've been added to the codebase (or something along those lines). So the new runtime also hashes these to break things down further. Things already referenced in the codebase are figured out using the `CodeLookup`
- Hashing the top-level let rec doesn't get rid of all cycles. These are lifted after hashing, possibly causing a new top-level let rec for each hashed reference. These are then represented in the runtime as different numbered sections of a single referenced definition. There is a 'main entry point' that corresponds to the original definition, and 'local' sections that are only referred to from within the definition.
- Work on this revealed a problem with how the new runtime was handling some stack management. This has been fixed.
- `PAp` closures have been modified to not store the machine code they will execute. Rather, it stores the numbering necessary to look up that code. This means that the closures can be serialized independently of the code. Continuations need the same treatment, but they are a bit trickier.
- ANF code has been reworked to use `Reference` rather than a machine-specific numbering. It seems like something close to ANF could be transportable, but it isn't safe to be including ephemeral numbering schemes in it in that case.
- Some tagging in the runtime used to be able to recover `References` has been swapped with just storing the reference. This is simpler for some applications, but we'll need to investigate the performance of stuff a bit, to make sure it isn't a problem.
- The new code scheme probably made decompilation of partially applied functions a bit worse. I haven't had time to work out how to do that well.

There's also a tweak for better failing test output in UnisonSources, and a helper function for hashing closed terms.

This branch isn't complete, but I think this work is useful enough to be merged back (and Paul is waiting on the second-last bullet point).

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
@@ -186,13 +186,16 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
   evalUnisonFile :: PPE.PrettyPrintEnv -> UF.TypecheckedUnisonFile v Ann -> _
   evalUnisonFile ppe (UF.discardTypes -> unisonFile) = do
     let codeLookup = Codebase.toCodeLookup codebase
-    selfContained <- Codebase.makeSelfContained' codeLookup unisonFile
+    evalFile <-
+      if Runtime.needsContainment rt
+        then Codebase.makeSelfContained' codeLookup unisonFile
+        else pure unisonFile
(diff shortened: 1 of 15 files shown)


pchiusano (approved):
Great, thanks!

PR #1708, opened 2020-10-05
Wip/codebase2, not ready for review, just tracking progress & todos

aryairani:
- [x] 1. finish V1 -> V2 conversion
  - [x] convert term component
    - [x] hashable representation
      - [x] update canonical ordering of component
    - [x] build locally-indexed component 
    - [x] populate type & type-mentions indices
    - [x] populate dependents index
  - [x] convert type component 👈 
    - [x] hashable representation
    - [x] build locally-indexed component
  - [x] convert patch
  - [x] convert namespace
  - [x] convert causal
- [x] 2. ingest .base into uncompressed V2 format for sanity check (1-3 days, depending on bugs)
- [ ] 3. flesh out and implement switching between full and compressed/diffed representations of namespaces (optionally patches) (1 week w/ pairing maybe?)
- [ ] 4. ingest .base into compressed V2 format for sanity check (1-3 days, depending on bugs)
--- If all good: ---
- [ ] 5. switch out V2 Reference / V2 Codebase types across full application (1 week w/ pairing)
- [x] 6. implement V2 -> V2 sync (1.5 weeks)
--- Later (separate effort) ---
develop standalone V2 -> V2 sync daemon in coordination with Stew and/or Dan

### questions:
Q: What is the deal with this bit in Term hashing?
```haskell
-- So long as `Reference.Derived` ctors are created using the same
-- hashing function as is used here, this case ensures that references
-- are 'transparent' wrt hash and hashing is unaffected by whether
-- expressions are linked. So for example `x = 1 + 1` and `y = x` hash
-- the same.
  Ref (ReferenceDerived h 0 1) -> H.fromBytes (Hash.toBytes h)
  Ref (ReferenceDerived h i n) -> H.accumulate
    [ tag 1
    , hashed $ H.fromBytes (Hash.toBytes h)
    , H.Nat i
    ]
```
Q: How to handle type-of-term / type-of-constructor lookups? How do constructor types actually get into the typechecker?
A: Type-of-terms have their own table now; type-of-constructor will just come from Decls, as before.


aryairani:
[c2.zip](https://github.com/unisonweb/unison/files/6361931/c2.zip)
Here's a copy of the v2-version-of-base-containing codebase I'm using to debug git push/pull

parser-typechecker/src/Unison/Codebase/Conversion/Sync12.hs
```diff
@@ -0,0 +1,613 @@
...
+  (R m n a, S m n, Applicative m) =>
+  (m ~> n) ->
+  Generation ->
+  Entity m ->
+  n (TrySyncResult (Entity m))
+trySync t _gc e = do
+  Env _ dest _ <- Reader.ask
+  case e of
+    C h mc -> do
+      t (Codebase.branchExists dest h) >>= \case
```
aryairani: Is this the right test here? Or should I check status instead

aryairani:
I opened another one.

PR #1710, opened 2020-10-06
Do transitive dependency resolution in the new runtime

### LINKED ISSUE #1709: New runtime doesn't load needed transitive dependencies for execution
Here's a simple failing transcript. It runs a watch expression that depends indirectly on some hash, and that hash isn't loaded. It looks like it is only loading the immediate dependents of the terms being evaluated:

    ```unison
    id x = x

    id2 x =
      z = 384849
      id x
    ```

    ```ucm
    .scratch> add
    ```

    ```unison
    > id2 "hi"
    ```

This fails with -

```
unknown reference: #a6au3
  error, called at src/Unison/Runtime/Interface.hs:202:17 in unison-parser-typechecker-0.1-DahyA8UvBK0B5OVYJToEFT:Unison.Runtime.Interface
  ref, called at src/Unison/Runtime/Interface.hs:208:38 in unison-parser-typechecker-0.1-DahyA8UvBK0B5OVYJToEFT:Unison.Runtime.Interface
  compileTerm, called at src/Unison/Runtime/Interface.hs:186:5 in unison-parser-typechecker-0.1-DahyA8UvBK0B5OVYJToEFT:Unison.Runtime.Interface
  compileAllocated, called at src/Unison/Runtime/Interface.hs:171:17 in unison-parser-typechecker-0.1-DahyA8UvBK0B5OVYJToEFT:Unison.Runtime.Interface
```

Where the `#a6au3` hash is the hash of `id`.

dolio:
The existing code was only loading the immediate child dependencies.

Fixes #1709

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Interface.hs
+++ b/parser-typechecker/src/Unison/Runtime/Interface.hs
@@ -11,9 +11,11 @@ import GHC.Stack (HasCallStack)
 import Control.Exception (try)
 import Control.Monad (foldM, (<=<))
 
-import Data.Bifunctor (first,second)
+import Data.Bifunctor (first)
 import Data.Foldable
 import Data.IORef
+import Data.Set as Set (Set, (\\), singleton, map, notMember, filter)
+import Data.Traversable (for)
 import Data.Word (Word64)
 
 import qualified Data.Map.Strict as Map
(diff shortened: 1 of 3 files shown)


pchiusano (approved):
Great, thanks!

PR #1712, opened 2020-10-07
Cryptographic primitives for hashing and HMAC

### LINKED ISSUE #1694: Hashing and HMAC builtins proposal
Will start with an example usage:

```Haskell
ex : Bytes
ex = 
  hmac.new Sha3_512 secretKey
    |> append "uno"
    |> append (1,2,3)
    |> append [1,2,3,4]
    |> appendBytes (Text.toUtf8 "Oat bag, gonna get my oat bag now!")
    |> hmac.finish 
```

Here's the API.

## Hashing

These would go in a new namespace, `crypto.hash`. Based on https://hackage.haskell.org/package/cryptonite-0.27/docs/Crypto-Hash.html.

```Haskell
-- type used for incremental hashing
builtin type Sha3_512

-- create a new hashing context
Sha3_512.new : Sha3_512

-- add any value to the hash context, incorporating type information into the hash
Sha3_512.append : Sha3_512 -> a -> Sha3_512

-- add bytes to the hash context
Sha3_512.appendBytes : Sha3_512 -> Bytes -> Sha3_512

-- all done, convert to raw `Bytes`
Sha3_512.finish : Sha3_512 -> Bytes
```

We'd have very similar functions for `Sha3_256` and other hashing functions we want to support. No fancy types. If you want to abstract over the hashing function, you can do so with pure Unison code (just a record of functions will do).

## HMAC

These would go in `crypto`. Based on https://hackage.haskell.org/package/cryptonite-0.27/docs/Crypto-MAC-HMAC.html

```Haskell
unique type Hash h = 
  { new : h
  , append : forall a . h -> a -> h
  , appendBytes : h -> Bytes -> h
  , finish : h -> Bytes
  }

builtin type Hmac

Sha3_512 : Hash Sha3_512
Sha3_512 = Hash Sha3_512.new Sha3_512.append Sha3_512.appendBytes Sha3_512.finish

hmac.new : Hash h -> Key -> Hmac h
hmac.append : Hmac h -> a -> Hmac h
hmac.appendBytes : Hmac h -> Bytes -> Hmac h
hmac.finish : Hmac h -> Bytes
```

### LINKED ISSUE #1715: Switch to unpinned byte array type in `Bytes` implementation
The `Bytes` type is a finger tree of some byte array type that uses [this interface](http://hackage.haskell.org/package/memory-0.15.0/docs/Data-ByteArray.html). `Data.ByteString` is an instance, but that uses pinned memory. 

I'd like to switch to something unpinned. It looks like `Block Word8` uses unpinned memory and would be a good representation. It already has instances of `ByteArray` too.

pchiusano:
This PR adds support for hashing and HMAC. See [the transcript](https://github.com/unisonweb/unison/blob/topic/basic-hashing/unison-src/new-runtime-transcripts/hashing.output.md) for docs and tests. Note that the newly created `unison-src/new-runtime-transcripts` directory has transcripts that will only be run using the new runtime and there's a tweak to the test harness to support this.

Closes #1694, but see loose ends tickets below.

## Implementation notes

Followed the instructions in `adding-builtins.md`. Nothing too exciting. I hit a couple bugs that @dolio was kind enough to fix. I have some ideas on how to improve the handling of foreign data types in the runtime but will work on that separately.

I revamped the `Bytes` implementation to use a `ByteArray` interface provided by the [memory package](https://hackage.haskell.org/package/memory-0.15.0/docs/Data-ByteArray.html), to get various encoding/decoding functions (like converting to and from base 16 and base 32) and to switch to using unpinned memory. That package is already a dependency of cryptonite which we are using for crypto. See the diff in `Bytes.hs` for details.

See self-review.

I had wanted an incremental hashing (and HMAC) implementation but put that aside for now: you have to feed it the full `Bytes` or other value you are hashing. An incremental interface would look like:

```Haskell
Hash.new : HashAlgorithm -> Hash
Hash.add : Bytes -> Hash -> Hash
Hash.finish : Hash -> Bytes
```

This will let you do things like hash a large file without having to load the whole file contents in memory at once. The [cryptonite library](https://hackage.haskell.org/package/cryptonite) we are using supports incremental hashing, but the data type it uses to support it isn't safely serializable - it's literally just an opaque byte array internally, and the hashing algorithm (which is implemented in C) then makes assumptions based on these bytes. I was concerned that if you just deserialize any old byte array and do hashing with it, you could get buffer overruns and so on.

We could support an incremental interface later.

## Test coverage

The transcript has some basic tests. The various `Bytes` encoding functions could use more tests.

## Loose ends

* Bytes encoding/decoding tests: #1713
* Provide implementations of universal hashing and functions: #1714


parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1104,6 +1108,50 @@ mvar'try'read instr
...
+  = ([BX],)
+  . TAbss [b]
+  . TLet e UN (AFOp instr [b])
+  . TMatch e . MatchSum
+  $ mapFromList
+  [ (0, ([BX], TAbs ev $ TCon eitherReference 0 [ev]))
+  , (1, ([BX], TAbs ev $ TCon eitherReference 1 [ev]))
+  ]
+  where
+  [e,b,ev] = freshes 3
```
pchiusano, line 1153: I added these common cases and am recycling them in a few places for various builtins.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1388,6 +1436,46 @@ declareForeigns = do
...
+  declareHashAlgorithm "Sha2_512" Hash.SHA512
+  declareHashAlgorithm "Sha2_256" Hash.SHA256
+  declareHashAlgorithm "Blake2b_512" Hash.Blake2b_512
+  declareHashAlgorithm "Blake2b_256" Hash.Blake2b_256
+  declareHashAlgorithm "Blake2s_256" Hash.Blake2s_256
+
+  -- declareForeign ("crypto.hash") pfopbb . mkForeign $ \(HashAlgorithm _ref _alg, _a :: Closure) ->
+  --   pure $ Bytes.empty -- todo : implement me
+
+  declareForeign "crypto.hashBytes" pfopbb . mkForeign $
```
pchiusano, line 1457: Notice am using the `pfopbb : ForeignOp` convention here - it's a pure foreign operation that takes 2 boxed values. Probably a lot of the foreign operations in this file could be cleaned up to use a small number of `ForeignOp` types.

parser-typechecker/src/Unison/Runtime/Foreign/Function.hs
```diff
@@ -295,6 +296,10 @@ instance ForeignConvention (MVar Closure) where
   readForeign = readForeignAs (unwrapForeign . marshalToForeign)
   writeForeign = writeForeignAs (Foreign . Wrap mvarRef)
 
+instance {-# overlappable #-} BuiltinForeign b => ForeignConvention b where
+  readForeign = readForeignBuiltin
+  writeForeign = writeForeignBuiltin
```
pchiusano, line 301: This saves you from needing to provide a `ForeignConvention` for a type that's already a `BuiltinForeign`.

pchiusano (commented):
Self review.

parser-typechecker/src/Unison/Util/Bytes.hs
```diff
@@ -3,94 +3,205 @@
...
+dropView k (View i n bs) = View (i + (k `min` n)) (n - (k `min` n)) bs
+
+data View bytes = View
+  { viewOffset :: !Int
+  , viewSize   :: !Int
+  , unView     :: !bytes
+  }
+
+instance B.ByteArrayAccess bytes => Eq (View bytes) where
+  v1 == v2 = viewSize v1 == viewSize v2 && unsafeDupablePerformIO (
```
dolio, line 185: To be honest, this (`unsafeDupablePerformIO`) seems like the wrong way to be doing this now. GHC has primops for comparing byte arrays that don't require dangerous stuff like this.

dolio (approved):
Things overall look all right to me.

I think it'd be better to get those `unsafeDupablePerformIO` calls eliminated, but it's not really different than depending on a library that does it (but shouldn't).

stew (approved):

pchiusano:
I wonder if we could just build some utilities around Vector Word8 rather than using the ByteArray interface. I'm not really sold on that interface. But I'd also like to start moving away from ByteString as I think the use of pinned memory isn't what we want.

Going to merge this but we can revisit the Bytes internals anytime.

PR #1719, opened 2020-10-08
Assorted builtin fixes

dolio:
- The operations that produce sockets were just throwing them away completely, breaking the calling convention.
- The IOError wrapper was not obeying the calling convention either. It expected a boxed result, but it is returned as an enum tag.
- Helpful addition to Show instance for Foreign wrapped text to show the actual text.


pchiusano (approved):
LGTM. There's conflicts on this, can you sync with trunk?

pchiusano:
😎👍

PR #1720, opened 2020-10-09
Add multi-argument `cases` lambdas

pchiusano:
This PR adds syntax for defining multi-argument functions using `cases`:

```Haskell
merge : [a] -> [a] -> [a]
merge = cases
  []       ys         -> ys
  xs       []         -> xs
  (h +: t) (h2 +: t2) ->
    if h <= h2 then h  +: merge t (h2 +: t2)
    else            h2 +: merge (h +: t) t2
```

See [the transcript](https://github.com/unisonweb/unison/tree/topic/multiarg-cases/unison-src/transcripts/lambdacase.output.md) for usage and explanation of the desugaring. Like destructuring bind (#1700), this is just syntax sugar which is mirrored by the pretty-printer.

Testing coverage is pretty decent: the existing parser and pretty printer share 95% of the same codepath, so the fact that existing tests pass is a good sign, and I added some unit tests as well as the transcript.

@atacratic, I moved a couple pretty-printer specific stuff out of `Term` and into `TermPrinter`. I'd like your review on the `TermPrinter` changes if you have a chance.

@dolio, I could imagine supporting multi-argument pattern matching more natively in the runtime or AST. I don't think it matters too much what's in the AST but it might be nice if the runtime could eventually have a more optimized implementation of this. Not high priority I don't think though.

/cc @anovstrup

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -6,6 +6,8 @@ module Unison.TermPrinter where
 
 import Unison.Prelude
 
+import           Control.Monad.State            (evalState)
+import qualified Control.Monad.State           as State
 import           Data.List
 import qualified Data.Map                      as Map
 import qualified Data.Set                      as Set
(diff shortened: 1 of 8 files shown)


parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -1087,3 +1100,78 @@ unLetBlock t = rec t where
...
+-- shown to the left of the `-> cases`.
+--
+-- For instance, if given this term:
+--
+--   x y z -> match z with
+--     [] -> "empty"
+--     (h +: t) -> "nonempty"
+--
+-- this function will return Just ([x,y], [[] -> "empty", (h +: t) -> "nonempty"])
+-- and it would be rendered as `x y -> cases [] -> "empty"`
```
atacratic: missing second branch in example in comment

atacratic (approved):
Nice, great to have this!

One comment about a comment.  

unLamsMatch' is complicated, and I didn't manage to give it a deep review - but didn't spot any problems either.

Tested a bit trying to trip it up with edge cases, no luck :-)

So LGTM!

parser-typechecker/src/Unison/TermPrinter.hs
pchiusano: Thanks for spotting! fixed in latest commit

pchiusano:
@atacratic thanks for taking a look at this, merging now (once CI passes).

PR #1722, opened 2020-10-11
fix foreign call to socketReceive for new Bytes impl

stew:
## Overview

Currently socketReceive is broken, when we switched the Bytes implementation away from ByteString, we started making a call to a different hGet with a different return signature


dolio (approved):
Looks good to me.

PR #1726, opened 2020-10-13
Fix transcript runner to show the actual error when it fails

pchiusano:
This was bugging me. When the transcript failed, it wouldn't report the error, it would only tell you that "The transcript failed due to an error encountered in the stanza above." Um, okay, but what was the error?? Now it fails with a nice message, for instance:

```


  This looks like the start of an expression here 
  
      3 | List.map f as sldkfj;;; jklljf; =
  
  but at the file top-level, I expect one of the following:
  
    - A binding, like List.map = 42 OR
                      List.map : Nat
                      List.map = 42
    - A watch expression, like > List.map + 1
    - An `ability` declaration, like ability Foo where ...
    - A `type` declaration, like type Optional a = None | Some a
    - A `namespace` declaration, like namespace Seq where ...
```

Previously, "the error is: ..." part wouldn't be shown.

## Implementation notes and testing

Nothing interesting. All the transcripts pass still so we can be pretty confident this hasn't broken anything. The code I added only kicks in when the transcript is about to die anyway.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
+++ b/parser-typechecker/src/Unison/Codebase/TranscriptParser.hs
@@ -178,11 +178,9 @@ run newRt dir configFile stanzas codebase branchCache = do
                 case Map.lookup cmd patternMap of
                   -- invalid command is treated as a failure
                   Nothing ->
-                    dieWithMsg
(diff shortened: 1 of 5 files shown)


stew (approved):
love this! TYVM

PR #1730, opened 2020-10-19
Nat.popCount and Int.popCount

pchiusano:
Added these builtins for counting the number of set bits in a `Nat` or `Int`.

```Haskell
Nat.popCount : Nat -> Nat
Int.popCount : Int -> Nat
```

For instance, `Nat.popCount 5 == 2` (since `5` is `101` in binary).

This is a handy thing to have around, and I wanted it for a [splitmix random number generator implementation](https://gist.github.com/pchiusano/a987b710b71769ab7faafb8701a35264).

See [this test harness](https://github.com/unisonweb/unison/blob/feature/popcount/unison-src/new-runtime-transcripts/builtins.output.md). I think that adding to this transcript should be the new preferred way of adding unit tests for new builtins and I figured out a nice idiom for getting all the tests to run even if some of them fail - basically, add all the tests and only `.> test` at the very end.

### Other items

* There were some bitwise functions on `Int` that were missing in the new runtime. I filled these in and included tests for them and all other `Int` and `Nat` builtins.
* Possibly controversial: I used a new instruction for these, rather than using foreign functions. I was mostly just following the convention of some of the other bit twiddly operations (like `complement`, `xor`, etc), which have their own instruction as well.

@dolio can you take a look?


parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -247,7 +247,6 @@ builtinsSrc =
   , B "Int.signum" $ int --> int
   , B "Int.leadingZeros" $ int --> nat
   , B "Int.negate" $ int --> int
-  , B "Int.negate" $ int --> int
```
pchiusano, line 250: For some reason there were 2 copies of `Int.negate` in this list. Which has no effect.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -242,10 +242,6 @@ gei = cmpopb LEQI Ty.intRef
 gtn = cmpopn LEQN Ty.intRef
 gen = cmpopb LEQN Ty.intRef
 
-neqi, neqn :: Var v => SuperNormal v
-neqi = cmpopn EQLI Ty.intRef
-neqn = cmpopn EQLN Ty.intRef
```
pchiusano, line 247: Deleted these as there isn't currently a `Nat.!=` or `Int.!=` builtin and they weren't being used.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -254,17 +250,24 @@ sgni, negi :: Var v => SuperNormal v
...
+
+andn, orn, xorn, compln, andi, ori, xori, compli :: Var v => SuperNormal v
 andn = binop ANDN Ty.natRef
 orn = binop IORN Ty.natRef
 xorn = binop XORN Ty.natRef
 compln = unop COMN Ty.natRef
+andi = binop ANDN Ty.intRef
+ori = binop IORN Ty.intRef
+xori = binop XORN Ty.intRef
+compli = unop COMN Ty.intRef
```
pchiusano, line 270: These were missing from the new runtime implementation.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1185,10 +1187,15 @@ builtinLookup
   , ("Int.shiftRight", shri)
   , ("Int.trailingZeros", tzeroi)
   , ("Int.leadingZeros", lzeroi)
+  , ("Int.and", andi)
+  , ("Int.or", ori)
+  , ("Int.xor", xori)
+  , ("Int.complement", compli)
```
pchiusano, line 1193: These were missing previously, so programs that used these would crash with the new runtime. Fixed and added usages of these to the test harness.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1411,8 +1418,8 @@ declareForeigns = do
   declareForeign "IO.socketSend" socket'send
     . mkForeignIOE $ \(sk,bs) -> SYS.send sk (Bytes.toArray bs)
   declareForeign "IO.socketReceive" socket'receive
-    . mkForeignIOE $ \(hs,n) -> 
-       fmap (maybe Bytes.empty Bytes.fromArray) $ SYS.recv hs n 
+    . mkForeignIOE $ \(hs,n) ->
+       fmap (maybe Bytes.empty Bytes.fromArray) $ SYS.recv hs n
```
pchiusano, line 1422: Just trailing whitespace cleanup (my editor strips these automatically).

pchiusano (commented):
Self review.

stew (approved):

dolio (approved):
Looks good to me.

I think these definitely needed to be instructions for performance reasons, so no controversy on my end. :)

PR #1732, opened 2020-10-20
Allow suffix-based name resolution to refer to definitions in the current file

### LINKED ISSUE #1578: Suffix-based name resolution uses definitions from the codebase instead of definitions from the current file

pchiusano:
See [the transcript](https://github.com/unisonweb/unison/blob/fix/1578/unison-src/transcripts/fix1578.output.md) for an explanation of the new behavior and tests.

Possibly controversial decision: suffix-based resolution prefers file definitions to matches from the codebase. So if your file defines `foo.bar`, and you say `bar` elsewhere in your file, it will refer to `foo.bar`. Previously, it might leave `bar` as free and attempt TDNR possibly involving definitions from the codebase. The trouble with the old behavior is we couldn't do it consistently: TDNR doesn't apply to patterns or types, only terms, leading to surprises of when the imports or FQNs were needed. And TDNR isn't very reliable when when there's multiple things needing TDNR in a block of code which might have type errors.

The new rule of suffix-based resolution is I think clear and easy to work with (you can always refer to things outside the file using explicit imports or longer names):

* If a symbol, `s`, is a suffix of exactly one definition `d` in the file, then `s` refers to `d`. 
* Otherwise, if `s` is a suffix of exactly one definition `d` in the codebase, then `s` refers to `d`.
* Otherwise, if `s` is a suffix of multiple definitions in the file or the codebase, then (at least for terms) type-directed name resolution will be attempted to figure out which definition `s` refers to.

See [the transcript](https://github.com/unisonweb/unison/blob/fix/1578/unison-src/transcripts/fix1578.output.md), which explores the behavior with some illustrative examples.

Other notes:

* The `FileParser` changes were not much code, but a little hairy. I added lots of comments to hopefully explain. The whole way we do name resolution during parsing could use a rethink at some point.
* The tests are pretty good. I explored a number of corner cases.

See self review.


unison-src/tests/cce.u
```diff
@@ -62,7 +62,8 @@ Monoid.orElse m = cases
 
 merge : (a -> a -> Boolean) -> [a] -> [a] -> [a]
 merge lte a b =
-  go acc a b = match List.at 0 a with
+  use List at
+  go acc a b = match at 0 a with
```
pchiusano, line 66: Here's an example where preferring local definitions isn't necessarily a win. There's a `Remote.at` defined in this file, so just `at` would otherwise refer to `Remote.at` (instead of `List.at`). Before, TDNR was able to figure this one out. Now you need to say `List.at` if you want to refer to that in a file where you've defined something else called `at`.

unison-src/transcripts/blocks.output.md
```diff
@@ -253,9 +253,9 @@ ex n =
 
 ```ucm
 
-  These definitions depend on each other cyclically but aren't guarded by a lambda: zap19, zap210
+  The expression in red needs the {SpaceAttack} ability, but this location does not have access to any abilities.
+  
       5 |   zap1 = launchMissiles "neptune" + zap2
```
pchiusano, line 258: I was puzzled about why this transcript output changed. See this comment if you're curious. The new output is correct: https://github.com/unisonweb/unison/pull/1732/commits/608ad4830c2059c6f21b79d65089148219009f47

pchiusano (commented):
Self review

pchiusano:
This fixes #1578

unison-src/tests/cce.u
aryairani: Err, the previous version called `List.at`, not `at`, which makes me not understand the comment.

unison-src/tests/soe.u
```diff
@@ -61,7 +61,8 @@ Monoid.orElse m = cases
 
 merge : (a -> a -> Boolean) -> [a] -> [a] -> [a]
 merge lte a b =
-  go acc a b = match List.at 0 a with
+  use List at
+  go acc a b = match at 0 a with
```
aryairani, line 65: Same here; why change the test at all?

aryairani (approved):
If the they idea here is to just have some rule that is easy to explain, and if a complete rethink is already on the agenda, then I'm good with this. Otherwise, I have lots of questions about the current behavior, the new behavior, the desired behavior. :)

unison-src/tests/soe.u
pchiusano: In both these files, there was an unqualified `at` or two in the function, and there wasn't a `use` to pin down what `at` it was referring to (I think you need to view more than just the immediately surrounding lines since the unqualified usage might be later in the function). I could have either changed these all to `List.at` or what I instead did was added an explicit `use List at` and then changed all the use sites to in the block to use that unqualified `at`.

unison-src/tests/cce.u
aryairani: I see; so your comment was referring to the unqualified at on line 67/68, but the switch from `List.at` to `at` on line 65/66 threw me off.

runarorama (approved):

PR #1733, opened 2020-10-21
In-line handlers were not being enclosed before floating

dolio:
This should fix some issues that crop up when handlers are nested under things in the new runtime.


pchiusano (approved):
@dolio cool, how about adding a regression test to the unison-src/new-runtime-transcripts/ directory?

dolio:
Oh, oops. I just added it to the test cases rather than the transcripts.

The previous behavior was a crash, so I was thinking in terms of that.

pchiusano:
@dolio that's fine, it looks like all those files will also be run using the new runtime.

dolio:
Yeah, we're running most stuff on both runtimes in CI at this point.

PR #1734, opened 2020-10-22
Get rid of those weird synthetic variable names

### LINKED ISSUE #1390: Convert synthetic variable types to nice named vars before serialization
Let's get rid of these blackboard letters, used in synthetic variables:

```
.> find List.map

  1. base.List.map : (a ->{𝕖} b) -> [a] ->{𝕖} [b]
```

Just another thing to explain, not really helping anything IMO.

To do this properly we need to make a fix: right now synthetic variable types get converted to named variables during serialization, too late. I suspect this could result in capture if the user picks the same variable prefixes as are chosen for synthetic vars (like if you call a variable `𝕖`). The right thing to do is to pick names for these  synthetic variables just before serializing a term or type, when you have the full term/type and can do whatever naming scheme you want in a way that always avoids variable capture.

### Implementation notes

Rather than having `V1` work  on `Term Symbol ann` and `Type Symbol ann`, define the formats  in terms of `Term SimpleVar ann` and `Type SimpleVar ann`, where `SimpleVar` is just something isomorphic to `(Text, Word64)`.

No repo format changes here as the on disk format for symbols is already equivalent to `SimpleVar`.

* To implement, we'll need functions:
   * `Unison.Var.simplifyVars :: Var v => ABT.Term f v a -> ABT.Term f SimpleVar a`. Here we can pick whatever naming conventions we want for synthetic variables, no need to use the funny unicode chars.
   * `Unison.Type.simplifyVars :: Var v => Type v a -> Type SimpleVar a`
   * `Unison.Term.simplifyVars :: Var v => Term v a -> Term SimpleVar a`, defined in terms of `typeMap` and `Unison.Type.simplifyVars`.
* Just monomorphise `putType`, `putTerm`, to work with `Type SimpleVar ann`, `Term SimpleVar ann`. (Okay to leave `getType` etc alone)
* Define `putType'`, `putTerm'` to take in a `Term Symbol a` / `Type Symbol a`.
* Replace calls to `putType`, `putTerm` with the `'` versions. Everything should build now and work fine.

pchiusano:
This fixes #1390 

Previously, when inferring type variables for the user, we'd pick names like `𝕩` or `𝕖`, which was confusing and people would regularly ask if these variable names had any special meaning. With this PR, we generate type variables which are just regular ascii, and we even try to pick "nice" variable names depending on what type of synthetic variable it is and what's not being used elsewhere in the type. Take a look at the transcript diffs for lots of examples.

The implementation was straightforward and most of my notes in #1390 on how to do it turned out to be irrelevant. The right place to make the change is when top-level type signatures are being emitted from the typechecker. At that point, you have the full, closed type, and can do whatever alpha renaming you want to it before showing it to the user (and eventually, writing it to the codebase).

## Test coverage

Test coverage is decent given all the transcripts, and I added a separate transcript as well. The implementation is done is such a way that this can't accidentally introduce variable capture issues.

## Loose ends

I could imagine having a separate `Var` type for the types that come out of the typechecking... making it clear that the typechecker can't possibly produce signatures with synthetic type vars in them. I kinda like this idea for a later possible cleanup task.


parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -264,6 +266,49 @@ data InfoNote v loc
   | TopLevelComponent [(v, Type.Type v loc, RedundantTypeAnnotation)]
   deriving (Show)
 
+topLevelComponent :: Var v => [(v, Type.Type v loc, RedundantTypeAnnotation)] -> InfoNote v loc
+topLevelComponent = TopLevelComponent . fmap (over _2 removeSyntheticTypeVars)
```
pchiusano, line 270: This function is used when emitting type signatures for top-level definitions, and it does the `removeSyntheticTypeVars` pass.

unison-core/src/Unison/ABT.hs
```diff
@@ -140,6 +140,13 @@ vmap f (Term _ a out) = case out of
   Cycle r -> cycle' a (vmap f r)
   Abs v body -> abs' a (f v) (vmap f body)
 
+vmapM :: (Applicative m, Traversable f, Foldable f, Ord v2) => (v -> m v2) -> Term f v a -> m (Term f v2 a)
+vmapM f (Term _ a out) = case out of
+  Var v -> annotatedVar a <$> f v
+  Tm fa -> tm' a <$> traverse (vmapM f) fa
+  Cycle r -> cycle' a <$> vmapM f r
+  Abs v body -> abs' a <$> f v <*> vmapM f body
```
pchiusano, line 148: Helper function for mapping variables in an ABT using an applicative function.

pchiusano (commented):
Self review.

dolio (approved):
Looks fine to me.

PR #1736, opened 2020-10-23
Fix to TDNR to verify that the suffix matches for any TDNR replacement 

### LINKED ISSUE #845: TDNR too aggressive: resolves Blargh.== as Universal.==
The following successfully typechecks (but shouldn't):

```haskell
foo a b = a Blargh.== b
```

```
  I found and typechecked these definitions in ~/projects/unison/foo.u. If you do an `add` or
  
      foo : a -> a -> builtin.Boolean
```

The `Blargh.==` was resolved to `Universal.==`:

```
.> add

    foo : a -> a -> builtin.Boolean

.> view foo

  foo : a -> a -> builtin.Boolean
  foo a b =
    use builtin.Universal ==
    a == b

.> 
```

pchiusano:
Previously, `Nat.zoink` could actually be resolved by TDNR to `Float.zoink`. Fixes #845 

This fixes that. There were a number of tests accidentally relying on the old behavior which have been updated.

Updates a small amount of code in the TDNR setup and implementation. The key idea is to populate the TDNR environment with definitions whose full name has a suffix which corresponds to a free term variable in the file. So if the namespace has `Float.zoink` in it, and the file has `zoink` as a free variable, `Float.zoink` gets added to the TDNR environment. But if the free variable is `Nat.zoink`, then that's not a suffix of `Float.zoink`, so `Float.zoink` shouldn't be considered.

Test coverage is probably pretty decent just from existing transcripts and the regression test I added but I'm going to noodle on some more subtle tests.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/IOSource.hs
+++ b/parser-typechecker/src/Unison/Runtime/IOSource.hs
@@ -166,9 +166,9 @@ unique[b28d929d0a73d2c18eac86341a3bb9399f8550c11b5f35eabb2751e6803ccc20] type
 d1 Doc.++ d2 =
   use Doc
   match (d1,d2) with
-    (Join ds, Join ds2) -> Join (ds Sequence.++ ds2)
-    (Join ds, _) -> Join (ds `Sequence.snoc` d2)
-    (_, Join ds) -> Join (d1 `Sequence.cons` ds)
+    (Join ds, Join ds2) -> Join (ds List.++ ds2)
+    (Join ds, _) -> Join (ds `List.snoc` d2)
+    (_, Join ds) -> Join (d1 `List.cons` ds)
     _ -> Join [d1,d2]
 
 unique[q1905679b27a97a4098bc965574da880c1074183a2c55ff1d481619c7fb8a1e1] type
--- a/unison-src/tests/324.u
+++ b/unison-src/tests/324.u
@@ -1,5 +1,5 @@
 foo a b =
-  if a Text.== "" then
+  if a `Text.eq` "" then
     match Text.size b with
       1 -> false
       _ -> true
--- a/unison-src/tests/handler-stacking.u
+++ b/unison-src/tests/handler-stacking.u
@@ -11,7 +11,7 @@ main = '(tell get)
 
 replicate : Nat -> '{e} () -> {e} ()
 replicate n x =
-  if n Nat.== 0 then () else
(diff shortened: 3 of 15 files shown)


parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -166,9 +166,9 @@ unique[b28d929d0a73d2c18eac86341a3bb9399f8550c11b5f35eabb2751e6803ccc20] type
 d1 Doc.++ d2 =
   use Doc
   match (d1,d2) with
-    (Join ds, Join ds2) -> Join (ds Sequence.++ ds2)
-    (Join ds, _) -> Join (ds `Sequence.snoc` d2)
-    (_, Join ds) -> Join (d1 `Sequence.cons` ds)
+    (Join ds, Join ds2) -> Join (ds List.++ ds2)
+    (Join ds, _) -> Join (ds `List.snoc` d2)
+    (_, Join ds) -> Join (d1 `List.cons` ds)
```
pchiusano, line 171: The function was renamed to `List.cons` and `List.snoc`, etc, but this code was working due to the bug!

unison-src/tests/324.u
```diff
@@ -1,5 +1,5 @@
 foo a b =
-  if a Text.== "" then
+  if a `Text.eq` "" then
```
pchiusano, line 2: `Text.==` is not a function, it's called `Text.eq` if you want the monomorphic version.

unison-src/tests/handler-stacking.u
```diff
@@ -11,7 +11,7 @@ main = '(tell get)
 
 replicate : Nat -> '{e} () -> {e} ()
 replicate n x =
-  if n Nat.== 0 then () else
+  if n `Nat.eq` 0 then () else
```
pchiusano, line 14: It's `Nat.eq` not `Nat.==` (or we could use `Universal.==`)

unison-src/tests/inner-lambda1.u
```diff
@@ -1,4 +1,4 @@
-use Nat drop >=
+use Nat drop
```
pchiusano, line 1: It's `Nat.gteq` if you want the monomorphic version. Here I just used universal `>=`

pchiusano (commented):
Self review.

runarorama (approved):

Issue #1728, opened 2020-10-14
Code serialization builtins


pchiusano:
**See [this comment](https://github.com/unisonweb/unison/issues/1728#issuecomment-712426285) for the latest design.**

The rest of this comment was the starting design and can be ignored unless you're just interested in following along with the design process that led to the above. 

---

Here's an API sketch for primitives for shipping code around at runtime. This would be used as a building block for network protocols that sync needed dependencies on the fly. This is a low-level, dynamically-typed API, not something you'd directly program against. These are the sort of primitives you'd use when providing an implementation of a `Remote` ability or some such:

```Haskell
builtin type Code

-- Convert any value to a `Code`, discarding all type information
Code.code : a -> Code
-- Is the local runtime missing a copy of the provided `Link`?
Code.isMissing : Link ->{IO} Boolean

-- Obtain the dependencies of a `Code`
Code.dependencies : Code -> [Link]

-- Serialization and deserialization
Code.serialize : Code -> Bytes
Code.deserialize : Bytes -> Either Text Code

-- Load a `Code` value - you better know what you're doing as this is effectively a cast
Code.load : Code ->{IO} (Optional a)
```

So any value can be converted to `Code`, the dependencies of a `Code` can be enumerated, and `Code` can be serialized and deserialized. A typical protocol might involve converting a value to `Code`, sending that over the network. The other side deserializes the `Code`, checks that it `Code.has` all the dependencies of that `Code`, and if yes, calls `Code.load`, and if no, contacts the sender (or perhaps a central `Code` cache) to retrieve any needed dependencies. 

With these protocol, the recipient will be assuming it's receiving a value of some particular type (like `'{Blah} ()`) and will be acting on that assumption.

pchiusano:
@dolio I'm not totally clear on whether we even care about type dependencies, or if we just care about term dependencies. Like if Alice sends Bob a function `f` that references the `Foo` type, does Bob need a copy of `Foo`? Or does he just need a copy of the other functions that `f` references?

I think I might have answered my own question - the runtime will probably need to know at least the arities of all constructors / ability operations in order to generate the correct code.

dolio:
So far I've been operating under the assumption that anything that would require knowing about constructor arities and whatnot would already be done before getting to the level that this `Code` type exists at.

However, I'm not certain that the various 'works for everything somehow' operations don't need to know about types.

What would cause `Code.load` to return `None`?

pchiusano:
[...]

PR #1748, opened 2020-11-10
Finish implementing code serialization support

dolio:
This is a pull request to complete the code saving/reloading support in the new runtime. The API provided is the one designed in #1728 

I haven't added any tests yet, as I'm not sure what sort of coverage is desirable. I'll at least add some simple ones for trivial behavior, though.


pchiusano:
Exciting!

In general I'd try to do round trip testing for a bunch of definitions. I don't think you need to go crazy as a small number of definitions will exercise all the code paths of the serialization and deserialization logic.

This could be done with a transcript if we have a way of doing I/O tests with transcripts, which @stew might have some thoughts on.

unison-src/transcripts/alias-many.output.md
```diff
@@ -41,285 +41,298 @@ Let's try it!
...
-  190. crypto.HashAlgorithm.Sha2_256 : HashAlgorithm
-  191. crypto.HashAlgorithm.Sha2_512 : HashAlgorithm
-  192. crypto.HashAlgorithm.Sha3_256 : HashAlgorithm
-  193. crypto.HashAlgorithm.Sha3_512 : HashAlgorithm
-  194. crypto.hash : HashAlgorithm -> a -> Bytes
-  195. crypto.hashBytes : HashAlgorithm -> Bytes -> Bytes
-  196. crypto.hmac : HashAlgorithm -> Bytes -> a -> Bytes
-  197. crypto.hmacBytes : HashAlgorithm
+  24.  builtin type Code
+  25.  Code.cache_ : [Tuple Term Code] ->{IO} [Term]
```
pchiusano: Noting this signature is off based on earlier comment on signature given in `Builtin.hs`.

parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -503,6 +506,22 @@ mvarBuiltins =
   mvar :: Type v -> Type v
   mvar a = Type.ref () Type.mvarRef `app` a
 
+codeBuiltins :: forall v. Var v => [(Text, Type v)]
+codeBuiltins =
+  [ ("Code.dependencies", code --> list termLink)
+  , ("Code.isMissing", termLink --> io boolean)
+  , ("Code.serialize", code --> bytes)
+  , ("Code.deserialize", bytes --> eithert text code)
+  , ("Code.cache_", list (pair termLink code) --> io (list termLink))
```
pchiusano: The definition should have type `[(Link.Term,Code)] ->{IO} [Link.Term]`. Tuples are right-nested pairs terminated by a `()`.

Just use this handy helper function in `Builtin.hs`. It's used by `Text.uncons` if you want to see an example. 

```suggestion
  , ("Code.cache_", list (tuple [termLink, code]) --> io (list termLink))
```

```Haskell
tuple :: Ord v => [Type v] -> Type v
tuple [t] = t
tuple ts = foldr pair (DD.unitType ()) ts
```

parser-typechecker/src/Unison/Runtime/ANF/Serialize.hs
```diff
@@ -0,0 +1,560 @@
+{-# language LambdaCase #-}
+{-# language BangPatterns #-}
+
+module Unison.Runtime.ANF.Serialize where
```
pchiusano, line 4: This file must have been fun to write. :)

pchiusano (approved):
LGTM. Reviewing this I am realizing we need more high level docs for runtime implementation stuff.

Before merging, can you fix the signature on `cache_`? See my inline comment.

I'd probably work on some tests part of this PR, but if you want to merge this sooner and create a ticket for adding tests and get on that ASAP that's okay with me.

@stew am a little worried this PR will be a pain to reconcile with your branch you're working on.

parser-typechecker/src/Unison/Runtime/ANF/Serialize.hs
dolio: Super fun.

unison-src/new-runtime-transcripts/codeops.md
```diff
@@ -0,0 +1,53 @@
...
+  Right v -> match Value.load v with
+    Left _ -> bug "missing deps"
+    Right func -> func
+```
+
+```ucm
+.> add
+```
+
+This takes advantage of an exhaustiveness loophole to run the IO code.
```
pchiusano: Lol. Epic. I guess fine for now.

unison-src/new-runtime-transcripts/codeops.md
```diff
@@ -0,0 +1,53 @@
+
+Test for code serialization operations.
+
+```ucm:hide
+.> builtins.merge
+.> cd builtin
+```
+
+Define a function, serialize it, then deserialize it back to an actual
```
pchiusano, line 9: How about some more tests for this transcript (that round trip is `Universal.==` for a variety of definitions), following https://github.com/unisonweb/unison/blob/trunk/unison-src/new-runtime-transcripts/builtins.output.md for inspiration. 

Basically, you create a bunch of `test>` watch expressions, add them all to the codebase, then do one `.> test` at the end.

dolio:
FYI: don't merge this yet. The transcript I wrote used an IO loophole that won't work once the 1696 fix is merged.

unison-src/new-runtime-transcripts/codeops.md
dolio: I've added some more complexity to the thing that's serialized. I can't run deserialization in a test block, though, because it involves IO, no?

pchiusano: That's right. 

@stew has an implementation of a `io.test` command that lets you run I/O-based tests, so let's leave this until that merges. I'll create an issue to track.

PR #1749, opened 2020-11-17
Use cases to inform ability matching type.

### LINKED ISSUE #1696: IO ability not inferred
The watch expression in this code successfully evaluates, although it requires the `IO` ability. It prompts once for a line during the evaluation of the watch expression and then keeps this result, even after closing and opening the UCM session. The inferred type of `dialog` is just `Text` but should be `{IO} Text` and therefore should be rejected.

Code: 
```
Ask.provide : '{IO} a -> '{Ask a} r -> r
Ask.provide answer asker =
  h = cases
    {r}                 -> r
    {Ask.ask -> resume} -> handle resume !answer with h
  handle !asker with h

dialog = Ask.provide readLine '("Hi, " ++ Ask.ask ++ "!")

> dialog
```

Output of UCM on file save after entering `abc` for readLine during the evaluation of the watch expression:
```
  I found and typechecked these definitions in ~/src/unison-expect/ask.u. If you do an `add` or
 
      Ask.provide : '{IO} a -> '{Ask a} r -> r
      dialog      : Text

    10 | > dialog
           ⧩
           "Hi, abc!"
```

dolio:
This change inspects the patterns all at once to determine all the abilities being matched on in a case analysis, and uses it to inform the scrutinee type before checking any of the case bodies. Previously ability matching actually had no influence on the scrutinee type, which is clearly erroneous.

I implemented this as a sort of separate behavior that only applies to well-formed sets of `Request` matches. If some weird situation is checked, like a mix of ability and data matches against the same value, it will just have the old behavior. It could probably be factored a bit more if desired, but I made more of a minimal change so far.

This also prevents code like the following from working:

```
unsafePerformIO : Request {IO} a -> a
unsafePerformIO = cases
  { x } -> x
```

Since there are no `IO` cases being matched against, the type checker decides the `Request` is not actually handling `IO`, I guess. Incomplete matching would still give erroneous results, though.

Fixes #1696 at least, assuming the transcript output is expected.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -51,6 +51,7 @@ import           Data.Bifunctor                 ( first
                                                 , second
                                                 )
 import qualified Data.Foldable                 as Foldable
+import           Data.Functor.Compose           ( Compose(..) )
 import           Data.List
 import           Data.List.NonEmpty             ( NonEmpty )
 import qualified Data.Map                      as Map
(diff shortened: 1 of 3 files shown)


pchiusano (approved):
Nice!

pchiusano:
Q: will this work for multihandlers as well? It seems like it will, but a test for this might be good.

dolio:
Yeah, it works for multi-handlers. Should I just tweak the 1696 transcript, or put it in another one or what?

pchiusano:
Sure, you can put it in the same transcript.

dolio:
Actually, we already have a multi-handler test in the methodical `abilities.u`

pchiusano:
Cool. Okay, merging then.

PR #1750, opened 2020-11-19
Rewritten lexer using Megaparsec supporting hex and bytes literals

pchiusano:
(You can start reviewing, but hold off on merging: I have one todo left, making lexer errors print out nicely.)

This is a drop in replacement for the old lexer, totally rewritten using Megaparsec. This was done to support new documentation syntax which will require some lexer fanciness. The code isn't really much shorter, but it's now more straightforward parser combinator code. The layout handling, however, is still pretty finicky.

I added a couple features during the rewrite:

* Hexadecimal and octal literals, like `0xdeadbeef` or `0o012723`, which are parsed as the corresponding `Nat`.
* Bytes literals, which start with `0xs`, for instance `0xs9ab38bcf83ffabb9012`, which are used by the pretty-printer.

I was going to add a syntax for identifiers that include spaces or other special characters, using surrounding double backticks (or any number of enclosing backticks that's greater than 1). The lexer change was very easy, but the pretty-printer change seemed annoying and hard to do properly because of how names are just an opaque text value currently. So I left this feature commented out for now.

## Test coverage

There's good coverage from all the existing tests and transcripts. The new bytes literals feature has some testing from existing transcripts. I added some tests of the hex and octal literals as well to the term printer test suite.

I added a new transcript (`error-messages.md`) demonstrating some of the new error messages.

## Loose ends

I wasn't able to delete some of the old wordy and symboly id parsers which are currently being used in path parsing which I didn't want to mess with for this PR. But I'd like to clean that up at some point.

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -345,7 +345,15 @@ lexemes = P.optional space >> do
         matchWith = openKw "match" <|> close' (Just "match-with") withOpens "with"
         ifElse = openKw "if" <|> close' (Just "then") ["if"] "then" <|> close' (Just "else") ["then"] "else"
         handle = openKw "handle" <|> close' (Just "with") withOpens "with"
-        typ = openKw1 "unique" <|> openKw1 "type" <|> openKw1 "ability"
+        typ = openKw1 "unique" <|> openTypeKw1 "type" <|> openTypeKw1 "ability"
+
+        -- In `unique type` and `unique ability`, only the `unique` opens a layout block,
+        -- and `ability` and `type` are just keywords.
+        openTypeKw1 t = do
+          b <- S.gets (topBlockName . layout)
+          case b of
+            Just "unique" -> wordyKw t
+            _ -> openKw1 t
 
         -- layout keyword which bumps the layout column by 1, rather than looking ahead
         -- to the next token to determine the layout column
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -268,7 +268,18 @@ lexemes = P.optional space >> do
     where
       sep = separated (\c -> isSpace c || not (isAlphaNum c))
       intOrNat = P.try $ num <$> sign <*> LP.decimal
-      float = P.try $ fnum <$> sign <*> LP.float
+      float = P.try $ do
+        sign <- fromMaybe "" <$> optional (lit "+" <|> lit "-")
+        base <- P.takeWhile1P (Just "base") isDigit
+        decimals <- P.optional $
+          liftA2 (<>) (lit ".") (P.takeWhile1P (Just "decimals") isDigit)
+        exp <- P.optional $ do
+          e <- map toLower <$> (lit "e" <|> lit "E")
+          sign <- fromMaybe "" <$> optional (lit "+" <|> lit "-")
+          exp <- P.takeWhile1P (Just "exponent") isDigit
+          pure $ e <> sign <> exp
+        pure $ Numeric (sign <> base <> fromMaybe "" decimals <> fromMaybe "" exp)
+
       bytes = P.try $ do
         _ <- lit "0xs"
         s <- map toLower <$> P.takeWhileP (Just "hexidecimal character") isHex
@@ -283,7 +294,6 @@ lexemes = P.optional space >> do
       num :: Maybe Char -> Integer -> Lexeme
       num sign n = Numeric (signStr sign <> show n)
       signStr = maybe "" (:[])
-      fnum sign n = Numeric (signStr sign <> show @Double n)
       sign = P.optional (char '+' <|> char '-')
 
   hash = Hash <$> P.try shorthash
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -5,7 +5,15 @@
 {-# LANGUAGE DeriveFoldable #-}
 {-# LANGUAGE DeriveTraversable #-}
 
-module Unison.Lexer where
+module Unison.Lexer (
+  Token(..), Line, Column, Err(..), Pos(..), Lexeme(..),
+  lexer, simpleWordyId, simpleSymbolyId,
+  line, column,
+  debugLex', debugLex'', debugLex''', showEscapeChar, touches,
+  -- todo: these probably don't belong here
+  wordyIdChar, wordyIdStartChar,
+  wordyId, symbolyId, wordyId0, symbolyId0)
+  where
 
 import Unison.Prelude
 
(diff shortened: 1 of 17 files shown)


runarorama (approved):

aryairani:
Sounds good.  Agree we should change Name to something structured and then support escaping segments.  And then pull out syntax into an interface :)

pchiusano:
Okay, I think this is all set. 

I did some work to improve the error reporting and added a transcript to demonstrate (error-messages.md). Feel free to add other failing code to that transcript to check for sensible error messages.

@mitchellwrosen you can take a look if you like, otherwise I will merge in a day or two assuming no objections from anyone else.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -21,11 +29,38 @@ import           GHC.Exts (sortWith)
...
+
+data Token a = Token {
+  payload :: a,
+  start :: !Pos,
+  end :: !Pos
+} deriving (Eq, Ord, Show, Functor)
+
+data ParsingEnv =
+  ParsingEnv { layout :: !Layout -- layout stack
+             , opening :: Maybe String } -- `Just b` if a block of type `b` is being opened
```
mitchellwrosen: Comment mentions a type var that looks like it was monomorphized to `String` afterwards.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -60,74 +97,440 @@ type IsVirtual = Bool -- is it a virtual semi or an actual semi?
...
 makePrisms ''Lexeme
 
+space :: P ()
+space = LP.space CP.space1 (fold <|> LP.skipLineComment "--")
+                           (LP.skipBlockCommentNested "{-" "-}")
+  where
+  fold = P.try $ lit "---" *> P.takeRest *> pure ()
+
+lit :: String -> P String
+lit = P.try . LP.symbol (pure ())
```
mitchellwrosen, line 109: Just curious, did you find the `try` here was necessary?

I had thought that e.g. `symbol "foo"` would not advance the parser state unless all of "f", "o", and "o" were matched.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -60,74 +97,440 @@ type IsVirtual = Bool -- is it a virtual semi or an actual semi?
...
+-- successful parse, and also takes care of emitting layout tokens
+-- (such as virtual semicolons and closing tokens).
+token' :: (a -> Pos -> Pos -> [Token Lexeme]) -> P a -> P [Token Lexeme]
+token' tok p = LP.lexeme space (token'' tok p)
+
+-- Committed failure
+err :: Pos -> Err -> P x
+err start t = do
+  stop <- pos
+  _ <- void CP.anyChar <|> P.eof
```
mitchellwrosen, line 131: Do you recall what this line is for? Looks like it's just for adjusting the underlying parser's state, but I'm not totally sure what that would be for if we're reporting a failure. Maybe it's for Megaparsec's multi-error support?

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -60,74 +97,440 @@ type IsVirtual = Bool -- is it a virtual semi or an actual semi?
...
+
+-- Token parser implementation which leaves trailing whitespace and comments
+-- but does emit layout tokens such as virtual semicolons and closing tokens.
+token'' :: (a -> Pos -> Pos -> [Token Lexeme]) -> P a -> P [Token Lexeme]
+token'' tok p = do
+  start <- pos
+  -- We save the current state so we can backtrack the state if `p` fails.
+  env <- S.get
+  layoutToks <- case opening env of
+    -- If we're opening a block of type b, we push (b, currentColumn) onto
```
mitchellwrosen: Here again is a reference to the "block of type b" that looks like it was monomorphised to "String" above. Should this comment be adjusted too or does it still make sense?

parser-typechecker/src/Unison/Lexer.hs
mitchellwrosen: Should that `String` be `BlockName`?

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -34,6 +69,7 @@ data Err
   | InvalidEscapeCharacter Char
   | LayoutError
   | CloseWithoutMatchingOpen String String -- open, close
+  | Opaque String
```
mitchellwrosen: Is this just a convenient catch-all for when you would rather not make up a new data constructor and pipe it through the rest of the code?

parser-typechecker/src/Unison/Lexer.hs
pchiusano: Good catch, yes.

parser-typechecker/src/Unison/Lexer.hs
pchiusano: I wasn't actually sure but I kinda like it better with the `try`, since you could certainly imagine a version of `symbol` that didn't `try`.

parser-typechecker/src/Unison/Lexer.hs
pchiusano: That line consumes a token and therefore produces a committed failure. So you can have things like `try p <|> err s t <|> p2` and it will not backtrack to trying `p`. I'll add a comment to this effect.

parser-typechecker/src/Unison/Lexer.hs
pchiusano: block type just meant like "handle" or "if", that sort of thing, not like the static type of the block. Maybe I'll try to think of a better wording. :)

pchiusano: came up with maybe better wording, said "block name" instead of "block type"

parser-typechecker/src/Unison/Lexer.hs
pchiusano: okay, added a comment on this

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -60,74 +99,442 @@ type IsVirtual = Bool -- is it a virtual semi or an actual semi?
...
+      octal = do start <- pos
+                 commitAfter2 sign (lit "0o") $ \sign _ ->
+                   fmap (num sign) LP.octal <|> err start InvalidOctalLiteral
+      hex = do start <- pos
+               commitAfter2 sign (lit "0x") $ \sign _ ->
+                 fmap (num sign) LP.hexadecimal <|> err start InvalidHexLiteral
+
+      num :: Maybe String -> Integer -> Lexeme
+      num sign n = Numeric (fromMaybe "" sign <> show n)
+      sign = P.optional (lit "+" <|> lit "-")
```
mitchellwrosen, line 397: I think the result of this is always `fromMaybe ""`'d - if so, it could be refactored to
```haskell
sign = lit "+" <|> lit "-" <|> pure ""
```

mitchellwrosen:
LGTM! My main feedback is that the implementation looks a little bit "stringy" - like if you had to get in there and amend something, you might have to lean on a comprehensive test suite rather than the type checker to make sure you've handled all the cases. But for a lexer, that's probably expected :)

Issue #1616, opened 2020-06-17
add mac builds to CI


aryairani:
We want to avoid surprises like the mac build fail in #1611 / #1615.

Maybe useful references:
* https://docs.travis-ci.com/user/multi-os/
* https://docs.travis-ci.com/user/reference/osx/
* #1614 "Add windows to CI"

mitchellwrosen:
Btw, @aryairani, any opinions on GitHub Actions? I was messing around with them the other week on `mitchell/github-actions`, you can see the results here. https://github.com/unisonweb/unison/actions

Mostly, I was interested in the better caching - you can have "fallback" keys, so (for example) commit `A'` can reuse commit `A`'s build artifacts of local packages (not just cached dependencies).

The `ci.yaml` file in that branch isn't done, I think I've learned a couple things since then, but adding windows/mac support is easy. (I'm sure it's easy in Travis, too).

sharno:
@mitchellwrosen Actuall Actions might be a good solution for Windows as well, Travis has an experimental Windows support and it's giving me a weird message in #1614 
> Automatic restarts limited: Please try restarting this job later or contact support@travis-ci.com.

aryairani:
@mitchellwrosen No strong opinion yet.  I'd be interested in understanding the caching thing better?

Side note: I'm not sure it's a good idea to cache `.stack-work`, as it can sometimes become busted, requiring a `stack clean`.  Caching `~/.stack` seems safe, thankfully.

PR #1751, opened 2020-11-22
Add .github/workflows/ci.yaml for GitHub Actions

mitchellwrosen:
## Overview

This PR adds a .github/workflows/ci.yaml file for GitHub actions. I tried to make it perform roughly the same actions as the Travis CI config (build and run tests with both runtimes, then cache stuff).

I cached the local packages' build artifacts in `.stack-work` their directories, in addition to `~/.stack`, which caches dependencies. The intention here is for CI to be similar in speed to locally editing a file and rebuilding - presumably not every local module will need to be recompiled.

However, there may be some downsides to caching `.stack-work` that I haven't thought of, and I didn't measure whether caching it actually produced a measurable speedup.

## Loose ends

Only an Ubuntu build is configured at the moment, but macOS shouldn't be too hard to add. The main reason I didn't add it in this first pass is I had to download `stack` manually as part of the build, because the official `setup-haskell` action is broken right now (it uses a deprecated mechanism for setting environment variables). So, the build explicitly downloads a `linux` flavor of `stack`, but that could switch on the runner's OS instead.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,56 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  build:
+    name: build (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest] # , macOS-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        name: cache ~/.stack
+        with:
+          path: ~/.stack
+          key: x-${{runner.os}}-stack-${{hashFiles('stack.yaml')}}
+      - uses: actions/cache@v2
+        name: cache .stack-work
+        with:
+          path: |
+            .stack-work
+            parser-typechecker/.stack-work
+            unison-core/.stack-work
+            yaks/easytest/.stack-work
+          key: x-${{runner.os}}-stack_work-${{hashFiles('stack.yaml')}}-${{github.sha}}
+          restore-keys: x-${{runner.os}}-stack_work-${{hashFiles('stack.yaml')}}-
+      - name: install stack
+        run: |
+          echo $PATH
+          curl -L https://github.com/commercialhaskell/stack/releases/download/v2.5.1/stack-2.5.1-linux-x86_64.tar.gz | tar -xz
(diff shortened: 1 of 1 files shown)


aryairani:
Re: caching, I'm scared of the situation where CI is succeeding but a clean build fails, because it actually depends on something old in the cache.  I'd rather have CI be slower (within reason) but more reproducible.  Thoughts?

mitchellwrosen:
@aryairani Makes sense to me, but just to clarify, this would be defending against a potential bug in `stack`. Is that correct? (And by "within reason" do you mean we should still cache `~/.stack` but not `.stack-work`?)

pchiusano:
Definitely keep at least `~/.stack` otherwise you're rebuilding all your dependencies every time, right? And builds would take like 2 hours. :)

And I'd also vote to keep `~/.stack-work` except for releases - for releases let's do a clean build. I'm thinking of the day-to-day quality of life improvement of having near instant instant CI results vs the theoretical possibility of a stack caching bug.

I can't recall the last time I've had to do a `stack clean` honestly... so while in principle there could be some bugs with stack's caching, I don't think we should be worrying about that too much for day-to-day work merging to trunk.

aryairani:
Would it be possible to have both of these?
*  a quick, caching CI task that we use for development branches & PRs
*  a long CI task that builds everything from scratch — simulating building as a first-time contributor, or building on a new machine.  This sanity-checks that newer minor releases of dependencies are still compatible, new releases of stack, etc.  But this one wouldn't have to complete before merging a PR, for example.  Basically making sure that anyone who follows the CI steps on their own machine can build the project.

aryairani:
@mitchellwrosen I'd say if you can add the mac build, then let's merge it — we want to avoid surprises like #1616.  The travis.yaml should show how to set up tools for a mac build; or is it missing something?

I'd like to run the build on multiple linuxes, multiple macoses, and maybe even some windows.  Do you know if it's possible to designate certain images to be required successes and others to be optional successes?

pchiusano:
Okay, we're going to merge this and delete the Travis build since Travis has become unusable with hours long wait times and lots of spurious failures, but @mitchellwrosen can you work on adding the macos build at least?

pchiusano (approved):

pchiusano:
Thanks for looking into this @mitchellwrosen !

mitchellwrosen:
@pchiusano, yup, it's on my list of things to do today in fact!

PR #1752, opened 2020-11-23
Avoid parsing erroneous patterns

### LINKED ISSUE #1731: Naming a cases branch variable the same as an ability constructor leads to empty ucm output
My scratch file solely consists of:
```
repro : Text -> ()
repro = cases
  input -> ()
```

If I save, regardless of whether I change anything (except for the name `input`), `ucm` gives me empty output.
Eg. after 3 saves I see:
```
.>
.>
.>
```

If I change the name `input` to anything else, I get the expected "These new definitions are ok to `add`").

I believe this is somehow colliding with the following saved ability, and confusing `ucm`:
```
.> view CLI                                                                                                                                                                                                                    
  ability CLI where
    print : Text ->{CLI} ()
    input : {CLI} Text
```

dolio:
Fixes #1731 

This also includes a new type error, in case we still parse a data pattern with ability references in it, or vice versa. It also contains actual printing for compiler bugs, which was the reason for the completely silent behavior before.

The approach I have so far actually rules out all parsing of erroneous patterns. For instance, if the example in 1731 were changed to explicitly say `CLI.input`, unison would just complain it doesn't know about a data constructor with that name (which is accurate).

This pull request is not ready quite yet, but I opened it to discuss something. There is a test case that fails currently, in `Unison.Test.TermParser` we have:

```
  match x with
    {Tuple.Cons x y} -> 1
    {Optional.Some 42 -> k} -> k 42
```

I'm not sure what this is supposed to mean, since there is no effect being matched on. Was this just a work-around to test the parser without having any ability declared?

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -273,6 +273,11 @@ loop = do
             ppe <- prettyPrintEnv =<< makeShadowedPrintNamesFromHQ hqs errNames
             respond $
               TypeErrors text ppe [ err | Result.TypeError err <- toList notes ]
+            respond . CompilerBugs text ppe
+              $ [ bug
+                | Result.CompilerBug (Result.TypecheckerBug bug)
+                    <- toList notes
+                ]
           Just (Right uf) -> k uf
       loadUnisonFile sourceName text = do
         let lexed = L.lexer (Text.unpack sourceName) (Text.unpack text)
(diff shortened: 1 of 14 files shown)


pchiusano (approved):
LGTM.

pchiusano:
> The approach I have so far actually rules out all parsing of erroneous patterns.

I think it would be nice if this could be a type error rather than a syntax error, but I think you are constrained by the current AST. So your approach seems fine - I'd just delete or alter that failing test, which is just nonsense anyway.

PR #1759, opened 2020-11-26
Fix typechecking of patterns for polymorphic effects

### LINKED ISSUE #693: Typechecker or runtime soundness bug in this example
Atop current unisonbase:

```Haskell
use .base
use .base.Optional
use .Pair
ability Parse where
  anyToken : Nat
  choose : '{Parse} a -> '{Parse} a -> {Parse} a
runTextParser : List Nat -> Request Parse a -> Optional a
runTextParser xs req =
  case req of
    { Parse.anyToken -> next } ->
      case xs of
        [] -> None
        (x +: xs) ->
          handle runTextParser xs in next x
    { Parse.choose l r -> next } ->
      lResult = handle runTextParser xs in !l
      case lResult of
        None -> handle runTextParser xs in !r
        Some x -> Some x
    { a } -> Some a
many : '{Parse} a -> {Parse} [a]
many p =
  unfold () '(choose '(Some (!p, ())) 'None)
bogusValue : Optional [Nat]
bogusValue =
  handle runTextParser [1,2,3] in many 'Parse.anyToken

> bogusValue
```

This returns `Some (Some (1, ()))`.

Thnaks to @zenhack for finding and minimizing this one.

dolio:
Fixes #693 

This actually fixes an additional problem I noticed while experimenting with the 693 example, which is that code like the following would work:

```
ability A t where
  a : t -> r

h : Request {A t} u -> u
h = cases
  { a x -> _ } -> x
  { y } -> y
```

because nothing was ensuring that `x` had type `t`, and its type was just allowed to unify with `u`. This case is covered in the test transcript.

Skolemization is done properly for data types as well, but I don't think it's possible to create a type where this matters yet.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -1215,6 +1215,13 @@ checkPattern scrutineeType0 p =
         -- an effect ctor should have exactly 1 effect!
         Type.Effect'' [et] it -> do
           -- expecting scrutineeType to be `Effect et vt`
+
+          -- ensure that the variables in `et` unify with those from
+          -- the scrutinee.
+          lift $ do
+            res <- Type.flattenEffects <$> applyM (existentialp loc e)
+            abilityCheck' res [et]
+
           st <- lift $ applyM scrutineeType
           case st of
             Type.App' _ vt ->
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -1160,7 +1160,7 @@ checkPattern scrutineeType0 p =
       lift $ subtype (Type.char loc) scrutineeType $> mempty
     Pattern.Constructor loc ref cid args -> do
       dct  <- lift $ getDataConstructorType ref cid
-      udct <- lift $ ungeneralize dct
+      udct <- lift $ skolemize forcedData dct
       unless (Type.arity udct == length args)
         . lift
         . failWith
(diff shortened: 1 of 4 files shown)


parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -1369,6 +1376,35 @@ ungeneralize' (Type.Forall' t) = do
...
+  appendContext $ uctx ++ sctx
+  pure $ foldl (flip $ uncurry ABT.substInheritAnnotation) ty rn
+  where
+  fovs = forced ty
+  (uvs, svs) = partition (`Set.member` fovs) vs
+skolemize _ ty = pure ty
+
+forcedEffect :: Type v loc -> Set (TypeVar v loc)
+forcedEffect (Type.Arrow' _ o) = forcedEffect o
+forcedEffect (Type.Effect1' es _) = Type.freeVars es
```
pchiusano, line 1401: I was thinking this wasn't quite right if arrows other than the final one have effects, but that shouldn't be a valid ability declaration (though we don't enforce that).

pchiusano (approved):
LGTM

parser-typechecker/src/Unison/Typechecker/Context.hs
dolio: Is that because `X ->{E} Y ->{F} Z` is represented like `Arrow X (Effect E (Arrow Y (Effect F Z)))`?

pchiusano:
Mac build randomly failed again..

parser-typechecker/src/Unison/Typechecker/Context.hs
pchiusano: Right

dolio: So, I was thinking about this randomly. My question is: how is this actually supposed to work? `X ->{E} Y ->{F} Z` is *not* the type of an `F` effect. It is of an `E` effect that produces a function with an `F` effect. If it is being put in the `F` ability, then that is wrong. The type checker might not prevent someone from writing it that way, but that means it is allowing something that doesn't make sense, I think.

pchiusano: Right. It doesn't make sense. The typechecker should probably give a nice error for this.

PR #1760, opened 2020-11-27
Format the codebase with ormolu

### LINKED ISSUE #1653: Auto-format codebase
When Unison started none of the Haskell auto-formatters were that amazing. So I understand why it's hand-formatted.

However, the latest generation of formatters are really quite good. I find [ormolu](https://hackage.haskell.org/package/ormolu) especially nice, though it doesn't really matter which one you pick. Sure, sometimes they make ugly decisions, but in return you get:

**(a)** A faster coding experience (no hand-formatting imports!) and

**(b)** consistency (whether you like the decisions or not, at least it's the same everywhere)

Is this something the core Unison devs are interested in?

PS: If you do go with ormolu you can still hand-format specific parts of code by wrapping it with `{- ORMOLU_DISABLE -}` and `{- ORMOLU_ENABLE -}`. This is occasionally helpful for things like enormous pattern matches where hand-formatting the columns can help readability.

mitchellwrosen:
## Overview

This PR formats the entire codebase with ormolu 0.1.4.1. It also adds a `./scripts/ormolu` file which can be run as follows:

To format all .hs files in the repo,
```
./scripts/ormolu.sh
```

To check that all .hs files in the repo are formatted,
```
./scripts/ormolu.sh check
```

Fixes #1653 

I removed `TypeApplications` from the list of default extensions in `unison-parser-typechecker.cabal`. Where necessary, `{-# LANGUAGE TypeApplications #-}` pragmas were added instead. This was done because `ormolu` does not enable any syntax-stealing extensions by default, so it only works out of the box on modules that enable such extensions explicitly.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -89,6 +89,16 @@ jobs:
           curl -L https://github.com/commercialhaskell/stack/releases/download/v2.5.1/stack-2.5.1-osx-x86_64.tar.gz | tar -xz
           echo "$HOME/stack-2.5.1-osx-x86_64/" >> $GITHUB_PATH
 
+      # install ormolu
+      - name: install ormolu (Linux)
+        if: runner.os == 'Linux'
+        run: curl -L https://github.com/tweag/ormolu/releases/download/0.1.4.1/ormolu-Linux --output "$HOME/.local/bin/ormolu"
+
+      # run ormolu
+      - name: run ormolu (Linux)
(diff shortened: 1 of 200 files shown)


mitchellwrosen:
I'm going to close this for now, I did get the go-ahead in Slack to put up this PR, but there was never really a convenient time to pull the trigger on it, I guess. Let me know if I should re-open at some point :)

PR #1761, opened 2020-11-27
Fix caching issue in CI that affects macOS

mitchellwrosen:
## Overview

This patch does two things:
- Use `ghcup` to install GHC, rather than relying on `stack`'s ability to do so
- Delete a folder in `~/.stack` to fix a weird caching issue that only affects macOS builds

The first change is unrelated to the bug; I only implemented it as a first attempt to diagnose the bug, but it turned out not to be necessary. However, I figured it doesn't hurt to include, as it may be helpful some day for installing GHC for a `cabal`-based build, which of course can't install GHC by itself as `stack` can.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,10 +23,21 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      # The random number towards the beginning of the cache keys below are meant to be bumped as a crude means to clear
+      # a cache. GitHub will automatically delete caches that haven't been accessed in 7 days, but there is no way to
+      # purge one manually.
+
+      # Cache ghc installed by ghcup
+      - uses: actions/cache@v2
+        name: cache ghc installed by ghcup
+        id: cache-ghc
+        with:
+          path: |
+            ~/.ghcup/bin/ghc-8.8.3
+            ~/.ghcup/ghc/8.8.3/bin/ghc
+          key: ghcup-0_${{matrix.os}}
+
       # Cache ~/.stack, keyed by the contents of 'stack.yaml'.
-      # The "0" in "stack-0" is meant to be bumped as a crude means to clear a cache, in case it's misbehaving.
-      # GitHub will automatically delete caches that haven't been accessed in 7 days, but there is no way to purge one
-      # manually.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -43,10 +53,28 @@ jobs:
             parser-typechecker/.stack-work
             unison-core/.stack-work
             yaks/easytest/.stack-work
-          # Main hash key: the branch + commit hash. This should always result in a cache miss...
-          key: stack-work-0_${{matrix.os}}-${{github.ref}}-${{github.sha}}
-          # ...but then fall back on the latest cache stored on that branch.
-          restore-keys: stack-work-0_${{matrix.os}}-${{github.ref}}-
+            # Main cache key: commit hash. This should always result in a cache miss...
+          key: stack-work-0_${{matrix.os}}-${{github.sha}}
```
mitchellwrosen, line 57: I changed the caching key here because I learned from the docs of the `cache` action that it's already branch-based, so there's no reason to include `${{github.ref}}`.

.github/workflows/ci.yaml
```diff
@@ -65,25 +93,29 @@ jobs:
       - name: set git username
         run: git config --global user.name "GitHub Actions"
 
+      - name: remove ~/.stack/setup-exe-cache on macOS
```
mitchellwrosen, line 96: Here's the actual fix to the "cannot execute binary file" bug

.github/workflows/ci.yaml
```diff
@@ -65,25 +93,29 @@ jobs:
...
+        if: runner.os == 'macOS'
+        run: rm -rf ~/.stack/setup-exe-cache
+
       # Build deps, then build local code. Splitting it into two steps just allows us to see how much time each step
       # takes.
       - name: build dependencies
-        run: stack --no-terminal build --fast --only-dependencies
+        run: stack --no-terminal --system-ghc build --fast --only-dependencies
```
mitchellwrosen, line 103: Use `--system-ghc` because now GHC is installed by `ghcup`

pchiusano (approved):
Great, thanks!

PR #1763, opened 2020-12-02
Introduce TLS builtins, Failure type, io.test

stew:
## Overview

This PR introduces TLS builtins to enable the ability to promote TCP Handles to negotiate TLS (SSL) connections.

In the process of implementing these bultins, we made a decision to introduce a Failure type which will be used in the builtins which might result in an IO failure, and it reimplements the `io2.IO` builtin functions to return a `Failure` as a result if an IO exception.

The IO builtins were mostly untested and have been buggy in the past. We haven't in the past had a good way of testing functions that require IO. This PR also adds a new verb to ucm: `io.test`. It will allow you to run a named test that has the type: `() -> {IO} [Result]`. With this we can include tests requiring IO in transcripts, and failures in the IO tests will fail the transcripts. Unlike regular tests, the results from running a test are _not_ cached

## Implementation notes

I made a big refactoring to how most of the IO / MVar functions are implemented, trying to make the implementations more generic and more readable.

We believe that we should enhance the `test` and `io.test` so that the *both* are able to either run all known tests or to run a subset of tests, but for now, we only added the ability for `io.test` to run a single test. We also think that it will be useful to 

## Test coverage

This includes new tests for the many but not all of the IO / MVar functions. We should be implementing the rest.

## Loose ends

This is only the beginning for TLS, as this does not yet include tests for the TLS functions. Those will come later, we have to enhance our TLS implementation before it is easily testable to allow for things like accepting self signed certs or certs from alternative CAs. Those enhancements and tests would come in a subsequent PR. we wanted to get this PR process started so that we don't have this branch which its large changes to the new runtime being long lived.


parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -433,8 +438,19 @@ builtinsSrc =
...
+  where
+    newName :: Text -> Text
+    newName name = newName' name $ RE.matchOnceText regex name
+
+    newName' _ (Just (before, _, _)) = prefix <> "." <> before
+    newName' name Nothing = prefix <> "." <> name
+    r :: String
+    r = "\\.v[0-9]+"
+    regex :: RE.Regex
+    regex = RE.makeRegexOpts (RE.defaultCompOpt { RE.caseSensitive = False }) RE.defaultExecOpt r
```
pchiusano: I think stripping out the `v2` doesn't belong in this function. I'd keep this function the same as it was, and then add a `stripVersionNumber` function that goes from `[BuiltinDSL v] -> [BuiltinDSL v]`.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -1320,6 +1320,22 @@ execute = InputPattern
     _   -> Left $ showPatternHelp execute
   )
 
+ioTest :: InputPattern
+ioTest = InputPattern
+  "io.test"
+  []
+  []
+  (P.wrapColumn2
+    [ ( "`io.test mystest`"
```
pchiusano: typo - `mystest`
```suggestion
    [ ( "`io.test mytest`"
```

unison-src/new-runtime-transcripts/utf8.md
```diff
@@ -36,7 +37,7 @@ We can check that encoding and then decoding should give us back the same `Text`
 checkRoundTrip: Text -> [Result]
 checkRoundTrip t = 
   bytes = toUtf8 t
-  match fromUtf8 bytes with 
+  match Text.fromUtf8 bytes with
```
pchiusano: Just wondering if this change was needed for some reason... it shouldn't be I don't think, unless the old `fromUtf8` is still in the starting namespace.

pchiusano (approved):
Nice! Left a few minor comments. Appreciate the cleanup in `Runtime.Builtin` and that we now have actual tests for the I/O stuff. :)

I'd like to give `Failure` an `Any` argument in addition to the failure type and message, but debating whether to address that separately. `Any` provides `Any : a -> Any`, basically an existential value. You can't do anything with it, but it saves you from having to do a bunch of text munging just to include some sort more informative payload with the failure. It's also nice in that when it is printed out by ucm it will use the latest names for things.

If we address adding `Any` with a separate PR we should then bump the `v2` suffix to `v3` just to be safe at that point. If bumping v2 to v3 is nbd then let's just do a separate PR. If it's super annoying, let's try to address now or figure out a way to make it less annoying.

```Haskell
type Any = Existential (forall e r . (forall a . a ->{e} r) ->{e} r) 

Any : a -> Any
Any a = Any.Any (k -> k a)
```

PR #1767, opened 2020-12-08
Fix a potential loop during type checking

### LINKED ISSUE #1695: Typechecker hang in this example
Possibly same as or related to #861 

```
unique ability S a where
  get : a 
  put : a -> ()

step : a -> '{S (a -> b)} b
step in _ = 
  -- x : a -> b  -- adding this signature causes it to work
  x = S.get
  x in
```

This hangs during typechecking and leaks memory.

dolio:
- `instantiateR` assumes that it will never be called in a situation
  that would create a cyclic type. However, `abilityCheck'` was doing
  so in some situations. Instead, move immediately to a fall-back
  behavior in such scenarios.

Fixes #1695


pchiusano:
LGTM but can you add a regression test for this? Could just be that file that previously looped, just add it to unison-src/tests/fix1695.u

pchiusano (approved):

pchiusano:
Great! Merging.

PR #1769, opened 2020-12-11
Implement crypto.hash and crypto.hmac

### LINKED ISSUE #1714: Universal hashing and HMAC implementations
These are stubbed out and have type signatures but are currently unimplemented:

```Haskell
hash : HashAlgorithm -> a -> Bytes
hmac : HashAlgorithm -> Bytes -> a -> Bytes
```

I wanted to do some refactoring of how foreign data types are handled before implementing this, and also it needs to have a story for how continuations are hashed (which is probably similar to how continuations get serialized). I thought it was out of scope for #1712 so punted it to this issue.

dolio:
This uses the serialization operation to get a bytes suitable for hashing from an arbitrary value. The implementations are otherwise quite similar to `hashBytes` and `hmacBytes`. The serialization is done to a lazy byte string so that cases where the value is large can be chunked.

Serialization for some missing builtin 'foreign' values has been implemented (text, lists, ...), since otherwise they wouldn't be hashable.

Examples have been added to the transcripts. I only added one call to each, since they're entirely uniform except for the output of the serialization, which can be tested separately. There's also no standard list of reference tests for unison values like there is for the raw bytes.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/ANF.hs
+++ b/parser-typechecker/src/Unison/Runtime/ANF.hs
@@ -52,6 +52,7 @@ module Unison.Runtime.ANF
   , GroupRef(..)
   , Value(..)
   , Cont(..)
+  , BLit(..)
   , packTags
   , unpackTags
   , ANFM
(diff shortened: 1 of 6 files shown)


pchiusano (approved):
Looks good.

PR #1770, opened 2020-12-11
`Any` builtin type

pchiusano:
This adds a new type, `Any`, with a function `Any : forall a . a -> Any`. Its main purpose is for things like logging and error payloads: you want to log a value but don't want to have to do a bunch of string munging to convert that value to something printable. `Any` gives you a way to dump an arbitrary value to an untyped representation that can still be intelligently printed out by UCM later, using whatever names you have in scope in your codebase. 

The plan is to start using this for the basic exception type, and I'd like to switch the test result type to use `Any` for failures. It will also get used in the new documentation type, to embed arbitrary expressions and live examples.

While you can certainly write things like `[Any "hi", Any 42, Any ("a", "b")]`, you can't compute with the values in the list as they are existential and you don't have access to runtime type info.

An `Any` type could also be defined using pure Unison code, like so: `type Any = Any (forall e r . (forall a . (a ->{e} r) ->{e} r))` but values of that type are pretty annoying to produce from Haskell code (like inside `IO` functions), so I went with making it an opaque builtin.

Added a simple transcript test to `builtin.md`.


pchiusano:
@dolio tagged you just to take a look at the diff on `universalCompare`.

@stew I'd like to integrate this into the `Failure` type as we discussed.

dolio (approved):
Looks fine to me.

atacratic:
@pchiusano This worked for me with the new runtime, but with the old runtime when I do 

`> .builtin.Any.Any 3`

I get 

```
The program being compiled referenced this definition ##Any.Any
```
Intended?  Shout and I can raise an issue.

pchiusano:
It's intended, in the sense that we've stopped trying to backport all new builtins to the old runtime and the next release will get rid of the old runtime. 😀

PR #1777, opened 2020-12-22
update to ghc 8.10.3

aryairani:
## Loose ends

One of the hashes displayed by a transcript changed in an unexpected way.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/IR.hs
+++ b/parser-typechecker/src/Unison/Runtime/IR.hs
@@ -595,8 +595,6 @@ compile0 env bound t =
                                   "This code should never be reached as this constraint is " <>
                                   "applied in the typechecker."
 
-        _ -> error $ "todo - compilePattern " ++ show pat
-
 type DS = StateT (Map Symbol (Term Symbol), Set RefID) IO
 
 runDS :: DS (Term Symbol) -> IO (Term Symbol)
--- a/parser-typechecker/src/Unison/Runtime/MCode.hs
+++ b/parser-typechecker/src/Unison/Runtime/MCode.hs
@@ -894,13 +894,13 @@ emitLet
   -> Emit Section
 emitLet _   _   _   _ _   _   (TLit l)
   = fmap (Ins $ emitLit l)
-emitLet rns grp _   _ _   ctx (TApp (FComb r) args)
-  -- We should be able to tell if we are making a saturated call
-  -- or not here. We aren't carrying the information here yet, though.
-  | False -- not saturated
-  = fmap (Ins . Name (Env n 0) $ emitArgs grp ctx args)
-  where
-  n = cnum rns r
+-- emitLet rns grp _   _ _   ctx (TApp (FComb r) args)
+--   -- We should be able to tell if we are making a saturated call
+--   -- or not here. We aren't carrying the information here yet, though.
+--   | False -- not saturated
+--   = fmap (Ins . Name (Env n 0) $ emitArgs grp ctx args)
+--   where
+--   n = cnum rns r
 emitLet _   grp _   _ _   ctx (TApp (FCon r n) args)
   = fmap (Ins . Pack r (rawTag n) $ emitArgs grp ctx args)
 emitLet _   grp _   _ _   ctx (TApp (FPrim p) args)
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -804,7 +804,6 @@ suffixCounterTerm :: Var v => PrettyPrintEnv -> Term2 v at ap v a -> PrintAnnota
 suffixCounterTerm n = \case
     Var' v -> countHQ $ HQ.unsafeFromVar v
     Ref' r -> countHQ $ PrettyPrintEnv.termName n (Referent.Ref r)
-    Ref' r -> countHQ $ PrettyPrintEnv.termName n (Referent.Ref r)
     Constructor' r _ | noImportRefs r -> mempty
     Constructor' r i -> countHQ $ PrettyPrintEnv.termName n (Referent.Con r i CT.Data)
     Request' r i -> countHQ $ PrettyPrintEnv.termName n (Referent.Con r i CT.Effect)
@@ -1189,4 +1188,3 @@ toBytes (App' (Builtin' "Bytes.fromList") (Sequence' bs)) =
   where go (Nat' n) = Just n
         go _ = Nothing
 toBytes _ = Nothing
-
--- a/parser-typechecker/src/Unison/Util/AnnotatedText.hs
+++ b/parser-typechecker/src/Unison/Util/AnnotatedText.hs
@@ -18,6 +18,7 @@ import           Unison.Lexer       (Line, Pos (..))
 import           Unison.Util.Monoid (intercalateMap)
 import           Unison.Util.Range  (Range (..), inRange)
 import qualified Data.ListLike      as LL
+import qualified GHC.Exts
 
 -- type AnnotatedText a = AnnotatedText (Maybe a)
 
@@ -197,3 +198,8 @@ instance IsString (AnnotatedText a) where
 
 instance IsString (AnnotatedExcerpt a) where
   fromString s = AnnotatedExcerpt 1 s mempty
+
+instance GHC.Exts.IsList (AnnotatedText a) where
+  type Item (AnnotatedText a) = Char
+  fromList s = fromString s
+  toList (AnnotatedText s) = join . Foldable.toList $ fmap fst s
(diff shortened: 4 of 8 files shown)


parser-typechecker/src/Unison/Runtime/IR.hs
```diff
@@ -595,8 +595,6 @@ compile0 env bound t =
                                   "This code should never be reached as this constraint is " <>
                                   "applied in the typechecker."
 
-        _ -> error $ "todo - compilePattern " ++ show pat
```
aryairani, line 598: 8.10 no longer needs or wants this catch-all.

parser-typechecker/src/Unison/Runtime/MCode.hs
```diff
@@ -894,13 +894,13 @@ emitLet
   -> Emit Section
 emitLet _   _   _   _ _   _   (TLit l)
   = fmap (Ins $ emitLit l)
-emitLet rns grp _   _ _   ctx (TApp (FComb r) args)
-  -- We should be able to tell if we are making a saturated call
-  -- or not here. We aren't carrying the information here yet, though.
-  | False -- not saturated
```
aryairani, line 900: We can no longer use a `| False` pattern guard to fake out the dead code detector.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -804,7 +804,6 @@ suffixCounterTerm :: Var v => PrettyPrintEnv -> Term2 v at ap v a -> PrintAnnota
 suffixCounterTerm n = \case
     Var' v -> countHQ $ HQ.unsafeFromVar v
     Ref' r -> countHQ $ PrettyPrintEnv.termName n (Referent.Ref r)
-    Ref' r -> countHQ $ PrettyPrintEnv.termName n (Referent.Ref r)
```
aryairani, line 807: Duplicate case

parser-typechecker/src/Unison/Util/AnnotatedText.hs
```diff
@@ -197,3 +198,8 @@ instance IsString (AnnotatedText a) where
 
 instance IsString (AnnotatedExcerpt a) where
   fromString s = AnnotatedExcerpt 1 s mempty
+
+instance GHC.Exts.IsList (AnnotatedText a) where
```
aryairani, line 202: `ListLike` now requires `IsList`.

unison-core/src/Unison/Name.hs
```diff
@@ -57,15 +57,15 @@ sortNamed by = sortByText (toText . by)
...
   comp (_,s) (_,s2) = RFC5051.compareUnicode s s2
   in fst <$> sortBy comp as'
 
 -- | Like sortNamed, but takes an additional backup comparison function if two
 -- names are equal.
 sortNamed' :: (a -> Name) -> (a -> a -> Ordering) -> [a] -> [a]
 sortNamed' by by2 as = let
-  as' = [ (a, Text.unpack (toText (by a))) | a <- as ]
+  as' = [ (a, toText (by a)) | a <- as ]
   comp (a,s) (a2,s2) = RFC5051.compareUnicode s s2 <> by2 a a2
```
aryairani, line 69: `RFC5051` package now takes `Text` instead of `String`

unison-src/new-runtime-transcripts/hashing.output.md
```diff
@@ -137,11 +137,11 @@ ex5 = crypto.hmac Sha2_256 mysecret f |> hex
   
     25 | > ex4
            ⧩
-           "30eb272078faa05fec07601353cb007c90977778965bb31d0c097c7191deb22f"
+           "43cccb245b4b50964e71dff9a53c330f38320673252ab9110e4bac8c3a965354"
   
     26 | > ex5
            ⧩
-           "978e69198db0ff7b2b17a976a681eceae26d5ed5642f96987b5ee7b10ca01274"
+           "8f8785a7442b8363df1f2ed560f6ea42132a63824984b531a23187ee94c110c4"
```
aryairani, line 144: Why?

parser-typechecker/src/Unison/Runtime/MCode.hs
mitchellwrosen: `not True` probably works :)

unison-src/new-runtime-transcripts/hashing.output.md
pchiusano: Would like to understand this one.

pchiusano (commented):
I'd like to understand the hashing output before merging. Are any other hashes affected? Like the worst thing would be if this silently changes all the hashing we are doing with cryptonite everywhere, including for hashing terms and types, etc.

Also there are a couple Mac CI targets failing, do you know what that is?

unison-src/new-runtime-transcripts/hashing.output.md
pchiusano: Okay, this isn't affecting term or type hashing since that would have changed other transcripts.

@dolio this looks to be a change to the output of universal hashing - it is hashing the identity function in both cases I believe. My guess is that there is something about the implementation of universal hashing that depends on some implementation detail of one of the libraries we are using, and that implementation detail changed between library versions. Any ideas on what that might be?

We want the hashing algorithm to be stable and not change unintentionally, so I'd like to get to the bottom of this.

parser-typechecker/src/Unison/Runtime/MCode.hs
aryairani: Hehe I thought about doing that; I commented it out instead.

unison-src/new-runtime-transcripts/hashing.output.md
aryairani: > I'd like to understand the hashing output before merging. Are any other hashes affected? Like the worst thing would be if this silently changes all the hashing we are doing with cryptonite everywhere, including for hashing terms and types, etc.

Agreed.

> Also there are a couple Mac CI targets failing, do you know what that is?

It's failing to build in Big Sur on GitHub CI; it doesn't even start the VM.
> "GitHub Actions has encountered an internal error when running your job."

It does, however, build with `stack` on Big Sur, so if we can find a satisfactory answer to the transcript/hash thing, I'd just revert the last two CI-related commits and deal with it in a separate ticket.

aryairani:
> I'd like to understand the hashing output before merging. Are any other hashes affected? Like the worst thing would be if this silently changes all the hashing we are doing with cryptonite everywhere, including for hashing terms and types, etc.

Agreed.

> Also there are a couple Mac CI targets failing, do you know what that is?

Some Github misconfiguration; I reverted the attempt to build on the Big Sur image, and will deal with it in a separate 
ticket.

unison-src/new-runtime-transcripts/hashing.output.md
dolio: The serialization output seems to be changing slightly every time I restart ucm with the watch cache disabled, so something funky is going on.

dolio: Okay, I think I figured out what's going on. The translation of combinator references from one level to another is just like `f (CIx r m n) = GR r m n`. However, this makes no sense, because one of those numbers (`m`) corresponds to the locally allocated number for the combinator. So it should be thrown out when serializing and figured out from the reference when deserializing.

pchiusano: @dolio Good sleuthing. So basically, the issue is present in trunk, it's just that you wouldn't notice it unless you rerun the transcripts with the watch cache disabled.

I was trying to figure out why CI passes but this might be the first PR merged since the universal hashing landed. Also possible that the transcript runner is still deterministic since even if the hashing depends on implementation details, it is always running the transcripts in the exact same order.

In any case, @dolio can you create an issue to track and @aryairani I think it is fine to merge this now.

pchiusano (approved):

unison-src/new-runtime-transcripts/hashing.output.md
pchiusano: Also I thought of an idea for when you fix this @dolio - randomize the order that transcripts are run in the test harness - I'm pretty sure it just takes a directory listing so you can just shuffle that before iterating over it. It shouldn't affect any of the transcript output and might catch some issues like this.

dolio: Will that do anything? I thought every transcript was run in its own execution, essentially, so they're all in separate runtime instances. Randomizing the order would only matter for this if multiple transcripts were getting loaded in a single runtime environment.

I'm not really sure why we get nondeterminstic results here, even. I guess a bunch of definitions from a file can end up represented in slightly different orders or something.

pchiusano: Oh... I was under the impression that the same runtime instance gets recycled for each of the transcripts, but I could be wrong about that.

dolio: I think the transcript tests just system call out to `unison transcript whatever.md` for each file.

PR #1785, opened 2021-01-08
Add more comprehensive serialization tests

dolio:
This adds IO-based tests to the serialization transcript, to verify that round-trip through serialization actually yields the same result.

It also adds serialization for `Bytes` values, since that was missing.

It also fixes a problem where the serialization format was including session-local information, which caused the serialization and thus hashing to be non-deterministic.


pchiusano (approved):
Cool. 👍 To fix the conflict, you might have to merge trunk into this branch then rerun the transcripts.

PR #1786, opened 2021-01-09
Add a codebase server

runarorama:
## Overview

This adds a server to UCM, listening on port 8081. The server conforms to an OpenAPI schema which it serves at `/openapi.json`.

There are currently two endpoints: `/list` which lists the contents of a namespace, and `/getDefinition` which returns the definition of a type or a term.

## Implementation notes

A bunch of code was moved from `HandleInput.hs` and friends into a new package called `Unison.Server`. The module `Unison.Server.Backend` is supposed to contain shared code that's called from both the HTTP server endpoints and the UCM command line tool. The idea is that we will ultimately have a bunch of front-ends that all call into this server backend.

## Test coverage

It would be beneficial to add integration tests to check that the HTTP server behaves according to its specification. Since the server is actually just a bunch of Haskell functions, these could be tested without firing up the server.

The server backend is well covered by existing transcripts.

## Loose ends

I had to move the `mvar.md` transcript, since it was failing. Will open a separate ticket for this.

We need to add more endpoints to make this server more useful.


unison-src/new-runtime-transcripts/tls.output.md
```diff
@@ -130,7 +130,7 @@ test> match (decodeCert (toUtf8 self_signed_cert_pem) with
   
     ⍟ These new definitions are ok to `add`:
     
-    test.ckc3ihvvem (Unison bug, unknown term)
+    test.faj5ui4d7r (Unison bug, unknown term)
```
pchiusano: For an unnamed test watch, it just makes up a GUID for the test name. I don't think we should rely on that in the transcripts, should use named tests. I'm guessing this difference is just due to the GUID generation not being sufficiently predictable, so I wouldn't worry about it. (I would fix the transcript to use a named test though)

unison-src/new-runtime-transcripts/utf8.output.md
```diff
@@ -90,7 +90,7 @@ test> checkRoundTrip greek
     
       checkRoundTrip : Text -> [Result]
       greek          : Text
-    test.kqfpde2g5a  (Unison bug, unknown term)
+    test.tiucpnku7r  (Unison bug, unknown term)
```
pchiusano: Same deal as [this comment](https://github.com/unisonweb/unison/pull/1786/files#r567961473)

unison-src/transcripts/emptyCodebase.output.md
```diff
@@ -9,33 +9,6 @@ BEHOLD!!!
...
-
-.foo> ls
-
-  1. builtin/ (491 definitions)
-
-```
-More typically, you'd start out by pulling `base.
+  ⚠️
+  
+  The namespace . doesn't exist.
```
pchiusano: This one is weird - the transcript just [cuts out](https://raw.githubusercontent.com/unisonweb/unison/454fb392e81e5423cbce654931aef1956ce8f642/unison-src/transcripts/emptyCodebase.output.md). It doesn't even close the backticks of [that first output block](https://raw.githubusercontent.com/unisonweb/unison/454fb392e81e5423cbce654931aef1956ce8f642/unison-src/transcripts/emptyCodebase.md). 

@runarorama If I had to guess, it looks like the transcript marks that first block as `ucm:error`, but the new output for `ls` of an empty namespace isn't considered an error. You can look at `Codebase.Editor.Output`, the `isFailure` function, and make sure that `ls` of an empty namespace is considered an error by your branch.

unison-src/transcripts/propagate.output.md
```diff
@@ -37,13 +37,13 @@ And then we add it.
 
 .subpath> find.verbose
 
-  1. -- #qae64o6am81hoadf7eabd909gojboi5iu3g9deip79ro18f11bbhir2vg51grg4m72kr5ikdovi6aupttet0nsqil7f0df9nqr10hqg
+  1. -- #fnvv5qd8qafiki64lnsuggdiuattl1tnso1la5ko2b25alvir0bt5jqpksj45qf3u1ckj7qpnps26rlq45dsvaeb98es5q4ike1fct8
```
pchiusano: Not sure on this one, but I'd maybe come back to it after making the GUID generation more predictable?

unison-src/transcripts/hello.output.md
```diff
@@ -83,10 +83,15 @@ hmm = "Not, in fact, a number"
...
-      1 | hmm : .builtin.Nat
-      2 | hmm = "Not, in fact, a number"
+    ❓
+    
+    I couldn't resolve any of these symbols:
+    
+        1 | hmm : .builtin.Nat
+    
+    Using these fully qualified names: .builtin.Nat
+
```
pchiusano: @runarorama This is indicative of a problem, I think - the transcript calls `builtins.merge` at the start so it should be finding `builtin.Nat`. 

To debug, you can add a `.builtin> find` or `.builtin> ls` after the `builtins.merge`, just to see what the namespace looks like.

runarorama: This looks like an issue with name lookup in general of FQNs with absolute paths. E.g. if `.base.Nat` exists, it doesn't find it, but it does find `base.Nat`.

pchiusano:
Nice!!

This is a mammoth PR and I feel like we would benefit from doing a synchronous walkthrough together.

@runarorama could you also do a self review of the diff for this and make notes on any changes that you think are worth calling out (and if your explanation doesn't pertain to the diff, you can add it instead as a comment in the actual code). Then we can take a look through and schedule a review session, maybe tomorrow morning?

unison-core/src/Unison/HashQualified'.hs
```diff
@@ -17,110 +18,108 @@ import           Unison.ShortHash               ( ShortHash )
 import qualified Unison.ShortHash              as SH
 import qualified Unison.HashQualified          as HQ
 
-data HashQualified' n = NameOnly n | HashQualified n ShortHash
-  deriving (Eq, Functor)
+data HashQualified n = NameOnly n | HashQualified n ShortHash
```
runarorama, line 21: The changes to `HashQualified` here are just to make the code a little easier to read. The `HashQualified` type was just an alias for `HashQualified' Name`, so I've removed the `'` from that type and removed the alias.

In particular I wanted to use `HashQualified NameSegment` in some places, and I had a hard time keeping a handle on what all the `'` were for.

parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -57,8 +70,28 @@ type TypecheckingResult v =
          (Either Names0 (UF.TypecheckedUnisonFile v Ann))
 
 data Command m i v a where
+  -- Escape hatch.
   Eval :: m a -> Command m i v a
 
+  HQNameQuery
```
runarorama, line 76: These commands serve to move some of the backend logic from `HandleInput` into `HandleCommand`, so we can call into `Backend` from those commands. `HandleInput` doesn't have access to a `Codebase` and so it can't call the `Backend`.

parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -190,3 +223,52 @@ data Command m i v a where
 
   RuntimeMain :: Command m i v (Type v Ann)
   RuntimeTest :: Command m i v (Type v Ann)
+
+commandName :: Command m i v a -> String
```
runarorama, line 227: Purely a debug tool. This is useful in case you want to print out a trace of all the commands.

parser-typechecker/src/Unison/Codebase/Editor/DisplayObject.hs
```diff
@@ -0,0 +1,15 @@
+{-# LANGUAGE DeriveGeneric #-}
+
+module Unison.Codebase.Editor.DisplayObject where
+
+import Unison.Prelude
+import Unison.ShortHash
+
+data DisplayObject a = BuiltinObject | MissingObject ShortHash | UserObject a
```
runarorama, line 8: Moved from `DisplayThing`, which was a dumb name we came up with in a moment of sleep-deprived mania.

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -144,24 +141,11 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
     GetTermsMentioningType ty -> Codebase.termsMentioningType codebase ty
     CodebaseHashLength -> Codebase.hashLength codebase
     -- all builtin and derived type references
-    TypeReferencesByShortHash sh -> do
-      fromCodebase <- Codebase.typeReferencesByPrefix codebase sh
-      let fromBuiltins = Set.filter (\r -> sh == Reference.toShortHash r)
-            $ B.intrinsicTypeReferences
-      pure (fromBuiltins <> Set.map Reference.DerivedId fromCodebase)
+    TypeReferencesByShortHash sh -> Backend.typeReferencesByShortHash codebase sh
```
runarorama: Eventually we want most (all?) commands to delegate to `Backend`.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -100,14 +102,13 @@ prefix (Absolute (Path prefix)) (Path' p) = case p of
...
 parsePath' p = case parsePathImpl' p of
-  Left  e       -> Left e
-  Right (p, "") -> Right p
-  Right (p, rem) ->
-    case (first show . (Lexer.wordyId0 <> Lexer.symbolyId0) <> unit') rem of
-      Right (seg, "") -> Right (unsplit' (p, NameSegment . Text.pack $ seg))
-      Right (_, rem) ->
-        Left ("extra characters after " <> show p <> ": " <> show rem)
-      Left e -> Left e
+  Left  e        -> Left e
```
runarorama, line 105: This should be replaced with Megaparsec sometime. The implementation is hard to understand.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -131,7 +132,15 @@ parsePathImpl' p = case p of
     Right (segs, rem) ->
       Left $ "extra characters after " <> segs <> ": " <> show rem
     Left e -> Left e
-  segs p = go (first show . (Lexer.symbolyId <> Lexer.wordyId) <> unit') p
+  segs p = go parseSegment p
+
+parseSegment :: String -> Either String (String, String)
```
runarorama, line 137: This function adds better error reporting and a little code re-use.

parser-typechecker/src/Unison/CommandLine/Main.hs
```diff
@@ -74,23 +74,23 @@ getUserInput
...
-                go
-              Right i -> pure i
+      Just l  -> case words l of
+        [] -> go
+        ws ->
+          case parseInput patterns . (>>= expandNumber numberedArgs) $ ws of
+            Left msg -> do
+              liftIO $ putPrettyLn msg
+              go
+            Right i -> pure i
```
runarorama, line 93: Formatting changes only. There was a change here but it got changed back at some point.

parser-typechecker/src/Unison/CommandLine/Main.hs
```diff
@@ -228,7 +228,7 @@ main dir defaultBaseLib initialPath (config,cancelConfig) initialInputs startRun
                 writeIORef pageOutput True
                 pure x) `catch` interruptHandler
       interruptHandler (asyncExceptionFromException -> Just UserInterrupt) = awaitInput
-      interruptHandler _ = pure $ Right QuitI
+      interruptHandler e = error (show e)
```
runarorama, line 231: This crashes loudly on receiving an asynchronous error, instead of silently quitting.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -323,39 +331,6 @@ loop = do
...
-            (misses, hits) =
-              partition (\(_, results) -> null results) (zip hqs resultss)
-            results =
-              List.sort
-                .   uniqueBy SR.toReferent
-                $   (hits ++ termResults ++ typeResults)
-                >>= snd
-          pure (missingRefs ++ (fst <$> misses), results)
-        hqNameQuery = hqNameQuery' False
-        hqNameQuerySuffixify = hqNameQuery' True
```
runarorama, line 359: Moved to `Backend`

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1628,12 +1539,12 @@ loop = do
...
           case e of
-            Left e -> respond $ EvaluationFailure e 
-            Right _ -> pure () -- TODO 
+            Left e -> respond $ EvaluationFailure e
+            Right _ -> pure () -- TODO
 
       IOTestI main -> do
         testType <- eval RuntimeTest
-        parseNames0 <- (`Names3.Names` mempty) <$> basicPrettyPrintNames0
+        parseNames0 <- (`Names3.Names` mempty) <$> basicPrettyPrintNames0A
```
runarorama, line 1547: I'm not sure where this `A` comes from. This is not my change.

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
```diff
@@ -204,38 +204,14 @@ data Output v
...
-       ShallowTypeEntry _ hq   -> HQ'.toName hq
-       ShallowBranchEntry ns _ -> ns
-       ShallowPatchEntry  ns   -> ns
-     toHash :: ShallowListEntry v a -> Maybe ShortHash
-     toHash = \case
-       ShallowTermEntry _ hq _ -> HQ'.toHash hq
-       ShallowTypeEntry _ hq   -> HQ'.toHash hq
-       ShallowBranchEntry _  _ -> Nothing
-       ShallowPatchEntry _     -> Nothing
-
```
runarorama, line 240: Moved to the `Server` namespace.

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1776,49 +1792,60 @@ noResults = P.callout "😶" $
...
+  missingTypes =
+    nubOrdOn snd
+      $  [ (HQ'.toHQ name, r) | SR'.Tp name (MissingObject r) _ _ <- results ]
+      <> [ (HQ'.toHQ name, Reference.toShortHash r)
+         | SR'.Tm name Nothing (Referent.toTypeReference -> Just r) _ <- results
+         ]
   missingBuiltins = results >>= \case
-    SR'.Tm name Nothing r@(Referent.Ref (Reference.Builtin _)) _ -> [(HQ'.toHQ name,r)]
+    SR'.Tm name Nothing r@(Referent.Ref (Reference.Builtin _)) _ ->
+      [(HQ'.toHQ name, r)]
```
runarorama, line 1866: This is just a formatting change. Part of a more substantial change that got aborted.

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -1218,6 +1219,13 @@ prettyParseError s = \case
     , "but there wasn't one.  Maybe check your indentation:\n"
     , tokenAsErrorSite s tok
     ]
+  go Parser.EmptyMatch = mconcat
```
runarorama, line 1222: I decided to add this corner case handling to the parser since I ran up against it in my testing.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -0,0 +1,557 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Unison.Server.Backend where
```
runarorama, line 6: A lot of code from `HandleInput` was moved to this module. This is supposed to serve as the "backend" for both the HTTP server and the UCM command line tool. Probably needs to be broken up into smaller modules later.

parser-typechecker/src/Unison/Server/CodebaseServer.hs
```diff
@@ -0,0 +1,102 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeOperators #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Unison.Server.CodebaseServer where
```
runarorama, line 7: This module defines the actual Servant server and its HTTP API.

parser-typechecker/src/Unison/Server/Endpoints/GetDefinitions.hs
```diff
@@ -0,0 +1,108 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeOperators #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Unison.Server.Endpoints.GetDefinitions where
```
runarorama, line 7: This module contains the `getDefinitions` endpoint.

parser-typechecker/src/Unison/Server/Endpoints/GetDefinitions.hs
```diff
@@ -0,0 +1,108 @@
...
+  rel <- fmap Path.fromPath' <$> traverse (parsePath . Text.unpack) relativePath
+  ea  <- liftIO . runExceptT $ do
+    root <- traverse (Backend.expandShortBranchHash codebase) mayRoot
+    Backend.prettyDefinitionsBySuffixes rel root width codebase
+      $   HQ.unsafeFromText
+      <$> hqns
+  errFromEither backendError ea
+ where
+  parsePath p = errFromEither (`badNamespace` p) $ Path.parsePath' p
+  errFromEither f = either (throwError . f) pure
```
runarorama, line 108: This function does the actual work of serving definitions over the HTTP server.

parser-typechecker/src/Unison/Server/Endpoints/ListNamespace.hs
```diff
@@ -0,0 +1,218 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeOperators #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Unison.Server.Endpoints.ListNamespace where
```
runarorama, line 10: This module defines the `list` endpoint for getting the contents of a namespace.

parser-typechecker/src/Unison/Server/Errors.hs
```diff
@@ -0,0 +1,71 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Unison.Server.Errors where
```
runarorama, line 6: Various errors for the Unison server.

parser-typechecker/src/Unison/Server/QueryResult.hs
```diff
@@ -0,0 +1,11 @@
+module Unison.Server.QueryResult where
+
+import qualified Unison.HashQualified          as HQ
+import           Unison.Name                    ( Name )
+import qualified Unison.Server.SearchResult as SR
+
+data QueryResult = QueryResult
```
runarorama, line 7: This used to be just a pair, but I thought it warranted its own data type.

parser-typechecker/src/Unison/Server/Types.hs
```diff
@@ -0,0 +1,122 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Unison.Server.Types where
```
runarorama, line 7: Various types common to server endpoints.

parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -123,16 +123,18 @@ blockTerm = lam term <|> infixAppOrBooleanOp
...
-  _ <- P.try (openBlockWith "with") <|> do
-         t <- anyToken
-         P.customFailure (ExpectedBlockOpen "with" t)
+  _         <- closeBlock
+  _         <- P.try (openBlockWith "with") <|> do
+    t <- anyToken
+    P.customFailure (ExpectedBlockOpen "with" t)
   (_arities, cases) <- unzip <$> sepBy1 semi matchCase
-  -- TODO: Add error for empty match list
+  when (null cases) $ P.customFailure EmptyMatch
```
runarorama, line 133: Decided to fix this TODO

parser-typechecker/src/Unison/Util/AnnotatedText.hs
```diff
@@ -7,71 +9,74 @@
...
 import qualified Data.Sequence      as Seq
-import           Data.Tuple.Extra   (second)
 import           Unison.Lexer       (Line, Pos (..))
 import           Unison.Util.Monoid (intercalateMap)
 import           Unison.Util.Range  (Range (..), inRange)
 import qualified Data.ListLike      as LL
 import qualified GHC.Exts
 
--- type AnnotatedText a = AnnotatedText (Maybe a)
+data Segment a = Segment { segment :: String, annotation :: Maybe a }
```
runarorama, line 23: Added this `Segment` type mainly so it could be a functor in `a`.

parser-typechecker/unison/Main.hs
```diff
@@ -151,6 +155,7 @@ main = do
   case restargs of
     [] -> do
       theCodebase <- FileCodebase.getCodebaseOrExit branchCache mcodepath
+      _ <- forkIO $ Server.start theCodebase 8081
```
runarorama: This line starts the HTTP server. Any obvious problems with this?

unison-core/src/Unison/Referent.hs
```diff
@@ -19,17 +20,17 @@ import qualified Unison.ConstructorType as CT
...
+pattern Con r i t = ConRef r i t
 {-# COMPLETE Ref, Con #-}
 
-type Id = Referent' R.Id
+type Id = TermRef R.Id
 
-data Referent' r = Ref' r | Con' r Int ConstructorType
-  deriving (Show, Ord, Eq, Functor)
+data TermRef r = IdRef r | ConRef r Int ConstructorType
+  deriving (Show, Ord, Eq, Functor, Generic)
```
runarorama: The names for this type and its constructors were pretty terrible, so I improved them a little.

unison-src/new-runtime-transcripts/stm.output.md
```diff
@@ -134,7 +134,7 @@ display m n s =
   "mismatch: " ++ toText m ++ " + " ++ toText n ++ " /= " ++ toText s
 
 nats : [Nat]
-nats = [89,100,116,144,169,188,200,233,256,300]
+nats = [8,10,11,14,16,18,20,23,25,30]
```
runarorama, line 93: This makes the test run faster, but I don't really know what I'm doing here.

unison-src/transcripts/propagate.output.md
runarorama: These GUIDs depend on the number of commands in the transcript before the point where the GUIDs are generated. So since we added some commands, they're obviously going to change. If we want transcript GUIDs to survive implementation changes, we need to implement the state machine that generates them some other way.

parser-typechecker/src/Unison/Server/Endpoints/GetDefinitions.hs
pchiusano: I'd put any sort of module overview as a comment in the code, although in this case, I feel like the module name is sufficient.

parser-typechecker/tests/Suite.hs
```diff
@@ -66,7 +65,6 @@ test rt = tests
   , Typechecker.test
   , UriParser.test
   , Context.test
-  , Git.test
```
pchiusano: Was there a reason this test was removed?

parser-typechecker/unison/Main.hs
pchiusano: Should we pick the port dynamically, otherwise you can't run two instances of UCM on the same machine?

Security concern: you don't necessarily want your codebase being served over http for anyone who can see your machine on the network. So I'd say for now at least put it behind a flag until we can add some authorization control to it.

For authorization, one simple approach would be to make the URL unguessable and is printed out on startup. Like you need to hit `localhost:<port-number>/<256 bit random number>/<endpoint>` where the random number is generated on startup. Any endpoints not starting with that random number are ignored. You get a message on startup:

    Codebase server is running at <the url>. Use `help server` to see this again or learn more.

Maybe @stew will have other ideas.

unison-core/src/Unison/HashQualified'.hs
pchiusano: Yeah it is confusing, and I think this is a positive change. 

However, I would have tried to address this and only this with a separate PR, which would have been totally mechanical and could have been merged quickly (including into your ongoing branch, so it need not have slowed you down). This change resulted in a massive diff, and switches the meaning of a symbol--always confusing if anyone has concurrent edits to the same files, which becomes more likely given how long this PR was in progress for. Throw in some formatting changes and it can become pretty brain bending to resolve conflicts. In particular, I am a little worried about conflicts with @aryairani branch, since I think he is also dealing with that type a lot. @aryairani are we going to have a disaster on our hands when attempting to merge this?? 🤞

unison-core/src/Unison/Referent.hs
pchiusano: Seems fine, but I'd also have done this as a separate PR. 

This one isn't as bad since the names were just changed, not repurposed so the conflict resolution process is simpler: you just fix the unresolved symbol errors. Repurposing names can lead to a lot more confusion since you can end up with type errors very far from the place you need to actually change.

pchiusano (commented):
Sweet!

I think the note about putting the server startup behind a flag and/or a `serve` command is the main thing I'd like to see changed before merging this. 

Also see my notes on git PR hygiene. I'm hoping in this case it doesn't generate conflict hell for @aryairani but even if it's fine in this instance, would like it on your radar for the future. :)

One thing: I noticed a bunch of types have `Generic` instances, is that so we can autoderive JSON instances for them? If yes, will that break front-end code if we change those types? It seems fine for now but how do you think we should handle this going forward? It does seem a shame to write a bunch of manual boilerplate.

pchiusano:
Also just adding a comment for next reviewer, elaborating on @runarorama explanation. The actual server is pretty boilerplate-y Servant stuff and isn't super interesting - if you read the [Servant docs](https://www.servant.dev/) the boilerplate will make more sense. The bulk of the change was just moving logic out of `HandleInput` (which is specific to a CLI) into `Command` and `Backend` (which is not specific to a `CLI`) so that it could be used by both `HandleInput` and this server. This wasn't as straightforward as it should have been due to `HandleInput` being a tangled mess.

parser-typechecker/tests/Suite.hs
runarorama: The test relies on specific GUIDs being generated for some types it creates. Should open a separate issue to refactor the test to not do that.

runarorama:
> One thing: I noticed a bunch of types have `Generic` instances, is that so we can autoderive JSON instances for them? If yes, will that break front-end code if we change those types? It seems fine for now but how do you think we should handle this going forward? It does seem a shame to write a bunch of manual boilerplate.

In the long term we should have separate types for the HTTP frontend vs the internals, even if that means we have to duplicate a bunch of code.

parser-typechecker/unison/Main.hs
stew: I feel pretty strongly like you need to let the environment dictate what port we bind to and when. I think it should be something like a `--listen 1.2.3.4:8080` or at least just a `--listen 8080`

aryairani: For comparison, here's what jupyter notebooks do:
```
arya@jrrr j % /opt/homebrew/anaconda3/bin/jupyter notebook
[I 14:30:29.525 NotebookApp] Writing notebook server cookie secret to /Users/arya/Library/Jupyter/runtime/notebook_cookie_secret
[I 14:30:30.956 NotebookApp] JupyterLab extension loaded from /opt/homebrew/anaconda3/lib/python3.8/site-packages/jupyterlab
[I 14:30:30.956 NotebookApp] JupyterLab application directory is /opt/homebrew/anaconda3/share/jupyter/lab
[I 14:30:30.958 NotebookApp] Serving notebooks from local directory: /private/tmp/j
[I 14:30:30.958 NotebookApp] Jupyter Notebook 6.1.4 is running at:
[I 14:30:30.958 NotebookApp] http://localhost:8888/?token=a48d7a48677f484ac72f7764329a0df1b83a44c3524034e5
[I 14:30:30.958 NotebookApp]  or http://127.0.0.1:8888/?token=a48d7a48677f484ac72f7764329a0df1b83a44c3524034e5
[I 14:30:30.958 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 14:30:30.970 NotebookApp] 
    
    To access the notebook, open this file in a browser:
        file:///Users/arya/Library/Jupyter/runtime/nbserver-67793-open.html
    Or copy and paste one of these URLs:
        http://localhost:8888/?token=a48d7a48677f484ac72f7764329a0df1b83a44c3524034e5
     or http://127.0.0.1:8888/?token=a48d7a48677f484ac72f7764329a0df1b83a44c3524034e5
```
A fixed port, binding on loopback (only, I suppose), and a security token that only the console user can see.  It also somehow launched the URL in my web browser automatically. ¯\\\_(ツ)\_/¯

unison-core/src/Unison/Referent.hs
aryairani: For better or worse, `Id` in the context of references has meant "statically known to not be a Builtin", which isn't the case for this `IdRef`, so I think this will be pretty confusing.  I guess if the previous names are intolerable, I would just switch them to `Ref` and `Con`, and give up on the patterns that preserved the original types.  I don't think `IdRef` is more meaningful than `Ref'`, is it?  

IIRC this nomenclature shows up in a few other places, so even if we decide to change it, I'd request that we revert this change for now, and then do a more comprehensive terminology switch after getting the codebase2 branch where we want it.

unison-core/src/Unison/Var.hs
```diff
@@ -137,7 +137,7 @@ reset :: Var v => v -> v
 reset v = typed (typeOf v)
 
 unqualifiedName :: Var v => v -> Text
-unqualifiedName = last . Name.segments' . name
+unqualifiedName = fromMaybe "" . lastMay . Name.segments' . name
```
aryairani, line 140: I agree we don't want it crashing, but `""` isn't a valid name, right?

parser-typechecker/src/Unison/Codebase/Causal.hs
```diff
@@ -39,7 +41,7 @@ import qualified Data.Set                      as Set
 -}
 
 newtype RawHash a = RawHash { unRawHash :: Hash }
-  deriving (Eq, Ord)
+  deriving (Eq, Ord, Generic)
```
aryairani, line 44: I'm not super familiar with `Generic`; does it mean that if we rename a field here, that it will change the HTTP API too?

unison-core/src/Unison/Referent.hs
pchiusano: +1 on reverting this and doing a terminology review at a more convenient time.

parser-typechecker/tests/Suite.hs
pchiusano: Gotcha. Yeah, please create a ticket to track.

aryairani: Tests still need to be able rely on specific GUIDs being generated; short of hard-coding the seeds for each unique type, refactoring this sort of test to not do that isn't possible, I don't think.

Currently, the transcript runner generates deterministic GUIDs according to the number of `Command`s that have been executed. This PR changed the sequence of `Command`s, which affected the GUID sequence, which changed the "deterministic" sequence, breaking the test's assumption.

We need to open a separate issue to refactor the deterministic GUID generation to depend on the number of typechecks (or something), rather than the number of commands, and then recreate this test to rely on the new, robust, deterministic GUID sequence.

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
aryairani: How do we know which ones should or shouldn't be?

parser-typechecker/src/Unison/Util/SyntaxText.hs
```diff
@@ -1,25 +1,29 @@
...
+import Unison.Referent (TermRef)
 import Unison.HashQualified (HashQualified)
 import Unison.Pattern (SeqOp)
 
-import Unison.Util.AnnotatedText      ( AnnotatedText(..), annotate )
+import Unison.Util.AnnotatedText      ( AnnotatedText(..), annotate, segment)
 
-type SyntaxText = AnnotatedText Element
+type SyntaxText = SyntaxText' Reference
+type SyntaxText' r = AnnotatedText (Element r)
```
aryairani, line 15: Hehe, I think this is the same `'` generalization pattern that this PR tries to eliminate from `HashQualified'` and `Referent`.  I feel vindicated. :)

parser-typechecker/unison/Main.hs
aryairani: (That said, I don't care for this PR.)

unison-core/src/Unison/HashQualified'.hs
aryairani: I don't think codebase2 deals with `HashQualified` at all (though `Referent` is a different story), but I do feel like I'm absolutely drowning in renames in this PR, and concerned that I didn't really have brain cycles remaining for any of the interesting codebase server stuff.

unison-src/new-runtime-transcripts/stm.output.md
aryairani: /cc @dolio was there anything special about the original numbers?

unison-src/new-runtime-transcripts/utf8.output.md
```diff
@@ -90,12 +90,12 @@ test> checkRoundTrip greek
     
       checkRoundTrip : Text -> [Result]
       greek          : Text
-    test.kqfpde2g5a  (Unison bug, unknown term)
+      greekTest      : [Result]
```
aryairani: What was the story behind the bug this no longer displays again?

pchiusano: There's some bug with anonymous test watches, which is a little known feature where you can get Unison to make up a GUID name for a test. But since you generally want your tests to have names, we've never gotten around to fixing. I think there's a ticket for it somewhere...

unison-core/src/Unison/Var.hs
runarorama: If the incoming `v` has an empty name, this isn't the place to handle that.

parser-typechecker/src/Unison/Codebase/Causal.hs
runarorama: Yes it does mean that. This is a good reason to duplicate these types into the HTTP API.

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
runarorama: Things that are super specific to the CLI probably should not go into `Backend`.

parser-typechecker/src/Unison/Util/SyntaxText.hs
runarorama: I kind of hate it and it would be nice to have better names for these aliases. Maybe later.

unison-core/src/Unison/Var.hs
aryairani: Maybe crashing loudly would be better?

pchiusano: Yeah, I'd replace that `fromMaybe ""` with `fromMaybe (error "unqualifiedName of empty name")`. That will preserve the old behavior rather than returning what the caller might consider a bogus value and leading to issues elsewhere.

```suggestion
unqualifiedName = fromMaybe (error msg) . lastMay . Name.segments' . name
  where msg = "Var.unqualifiedName applied to a variable whose name is the empty string"
```

In principle there's nothing wrong with a function like this silently returning the empty string (much like `take 10` may return a list with fewer than 10 elements and that's all fine and expected) but it's a change to the conventions previously established for this function, so unless you want to audit all the use sites to make sure they are okay with the new convention I'd leave it alone.

unison-core/src/Unison/Referent.hs
runarorama: The real issue is that `'` can't appear in a type that you serve out with Servant. So the solution is either:

1. Rename these things.
2. Copy-paste them into the HTTP module with different names.

I think 2 is the solution we'll have to end up with eventually.

unison-core/src/Unison/HashQualified'.hs
runarorama: Needed to remove the `'` here to satisfy Servant. The alternative is to duplicate this type into the Server module.

unison-core/src/Unison/Var.hs
runarorama: This function throwing an error was the source of a bug in `ls` which caused UCM to crash. We have a type system though, so we can just force callers to give us at least one segment here.

pchiusano: Gotcha. I guess you could have this function return `Maybe` and update the callers, but I think your original change is fine too.

unison-core/src/Unison/Referent.hs
pchiusano: @runarorama Isn't there a third option: don't use the `Generic` autoderived JSON instance, just write a `From/ToJSON` manually (or whatever typeclasses you have to give to be served up by Servant). Assuming yes, I'd revert the names and do this.

Giving manual JSON instances (possibly as orphan instances in the Server packages) also is another way to  avoid having to pessimistically fork copies of every type that the server works with.

pchiusano (commented):
We just did a review of this with @aryairani @runarorama and myself. It's looking good but quick summary of outstanding work before we merge:

* On starting the codebase server, it's probably okay for now to just start the server on a random port on loopback (what it's doing now as of latest commit) but Rúnar indicated it's pretty easy to add a random GUID to the base url or as a query token so he'll likely just do that as part of this PR. If punting on this for this PR, just create a ticket to track. The token should be in place before we do beta release.
  * Something that @hojberg has brought up is if at some point we want something like an electron app for the codebase UI, rather than having to run something in the browser by copy-pasting a weird URL. I can imagine a few approaches to this but overall it needs more thought. I think it's out of scope for this PR though.
* Rúnar is going to revert the naming of `Referent` constructors back to what it was before to minimize conflicts with codebase 2 work. To work around Servant's issues with serving up types with `'` in their name, he'll either make a copy of `Referent` in the server package so it can evolve independently, or else add a custom orphan JSON-ify instance in the server package for now, whatever he thinks makes the most sense.
* Fix the conflicts in stack.yaml and get CI to pass.

Other random discussion: 

* For the future, like if this were running on unisonweb.org, we'd likely add more advanced ways of doing authentication (like authenticate using any OAuth provider) and it's possible that the URL scheme will change. But it should be easy enough to adapt the server if/when we support that. Basically, out of scope for this PR.
* We discussed converting to opt-parse-applicative (possibly revive #1426) which I'm totally cool with as a separate PR.

pchiusano (approved):
Sweet, looks like this is all set. @runarorama looks like you just have to sync with trunk and refresh transcripts - I'm guessing that will fix the CI as well.

PR #1787, opened 2021-01-15
Implement STM builtins

dolio:
This merge adds `STM` and `TVar` stuff to the new runtime. The implementation is a pretty direct wrapping of GHC's API, and was more or less straight forward. Most operations are 'foreign' functions calling the GHC analogue while coercing the `STM` into `IO` if applicable. `atomically` is the only new primop, and is implemented by evaluating a closure with a temporary stack inside GHC's `atomically` implementation. This uses `unsafeIOToSTM` to do all the runtime operations necessary, but this should not be a problem as long as the unison code is well typed, because the only mutable state is local to the transaction.

There is a test transcript that should use all the operations currently implemented. I didn't try very hard to trigger transaction issues, but I think the implementation is sound.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Builtin.hs
+++ b/parser-typechecker/src/Unison/Runtime/Builtin.hs
@@ -1454,6 +1454,9 @@ declareForeigns = do
   declareForeign "TVar.readIO" boxDirect . mkForeign
     $ \(v :: STM.TVar Closure) -> STM.readTVarIO v
 
+  declareForeign "TVar.swap" boxBoxDirect . mkForeign
+    $ \(v, c :: Closure) -> unsafeSTMToIO $ STM.swapTVar v c
+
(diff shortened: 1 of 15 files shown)


pchiusano (approved):
Sweet! LGTM. It seems like the CI failure is spurious, it bombed during setup, fetching dependencies or something. I'm just rerunning.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1228,13 +1253,15 @@ builtinLookup
 
   , ("jumpCont", jumpk)
 
-  , ("IO.forkComp", fork'comp)
+  , ("IO.forkComp.v2", fork'comp)
```
pchiusano, line 1256: @stew FYI, not sure if you were also hitting this.

PR #1793, opened 2021-01-23
Working TLS, Tests for io2

### LINKED ISSUE #1693: TLS builtins proposal
Mostly following the [tls Haskell package](https://hackage.haskell.org/package/tls-1.5.4/docs/Network-TLS.html) API, which will be what's backing this.

Kind of interesting is that the API is the same for the client and server, you just specify different config to initialize the `Tls` in each case.

```Haskell
use io2

builtin type Tls.Config -- see below for the API

-- Create a new TLS instance, backed by the provided socket
Tls.new : Tls.Config -> Socket ->{IO} Tls

-- Initiate handshake with other party. The params control whether
-- the handshake is done in the client or server role  
Tls.handshake : Tls ->{IO} () 

-- Send some bytes over the connection
Tls.send : Tls -> Bytes ->{IO} ()

-- Receive some bytes from the connection
Tls.receive : Tls ->{IO} Bytes

-- Indicate to the other party you wish to terminate the connection.
-- This doesn't close the underlying socket, allowing the socket
-- to be reused for another connection (following a handshake I guess?)
Tls.terminate : Tls ->{IO} ()

-- Obtain the underlying socket
Tls.socket : Tls -> Socket
```

### Configuration

For the configuration API, there's reasonable defaults and a LOT of optional parameters. An ergonomic way of dealing with only the parameters you care about is to have a value `default : Tls.Config`, and then various functions of type `Tls.Config -> Tls.Config` to add non-default settings. Here's how this will look:

```Haskell
config : Tls.Config
config = 
  use Tls.Config
  default 
    |> server.set  
    |> versions.set [Tls13, Tls12] 
    |> ciphers.set Tls.Ciphers.strong
```

This produces a configuration for the TLS server role (default is client), allows only TLS 1.3 and TLS 1.2, and allows only the "strong" ciphersuite. Here's another example, configuring for the client role and setting the expected hostname to "google.com":

```Haskell
clientConfig : Tls.Config
clientConfig =
  use Tls.Config
  default
    |> client.set
    |> hostname.set "google.com"
```

Here's a sketch of the rest of the API. By representing the relevant types and terms as opaque builtins, we can progressively add more configuration options over time without breaking anyone's code.

```Haskell
-- conceptually a [Tls.Setting], with later settings overriding 
-- but represented as an opaque builtin type so we can add more 
-- optional settings over time
builtin type Tls.Config

-- default settings
Tls.Config.default : Tls.Config

-- configure for client-side role (this will be the default)
Tls.Config.client.set : Tls.Config -> Tls.Config

-- configure for server-side role
Tls.Config.server.set : Tls.Config -> Tls.Config

-- configure the allowed TLS versions, preferred version first
Tls.Config.versions.set : [Tls.Version] -> Tls.Config -> Tls.Config

-- configure the allowed ciphers (see below)
Tls.Config.ciphers.set : Tls.Ciphers -> Tls.Config -> Tls.Config

-- configure for client-side role
Tls.Config.client.set : Tls.Config -> Tls.Config

builtin type Tls.Ciphers

-- allow no ciphers
Tls.Ciphers.empty : Ciphers

-- allow unencrypted ciphers (for debugging only)
Tls.Ciphers.unencrypted : Ciphers

-- https://hackage.haskell.org/package/tls-1.5.4/docs/Network-TLS-Extra-Cipher.html
Tls.Ciphers.default : Ciphers
Tls.Ciphers.strong : Ciphers

-- We could later allow very fine-grained individual ciphers
-- https://hackage.haskell.org/package/tls-1.5.4/docs/Network-TLS-Extra-Cipher.html

-- AES cipher (128 bit key), RSA key exchange and SHA256 for digest
Tls.Ciphers.AES128_SHA256 : Ciphers

-- prefer to use earlier ciphers if the other side supports them
Tls.Ciphers.append : Ciphers -> Ciphers -> Ciphers

builtin type Tls.Version

Tls.Version.Tls13 : Tls.Version
Tls.Version.Tls12 : Tls.Version
Tls.Version.Tls11 : Tls.Version
Tls.Version.Tls10 : Tls.Version

-- configure the expected hostname
Tls.Config.hostname.set : Text -> Tls.Config -> Tls.Config

-- https://hackage.haskell.org/package/x509-validation-1.6.11/docs/Data-X509-Validation.html#t:ValidationChecks
Tls.Config.disableHostnameCheck : Tls.Config -> Tls.Config
Tls.Config.disableExpirationCheck : Tls.Config -> Tls.Config
Tls.Config.disableCAConstraintsCheck : Tls.Config -> Tls.Config
```

### Certificate stores

If we want to be able to customize certificate stores, it will require having some Unison type for representing X.509 certificates as Unison values, which is [a bit of a can of worms](https://hackage.haskell.org/package/x509-1.7.5/docs/Data-X509.html#t:Certificate).

Perhaps initially we don't need this (if it's just being used for client-side https, default system certificate store might be okay).

Here's a sketch:

```Haskell
-- https://hackage.haskell.org/package/tls-1.5.4/docs/Network-TLS.html#t:Shared
Tls.Config.certificates.set : CertificateStore -> Tls.Config -> Tls.Config

X509.CertificateStore.empty : Certificates
X509.CertificateStore.system : '{IO} Certificates
X509.CertificateStore.fromDirectory : Text ->{IO} Certificates
X509.CertificateStore.fromFile : Text ->{IO} Certificates
X509.CertificateStore.union : Certificates -> Certificates -> Certificates
X509.CertificateStore.fromList : [SignedCertificate] -> Certificates
X509.fromBytes : Bytes ->{Exception Text} SignedCertificate
```

A hacky idea is to have `SignedCertificate` just be an opaque blob of bytes. It can be converted to a proper certificate when needed using: https://hackage.haskell.org/package/x509-1.7.5/docs/Data-X509.html#v:decodeSignedCertificate

But I think the right thing to do is properly model the type in Unison, following [this type](https://hackage.haskell.org/package/x509-1.7.5/docs/Data-X509.html#t:Certificate).

stew:
closes #1693 

## Overview

This adds basic TLS support to the new runtime. It adds tests for TLS but also for a lot of the other parts of io2.

## Test coverage

TLS is covered, the rest of IO is maybe 80% covered?

## Loose ends

Still need to implement the user configuration of acceptable TLS Versions and Ciphers


pchiusano (approved):
👍

A couple notes -

- The friction of converting among exception types and Eithers is annoying. I know we have a plan for how to fix that in base but it's also pretty annoying in these tests. I think we can improve this. I'll write up a separate ticket with some ideas.
- You can pass multiple transcripts to the transcript command. They'll be run in sequence on the same codebase. It might be possible to share some of the preamble code that starts out each transcript in your tests this way. Could be cleanup for later.
- @dolio can you peep the Machine.hs and Runtime.hs changes if you haven't already? If it looks good you can merge.

dolio (approved):
Things look fine to me.

PR #1798, opened 2021-02-01
Feature/nice io names

stew:
## Overview

This PR renames the new IO builtin functions which return a `Either Failure a` to be appended by `.impl`. So, for example, the Built-in function previously named `io2.IO.readFile` is now `io2.IO.readFile.impl`. The reason being that we want the return type of a future function named readFile to be introduced which wraps this `.impl` and instead has a return type of `{Exception} a`

Paul and I had initially decided to include the new implementations which wrap the .impl verisons in IOSource.hs, but after revisiting this decision with Rúnar, we decoded that it would be more appropriate to move these definitions to base, so I'll be making a PR there as well. Having things defined in two places (IOSource.hs and base) is confusing, and if we are going to lean towards one or the other, it seems obvious that we should be aiming to put more stuff into base. This decision will lead to the problem of boilerplate in the transcripts, we think we should address this by adding some mechanism for allowing one transcript to include another, so that boilerplate that is creeping into more and more new transcripts can be consolidated instead of repeated.

## Test coverage

This stuff is pretty well covered by the transcripts which are updated here as well.

## Loose ends

Make sure we are tracking:
- [ ]  There will be a forthcoming PR to base which should merge AFTER this PR is merged.
- [ ]  We should add an ability to include one transcript in another


pchiusano:
@stew this is still in draft form but I notice you tagged me for review... should I be reviewing it now or what?

aryairani:
Oh hey — none of the `B` names need to change, and correspondingly, the runtime  also doesn’t need to change; and importantly, although it’s arguably okay to do right now because we haven’t “released” io2 yet, in general those are not allowed to change, because it will break existing codebases.

What you instead want to do is add a `Rename` directive somewhere after the `B`(uiltin) directive, which only affects where the builtin shows up in the namespace rather than redefining its identity.  

In general, we don’t want the old builtins to cease to exist, as this PR does, but just move them around in the namespace with `Rename`.

aryairani:
This is perhaps not super obvious, because `B` was factored out of `ioBuiltins` entirely; it should probably be put in to give more inline flexibility for this sort of thing.

pchiusano:
Yeah, what @aryairani said. 

If we're going the route of changing this stuff in base rather than in `IOSource` then I'd do this PR with a little `baselite.md` precursor transcript that can be run before every transcript. If you do `unison transcript t1.md t2.md` then `t2.md` will run on the codebase produced by t1, so that `baselite.md` could have whatever stuff in it to do renaming and introduce helper functions as appropriate.

aryairani:
@pchiusano — @stew and I chatted and decided it's okay to switch these up now, provided they don't change after the next official release, if it will mean simpler code long-term.

parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -386,8 +386,8 @@ io.stderr : io.Handle
 io.stderr = Handle "stderr"
 
 -- Throw an I/O error on the left as an effect in `IO`
-io.rethrow : (Either io.Error a) -> {IO} a
-io.rethrow x = match x with
+io.reraise : (Either io.Error a) -> {IO} a
+io.reraise x = match x with
     Either.Left e -> io.IO.throw e
     Either.Right a -> a
```
pchiusano: Is this change still relevant to this PR? I think this file could be left alone. If the concern is around name collisions between io and io2, the right thing to do I think is just a `delete.namespace io` at the start of any transcripts expecting to run with the new runtime.

pchiusano (approved):
@stew just left one comment for you in `IOSource`. There's some conflicts still, maybe just need to refresh the transcripts.

LMK if you want a hand with tweaking the transcript runner to use a common prelude. That could be done as a separate PR also.

parser-typechecker/src/Unison/Runtime/IOSource.hs
aryairani: Not sure if I'm understanding correctly, but IMO it would be better if old `io` is never added in the first place, when running with the new runtime -- but maybe that's a separate PR.

pchiusano: Sort of orthogonal.. even if that were done, old IO names could be in the namespace from a prior invocation of ucm. It's not as if you can just delete them then.

aryairani: Sorry, I meant for transcripts that start with an empty namespace and then add stuff to it, just add the correct stuff. 😅

pchiusano: Oh gotcha. Yeah that makes sense. I think you need to do `builtin.mergio` to get the stuff in `IOSource` anyway.

@stew I think the point still stands that you could do without this renaming in this PR.

stew: we certainly don't need this rename

PR #1806, opened 2021-02-10
Fix delete.term-replacement to support coinciding type and term names

### LINKED ISSUE #1804: remove.{term,type}-replacement reports an error (but doesn't terminate) if `hqNameQuery` returns both term and type hits for the query argument
Here is a transcript to reproduce the issue.  I don't think it matters whether we are using `delete.term-replacement` or `delete.type-replacement`, nor whether the argument is actually present in a patch.  It looks like the code assumes [and would be behave correctly if] `hqNameQuery` only returned a type _or_ a term for a given query, and not both, but that's not the case.

    ```unison
    bar = 3
    type bar = Foo
    ```

    ```ucm
    .> add
    ```

    ```unison
    type bar = Foo | Bar
    ```

    ```ucm
    .> update
    .> view.patch
    .> delete.type-replacement bar
    .> view.patch
    ```

```
🛑

    bar
```
Note that despite the error message, the type replacement *is* deleted from the patch.

IIRC, we're thinking these lines should be restructured:

          unless (null misses) $
            respond $ SearchTermsNotFound misses
          traverse_ go (if isTerm then tmRefs else tpRefs)

/cc @dolio for extra detail if needed, @runarorama in case we misunderstood

scott-christopher:
## Overview

(hopefully) resolves #1804

## Implementation notes

Removes names from `hqNameQuery` misses in `doRemoveReplacement` when the relevant type or term exists in both.

While the change implemented satisfies the new and existing transcripts, I'm not 100% confident that the behavior is what is desired and would welcome someone familiar with the issue to cast a critical eye over it.

## Test coverage

Extended existing `deleteReplacements.md` to include the transcript from #1804 

## Loose ends

n/a

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -370,11 +370,9 @@ loop = do
           (misses', hits) <- hqNameQuery [from]
           let tpRefs = Set.fromList $ typeReferences hits
               tmRefs = Set.fromList $ termReferences hits
-              tmMisses = misses'
-                         <> (HQ'.toHQ . SR.termName <$> termResults hits)
-              tpMisses = misses'
-                         <> (HQ'.toHQ . SR.typeName <$> typeResults hits)
-              misses = if isTerm then tpMisses else tmMisses
+              misses = Set.difference (Set.fromList misses') if isTerm
+                then Set.fromList $ HQ'.toHQ . SR.termName <$> termResults hits
+                else Set.fromList $ HQ'.toHQ . SR.typeName <$> typeResults hits
               go :: Reference -> Action m (Either Event Input) v ()
               go fr = do
                 let termPatch =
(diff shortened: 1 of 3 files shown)


runarorama (approved):

dolio:
I think this is actually wrong in the other direction, now.

My understanding of the old logic is that `hqNameQuery` only reports a 'miss' when it finds _neither_ a type _nor_ a term with the given name. So, if you try to look up a term, but it finds a type, it gives you back a type 'hit', not a miss. So now if you try to delete a type replacement, but give a term hash, it won't report any error because there are no misses (but it also won't do anything).

I think the best condition to test for an error message is if the relevant hits _are_ empty. Then the misses and opposite hits would be useful in a more informative error message (the message just reports `misses` as not existing, but it could tell you that it just found the wrong type of hit).

aryairani:
@scott-christopher Wanna take another stab at it? Possibly creating a custom informational message in Output.hs / OutputMessages.hs

scott-christopher:
Yep, sure thing.

> I think the best condition to test for an error message is if the relevant hits are empty.

I'm pretty sure I tried this first but one of the existing transcripts then failed. I'll take another look and see what I find.

dolio:
Might be the transcript that's actually wrong. :)

aryairani:
@scott-christopher Yeah — feel free to reach out here or in the Slack as needed to compensate for the vague task spec here :D

PR #1807, opened 2021-02-10
Fix boolean pattern compilation in the new runtime

### LINKED ISSUE #1711: New runtime fails with "Blank" when evaluating this watch expression
Here's a transcript:

    ```ucm:hide
    .> builtins.merge
    ```

    ```unison
    check : Boolean -> [Result]
    check = cases
      true  -> [Result.Ok "Passed."]
      false -> [Result.Fail "Failed."]

    > check true
    ```

This fails when evaluating the watch expression with:

```
  change:
  
      check : Boolean -> [Result]
  
  `>`)... Ctrl+C cancels.

  "Blank"
```

What is `"Blank"` in quotes like that?

dolio:
The case for handling boolean patterns was just missing in the pattern compiler. It has been added.

I also added some more pattern matching tests.

Fixes #1711


pchiusano:
Cool. Can you get CI passing? https://github.com/unisonweb/unison/pull/1807/checks?check_run_id=1874767087#step:16:3060

dolio:
I'm not actually sure why it doesn't pass. The tests work on my machine.

dolio:
Oh, I know what the problem is. The tests are pinned to a broken version of `containers` that causes ill-formed maps (`IntMap` I think) to be built sometimes. It's not broken for me locally because I use a newer version. The bug has been fixed for months, but stackage still has the broken version pinned.

Should I just remove the part of the test that fails?

pchiusano:
@dolio you can locally override which version of `containers` gets used by Stack I think in the stack.yaml? Or I'm not sure. But that's the correct thing to do, since at the moment stack is the official way we do builds and produce releases, so we want that to also use the correct version of containers.

pchiusano (approved):
LGTM

PR #1815, opened 2021-02-17
Fix a corner case in abilityCheck'

### LINKED ISSUE #1779: Can't unify an ability variable with {}
Unison doesn't seem to want to match the type `'{S (R g), g} ()` with `'{S (R {})} ()` in the definition of `run'` below, but I think it should?

```haskell
unique ability S a where
  s : a

unique type R g = R ('{g} ())

run : '{S (R g), g} ()
run _ = todo ()

run' : '{S (R {})} ()
run' = run

--  The expression in red  needs these abilities: {S (R {}), }, but this 
--  location only has access to the {S (R {})} ability,
--
--     10 | run' = run
```

(trunk@b586abb)

dolio:
It was possible to write code that would unify a variable to e.g. `{}`, while `abilityCheck'` was written under the premise that the check runs on a completely flattened list. This code catches that corner case.

Fixes #1779

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -1738,6 +1738,10 @@ abilityCheck' ambient0 requested0 = go ambient0 requested0 where
       Just amb -> do
         subtype amb r `orElse` die r
         go ambient rs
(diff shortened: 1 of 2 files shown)


pchiusano:
Sweet!

pchiusano (approved):

PR #1825, opened 2021-02-26
Get rid of nondeterministic failure in thread.md

### LINKED ISSUE #1824: nondeterministic new-runtime transcript failure in thread.md
This run failed, but it seems like it must be an older bug than the implicated commit.
https://github.com/unisonweb/unison/actions/runs/600816294

The output suggested re-running with:
```
EasyTest.rerunOnly 8643038112808378411 "thread.md"
```
but I forget what to actually do with this information.

@dolio Could you take a look and see if you think it's a Rt2 issue, vs maybe a bug in the transcript?

stew:
Fixes #1824

### DIFF
--- a/unison-src/new-runtime-transcripts/thread.md
+++ b/unison-src/new-runtime-transcripts/thread.md
@@ -25,10 +25,9 @@ See if we can get another thread to stuff a value into a MVar
 ```
 
 ```unison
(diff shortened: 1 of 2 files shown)


pchiusano (approved):
🙏

PR #1826, opened 2021-02-28
Allow listing namespace contents by hash

runarorama:
## Overview

This modifies the UCM backend so that the `list` server endpoint can handle namespace hashes.

## Implementation notes

The parser used to simply return an error when the namespace requested was only a hash, but now it looks up the namespace by hash in the codebase.

We don't in general know _the name_ of a namespace that we looked up by hash. There's a good reason for this: a namespace hash could have several names in the codebase, and names could be _quite far_ back in the codebase's history. So while we could potentially just list _every name this namespace hash has ever had_, this would be tremendously slow with the current codebase format.

We can revisit this with the new codebase format, where we can simply keep an index mapping namespace hashes to their historical names.


pchiusano (approved):
👍

PR #1827, opened 2021-03-03
add consistent list naming, issue #1822

### LINKED ISSUE #1822: The Unison `List` type is inconsistently called "list", "sequence", and "vector" in different parts of the Haskell code
Minor cleanup of nomenclature. In `Type.hs` we have:

```Haskell
vectorRef :: Reference
vectorRef = Reference.Builtin "Sequence"

vector :: Ord v => a -> Type v a
vector a = ref a vectorRef
```

We have to keep the `Builtin "Sequence"` as is, since otherwise that changes the identity of the type, but the names of `vectorRef` and `vector` should be changed to use the list nomenclature:

```Haskell
listRef :: Reference
listRef = Reference.Builtin "Sequence"

list :: Ord v => a -> Type v a
list a = ref a listRef
```

Meanwhile, in `Term.hs`, the constructor is called "Sequence" and the smart constructors are called `seq` and `seq'`. These should be changed as well to `list` and `list'` and resulting compile errors fixed.

nini-faroux:
## Overview

Resolves #1822

## Implementation notes

Made the initial changes to Type.hs and Term.hs and then followed the compiler errors to apply the fixes needed in other modules.

n/a

## Test coverage

Used the existing test coverage 

## Loose ends

Hope it's ok! Let me know if I need to make any other changes.

### DIFF
--- a/parser-typechecker/src/Unison/TermParser.hs
+++ b/parser-typechecker/src/Unison/TermParser.hs
@@ -323,8 +323,8 @@ boolean :: Var v => TermP v
 boolean = ((\t -> Term.boolean (ann t) True) <$> reserved "true") <|>
           ((\t -> Term.boolean (ann t) False) <$> reserved "false")
 
-seq :: Var v => TermP v -> TermP v
-seq = Parser.seq Term.seq
+list :: Var v => TermP v -> TermP v
+list = Parser.seq Term.list
 
 hashQualifiedPrefixTerm :: Var v => TermP v
 hashQualifiedPrefixTerm = resolveHashQualified =<< hqPrefixId
@@ -357,7 +357,7 @@ termLeaf =
     , link
     , tupleOrParenthesizedTerm
     , keywordBlock
-    , seq term
+    , list term
     , delayQuote
     , bang
     , docBlock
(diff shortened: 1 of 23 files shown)


parser-typechecker/src/Unison/TermParser.hs
```diff
@@ -324,7 +324,7 @@ boolean = ((\t -> Term.boolean (ann t) True) <$> reserved "true") <|>
           ((\t -> Term.boolean (ann t) False) <$> reserved "false")
 
 seq :: Var v => TermP v -> TermP v
-seq = Parser.seq Term.seq
+seq = Parser.seq Term.list
```
pchiusano: You could switch this one to `list` too.

pchiusano (commented):
Looks good, thanks so much! Left one minor comment, and then can you also add yourself to CONTRIBUTORS.md?

parser-typechecker/src/Unison/TermParser.hs
nini-faroux: No problem, sure I'll do this in an hour or so. Thanks!

pchiusano (approved):
Thanks so much, looks good.

PR #1829, opened 2021-03-04
Codebase server: indicate in namespace listing whether a term is a doc or a test, and if a type is an ability

runarorama:
## Overview

This adds new fields `termTag` and `typeTag` on namespace objects in the `list` endpoint. The term tag is optional and can be either `Doc` or `Test`. The type tag is non-optional and is always either `Ability` or `Data`.

This actually goes and retrieves the type declarations in order to determine if they're abilities or data types, so this is doing a lot more work now than it used to.

### DIFF
--- a/parser-typechecker/src/Unison/Server/Backend.hs
+++ b/parser-typechecker/src/Unison/Server/Backend.hs
@@ -204,14 +204,17 @@ findShallowInBranch codebase b = do
       b0 = Branch.head b
   termEntries <- for (R.toList . Star3.d1 $ Branch._terms b0) $ \(r, ns) -> do
     ot <- lift $ loadReferentType codebase r
+    -- A term is a doc if its type conforms to the `Doc` type.
     let isDoc = case ot of
           Just t  -> Typechecker.isSubtype t $ Type.ref mempty Decls.docRef
           Nothing -> False
+        -- A term is a test if it has a link of type `IsTest`.
         isTest =
           Metadata.hasMetadataWithType r (Decls.isTestRef) $ Branch._terms b0
         tag = if isDoc then Just Doc else if isTest then Just Test else Nothing
     pure $ ShallowTermEntry r (hqTerm b0 ns r) ot tag
   typeEntries <- for (R.toList . Star3.d1 $ Branch._types b0) $ \(r, ns) -> do
+    -- The tag indicates whether the type is a data declaration or an ability.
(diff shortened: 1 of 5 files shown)


parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -195,12 +204,23 @@ findShallowInBranch codebase b = do
       b0 = Branch.head b
   termEntries <- for (R.toList . Star3.d1 $ Branch._terms b0) $ \(r, ns) -> do
     ot <- lift $ loadReferentType codebase r
-    pure $ ShallowTermEntry r (hqTerm b0 ns r) ot
+    let isDoc = case ot of
+          Just t  -> Typechecker.isSubtype t $ Type.ref mempty Decls.docRef
```
runarorama, line 209: A term is a doc if its type conforms to `Doc`

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -195,12 +204,23 @@ findShallowInBranch codebase b = do
...
+          Nothing -> False
+        isTest =
+          Metadata.hasMetadataWithType r (Decls.isTestRef) $ Branch._terms b0
```
runarorama, line 213: A term is a test if it has metadata of type `IsTest`

parser-typechecker/src/Unison/Server/Backend.hs
aryairani: These comments might be better as code comments, wdyt?

aryairani (approved):

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -195,12 +204,23 @@ findShallowInBranch codebase b = do
...
+        tag = if isDoc then Just Doc else if isTest then Just Test else Nothing
```
hojberg, line 214: What do you think of having the non-doc, non-test case be "Term" or similar instead dealing with the Maybe?

hojberg (approved):
Sweet! Added a non-blocking comment.

parser-typechecker/src/Unison/Server/Backend.hs
runarorama: I can see this evolving to a list instead of a Maybe later on, or something more complex. Inclined to leave it alone for now.

PR #1814, opened 2021-02-16
provide more helpful error message when `run`ning an improperly-typed term

samgqroberts:
## Overview

Endeavoring to fix #1747 - see the output of the new transcript for what the new error message is and when it is shown.

## Implementation notes

Instead of `HandleInput.addRunMain` just returning a `Maybe` (I could do it, or I couldn't), it now returns a type that disambiguates whether it couldn't due to not being able to find the term, or whether the term has a bad type. That new case is subsequently handled in the error reporting module `OutputMessages.hs` to produce the new error message.

## Test coverage

I've included a new transcript `run.md` that attempts to `run` properly-typed, improperly-typed, and nonexistent terms to show the differences in behavior / error messages.

## Loose ends

- **This PR is obfuscated by another bug** - you'll see me call this out in the new transcript. I believe I'm running into #1800, and I've confirmed that without my changes that `runnable` function still fails to be found. I wanted to put this PR up optimistically to gather feedback while that issue is resolved.
- I didn't update the error reporting in `Execute.hs` to provide this distinction (partly due to my lack of familiarity of what experience that file is part of..)

In general, this being my first contribution, I'm not confident at all that the changes I've provided are idiomatic / sufficient. I'd be more than happy to add tests, move code, rename types. Thank you!

Issue #1800, opened 2021-02-06
Regression in `run` - fails to find the function to run


atacratic:
Using trunk@7ceadd, `run` fails:
![image](https://user-images.githubusercontent.com/21991324/107128793-833cc080-68b8-11eb-9820-dbba5aa2b993.png)

Possibly related commit [here](https://github.com/unisonweb/unison/commit/f0738b21d8bffac1f5b654446598f05a47b704de#diff-d190a1cdc57520446781fb2150b58a7d9677d5fe5e69265031126994896096e4).  

From slack:

> Arya Irani (he/him)  1 hour ago
> Just based on the output, it looks to me like the problem is a change in the type of suspended computations, but no corresponding change to main detection
> Arya Irani (he/him)  1 hour ago
> or maybe it’s just a pretty-printer improvement, and a bug in main detection

jaredly:
Did you have `--new-runtime` enabled? I'm also seeing this with new-runtime, and it also refuses to run when invoked with `ucm --new-runtime run foo`.

The output from `stack exec unison -- --new-runtime run .testbed.run_long`:
```
Just (#568rs -> (({[#fgaev]} #568rs)))
.testbed.run_long is not of type '{IO} ()
```
If I remove `--new-runtime`, it executes fine.

atacratic:
Interesting!  I was using the old runtime...

jaredly:
huh ok I'll make a separate ticket :)

samgqroberts:
@seagreen and I looked into this and I believe we made some progress. 

**TL;DR**: You can indeed `run` properly typed terms on trunk currently, but something seems off with automatic type inference - it gives a term as simple as `'(printLine "hello")` an unrunnable type.

<hr />

We confirmed that a term defined like this:
```
main = '(printLine "hello")
```
runs successfully on the M1m release, but gives a "this term has improper type" error on trunk (or, before #1814, a "i can't find this term" error).

We found that the `Typechecker.isSubtype` call in `HandleInput.addRunMain` that compares `mainType`, the most general runnable type, and `ty`, the type of the given term to run, is involved in the discrepancy - it returns that the `ty` of `main` is a subtype of `mainType` (as expected) on M1m, but not on trunk.

On trunk, we printed out the value of `mainType`:
```
(User "a". #568rs -> ({[#fgaev]} Var User "a"))
```
("for all `a`, a function from unit to `a` decorated with the `IO` ability")
and `ty`:
```
[...]

PR #1833, opened 2021-03-11
minor update to `run.md` transcript

samgqroberts:
## Overview

The `run.md` transcript was originally created (in #1814) assuming that #1800 meant that no terms could be `run` on `trunk` currently. Now it looks like terms can be run, but they need explicit type signatures (see https://github.com/unisonweb/unison/issues/1800#issuecomment-796820111). So this PR removes the #1800 concession from the `run.md` transcript.

Worthy to note though: the actual `printLine` output of the command in `run.md` is not showing up in `run.output.md`. Perhaps special treatment of IO output is required in the transcript runner for this.

### DIFF
--- a/unison-src/transcripts/run.md
+++ b/unison-src/transcripts/run.md
@@ -8,16 +8,15 @@ In a scratch file, we'll define two terms.
 - `badtype`: an improperly typed term that cannot be `run`
 
 ```unison scratch.u
+runnable : '{builtin.io.IO} ()
 runnable = '(printLine "hello!")
+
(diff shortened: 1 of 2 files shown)


aryairani:
> Worthy to note though: the actual `printLine` output of the command in `run.md` is not showing up in `run.output.md`. Perhaps special treatment of IO output is required in the transcript runner for this.

I think that's right; special treatment would be good.

aryairani (approved):
Thanks for the update!

PR #1834, opened 2021-03-13
Tests: explicit branch name in git init

ceedubs:
## Overview

Before this change, when I ran the tests I got the following "hints"
printed to the console:

```
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint:   git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint:   git branch -m <name>
```

I think that this means that the tests could potentially fail, depending
on the user's local git config and/or version.

I tried using the `trunk` branch and the tests failed. It seems that
the tests are dependent on the `master` branch being used, so I figured
it was probably be best to explicitly specify that branch name.

## Loose ends

This PR makes the dependency on the `master` branch name explicit, but is that the branch name that Unison wants to use? `trunk` is used elsewhere. Also I'm not sure whether this change will be relevant once the new repo format is released.

### DIFF
--- a/parser-typechecker/tests/Unison/Test/Git.hs
+++ b/parser-typechecker/tests/Unison/Test/Git.hs
@@ -175,7 +175,7 @@ testPull = scope "pull" $ do
 
   -- initialize git repo
   let repo = tmp </> "repo.git"
-  io $ "git" ["init", "--bare", Text.pack repo]
+  io $ "git" ["init", "--bare", "--initial-branch=master", Text.pack repo]
 
   -- run author/push transcript
   runTranscript_ tmp authorCodebase branchCache [iTrim|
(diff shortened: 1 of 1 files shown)


aryairani:
We want to use `trunk`, not `master`. I'm not sure where the dependency on the name `master` would be, I don't see it in the Haskell source, though it does show up in some transcripts, which is a separate issue.  This will still be relevant for some time after the new repo format is released; for as long as we are depending on Git for storage.

ceedubs:
@aryairani hmm I'm not sure where it's coming into play either. I tried explicitly passing `trunk` as the branch name to the `git clone` commands that I saw, but it didn't make the tests pass. It looks like maybe there could be some weirdness around checkout out an empty repo? I'm not sure though. I tried poking around a bit and this doesn't appear to be the quick fix that I thought that it was (at least with my limited knowledge of the Unison codebase). I'm not sure how worthwhile it is to go down too much of a rabbit hole on it?

aryairani (approved):
@ceedubs Yeah, I agree. Thanks for taking a look. I'm gonna merge it as is.  It does reduce the visibility of `master` but suppressing the error you saw.

PR #1836, opened 2021-03-13
Use mainline haskeline (0.8.1.1 instead of forked version)

ceedubs:
## Overview

Not needing to maintain a custom fork seems like a good thing. Also this is a step toward getting a Nix build of Unison that builds from source instead of simply wrapping a pre-built binary.

## Implementation notes

After fixing one minor change related to haskeline switching to using
`Control.Monad.Catch`, the tests are passing. I'm not sure whether there
was a more subtle reason that a custom haskeline was being used.

cc @maralorn who had brought this up

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/Main.hs
+++ b/parser-typechecker/src/Unison/CommandLine/Main.hs
@@ -9,6 +9,7 @@ import Unison.Prelude
 import Control.Concurrent.STM (atomically)
 import Control.Exception (finally, catch, AsyncException(UserInterrupt), asyncExceptionFromException)
 import Control.Monad.State (runStateT)
+import Control.Monad.Catch (MonadMask)
 import Data.Configurator.Types (Config)
 import Data.IORef
 import Data.Tuple.Extra (uncurry3)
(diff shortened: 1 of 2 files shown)


pchiusano:
I believe tab complete behavior in haskelline was busted in cases where the match wasn't a prefix of the completion. I think we reported a ticket so perhaps it's been fixed?

There's no unit tests for the tab complete behavior unfortunately - I'd try like, completing 'view List.map<tab>' to base.List.map.

aryairani:
I think ours [also?] has a patch for wrongly counting invisible ansi escape strings when aligning tab-complete listings: https://github.com/judah/haskeline/pull/115 (unmerged)

ceedubs:
Ah bummer. @aryairani it looks like that PR has been hanging around for quite some time, too :(

Okay well it was worth a try.

pchiusano:
Oh right! It's actually what Arya said, the thing I said we found a workaround that didn't require a patch to haskelline, I don't think.

aryairani:
@ceedubs Thanks though!  We could try to get that into mainline haskeline, but I'm not sure what obstacles there might be.

PR #1835, opened 2021-03-13
stack resolver nightly-2021-01-02 -> lts-17.5

ceedubs:
This also moves from ghc 8.10.3 to 8.10.4.

My main motivation was making some versions line up for a newer
packaging of haskell-language-server, but it seemed like it might
generally be good to move from a nightly to an lts version?

Issue #1782, opened 2021-01-05
use stack to install ghc in CI


aryairani:
I noticed when trying to update from 8.8.3 to 8.10.3 that I needed to update the version both in `stack.yaml` and in several spots in `ci.yaml`. Should we not just use `stack` to install whatever version of ghc is specified in the resolver?  Alternatively, can we set a variable with the ghc version in `ci.yaml`?

/cc @mitchellwrosen (Sorry I think you answered this once, but I lost it.)

mitchellwrosen:
That seems reasonable, the original reason for switching to `ghcup` to install GHC is given here: https://github.com/unisonweb/unison/pull/1761

aryairani:
Oh right! — Thanks.

aryairani:
@mitchellwrosen Do you have some time to knock this one out?  We should prioritize stack since the cabal build still isn't officially maintained/supported.

mitchellwrosen:
Sure, I can take care of it

PR #1837, opened 2021-03-13
does the macOS-11.0 CI image work yet?

aryairani:
inspired by #1835, we need to update CI so that we would know if such a change would break builds


aryairani:
After this, I think we should also #1782 and then #1835.

ceedubs (approved):
Yay more testing!

PR #1838, opened 2021-03-16
name.segments test: don't pick from an empty list

### LINKED ISSUE #1648: random seed 5399549024366200317 of `name.segments` test failed
`stack exec tests -- 5399549024366200317 "name.segments"`

fails in EasyTest.hs:
```
-- | Sample uniformly from the given list of possibilities
pick :: [a] -> Test a
pick as = let n = length as; ind = picker n as in do
  i <- int' 0 (n - 1)
  Just a <- pure (ind i) -- <------ here
  pure a
```

ceedubs:
Resolves #1648.

Technically I think that this has slightly different RNG characteristics
than the old implementation, but I wouldn't expect it to matter for
the sake of the test.

### DIFF
--- a/parser-typechecker/tests/Unison/Core/Test/Name.hs
+++ b/parser-typechecker/tests/Unison/Core/Test/Name.hs
@@ -20,10 +20,8 @@ test = scope "name" $ tests
     , scope "terms named `.`" $ expectEqual (suffixes "base..") ["base..", "."]
     ]
   , scope "segments" $ do
-    numDots <- int' 0 10
-    numSegs <- int' 0 10
-    n       <- int' 0 10
-    segs <- listOf n . pick $ replicate numDots "." ++ replicate numSegs "foo"
+    n    <- int' 0 10
+    segs <- listOf n $ pick [".", "foo"]
     expectEqual (segments $ Name . pack $ intercalate "." segs)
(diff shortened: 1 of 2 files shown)


pchiusano:
Thank you!!!

pchiusano (approved):
LGTM, I'd say merge once CI passes.

aryairani:
Wait, do we wanna do anything for other/future callers of `pick`, like fail with a particular message when an empty list is given?

pchiusano:
Yeah, that could be a good idea @ceedubs if you want you can edit `EasyTest.hs` in yaks/ to at least bomb with a more meaningful error when `pick` is called with an empty list. 

Or I think it's okay to merge as is since it addresses the original issue of the test having spurious failures due to bad test generation.

ceedubs:
@pchiusano @aryairani That makes sense. I've pushed a commit to give a better error message when `pick` is called with an empty list.

In fact I originally had added an `_ <- expect (n > 0)` but I removed it when I realized that the error message wasn't any more helpful. I used an if/else expression here, but I was thinking that it might be handy to have something like `expect` but that takes an error message to `crash` with if the expectation isn't met.

aryairani:
Thanks @ceedubs!

PR #1839, opened 2021-03-16
Use stack to install GHC in CI (fixes #1782)

### LINKED ISSUE #1782: use stack to install ghc in CI

mitchellwrosen:
## Overview

This PR switches from installing GHC with `ghcup` to installing GHC with `stack` - https://github.com/unisonweb/unison/issues/1782

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,16 +26,6 @@ jobs:
       # a cache. GitHub will automatically delete caches that haven't been accessed in 7 days, but there is no way to
       # purge one manually.
 
-      # Cache ghc installed by ghcup
-      - uses: actions/cache@v2
-        name: cache ghc installed by ghcup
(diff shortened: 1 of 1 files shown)


aryairani:
😃

mitchellwrosen:
@aryairani Ok, I force-pushed just to see if the caching was even working (heh).

Looks good to go 👍

aryairani (approved):

aryairani:
Thanks!

PR #1845, opened 2021-03-17
allow port / token / ip for codebase server to come from environment …

stew:
## Overview

The current `ucm` starts up a web browser on a randomly picked port and only on 127.0.0.1, and with a randomly selected token. This PR effects `ucm` to check three new environment variables:

UCM_PORT - port to bind to (must be 1ish-65535ish)
UCM_HOST - an ip address or "*" meaning "all" or "!4" meaning "all ipv4" etc, as defined here: https://hackage.haskell.org/package/streaming-commons-0.2.2.1/docs/Data-Streaming-Network.html#t:HostPreference
UCM_TOKEN - the "secret" shared token

we should really look into unifying all the of the config into a thing where any of then can be statically configured, or configured with env vars or passed on the command line. so like if

    ucm --port 80

is a thing, than so is

    env UCM_PORT=80 ucm

as is `echo port=80 > .config/ucm/config.ini ; ucm` or something

## Test coverage

I swear it works for me

## Loose ends

to be tied up in forthcoming PRs that start making use of this feature in artifacts built by our forthcoming Continuous Deployment Server ™

### DIFF
--- a/parser-typechecker/src/Unison/Server/CodebaseServer.hs
+++ b/parser-typechecker/src/Unison/Server/CodebaseServer.hs
@@ -9,6 +9,7 @@ module Unison.Server.CodebaseServer where
 import           Data.Aeson                     ( )
 import qualified Data.ByteString.Lazy          as Lazy
 import qualified Data.ByteString               as Strict
+import qualified Data.ByteString.Char8         as C8
 import           Data.OpenApi                   ( URL(..)
                                                 , Info(..)
                                                 , License(..)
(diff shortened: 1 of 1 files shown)


pchiusano (approved):
Looks good. Does this revert to the old behavior if those environment variables aren't given? I think that will be a common case when running within ucm for local dev.

stew:
yeah, if the env vars are not present, it goes back to the current version, binding to 127.0.0.1, picking a random token and a random port

hojberg:
💯 for a a more uniform configuration scheme — I'd been planning to use dotenv for the UI, not sure it applies here, but wanted to bring it on the radar at least.

PR #1848, opened 2021-03-18
Fix decompilation of 'large' constructors + extra

### LINKED ISSUE #1844: Decompilation in new runtime reverses output of constructors?
See this minimized example: https://github.com/unisonweb/unison/issues/1844#issuecomment-801344682

----

Original report, disregard -

The new runtime seems to mix up bound variables during pattern matching. The issue does not arise with the old runtime.

```Haskell
type One a m = One a m 
type Deep l m r x = Deep l m r x
unique type Z = Z

snoc f k aN mN = 
  go k aN mN = match k with 
    One a0 m0 -> Deep (One a0 m0) [] (One aN mN) (f 0 m0 mN)
  go k aN mN

> snoc (_ _ _ -> Z) (One 1 Z) 2 Z
```

Yields: 

```
    12 | > snoc (_ _ _ -> Z) (One 1 Z) 2 Z
           ⧩
           Deep Z (Cons 2 Z) [] (Cons 1 Z)
```

The example above is the result of some minimization - it isn't triggered in the obvious cases, and I'm not sure what's unique about this case.

dolio:
This fixes decompilation of data type constructors of arity greater than 3 by fixing a pattern synonym that that uses. The synonym was also used in serialization, so strange results could have been observed if the binary format were inspected directly, although it would otherwise be consistent.

This also includes some modifications to make the transcript test runnable with a `--new-runtime` flag. Previously this would do nothing different, and I think it was omitted because there was one old IO-based transcript. I moved that into an `old-runtime-transcripts` directory and made the test always run that directory with the old runtime, similarly to the `new-runtime-transcripts` directory.

Since transcripts are now actually run with the new runtime, I also needed to fix the `bug` and `todo` implementation in the new runtime to give the same error message as the old runtime, since there is a transcript that uses that.


pchiusano (approved):
Great!

PR #1849, opened 2021-03-21
Allow filtering of transcript tests by prefix

bontaq:
## Overview

This allows people to use `stack exec transcripts -- prefix-of-test` to only run transcripts which match the prefix.

## Implementation notes

Adds a `TestConfig` and parses arguments into it.  Order for `--new-runtime` and `somePrefix` shouldn't matter, 
`stack exec transcripts -- --new-runtime prefix` and `stack exec transcripts -- prefix --new-runtime` both work.

I made a slight change to `buildTests` because, if in the first directory it searched there were no matching transcripts, it would exit early and skip searching the rest of the transcript folders.  So now we just `pure ()` if there are no matches, which allows it to continue.

Nothing too controversial, code-style wise it's up to whichever is preferred for passing around configs.  It'd be easy to destructure the `TestConfig` and pass in only `matchPrefix` and `useNewRuntime` to `buildTests` and `testBuilder`, respectively.

I also wasn't sure if I should hoist `handleArgs` somewhere so that `parser-typechecker/tests/Suite.hs` and `parser-typechecker/transcripts/Transcripts.hs` can share the same args parsing.  Since it's only in two places, I left it local.

## Test coverage

Tested locally with 
* `stack exec transcripts`

image
... the usual ... and:
image

* `stack exec transcripts -- todo`

image

* `stack exec transcripts -- ucm --new-runtime`

image

* `stack exec transcripts -- --new-runtime ucm`

image

* `stack exec transcripts -- wigwam` (to see how it looks if nothing matches)

image

I think it's currently untested, so it'd be good to test, but tricky to implement.

### DIFF
--- a/development.markdown
+++ b/development.markdown
@@ -11,8 +11,12 @@ To get cracking with Unison:
 * [Install `stack`](https://docs.haskellstack.org/en/stable/README/#how-to-install).
 * Build the project with `stack build`. This builds all executables.
 * After building, `stack exec unison` will fire up the codebase editor, create a codebase in the current directory, and watch for `.u` file changes.  If you want to run it in a different directory, just add `unison` to your `PATH`, after finding it with `find .stack-work -name unison -type f`.  (For me, this finds two, they both work, but have different contents.  ¯\\\_(ツ)\_/¯ )
+
+## Running Tests
+
 * `stack exec tests` runs the tests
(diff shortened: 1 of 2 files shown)


aryairani (approved):
Awesome! I like the `TestConfig`. I wish we had something like that for ucm’s Main :) *hint hint*

PR #1850, opened 2021-03-23
Allow redundant commas in list parsing

### LINKED ISSUE #1841: Allow multiple commas between elements of a list, and allow a trailing comma
This comes up when rearranging or deleting elements of a list. I can't see any reason not to just allow ignored optional commas. The current UX is you get an annoying or cryptic parse error. Once the code gets added to the codebase, it will get nicely formatted with the proper number of commas but it's okay to have some sloppiness at parse time.

You can check out `Parser.seq` in `Parser.hs` which is the only place that would need modification I think.

```Haskell
seq :: Ord v => (Ann -> [a] -> a) -> P v a -> P v a
seq f p = f' <$> reserved "[" <*> elements <*> trailing
  where
    f' open elems close = f (ann open <> ann close) elems
    trailing = optional semi *> reserved "]"
    sep = P.try $ optional semi *> reserved "," <* optional semi
    elements = sepBy sep p
```

JohannesHuster:
## Overview

Possibly resolves https://github.com/unisonweb/unison/issues/1841

Allows multiple commas leading, trailing and in between elements of lists.

## Implementation notes

I wasn’t certain about the rules regarding virtual semicolons. Duplicate semicolons are currently allowed in my solution. I did not see a problem with it, but it might be something Unison wants to disallows in general?

I don’t have a lot of experience with Haskell. If you have feedback on how to improve the solution, I’m happy to have another go!

## Test coverage

I’m happy to add some tests for the new cases, but wanted to see first, if the behavior is what you had in mind. For now, I tried to break the parsing every way I thought of.

## Loose ends

See Test Coverage.


pchiusano:
Great! For test coverage, I'd just add a few examples to the file unison-src/tests/methodical/literals.u and update the corresponding literals.ur. 

And then add yourself to CONTRIBUTORS.markdown. :)

JohannesHuster:
I’ve added a few tests. Thanks for the pointers :-)

pchiusano (approved):
Great, thanks!

PR #1851, opened 2021-03-29
Avoid unnecessary floating of some lambdas

dolio:
- The old floater would float even
     `x -> e`
  into:
  ```
  let f x = e
      f
  ```
  which is quite unnecessary.

- It is also unnecessary to float a lambda immediately in the body of a
  let rec, because the entry point for a compiled let rec can take
  variables.

- The old floating behavior was essentially ensuring that lambda
  expressions could not be decompiled (currently), because they are
  turned into references to internal bindings of a top level combinator.

- Note that the transcript change is because we are serializing a
  different value, not because serialization has changed.


pchiusano (approved):
Sweet, thanks man!

PR #1852, opened 2021-03-31
Keep track of more decompilation information.

dolio:
In addition to the whole term for a combinator, each floated expression
is remembered so it can be used in decompilation when they are partially
applied.

@pchiusano Let me know if this works well enough for your doc work, or if it needs additional tweaks.


pchiusano (approved):
Sweet. I'll give it a go. Appreciate the quick turnaround.

Issue #1256, opened 2020-02-19
The @[source] documentation tag does not align the source


pete-ts:
Minor/aesthetic only

```
## Example
  @[source] base.tree.inOrder
```
does not count the whitespace in front of the @[source] tag, ending up in misalignment.

```
  ## Example
    base.tree.inOrder : Tree a -> [a]
  base.tree.inOrder t =
  -- the code is not indented in the same depth as @[source]
```

ceedubs:
I believe that this was resolved by #1853.

Issue #1051, opened 2019-12-11
Unison silently uses stale documentation when a Doc value is updated


anovstrup:
When the documentation linked to a particular definition is updated, Unison should automatically update the link (consistent with the behavior for type-compatible changes) or at least trigger refactoring to let the user know that the link is stale.

anovstrup:
Transcript here: https://gist.github.com/anovstrup/fe0fd0fb8ffd264ab7ff03c05639b6d3

Issue #1050, opened 2019-12-10
Updating documentation results in broken state


anovstrup:
In M1g, updating the definition of a Doc term associated with a definition results in a state where the `docs` command cannot be used on the definition (due to the name conflict with the old Doc definition) but where the old Doc definition can neither be deleted nor unlinked from the documented definition.

To reproduce, first create a definition with documentation:

```unison
docs.foo = [: foo is the number 3.14 :]
foo = 3.14
```

Now add the definitions, link foo to docs.foo, and view the docs:

```ucm
.> add
.> link foo docs.foo
.> docs foo
```

Now edit the docs:

```unison
docs.foo = [: foo is the constant 3.14, an estimate of pi :]
```

Update the definition, update the link, and try to view the docs:
```ucm
.> update
.> link foo docs.foo
.> docs foo
```

Unlink the old definition and try to view the docs (this fails):
```ucm
.> unlink foo docs.foo#qldo40h52u
.> docs foo
```

Delete the old definition and try to view the docs (this also fails):
```ucm
.> delete.term docs.foo#qldo40h52u
.> docs foo
```

anovstrup:
Note: the issue does not arise if `unlink` is performed _before_ `update`.

anovstrup:
gist of the transcript: https://gist.github.com/anovstrup/7df1853fbf7368906faa07cf5e476b79

PR #1853, opened 2021-04-01
New documentation format

### LINKED ISSUE #1771: Documentation v2 Design Writeup
We've learned a lot from the existing documentation support and are ready for version 2. The new documentation type will be more structured and thus easy to give good feedback and errors while authoring, and easy to render to different targets (like rendering as plain text at various widths or HTML or Javascript-y HTML), and will be seamless to write and update. We want it to be delightful experience!

In this design, documentation can be unnamed, added inside double braces `{{ }}` before any definition. 

Within each `{{ }}`, opening another `{{ }}` escapes out to regular Unison code, making it easy to generate documentation fragments programmatically. There's no limit to the nesting, so this is confusing but totally valid: 

    {{ This is a {{ (x -> x) {{ nested }} }} documentation block. }}

Which is equivalent to:

    {{ This is a nested documentation block. }}
    
Here's a more realistic example showing how you might document a definition using an anonymous documentation block. Unlike before, you won't have to add `List.map` to the codebase first before writing the docs:

```
{{
 `List.map f xs` applies `f` to each element of `xs`.
 
 ## Examples
 
 {{ example '(List.map (x -> x*10) [1,2,3,4]) }}
 
 {{ example '(List.map (x -> "hi") []) }}
 
 ## Also see
 
 * @List.filter
 * @List.fold
}}
List.map : (a -> b) -> [a] -> [b]
List.map f xs = 
  go acc = cases
    [] -> acc
    h +: t -> go (acc :+ f h) t
  go [] xs
```

The anonymous documentation before a binding, `List.map` is sugar for `List.map.doc = {{ .. }`, and the `docs List.map` command will look for `List.map.doc` if no documentation is explicitly linked to `List.map`. This avoids a tedious linking step for the most common case where you're following the standard convention of where to put the docs.

In addition, saving a scratch file `blah.u`, UCM (re)generates `blah.u.{md,html}` containing markdown and html rendered versions of any documentation elements declared in the file. So you can easily preview your work without needing to add anything to the codebase.

### Syntax

Within documentation blocks, we do a full parse of the text into an AST like the below (DRAFT, still tweaking). Unlike the current `Doc` type, it's very structured, so it's easy to generate good feedback during authoring and also target multiple backends. I want parse and type errors at the time you're authoring the documentation, not some nonsense or runtime error later when you go to view the documentation.

<details><summary>Proposed documentation AST, still being evolved</summary>

```Haskell
unique[e88630dc5277b2e84086cca85565494ae802862c] type Doc2 f
  = TypeLink Link.Type
  | TypeSource Link.Type
  | TermLink Any 
  | TermSource Any
  | TermSignature Any
  | Evaluate Any
  | Trace Any
  | Example Nat Any -- Example 3 (Any (x y z -> x + y + z)) will render as `x + y + z`.
  | Element (f (Doc2 f))
  -- no need for transclusion, just evaluation will do it
  -- what about anchors?

unique[fadb42d14e8e64c65713c237faf6604fd8aaea40] type Doc2.Element a 
  = Word Text
  | Code a 
  | Bold a 
  | Italic a
  | Quote a
  | Embed Any -- Embed (Any (Img "Rad image alt text" "https://xkcd.com/1433/")) 
  | Block a
  | Blankline
  | Aside (Optional a) a -- Aside title content
  | Folded Boolean a a -- Folded isFolded whenFolded whenUnfolded
  | Link a a -- Link destination title
  | Paragraph [a]
  | Unordered [a]
  | Ordered Nat [a] -- `Nat` is the starting number
  | Section a [a] -- Section title children
```
</details>

Some interesting stuff to note:

* Whitespace is ignored within paragraphs, though blanklines are preserved. For instance `{{ the   dog }}` is just parsed as `paragraph [word "the", word "dog"]`. 
* Inline examples are actually parsed and typechecked. When you write inline `List.map id xs`, this actually gets parsed and when you view docs that make use of inline examples, it will use the latest names and you can be assured as a reader that what's there isn't nonsense. Notes:
  * For raw inline code formatting, perhaps put an `r` immediately before the backticks, as in: ``r`List.map id xs` ``
  * You can start these with multiple backticks and finish with the same number of backticks.
  * In `List.map id xs`, the parser will presume `List.map`, `id`, and `xs` are actual values and during typechecking will complain if they don't exist. If you want any of these to just be variables local to the example, you can say `List.map id $xs`, prefixing that variable with a `$`. (Also idea: treat the function being called as a literal and the arguments as variables by default?)
* The `Embed` element allows open ended embedding of values of various types. Not all renderers will support all embedded elements, and rendering is best effort.

__An important point:__ the parser won't actually generate the AST directly, instead, it will call named functions (like `Doc.paragraph` and `Doc.word`), allowing this syntax to be overridden (even local to a function) for other documentation-like types. We'd have these names be by default bound to functions that produce the above AST though that could be changed by the user if they want.

We can add more syntax and special elements over time. For instance, even with the base syntax, you could do `{{ bold {{hi}} }}` to generate some bold text within a document, but nothing stops us from adding `__hi__` as syntax for that now or later (I'd probably add this one now since it's familiar from markdown and pretty handy, but you get the idea).

## Other notes

The pretty-printer should render documentation source as appropriate.

Open to suggestions on syntax or other bikeshedding!

Since the documentatation is a fully parsed tree, rendering it a different widths or indented is easy (unlike currently, where that logic has gotten pretty complicated and still doesn't work that great). 

We can also render it to HTML or whatever.

What happens to the old syntax? We can keep it around if we want, or delete it. I'd probably keep it around for a deprecation cycle, since it's easier to convert from the old documentation syntax to the new one vs the old documentation AST to the new one (the AST form is hard to edit).

Why `{{`? Any matching brace-like thing will do. The double braces is available and is easy to type, and it looks pretty good to me. Technically, the syntax conflicts with a nested ability list, like you could write `a ->{{IO},{Store Nat}} b`, but that has the same meaning as `a ->{IO, Store Nat} b` so I don't see a downside to stealing this syntax.

pchiusano:
This PR adds the new documentation format! 

It's been a joy to use. Docs now have a lot more features and no longer need to be explicitly linked to definitions. Instead, `docs foo` will look up `foo.doc`, either in the current file or in the codebase. (Though you can still do explicit linking if you prefer)

__I highly recommend [checking out this transcript](https://github.com/unisonweb/unison/blob/f4e99cbfc2bbb6b7a7a47834286f3ae616f020a3/unison-src/new-runtime-transcripts/doc.output.md) for an overview.__

Here's the current console rendering, but @hojberg will be working on rendering this in the front-end using rich HTML/JS:

Screen Shot 2021-04-08 at 2 47 42 PM

Closes #1771 #1497 #1050 #1051 #1256

## Implementation notes

See detailed self review.

A couple misc other notes:

There isn't really a clear split between parsing and lexing with this sort of syntax, because you can do arbitrarily deep escaping and unescaping. For instance, consider:

```Haskell
{{ This is a doc, but {{ someUnisonFn {{ which takes a doc as an arg }} }} }}
```

So it was sort of awkward to split the parsing task up into these arbitrary two phases. I came up with an okay way of doing it where most of the work is done by the lexer, and the parser is pretty straightforward but has some duplicate structure.

Another interesting decision is that console formatting is done with Unison code (currently lives in `IOSource`), so aside from special elements (like source rendering, which is provided by the Haskell code), it can be customized pretty easily. See the function `formatConsole` in `IOSource`.

## Test coverage

Testing is pretty decent. I added transcripts to exercise everything. That said I wouldn't be surprised to find parsing or pretty-printing bugs later as people start using it more.

## Loose ends

- [x] Including source that's only in the scratch file but not the codebase doesn't work
- [x] Display gets mangled in nested documents, likely needs some grouping added
- [x] Investigate possible regression when displaying explicitly linked docs

Tracking remaining items in #1864.


parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -381,6 +381,10 @@ builtinsSrc =
   , B "Text.take" $ nat --> text --> text
   , B "Text.drop" $ nat --> text --> text
   , B "Text.size" $ text --> nat
+  , B "Text.repeat" $ nat --> text --> text
+  , B "Text.alignLeftWith" $ nat --> char --> text --> text
+  , B "Text.alignRightWith" $ nat --> char --> text --> text
+  , B "Text.alignCenterWith" $ nat --> char --> text --> text
```
pchiusano: Added some unit tests in [builtins.output.md](https://github.com/unisonweb/unison/pull/1853/files#diff-ea4a4a67c330ab1608e48937be8fde657c8329fa0fbd2019daa6df1fdb8441fe).

parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -149,12 +154,10 @@ data Command m i v a where
 
   Evaluate :: PPE.PrettyPrintEnv
            -> UF.TypecheckedUnisonFile v Ann
-           -> Command m i v (Either Runtime.Error
-                ([(v, Term v ())], Map v
-                (Ann, UF.WatchKind, Reference, Term v (), Term v (), Runtime.IsCacheHit)))
+           -> Command m i v (Either Runtime.Error (EvalResult v))
```
pchiusano, line 157: I gave this a type alias since I needed to reference it elsewhere.

parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -149,12 +154,10 @@ data Command m i v a where
...
 
   -- Evaluate a single closed definition
-  Evaluate1 :: PPE.PrettyPrintEnv -> Term v Ann -> Command m i v (Either Runtime.Error (Term v Ann))
+  Evaluate1 :: PPE.PrettyPrintEnv -> UseCache -> Term v Ann -> Command m i v (Either Runtime.Error (Term v Ann))
```
pchiusano, line 160: You can now specify whether `Evaluate1` should use the watch expression cache or not. Stuff that uses `Evaluate1` to do I/O passes `False`, and stuff that doesn't can (if desired) pass `True`. 

Docs that contain embedded terms actually use the cache to avoid repeatedly evaluating those terms.

parser-typechecker/src/Unison/Codebase/Editor/Command.hs
```diff
@@ -224,6 +227,16 @@ data Command m i v a where
...
 
+type UseCache = Bool
+
+type EvalResult v =
+  ( [(v, Term v ())]
+  , Map v (Ann, UF.WatchKind, Reference, Term v (), Term v (), Runtime.IsCacheHit)
+  )
+
+lookupEvalResult :: Ord v => v -> EvalResult v -> Maybe (Term v ())
+lookupEvalResult v (_, m) = view _5 <$> Map.lookup v m
```
pchiusano, line 238: This is the most common thing you are doing with the result of evaluation: looking up a watch expression result by variable name. Just added a helper function for it to clean things up.

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -178,10 +178,21 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
       lift . runExceptT $ Backend.definitionsBySuffixes mayPath branch codebase query
     FindShallow path -> lift . runExceptT $ Backend.findShallow codebase path
 
-  eval1 :: PPE.PrettyPrintEnv -> Term v Ann -> _
-  eval1 ppe tm = do
+  watchCache (Reference.DerivedId h) = do
+    m1 <- Codebase.getWatch codebase UF.RegularWatch h
+    m2 <- maybe (Codebase.getWatch codebase UF.TestWatch h) (pure . Just) m1
+    pure $ Term.amap (const ()) <$> m2
+  watchCache Reference.Builtin{} = pure Nothing
```
pchiusano, line 185: This `watchCache` function was previously just used by `evalUnisonFile`, but now it's also used by `eval1` so just pulled it out.

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -178,10 +178,21 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
...
+  eval1 :: PPE.PrettyPrintEnv -> UseCache -> Term v Ann -> _
+  eval1 ppe useCache tm = do
     let codeLookup = Codebase.toCodeLookup codebase
-    r <- Runtime.evaluateTerm codeLookup ppe rt tm
+        cache = if useCache then watchCache else Runtime.noCache
+    r <- Runtime.evaluateTerm' codeLookup cache ppe rt tm
+    when useCache $ case r of
+      Right tmr -> Codebase.putWatch codebase UF.RegularWatch (Term.hashClosedTerm tm)
+                                     (Term.amap (const Parser.External) tmr)
+      Left _ -> pure ()
```
pchiusano, line 195: Just plumbing to use and populate the cache if `useCache` is on.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -278,7 +280,8 @@ loop = do
           Nothing -> respond $
             ParseErrors text [ err | Result.Parsing err <- toList notes ]
           Just (Left errNames) -> do
-            ppe <- prettyPrintEnv =<< makeShadowedPrintNamesFromHQ hqs errNames
+            ns <- makeShadowedPrintNamesFromHQ hqs errNames
+            ppe <- prettyPrintEnv (Names3.suffixify ns)
```
pchiusano, line 284: These two lines mean that error messages use suffixed names rather than FQNs. A big improvement IMO.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -291,10 +294,9 @@ loop = do
         let lexed = L.lexer (Text.unpack sourceName) (Text.unpack text)
         withFile [] sourceName (text, lexed) $ \unisonFile -> do
           sr <- toSlurpResult currentPath' unisonFile <$> slurpResultNames0
-          names <- makeShadowedPrintNamesFromLabeled
-                      (UF.termSignatureExternalLabeledDependencies unisonFile)
-                      (UF.typecheckedToNames0 unisonFile)
-          ppe <- PPE.suffixifiedPPE <$> prettyPrintEnvDecl names
+          names <- displayNames unisonFile
+          pped <- prettyPrintEnvDecl names
+          let ppe = PPE.suffixifiedPPE pped
```
pchiusano, line 299: Using `displayNames` helper function instead of that `makeShadowedPrintNamesFromLabeled` complicated expression, which was inlined all over the place in this file. :)

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -630,6 +630,37 @@ loop = do
               diffHelper (Branch.head root') (Branch.head root'') >>=
                 respondNumbered . uncurry ShowDiffAfterDeleteDefinitions
             else handleFailedDelete failed failedDependents
+
+        displayI outputLoc hq = do
```
pchiusano, line 634: This function is the logic of the `display` command, but it's pulled out as a function because functions are fun! 

And it gets reused by `docs`. So `docs foo` will end up calling `displayI ConsoleLocation foo.doc`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1026,15 +1057,47 @@ loop = do
...
+              numberedArgs .= fmap (HQ.toString . view _1) out
+              respond $ ListOfLinks ppe out
+
+        codebaseByName = do
+          parseNames <- Names3.suffixify0 <$> basicParseNames0
+          case Names3.lookupHQTerm dotDoc (Names3.Names parseNames mempty) of
+            s | Set.size s == 1 -> displayI ConsoleLocation dotDoc
+              | Set.size s == 0 -> respond $ ListOfLinks mempty []
+              | otherwise       -> -- todo: return a list of links here too
+                respond $ ListOfLinks mempty []
```
pchiusano, line 1104: This todo needs a better treatment. I'll have a go at that, either now or in follow up PR.

parser-typechecker/src/Unison/Codebase/Path.hs
```diff
@@ -454,3 +454,12 @@ instance Resolve Absolute HQSplit HQSplitAbsolute where
...
   resolve a (Path' (Right r)) = resolve a r
+
+instance Convert [NameSegment] Path where convert = fromList
+instance Convert Path [NameSegment] where convert = toList
+instance Convert HQSplit (HQ'.HashQualified Path) where convert = unsplitHQ
+instance Convert Path Name where convert = toName
+instance Convert Path' Name where convert = toName'
+instance Convert HQSplit' (HQ'.HashQualified Path') where convert = unsplitHQ'
+instance Parse Name HQSplit' where parse = hqSplitFromName'
+instance Parse Name Split where parse = splitFromName
```
pchiusano, line 465: I was going crazy trying to figure out how to convert between all the various name and path-like things, so added this. `Name.convert thing1 :: Path` converts `thing1` to a `Path`, if that's a reasonable thing.

There's also `Name.parse thing1 :: Maybe Path` if converting `thing1` to a `Path` isn't guaranteed to succeed.

parser-typechecker/src/Unison/Codebase/Runtime.hs
```diff
@@ -120,21 +119,33 @@ evaluateWatches code ppe evaluationCache rt uf = do
...
-             (Map.singleton UF.RegularWatch [(Var.nameds "result", tm)])
-  runnable <-
-    if needsContainment rt
-      then Codebase.makeSelfContained' codeLookup uf
-      else pure uf
-  r <- evaluateWatches codeLookup ppe noCache rt runnable
-  pure $ r <&> \(_,map) ->
-    let [(_loc, _kind, _hash, _src, value, _isHit)] = Map.elems map
-    in value
+evaluateTerm' codeLookup cache ppe rt tm = do
```
pchiusano, line 130: Some wrangling here to allow evaluation of a single term to use the cache.

parser-typechecker/src/Unison/CommandLine/DisplayValues.hs
```diff
@@ -1,52 +1,251 @@
...
-           -> (Reference -> m (Maybe (Term v a)))
-           -> (Referent -> m (Maybe (Type v a)))
-           -> (Reference -> m (Maybe (Term v a)))
-           -> (Reference -> m (Maybe (DD.Decl v a)))
-           -> Term v a
+           -> (Reference -> m (Maybe (Term v ())))
+           -> (Referent -> m (Maybe (Type v ())))
+           -> (Term v () -> m (Maybe (Term v ())))
+           -> (Reference -> m (Maybe (DD.Decl v ())))
+           -> Term v ()
```
pchiusano, line 42: A bunch of rejiggering here to allow use of the watch cache for embedded terms inside a doc.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -602,18 +953,17 @@ stanzas = go [] where
...
-      Reserved "use" -> 0
-      _ -> 3 :: Int
+reorder = join . sortWith f . stanzas where
+  f [] = 3 :: Int
+  f (t0 : _) = case payload $ headToken t0 of
+    Open "type" -> 1
+    Open "unique" -> 1
+    Open "ability" -> 1
+    Reserved "use" -> 0
+    _ -> 3 :: Int
```
pchiusano, line 966: No relevant change here. 

I originally was exploring a different approach to allowing anonymous docs before type decls, which involved hacking up this function. I ended up coming up with a nicer way I think but didn't manage to exactly reproduce the old formatting when I changed it back.

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -188,14 +188,13 @@ renderTypeError e env src = case e of
...
-        , style Type1 (renderType' env expectedType)
-        , " and another is "
-        , style Type2 (renderType' env foundType)
-        , ":"]
+    [ Pr.lines
+        [ Pr.wrap preamble
+        , ""
+        , "Here, one   is:  " <> style Type1 (renderType' env expectedType)
+        , "and another is:  " <> style Type2 (renderType' env foundType)
+        , ""
```
pchiusano, line 196: Minor thing unrelated to this PR: 

When showing two types for comparison, this ensures they are left aligned. Applied similar treatment elsewhere in this file. Did this when staring at some cryptic type errors.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1003,6 +1004,33 @@ boxToBool = inBx arg result
...
+--          wordWordBoxDirect Ty.floatRef Ty.natRef
+wordWordBoxDirect :: Reference -> Reference -> ForeignOp
+wordWordBoxDirect wordType1 wordType2 instr
+  = ([BX,BX,BX],)
+  . TAbss [b1,b2,b3]
+  . unbox b1 wordType1 ub1
+  . unbox b2 wordType2 ub2
+  $ TFOp instr [ub1,ub2,b3]
+  where
+  (b1,b2,b3,ub1,ub2) = fresh5
```
pchiusano: Calling conventions used by new text builtins, though they are generally useful for other stuff.

parser-typechecker/src/Unison/Runtime/Decompile.hs
```diff
@@ -56,7 +57,7 @@ decompile topTerms (DataC rf ct [] bs)
   = apps' (con rf ct) <$> traverse (decompile topTerms) bs
 decompile topTerms (PApV (CIx _ rt k) [] bs)
   | Just t <- topTerms rt k
-  = substitute t <$> traverse (decompile topTerms) bs
+  = Term.etaReduceEtaVars . substitute t <$> traverse (decompile topTerms) bs
```
pchiusano, line 60: The runtime does eta expansion of constructors. The `etaReduceEtaVars` removes them again via eta reduction.

parser-typechecker/src/Unison/Runtime/Rt1IO.hs
```diff
@@ -461,7 +461,7 @@ runtime = Runtime terminate eval (nullaryMain External) (nullaryTest External) T
   terminate :: IO ()
   terminate = pure ()
   eval cl' ppe term = do
-    let cl = void (hoist (pure . runIdentity) IOSrc.codeLookup) <> cl'
+    let cl = cl' <> void (hoist (pure . runIdentity) IOSrc.codeLookup)
```
pchiusano, line 464: Swapping the order of these code lookups just ensured that `IOSrc` isn't loaded until it's needed. This code is going away soon anyway when we get rid of the old runtime.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -216,7 +220,7 @@ pretty0
                     in uses $ [pretty0 n (ac 0 Block im' doc) tm]
     App' x (Constructor' DD.UnitRef 0) ->
       paren (p >= 11) $ (fmt S.DelayForceChar $ l "!") <> pretty0 n (ac 11 Normal im doc) x
-    LamNamed' v x | (Var.name v) == "()" ->
+    Delay' x  ->
```
pchiusano, line 223: Keying off the variable name is a little fragile. Now we're using the `Delay'` pattern, which treats any unary lambda that ignores its argument as a delay.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -1201,3 +1209,285 @@ toBytes (App' (Builtin' "Bytes.fromList") (List' bs)) =
...
         go _ = Nothing
 toBytes _ = Nothing
+
+prettyDoc2
+  :: forall v . Var v
+  => PrettyPrintEnv
+  -> AmbientContext
+  -> Term3 v PrintAnnotation
+  -> Maybe (Pretty SyntaxText)
+prettyDoc2 ppe ac tm = case tm of
```
pchiusano, line 1219: Tedious pretty-printing code for docs. Nothing super interesting here.

parser-typechecker/src/Unison/Util/ColorText.hs
```diff
@@ -82,7 +92,16 @@ toANSI (AnnotatedText chunks) =
...
+  toANSI c = pure $ ANSI.setSGRCode (toANSI' c)
+
+  toANSI' :: Color -> [ANSI.SGR]
+  toANSI' c = case c of
+    Default  -> []
+    Background c c2 -> (setBg <$> toANSI' c) <> toANSI' c2 where
+      setBg (ANSI.SetColor ANSI.Foreground intensity color) =
+             ANSI.SetColor ANSI.Background intensity color
+      setBg sgr = sgr
+    Invert c -> [ANSI.SetSwapForegroundBackground True] <> toANSI' c
```
pchiusano, line 104: This change allows text that is both bold and underlined, or other combinations of ANSI terminal formatting.

unison-src/new-runtime-transcripts/_base.md
```diff
@@ -13,215 +13,7 @@ transcripts which contain less boilerplate.
...
-socketPort = compose reraise socketPort.impl
-newClient = compose2 reraise newClient.impl
-MVar.take = compose reraise take.impl
-MVar.put = compose2 reraise put.impl
-MVar.swap = compose2 reraise MVar.swap.impl
-```
-
-
-```ucm:hide
+.> load unison-src/new-runtime-transcripts/base.u
```
pchiusano, line 16: Switched to loading `base.u` from a file, which is handy when doing development of `base.u`.

unison-src/new-runtime-transcripts/codeops.md
```diff
@@ -104,7 +104,7 @@ f x = h zap x
 fVal : Value
 fVal = Value.value f
 
-fDeps : [Term]
+fDeps : [Link.Term]
```
pchiusano, line 107: The longer name here is due to there being `Doc2.Term`, used for embedded terms.

unison-src/transcripts/destructuring-binds.output.md
```diff
@@ -95,7 +95,7 @@ ex4 =
   
       2 |   (a,b) = (a Nat.+ b, 19)
   
-  Whatever it is, it has a type that conforms to builtin.Nat.
+  Whatever it is, it has a type that conforms to Nat.
```
pchiusano, line 98: Error messages use suffixed names now. :)

unison-src/transcripts/docs.output.md
```diff
@@ -163,22 +163,7 @@ Or there's also a convenient function, `docs`, which shows the `Doc` values that
...
-    List.take.ex1 = builtin.List.take 0 [1, 2, 3, 4, 5]
-    🔽
-    ex1 = []
-  
-  
-    List.take.ex2 = builtin.List.take 2 [1, 2, 3, 4, 5]
-    🔽
-    ex2 = [1, 2]
-  
+  docs.List.take
```
pchiusano: Possible regression here. I'll take a look.

unison-src/transcripts/fix1356.output.md
```diff
@@ -85,10 +85,6 @@ Let's check that we see both docs:
 ```ucm
 .> docs x
 
-  1. x.doc       : Doc
-  2. #v8f1hhvs57 : Doc
-  
-  Tip: Try using `display 1` to display the first result or
-       `view 1` to view its source.
+  I am the documentation for x, and I now look better
```
pchiusano, line 88: This change is just due to it looking in the file for docs, not just the codebase.

pchiusano:
This is ready for review! I have a couple loose ends I'm tracking down which are noted in the writeup but that shouldn't affect your review.

parser-typechecker/src/Unison/Builtin.hs
pchiusano: These were added to support the doc console renderer, which is written in Unison.

pchiusano:
Okay, I knocked out the issues that concerned me and opened #1864 to track a few remaining loose ends.

unison-src/transcripts/docs.output.md
pchiusano: This was a regression and has now been fixed.

parser-typechecker/src/Unison/Builtin.hs
aryairani: What do you think about having these be pure Unison functions?

parser-typechecker/src/Unison/Codebase/Path.hs
aryairani: Déjà vu of having done the same thing? Maybe in an alternate reality.  Or maybe I'm just thinking of Cons / Snoc.

parser-typechecker/src/Unison/PrintError.hs
aryairani: Minor suggestion to put the alignment padding after the `:` rather than aligning the `is`.

unison-src/new-runtime-transcripts/_base.md
aryairani: We have a working version of `base.u`??  Also what is `base.u`.

aryairani:
> This is ready for review! I have a couple loose ends I'm tracking down which are noted in the writeup but that shouldn't affect your review.

Is there anything in particular you recommend we look for?

parser-typechecker/src/Unison/Builtin.hs
pchiusano: I was doing that originally but I thought going through boxed lists of characters was way too inefficient. 

I could get rid of the alignment functions and just have Text.repeat builtin. Wdyt?

parser-typechecker/src/Unison/Codebase/Path.hs
pchiusano: There is a class called Resolve in Path, but that has to do with like resolving various kinds of relative to absolute paths I think.

parser-typechecker/src/Unison/PrintError.hs
pchiusano: Here they are for comparison -

```
Here, one is:    Foo x (Qux j Bat)
and another is:  Foo x (Qux g Nat)
```

```
Here,  one  is:  Foo x (Qux j Bat)
and another is:  Foo x (Qux g Nat)
```

My thinking was the second one draws focus to the aligned edge better, and the jagged edge is a little distracting. Wdyt? 

@hojberg You might have an opinion on this

unison-src/new-runtime-transcripts/_base.md
pchiusano: It's just common code used by multiple transcripts so they don't all have to start with a prelude of utility functions. @stew added it a little while back and I just pulled out the unison code into a separate .u file

pchiusano:
@aryairani @runarorama a thing to look for is are there any documentation elements you feel are missing, that can't be expressed in terms of existing elements. That will be more annoying to change later since adding a new constructor creates a new type. While it's very possible to support multiple types (we are still supporting the old doc type for now) it's more of a burden.

You can peep the Doc2 AST in IOSource, but I'm realizing IOSource needs more comments so if you want hold off until I mark it up.

I had a session with @hojberg where he thought of some elements to include, like images and tables, and @runarorama and I did the initial brainstorming, but maybe you've thought of some ideas since then.

unison-src/new-runtime-transcripts/_base.md
aryairani: Oh, I see, thanks.

pchiusano:
Okay, I did a pass through `IOSource` adding comments: https://github.com/unisonweb/unison/pull/1853/files#diff-6ef022a113eb26e9c8f263a6eea38cf971e27085a7c167463975cb1c606ee7e7R675

parser-typechecker/src/Unison/Builtin.hs
pchiusano: I went ahead and got rid of the text alignment builtins (made them pure Unison), but kept `Text.repeat` and added `Char.toText : Char -> Text` which was missing.

runarorama (approved):

parser-typechecker/src/Unison/Builtin/Decls.hs
```diff
@@ -285,10 +287,27 @@ builtinEffectDecls = []
...
+  :: Term.Term2 vt at ap v a
+  -> Maybe (Term.Term2 vt at ap v a)
+unRightTerm t = case t of
+  Term.App' (Term.Constructor' EitherRef EitherRightId) tm ->
+    Just tm
+  _                           -> Nothing
+unLeftTerm t = case t of
+  Term.App' (Term.Constructor' EitherRef EitherLeftId) tm ->
+    Just tm
+  _                           -> Nothing
```
aryairani, line 310: Optional suggestion: format some of these with ormolu 😬 
In vscode it's ⌘K ⌘F to "Format selection" (as opposed to whole file).

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1496,7 +1547,7 @@ loop = do
                   Nothing -> [] <$ respond (TermNotFound' . SH.take hqLength . Reference.toShortHash $ Reference.DerivedId rid)
                   Just tm -> do
                     respond $ TestIncrementalOutputStart ppe (n,total) r tm
-                    tm' <- eval $ Evaluate1 ppe tm
+                    tm' <- eval $ Evaluate1 ppe False tm
```
aryairani: Should this (and the one on line 1627) be `True`? Or is `False` correct?

parser-typechecker/src/Unison/PrintError.hs
aryairani: Hm, yeah, thanks for the side-by-side; I'm not sure.  They both look weird. 😅  I don't mind it staying as-is until there's a stronger opinion.

aryairani: A third option is like in the very next block:
```
  Here, one is:  Foo x (Qux j Bat)
and another is:  Foo x (Qux g Nat)
```

aryairani (approved):
Looks great!

parser-typechecker/src/Unison/PrintError.hs
pchiusano: I think I like your third option, will go with that!

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
pchiusano: I wish I could see more context for this. I see like 2 surrounding lines with no way to expand that on mobile...

pchiusano: This one is fine - the test watch is added to the cache explicitly later. I’ll check the other one when by a computer.

pchiusano: Yeah the other is okay too, it's an I/O test which you don't want to cache. I'm adding a comment in the code to this effect.

parser-typechecker/src/Unison/PrintError.hs
pchiusano: Disregard, the third option looks really weird in context, because that left edge should be aligned with the paragraph above it. 

The other example indents this portion of the message. There was some reason I didn't go with that here. I'm just going to leave as is for now, but am open to doing a more cohesive pass through the error message formatting.

PR #1855, opened 2021-04-01
Adding a config for UCM

bontaq:
## Overview
Just a start on some refactoring to clean up `Main`.  I'm opening it with some first steps done to see if I'm on the right path and where it might go.  Next step might be combining `ucmConfig` and the file-based `config` so that it's just one.

So lemme know what you think

## Implementation notes

Adds a config record with some of the shared variables


aryairani:
Sorry I didn't see this until now.  You may have seen that some things have been changing in Main; but if you want to make a new PR against the `wip/codebase2` branch, I'm open to cleaning that up too.

PR #1856, opened 2021-04-06
use hpack package.yaml to generate .cabal files

aryairani:
This PR switches the package definitions to hpack format, so I don't have to edit the package definitions every time I want to add a new module.  Feedback is welcome, but I hope we can merge it.

## Overview

`stack build` will automatically use [hpack](https://github.com/sol/hpack) to convert the more concise package.yaml files into .cabal files [slides](http://typeful.net/talks/hpack/#7), [blog](https://mmhaskell.com/blog/2020/1/27/hpack-a-simpler-package-format).  In particular, **this allows you to add modules to a package just by putting them on the filesystem**, without having to edit .cabal files, although you can still manual specify `exposed-modules` or `other-modules`, in which case the complementary set is inferred.

* There's disagreement on whether to include the generated .cabal file into the git repo.  I lean slightly "no", but included them in this PR anyway, and would like to get second opinions.  
  > Both cabal2nix and stack support `package.yaml` natively. For other build tools, the `hpack` executable can be used to generate a .cabal file from package.yaml.

* > Hpack allows the inclusion of common fields from a file on GitHub or a local file.

  which might simplify some things further — e.g. copyright & license info, default-extensions, etc. — but I didn't do this yet.

### DIFF
--- a/parser-typechecker/package.yaml
+++ b/parser-typechecker/package.yaml
@@ -1,0 +1,198 @@
+name: unison-parser-typechecker
+github: unisonweb/unison
+copyright: Copyright (C) 2013-2018 Unison Computing, PBC and contributors
+
+default-extensions:
+  - ApplicativeDo
+  - BlockArguments
+  - DeriveFunctor
+  - DerivingStrategies
+  - DoAndIfThenElse
+  - FlexibleContexts
+  - FlexibleInstances
+  - LambdaCase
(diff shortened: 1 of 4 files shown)


pchiusano (approved):
No objections from me and I have always disliked having to manually add modules to the cabal file.

PR #1857, opened 2021-04-06
Use liftIO instead of MonadUnliftIO

aryairani:
## Overview

By changing the Codebase function `rootBranchUpdates  :: m (m (), m (Set Branch.Hash))` to `m (IO (), IO (Set Branch.Hash))`, we can generalize our constraints from `MonadUnliftIO m`, of which there aren't a ton of instances, to `(MonadIO m, MonadCatch m)`, of which there are.

## History
IIRC, we had originally introduced `MonadUnliftIO` because the file-watching library wanted us to pass in a handler in `IO` when we only had `m` at that point.  So we asked “How do we go `m ~> IO`?” and the answer was `MonadUnliftIO`.

I autoformatted all of the imports in `FileCodebase.hs` using ormolu.

I didn’t remove the `unliftio` package dependency, because it also provides a bunch of wrappers around common `IO` functions, lifted to `MonadIO`.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -57,7 +57,7 @@ data Codebase m v a =
 
            , getRootBranch      :: m (Either GetRootBranchError (Branch m))
            , putRootBranch      :: Branch m -> m ()
-           , rootBranchUpdates  :: m (m (), m (Set Branch.Hash))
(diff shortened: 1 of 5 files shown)


pchiusano (approved):

aryairani:
I guess ready-to-merge doesn't merge anymore?

PR #1858, opened 2021-04-06
Further decompiling improvements in the new runtime

dolio:
This pull request fixes some bugs and makes it significantly more likely for values to be decompilable in the new runtime.

Partially applied lambdas weren't being decompiled very nicely. There was code that was supposed to substitute under the lambda, but it used the wrong pattern, and ended up never actually substituting. This is now fixed.

The binding group order ends up getting reversed when translating to supercombinators, so the last attempt to decompile local bindings was getting the numbering backwards. This turns out to still be relevant even with the changes below.

The floating pass now keeps track of original definitions for many floated bindings, and these are used for decompiling. The full floating pass now works as follows:

  1. Add arguments to all lambda expressions corresponding to their non-recursive free variables, together with applying the expanded lambda expression to the local variable
  2. Float every lambda expression to a top-level recursive let, remembering _closed_ lambda terms for decompilation

The latter part of step 2 works because step 1 has ensured that the only free variables in a lambda term correspond to (local, mutually) recursive functions. So every non-recursive local function is eligible for decompilation. The decompiled terms have already had step 1 performed, so they may have some redundant beta redexes. But those are equivalent to the original term, and it eliminates the decompilation causing ill scoped terms (I believe).

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/ANF.hs
+++ b/parser-typechecker/src/Unison/Runtime/ANF.hs
@@ -243,7 +243,7 @@ isStructured _ = True
 close :: (Var v, Monoid a) => Set v -> Term v a -> Term v a
 close keep tm = ABT.visitPure (enclose keep close) tm
 
-type FloatM v a r = State (Set v, [(v, Term v a)]) r
+type FloatM v a r = State (Set v, [(v, Term v a)], [(v, Term v a)]) r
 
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
Sweet! Thanks. I'll give it a go.

pchiusano:
@dolio this... has worked! 🙏

PR #1865, opened 2021-04-09
Add an endpoint for codebase search

runarorama:
## Overview

This adds a `find` endpoint to the codebase server with the following API:

#### GET Parameters:

- rootBranch
     - **Example**: *#abc123*
     - **Description**: The hash or hash prefix of the namespace root. If left absent, the most recent root will be used.

- relativeTo
     - **Example**: *., .base, foo.bar*
     - **Description**: The namespace relative to which the `names` parameter is to be resolved. If left absent, the root namespace will be used.

- renderWidth
     - **Example**: *80, 100, 120*
     - **Description**: The preferred maximum line width (in characters) of the source code of definitions to be rendered. If left absent, the render width is assumed to be 80.

- limit
     - **Example**: *1, 10, 20*
     - **Description**: The maximum number of results to return. Defaults to 10.

- query
     - **Example**: *foo, ff, td nr*
     - **Description**: Space-separated subsequences to find in the name of a type or term.

## Implementation notes

This uses the FuzzyFind library from https://github.com/runarorama/fuzzyfind

Doesn't return any definitions, whether for types or terms. Getting the definitions for the search results was slowing the search down by an order of magnitude or so.


parser-typechecker/src/Unison/Server/Endpoints/FuzzyFind.hs
```diff
@@ -0,0 +1,191 @@
...
+   toEncoding = genericToEncoding defaultOptions
+instance ToJSON FZF.ResultSegment where
+   toEncoding = genericToEncoding defaultOptions
+
+deriving instance ToSchema FZF.Alignment
+deriving instance ToSchema FZF.Result
+deriving instance ToSchema FZF.ResultSegment
+
+data FoundTerm = FoundTerm
+  { bestTermName :: HashQualifiedName
```
hojberg: @runarorama I noticed when integrating that we had named this one `bestTermName` and the other `bestFoundTypeName`. Should we name this one `bestFoundTermName` to match?

runarorama: good idea

parser-typechecker/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1557,11 +1557,17 @@ showDiffNamespace sn ppe oldPath newPath OBD.BranchDiffOutput{..} =
...
          <*> traverse prettyGroup (terms `zip` [length types ..])
     where
-        leftNamePad :: Int = foldl1' max $
-          map (foldl1' max . map HQ'.nameLength . toList . view _3) terms <>
-          map (foldl1' max . map HQ'.nameLength . toList . view _3) types
-        prettyGroup :: ((Referent, b, Set (HQ'.HashQualified Name), Set (HQ'.HashQualified Name)), Int)
-                    -> Numbered Pretty
+        leftNamePad :: P.Width =
+          foldl1' max
+            $  map (foldl1' max . map (P.Width . HQ'.nameLength) . toList . view _3)
```
runarorama, line 1562: I got burned by `Width` and `Int` getting confused, so the type changed here and in other places.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -473,67 +537,77 @@ prettyDefinitionsBySuffixes relativeTo root renderWidth codebase query = do
...
-          ( Text.pack
-          . Pretty.render width
-          . fmap SyntaxText.toPlain
-          . TypePrinter.pretty0 @v (PPE.suffixifiedPPE ppe) mempty (-1)
-          $ Type.ref () r
-          )
-        $ fmap mungeSyntaxText tp
-    typeDefinitions =
-      Map.mapWithKey mkTypeDefinition $ typesToSyntax width ppe types
+  let printNames =
```
runarorama, line 540: Parts of this function body got abstracted out to another function, `bestNameForTerm`

parser-typechecker/src/Unison/Server/Types.hs
```diff
@@ -97,6 +106,67 @@ data DefinitionDisplayResults =
     , missingDefinitions :: [HashQualifiedName]
     } deriving (Eq, Show, Generic)
 
+data TermTag = Doc | Test
```
runarorama, line 109: These types were moved here from endpoint-specific files since they get used in more than one place.

unison-core/src/Unison/Term.hs
```diff
@@ -474,7 +474,7 @@ var = ABT.annotatedVar
 var' :: Var v => Text -> Term0' vt v
 var' = var() . Var.named
 
-ref :: Ord v => a -> Reference -> Term2 vt at ap v a
+ref :: forall v a vt at ap. Ord v => a -> Reference -> Term2 vt at ap v a
```
pchiusano: Any particular reason for this one? I don't think it's helping.

unison-core/src/Unison/Type.hs
```diff
@@ -188,7 +188,7 @@ isArrow _ = False
 --vectorOf :: Ord v => a -> Type v a -> Type v
 --vectorOf a t = vector `app` t
 
-ref :: Ord v => a -> Reference -> Type v a
+ref :: forall v a. Ord v => a -> Reference -> Type v a
```
pchiusano: Same deal here... I'd prob revert.

pchiusano (approved):
LGTM. Left a couple minor comments.

I'm interested to incorporate type based search into this somehow that doesn't require typing out full type signatures, but that can be for later.

unison-core/src/Unison/Term.hs
runarorama: We wanted to pass `@t` to this function for the type parameter `v` but found that it was getting applied to the parameter `a`. Perhaps no longer relevant after some simplifications elsewhere.

unison-core/src/Unison/Type.hs
runarorama: Same deal here.

PR #1871, opened 2021-04-16
delete old runtime (fixes #1869)

### LINKED ISSUE #1869: Delete old 'n busted runtime; make new runtime the default
The time has come. Old runtime served its purpose, but we can rip it out now and remove the `--new-runtime` flag. Old runtime hasn't been kept up to date with latest builtins. The new runtime is faster and works great.

mitchellwrosen:
## Overview

This PR deletes the old runtime and old `io` type, and deletes the various `Bool` arguments and command-line flags that control whether the old or new runtime should be used.

Fixes #1869 

## Implementation notes

Mostly followed @pchiusano advice about deleting a couple modules and chasing down type errors.

"Old runtime" transcripts were all deleted. I didn't read them, but just assumed they are not longer needed. "New runtime" transcripts (`unison-src/new-runtime-transcripts`) were not moved.

## Test coverage

No new tests, only a couple deleted ones.

## Loose ends

- [x] ~io.BufferMode still exists, referenced [here](https://github.com/unisonweb/unison/blob/b0dd6e5496775a8b37b6271d1356a0a9c6041894/parser-typechecker/src/Unison/Runtime/Builtin.hs#L748)~
- [x] ~io.Mode still exists, referenced [here](https://github.com/unisonweb/unison/blob/b0dd6e5496775a8b37b6271d1356a0a9c6041894/parser-typechecker/src/Unison/Runtime/Builtin.hs#L852)~
- [x] One failing transcript (can't find `printLine`)
- [x] Need to make a separate PR to Arya's branch


aryairani:
I changed all these same lines to try to add a codebase2 flag alongside the runtime flag in `wip/codebase2`, so there's gonna be a bunch of conflicts. 😓  
Wanna make a PR to merge this into that branch too? 😃  
/cc @mitchellwrosen @pchiusano

mitchellwrosen:
@aryairani sure, no problem. There's still one regression I'm working through in this patch, any pointers? :smile: 

One transcript is failing with:

```
  I found this function:

    main : '{builtin.io.IO} ()

  but in order for me to `run` it it needs to have the type:

    main : '{builtin.io2.IO} a
```

To rerun: `stack exec tests IO.handleOps`

pchiusano:
@mitchellwrosen You can delete that test - io.IO is the old IO type. You can also delete it from IOSource.hs

mitchellwrosen:
Ok. I believe I deleted `io.IO` from `IOSource`, but left around other types in the `io` namespace, such as `io.BufferMode`, because even though there appears to be an `io2.BufferMode`, I think the old one is still in use.

Does that sound right? Or were you expecting `io.BufferMode`, `io.Socket`, `io.Handle`, etc. to be deleted as well?

mitchellwrosen:
One more confusion: in deleting `io.IO`, I also deleted all of the functions that mentioned the type from `IOSource.hs`, such as `io.printLine`.

But now the `diff.md` transcript is failing, because it calls `printLine`. Did I delete too much, or does that transcript need to be rewritten to use `io2.printLine`, somehow?

pchiusano:
There shouldn't be anything left in io, only io2.

Not sure on diff.md will have to look at that more closely.

mitchellwrosen:
@pchiusano Interesting, ok. So, one type I tugged on does seem used in a builtin.

`io.BufferMode`'s defined in Haskell here
https://github.com/unisonweb/unison/blob/b0dd6e5496775a8b37b6271d1356a0a9c6041894/parser-typechecker/src/Unison/Runtime/IOSource.hs#L77

and then a `getBuffering` primitive references this type here
https://github.com/unisonweb/unison/blob/b0dd6e5496775a8b37b6271d1356a0a9c6041894/parser-typechecker/src/Unison/Runtime/Builtin.hs#L738-L763

Initially I had tried deleting that, but grepping around the codebase for `getBuffering`, I don't see an equivalent defined for `io2`. All I found is what looks like the _type_ of `getBuffering` defined for `io2`:
https://github.com/unisonweb/unison/blob/b0dd6e5496775a8b37b6271d1356a0a9c6041894/parser-typechecker/src/Unison/Builtin.hs#L518

Could you help me understand how this all fits together?

aryairani:
@mitchellwrosen I see that the `bmode` you referenced leads back to `bufferModeRef = lookupDeclRef "io2.BufferMode"` in `parser-typechecker/src/Unison/Builtin/Decls.hs`. Is that what you're looking for, or is there something that's still referencing the original `io.BufferMode` that you're hesitant to delete?

mitchellwrosen:
@aryairani Sorry, I mistakenly pasted the first source code link twice :facepalm: 

Amending... done.

aryairani:
I wonder if it's: 
a) correctly relying on an old type and we should keep that type around, or
b) incorrectly relying on an old type and should be updated to use `bufferModeRef` instead of `bufferModeReference`. Tricky!

@dolio Do you know?

dolio:
I think `bufferModeRef` and `bufferModeReference` should actually be the same, because one is the result of copying the data type declaration from `IOSource` to `Unison.Builtin.Decls`. So it's probably just a mistake that doesn't actually matter.

pchiusano:
The one in `IOSource` is: 

```Haskell
-- If the buffer size is not specified,
-- use an implementation-specific size.
unique[e65de145a461a771de93d6c7885acae28552d77f8ae460bc8bf5de6f2a15ff77] type
  io.BufferMode = Line | Block (Optional Nat)
```

The one in `Decl.hs` is:

```Haskell
  bmode = DataDeclaration
    (Unique "7dd9560d3826c21e5e6a7e08f575b61adcddf849")
    ()
    []
    [ ((), v "io2.BufferMode.NoBuffering", var "io2.BufferMode")
    , ((), v "io2.BufferMode.LineBuffering", var "io2.BufferMode")
    , ((), v "io2.BufferMode.BlockBuffering", var "io2.BufferMode")
    , ((), v "io2.BufferMode.SizedBlockBuffering"
      , Type.nat () `arr` var "io2.BufferMode")
    ]
```

So they are different types and not isomorphic either - the one in `Decl` has a "no buffering" option while the one from `IOSource` does not. I guess it's unclear what the `Block None` means in the `IOSource` version, but it can't mean "use a system default block size" and also "no buffering" at the same time. :)

Seems like we want the one from `Decl.hs` but @mitchellwrosen you are saying that the one from `IOSource` is still being referenced?

@mitchellwrosen you can put together the PR deleting everything other than that one type if it's still being referenced, and then @dolio can work on ripping it out as a separate task.

dolio:
Ah, sorry. If they're not the same, then it's still just a mistake, and Paul is correct that the one from `Decl` should be used. It's just a matter of both references being imported into that file, and the wrong one being used.

mitchellwrosen:
> @mitchellwrosen you can put together the PR deleting everything other than that one type if it's still being referenced, and then @dolio can work on ripping it out as a separate task.

Sure!

mitchellwrosen:
Ok, I deleted what I could. Two things remain:

- [ ] io.BufferMode, referenced [here](https://github.com/unisonweb/unison/blob/b0dd6e5496775a8b37b6271d1356a0a9c6041894/parser-typechecker/src/Unison/Runtime/Builtin.hs#L748)
- [ ] io.Mode, referenced [here](https://github.com/unisonweb/unison/blob/b0dd6e5496775a8b37b6271d1356a0a9c6041894/parser-typechecker/src/Unison/Runtime/Builtin.hs#L852)

Finally, there's still the one transcript error that I'm unsure how to fix:

```


  I'm not sure what printLine means at line 5, columns 16-25

      5 | helloWorld = '(printLine "Hello, world!")

  Whatever it is, it has a type that conforms to Text -> 𝕣.
```

To reproduce: `stack run transcripts -- diff.md`

Is there no function called `printLine` in the new runtime, perhaps?

pchiusano:
@mitchellwrosen that's right, there's not a `printLine` currently (it can be defined in terms of `putBytes` and utf8 encoding), however that test never actually runs that expression, it's just an arbitrary term. You could just remove the call to `printLine` -

```Haskell
helloWorld = '(printLine "Hello, world!")
```

Just change that to

```Haskell
helloWorld = "Hello, world!"
```

After updating that, can you create a ticket for switching to `io2.{BufferMode,Mode}` and assign it to @dolio? Then we can get this merged.

pchiusano (approved):
Thanks @mitchellwrosen

One thing - I'd move all the `new-runtime-transcripts` directory into the existing `transcripts` directory. Otherwise it's sort of confusing to have both there even though there's no difference between the two directories.

This could be a separate PR if you want.

mitchellwrosen:
@pchiusano Sounds good, I can get that into this PR before merging.

I tried to do the obvious thing of just moving the files, and grepping around for `new-runtime-transcripts`, but now every transcript is failing with this message.

```
💥  transcript-parser-commands.md

  Transcript will be run on a new, empty codebase.

Initializing a new codebase in:
/private/var/folders/_m/wy1skj4147l3bnw5cwmvg0s40000gp/T/transcript-4a13ac8a170b487f

   _____     _
  |  |  |___|_|___ ___ ___
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|

  Running the provided transcript file...


⚙️   Processing stanza 2 of 8.# Base transcript

## Overview

This transcript is meant to be a transcript which can be run as a
prelude to other transcripts, creating helper functions, and including
a minimal subset of base in order to facilitate write nicer
transcripts which contain less boilerplate.

## Usage

``ucm
.> builtins.merge.> builtins.mergeio.> cd builtin.> load unison-src/base.u.> add
``


🛑

  line 1, column 1: Identifier segment can't be a keyword: namespace


Run `unison -codebase /private/var/folders/_m/wy1skj4147l3bnw5cwmvg0s40000gp/T/transcript-4a13ac8a170b487f` to do more work with it.

unitnamespace.md EXCEPTION!!!: callCommand: /Users/mitchell/haskell/unison/delete-old-runtime/.stack-work/install/x86_64-osx/c57e2487f6d55d96d0c746d3025d96eeb6bfd9c5068fedede2dd18e137c433a1/8.10.4/bin/unison transcript unison-src/transcripts/_base.md unison-src/transcripts/unitnamespace.md (exit 1): failed
```

This part looks weird to me, it's like all newlines have been deleted from this setup snippet.

```
.> builtins.merge.> builtins.mergeio.> cd builtin.> load unison-src/base.u.> add
```

And what's extra weird is I can run a single transcript, e.g. `stack exec transcripts -- addupdatemessages.md` and it passes. 😢 

Any idea what might be going on here?

pchiusano:
Hmm okay. I don't understand that error, but one thing that I realized is different is that `base.md` is run before all those transcripts in `new-runtime-transcripts`, and we don't necessarily want to do that for all transcripts. So there's actually still some value to having transcripts in separate directories.

So I'd say just rename that directory to `transcripts-using-base` and leave the rest alone and then we can merge.

mitchellwrosen:
@pchiusano Ok, if this passes CI it's good to go 🤞

pchiusano:
Great, thanks @mitchellwrosen!

I applied ready-to-merge label so it should get merged once CI finishes.

pchiusano:
I'm not sure why mergify didn't trigger but it looks like all the CI passed, so merging.

PR #1872, opened 2021-04-16
Don't wait for CTRL-D before starting up server

runarorama:
Fixes an issue where setting the `UCM_PORT` in the environment would cause you to have to send `EOF` to ucm in order to allow the server startup to complete.

### DIFF
--- a/parser-typechecker/src/Unison/Server/CodebaseServer.hs
+++ b/parser-typechecker/src/Unison/Server/CodebaseServer.hs
@@ -23,7 +23,6 @@ import           Network.Wai                    ( responseLBS
                                                 , queryString
                                                 )
 import           Network.Wai.Handler.Warp       ( withApplicationSettings
-                                                , runSettings
                                                 , defaultSettings
                                                 , Port
                                                 , setPort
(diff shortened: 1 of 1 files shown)


pchiusano (approved):
LGTM though I don't understand from the diff why the old code behaved as it did and why the new code does the right thing. 🙃

parser-typechecker/src/Unison/Server/CodebaseServer.hs
```diff
@@ -152,23 +152,18 @@ start
...
+  let settings = appEndo
+        (  foldMap (Endo . setPort)              envPort
+        <> foldMap (Endo . setHost . fromString) envHost
+        )
+        defaultSettings
       a = app codebase token
-
-  case envPort of
-    Just p -> do
-      (k token p)
```
runarorama, line 168: This starts up UCM and waits for it to finish (which it appears to do when it receives a CTRL-D for some reason.

parser-typechecker/src/Unison/Server/CodebaseServer.hs
```diff
@@ -152,23 +152,18 @@ start
...
-      runSettings (setPort p settings) a
```
runarorama, line 169: This starts the server after the UCM CLI has "finished".

parser-typechecker/src/Unison/Server/CodebaseServer.hs
```diff
@@ -152,23 +152,18 @@ start
...
-    Nothing ->
-      withApplicationSettings settings (pure a) (k token)
+  withApplicationSettings settings (pure a) (k token)
```
runarorama, line 166: Start the server and once it has started, fire up the CLI.

PR #1883, opened 2021-04-22
Add term tags and type tags to GetDefinitions endpoint

runarorama:
Adds `TermTag` and `TypeTag` to definitions in the `GetDefinitions` endpoint, so the UI can show which terms are tests/docs and which types are abilities.


hojberg (approved):
Sweet!

PR #1886, opened 2021-04-23
Automatically initialize codebase

runarorama:
## Overview

This removes the error message about the codebase not being initialized and replaces it with a message like the following:

```
  No codebase exists in the directory /Users/runar. Creating one now.

  If this is not what you wanted, run `unison -codebase <dir>`
   to load a codebase from the directory <dir>.

  Initializing a new codebase in: /Users/runar
```

Unison now just proceeds to initialize the codebase instead of giving an error.

I replaced all instances of "get the codebase or crash" with automatic initialization. That includes transcripts. Maybe that's not the correct thing to do. Please review.

### DIFF
--- a/parser-typechecker/unison/Main.hs
+++ b/parser-typechecker/unison/Main.hs
@@ -139,7 +139,6 @@ main = do
   args <- getArgs
   progName <- getProgName
   -- hSetBuffering stdout NoBuffering -- cool
-
   _ <- installSignalHandlers
   -- We need to know whether the program was invoked with -codebase for
   -- certain messages. Therefore we keep a Maybe FilePath - mcodepath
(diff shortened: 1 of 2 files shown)


parser-typechecker/unison/Main.hs
```diff
@@ -160,35 +159,37 @@ main = do
...
   case restargs of
     [] -> do
-      theCodebase <- FileCodebase.getCodebaseOrExit branchCache mcodepath
+      theCodebase <- FileCodebase.getCodebaseOrInit branchCache mcodepath
       Server.start theCodebase $ \token port -> do
-        PT.putPrettyLn . P.string $ "I've started a codebase API server at "
-        PT.putPrettyLn . P.string $ "http://127.0.0.1:"
-          <> show port <> "?" <> URI.encode (unpack token)
-        PT.putPrettyLn' . P.string $ "Now starting the Unison Codebase Manager..."
+        PT.putPrettyLn $ P.lines
```
runarorama, line 164: Cleaned this up to look more consistent with the rest of the output.

pchiusano:
I can't decide if this is the behavior we want or if we should ask for confirmation.

I think if you're going to do it without confirmation the message should say how to undo it. "You can delete the directory xyz if you didn't mean to create a new codebase there."

aryairani:
> I can't decide if this is the behavior we want or if we should ask for confirmation.
> 
> I think if you're going to do it without confirmation the message should say how to undo it. "You can delete the directory xyz if you didn't mean to create a new codebase there."

In terms of merge conflicts, I also refactored all the initialization stuff in order to construct unit tests that involved multiple runs of ucm (e.g. checking if some codebase change survives a reboot, which has often not been the case for us).  I could try to tease that out into a separate PR? Let me know.

> I replaced all instances of "get the codebase or crash" with automatic initialization. That includes transcripts. Maybe that's not the correct thing to do. Please review.

I'm not sure about this either.  In my refactor, there's a "create codebase" and there's an "open codebase" and neither of them crashes; though they can be called (and, to maintain the original behavior, are called) from a function that crashes. 😀   My preference would be to apply my patch first  (either after codebase2 or pulled out of it), but maybe I'm just shellfish 🦐

aryairani:
It's this sort of thing:
```haskell
data CreateCodebaseError
  = CreateCodebaseAlreadyExists
  | CreateCodebaseOther Pretty

data Init m v a = Init
  { -- | open an existing codebase
    openCodebase :: CodebasePath -> m (Either Pretty (Codebase m v a)),
    -- | create a new codebase
    createCodebase' :: CodebasePath -> m (Either CreateCodebaseError (Codebase m v a)),
    -- | given a codebase root, and given that the "codebase root" may have other junk in it, 
    -- like a README or different-version codebase, give the path to the "actual" files; 
    -- i.e. what a forked transcript should clone.
    codebasePath :: CodebasePath -> CodebasePath
  }

-- | basically turns `CreateCodebaseAlreadyExists` into a `Pretty`
createCodebase :: MonadIO m => Init m v a -> CodebasePath -> m (Either Pretty (Codebase m v a))

```

pchiusano:
I think let's just revisit this topic after new codebase format lands - it seems like this logic will be totally rewritten then anyway. 

I like the spirit of this PR though and want the "launch unison for the first time" process to be as painless as possible, including fetching the correct version of base.

aryairani:
This is superseded by @rlmark and @hojberg's welcome onboarding.

PR #1889, opened 2021-04-24
Fix issue with Bytes.at always returning `None` in new runtime

### LINKED ISSUE #1888: `Bytes.at 0` always returns None
*Version*: devel/M1l-1068-gb0dd6e549 (almost latest trunk)
Same results with `--new-runtime` and without it.

How to reproduce:

```
    1 | > Bytes.fromList [77, 13, 2] |> Bytes.at 0
          ⧩
          None
```

```
    1 | > Bytes.fromList [77, 13, 2] |> Bytes.at 1
          ⧩
          Some 13
```

pchiusano:
Fixes #1888 

@dolio can you review? I took a guess that the `None` tag was being written to the unboxed stack even when the index was found. But see my self review for questions on the fix.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Machine.hs
+++ b/parser-typechecker/src/Unison/Runtime/Machine.hs
@@ -1289,7 +1289,7 @@ bprim2 !ustk !bstk IDXB i j = do
     Just x -> do
       poke ustk $ fromIntegral x
       ustk <- bump ustk
-      ustk <$ poke ustk 0
+      ustk <$ poke ustk 1
   pure (ustk, bstk)
 bprim2 !ustk !bstk CATB i j = do
   l <- peekOffBi bstk i
--- a/parser-typechecker/tests/Unison/Test/Util/Bytes.hs
+++ b/parser-typechecker/tests/Unison/Test/Util/Bytes.hs
@@ -17,6 +17,11 @@ test = scope "util.bytes" . tests $ [
     ==
     Bytes.fromWord8s [0,1,2,3,4,5,6,7,8,9],
 
+  scope "at" $ do
+    expect' (Bytes.at 0 (Bytes.fromWord8s [77,13,12]) == Just 77)
+    expect' (Bytes.at 0 (Bytes.fromWord8s []) == Nothing)
+    ok,
+
   scope "consistency with ByteString" $ do
     forM_ [(1::Int)..100] $ \_ -> do
       n <- int' 0 50
--- a/parser-typechecker/tests/Unison/Test/Util/Bytes.hs
+++ b/parser-typechecker/tests/Unison/Test/Util/Bytes.hs
@@ -22,6 +22,12 @@ test = scope "util.bytes" . tests $ [
     expect' (Bytes.at 0 (Bytes.fromWord8s []) == Nothing)
     ok,
 
+  scope "at.cornerCases" $ do
+    let b = Bytes.drop 3 $ Bytes.fromWord8s [77,13,12] <> Bytes.fromWord8s [9,10,11]
+    expect' (Bytes.at 0 b == Just 9)
+    expect' (Bytes.at 0 (mempty <> Bytes.fromWord8s [1,2,3]) == Just 1)
+    ok,
+
   scope "consistency with ByteString" $ do
     forM_ [(1::Int)..100] $ \_ -> do
       n <- int' 0 50
--- a/unison-src/new-runtime-transcripts/builtins.md
+++ b/unison-src/new-runtime-transcripts/builtins.md
@@ -1,5 +1,11 @@
 # Unit tests for builtin functions
 
+```ucm:hide
+.> builtins.mergeio
+.> load unison-src/new-runtime-transcripts/base.u
+.> add
+```
+
 This transcript defines unit tests for builtin functions. There's a single `.> test` execution at the end that will fail the transcript with a nice report if any of the tests fail.
 
 ## `Int` functions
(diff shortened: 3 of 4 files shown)


parser-typechecker/tests/Unison/Test/Util/Bytes.hs
```diff
@@ -17,6 +17,17 @@ test = scope "util.bytes" . tests $ [
...
+  scope "at" $ do
+    expect' (Bytes.at 0 (Bytes.fromWord8s [77,13,12]) == Just 77)
+    expect' (Bytes.at 0 (Bytes.fromWord8s []) == Nothing)
+    ok,
+
+  scope "at.cornerCases" $ do
+    let b = Bytes.drop 3 $ Bytes.fromWord8s [77,13,12] <> Bytes.fromWord8s [9,10,11]
+    expect' (Bytes.at 0 b == Just 9)
+    expect' (Bytes.at 0 (mempty <> Bytes.fromWord8s [1,2,3]) == Just 1)
+    ok,
```
pchiusano, line 29: Added a few more unit tests to the `Bytes` type to figure out whether that was the problem here. These tests worked fine even before the fix.

unison-src/new-runtime-transcripts/builtins.md
```diff
@@ -177,6 +177,22 @@ test> Text.tests.alignment =
...
+        checks [
+          Bytes.at 1 bs == Some 13,
+          Bytes.at 0 bs == Some 77,
+          Bytes.at 99 bs == None
+        ]
+```
+
+```ucm:hide
+.> add
+```
```
pchiusano, line 194: Regression test.

parser-typechecker/src/Unison/Runtime/Machine.hs
```diff
@@ -1289,7 +1289,7 @@ bprim2 !ustk !bstk IDXB i j = do
     Just x -> do
       poke ustk $ fromIntegral x
       ustk <- bump ustk
-      ustk <$ poke ustk 0
+      ustk <$ poke ustk 1
   pure (ustk, bstk)
```
pchiusano, line 1293: @dolio a few questions:

* Can you briefly explain `ustk` vs `bstk`?
* Is the final `poke ustk 0` or `poke ustk 1` the tag for the data type? Like `0` means `None` and `1` means `Some`?
* Is the idea that `Some 8` is represented as `[8, 1]` on the top of the `ustk`? Or should it be `[1,8]`? (Basically, should the tag be on the top of the stack) I'm a little confused because I looked at `IDXS` (`List.at`) and it seems to be done differently.
* Can you explain how it works to store things to the unboxed stack? 
  * Suppose I do `match Bytes.at 0 b with Some x -> x + 99`. How does that code know that the x will be found on `ustk`?
  * Suppose I do `match List.at 0 [1,2,3] with Some x -> x + 99`. How does that code know that the `x` value will be found on the `bstk`?

dolio: The 'data' stack should be thought of as consisting of both `ustk` and `bstk`, although they are stored/indexed independently. `ustk` is where we store 64-bit unboxed stuff in a byte array, and `bstk` stores boxed stuff as an array of closures.

As you guessed, the final `poke ustk 1` is a sort of tag information indicating what is being returned. The builtin operation that makes use of this primop will test that to determine how to build the result. It's basically like an unboxed sum that the wrapper uses to build an actual sum value.

The ordering on the stack depends on how you're thinking about it. I have it set up so that they grow toward increasing array indices. So, the stack 'pointer' is a positive index into the relevant array, saying the highest index that has meaningful stack data. Offset positions in the stack are at lower indices, so position `i` on the stack is at index `p-i`, where `p` is the 'stack pointer' index. So as array indices, they are backwards, but the stack operations are set up so that you'd use the same numbering if the stack grew _down_, in which case things would be stored in the 'right' order.

None of your code examples do anything with the unboxed stack, except that pattern matching will use it for the implementation of testing the data type tag. `x` is not on the unboxed stack, it is a boxed integer. The builtins that implement addition will use the unboxed stack, but no value in surface-level unison is something on the unboxed stack, I think.

dolio (approved):
The change looks good to me.

parser-typechecker/src/Unison/Runtime/Machine.hs
pchiusano: > None of your code examples do anything with the unboxed stack, except that pattern matching will use it for the implementation of testing the data type tag. x is not on the unboxed stack, it is a boxed integer. 

@dolio hmm then can you explain this line, which seems to put the thing inside the `Just` onto the unboxed portion of the stack?

```
poke ustk $ fromIntegral x
```

dolio: You're looking at the implementation of the machine instruction. But it is wrapped by code in `Unison.Runtime.Builtin` that implements the surface-level unison functions, which are not directly calling the machine instruction.

The builtin code reads the tag and value off the unboxed stack and builds an appropriate data value, which is what the corresponds to the surface level.

pchiusano: Gotcha. So the machine instruction is technically returning an unboxed `Nat` with a tag of `0` or `1`, while the surface level function is calling that with some calling convention that boxes up the result appropriately.

dolio: Yeah, that's right. There's always a wrapper around these instruction things. Some of them just wrap the instruction in something that looks more like the representation of compiled code. But many of them do this unpacking and packing stuff.

pchiusano: Cool. Makes sense. Thanks for clarification.

PR #1895, opened 2021-04-28
v2 (sqlite) codebase format  

aryairani:
## Overview

This PR introduces a new codebase format, backed by sqlite instead of individual files.  Converting `unisonweb/base` from v1 to v2 format resulted in a 99.5% size reduction. 😮 
```
% du -hs .unison/v*
1.4G	.unison/v1
5.3M	.unison/v2
```

`stack exec -- unison [-codebase <root>]` to launch using a v2 codebase, like usual.
`stack exec -- unison [-codebase <root>] --old-codebase` to launch using a v1 codebase
`stack exec -- unison [-codebase <root>] upgrade-codebase` to upgrade from v1 to v2

## Loose ends
See #1915.


pchiusano (approved):
Discussed with @aryairani and we think this is good to merge and at least start getting ambient testing on trunk. @aryairani is going to create a separate ticket (can you link to it here) with miscellaneous loose ends that we'd like to address in the future.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1817,34 +1814,6 @@ loop = do
                 prettyDefn renderR (r, (Foldable.toList -> names, Foldable.toList -> links)) =
                   P.lines (P.shown <$> if null names then [NameSegment "<unnamed>"] else names) <> P.newline <> prettyLinks renderR r links
         void . eval . Eval . flip State.execStateT mempty $ goCausal [getCausal root']
-      -- DebugDumpNamespacesI -> do
```
aryairani, line 1820: This is just deleting a commented copy of the code, not the actual code.

PR #1897, opened 2021-04-28
Uncontroversial fix to rendering of tables in docs

pchiusano:
I noticed a bug with plain text rendering of tables with that span multiple lines -

```
  # Functions for lookahead and backtracking
  
    `_mark`       Pushes the current parser state onto a stack.
    `_reset i`    Pops the mark stack up to and including `i` and resets
                the parse state to what it was at mark `i`.
    `_unmark i`   Pops the mark stack up to and including `i`.
```

Notice the multi-line row second line is 2 characters off from being aligned. This PR fixes that and adds a regression test for it.

Looking at the diff, I did notice some weirdness in pretty-printing of lists of docs which is unrelated to this fix. I'd like to do a more thorough audit of pretty-printing. There are a number of cases where it produces weird or unparseable results.


aryairani (approved):
Transcript output looks good.

Issue #1880, opened 2021-04-21
Codebase UI local deployment


pchiusano:
Discussed with @stew @hojberg and @runarorama how UCM releases will embed the codebase UI. I think this writeup accurately captures the decisions from that, but pipe up if not. 

Here's the UX:

The UX is: User starts up UCM which provides a URL to visit. User visits that URL and gets the codebase UI. (Later we might have an electron app which bundles the codebase UI and UCM, however that's out of scope for now.)

Nice touch: the startup message should also say how to find that URL if the user forgets it. Like maybe `help ui` shows the URL as well.

Here's the plan:

- [x] @runarorama will change the codebase server to serve the API at `<token>/api` and the `<token>/ui`. The `/ui` endpoint will look for a file `/ui/index.html` by default or will serve whatever file is specified via a `--ui=/path/to/dev-ui/index.html` flag and/or environment variable. If no such file exists, it will 404 as normal.
- [x] @hojberg will get CI for the codebase-ui project to produce GitHub releases containing the HTML+JS for the built codebase UI. @stew to support this.
- [x] @stew will create a third project, maybe `ucm-release`, which will pull any two named versions of Unison and the codebase UI from GitHub releases and package them up into a release.

For later, @stew or someone else could work on doing daily or weekly releases, but I don't think that is MDP for M2. You can tag releases as 'daily' and then they don't show up in the main releases page, see for instance [this example](https://github.com/microsoft/BotFramework-WebChat/releases/daily)

PR #1900, opened 2021-05-01
Serve backend API from `/api` and frontend from `/ui`

runarorama:
## Overview

Addresses #1880 as it pertains to the codebase server. The API is now served at `http://127.0.0.1/api`. The frontend is served from `http://127.0.0.1/ui`.

This also adds some configurability. The options are as follows:

### Directory containing the UI

Environment variable: `UCM_WEB_UI`
Command line option: `--ui <DIR>`

This is the path to the directory containing the UI. Any request for a resource under `/ui` will get served the file `index.html` in that directory. An exception to this is any request for a resource under`/ui/static`, which will get served the contents of `$UCM_WEB_UI/static` as if it were a static website.

If left blank, this defaults to a `ui` subdirectory under UCM's working directory. If that directory doesn't exist, the UI will not be served, and `/ui` will give a 404 error.

### Server port

Environment variable: `UCM_PORT`
Command line option: `--port <PORT>`

Controls the port that the server listens on. If not specified, the server will choose an arbitrary available port.

### Host name

Environment variable: `UCM_HOST`
Command line option: `--host <STRING>`

Controls the hostname of the server. If not specified, the server will bind to the host`127.0.0.1`.

### Authentication token

Environment variable: `UCM_TOKEN`
Command line option: `--token <STRING>`

Sets the authentication token that must be present in any request to the server. If not specified, the server will generate a random string as the token.

The token has to be given as a query parameter. For example, if the token is `foo`, then a request for the UI has to be `http://host:port/ui?foo`.

I was not able to easily figure out how to make the token part of the URI path, e.g. `token/ui`. It was relatively easy to do with a query parameter, so I left it at that.

UCM no longer prints the `help` text if you start it with unknown arguments. Arguments not captured by `main` get forwarded to the codebase server which uses `optparse-applicative` to grab them. Therefore they're also not part of the `help` text.

## Loose ends

We should really convert all UCM command-line options to `optparse-applicative` and print a unified usage text that covers both the standard UCM options and the UCM server options.

### DIFF
--- a/parser-typechecker/package.yaml
+++ b/parser-typechecker/package.yaml
@@ -202,6 +202,7 @@ library:
     - unison-core
     - unliftio
     - unliftio-core
+    - utf8-string
     - util
     - unicode-show
     - vector
--- a/parser-typechecker/package.yaml
+++ b/parser-typechecker/package.yaml
@@ -158,6 +158,7 @@ library:
     - hashtables
     - haskeline
     - http-types
+    - http-media
     - io-streams
     - lens
     - ListLike
@@ -185,6 +186,7 @@ library:
     - servant-docs
     - servant-openapi3
     - servant-server
+    - servant-auth-server
     - shellmet
     - split
     - stm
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
LGTM. Agree on converting everything to optparse-applicative

PR #1909, opened 2021-05-05
Stop using old `io.Mode` and `io.BufferMode` types in favor of versions in `io2` package

### LINKED ISSUE #1890: Stop referencing io.BufferMode, io.Mode
In #1871, we discovered `io.BufferMode` and `io.Mode` are still being referenced from code (see comments there for links)

pchiusano:
DRAFT


pchiusano:
@dolio can I hand this off to you? The code compiles with the old types removed and there's a test in `io.md` that should start passing once it's fixed. I think the calling convention for `setBuffering` just needs tweaking like you said.

I'd also audit the `ForeignConvention BufferMode` - it's taking that block size on the unboxed stack which I'm not sure is correct.

dolio:
I think I've worked through the rest.  Are any more tests needed or anything like that?

pchiusano:
I think if that test passes that's pretty decent coverage.

pchiusano:
@dolio LGTM. You just have to approve the PR since technically I opened it.

pchiusano:
Actually, hang on, should we bump the version number of the builtins affected by this? I feel like yeah... any caller of `openFile` or `{get,set}Buffering` would start bombing after this change.

pchiusano:
@dolio okay all set, I bumped the version number of `openFile`, `getBuffering` and `setBuffering`. You just have to approve the PR and it'll merge.

dolio:
I don't think the version needed to be chaged. This only changed the internal implementation of these things, not the Unison API. The type, calling convention, etc. of these functions is the same as before; they're just no longer implemented incorrectly.

pchiusano:
Just recording for posterity, yes @dolio was right - 

```
.> find openFile.impl getBuffering.impl setBuffering.impl
  1. builtin.io2.IO.getBuffering.impl : io2.Handle ->{io2.IO} Either Failure io2.BufferMode
  2. builtin.io2.IO.openFile.impl : Text -> FileMode ->{io2.IO} Either Failure io2.Handle
  3. builtin.io2.IO.setBuffering.impl : io2.Handle -> io2.BufferMode ->{io2.IO} Either Failure ()
```

`FileMode` is from `io2`, the old type was called `io.Mode`.

dolio (approved):
I think this looks good now.

Issue #1902, opened 2021-05-02
missing data case + dumpData: bad closure


aryairani:
I didn't spot an existing ticket for this, though something similar seems to happen in #1901.

I encounter this on trunk (c5d8598b98d77573df7b4e265a5a51f3ce667844) when running the `test` command on a codebase that contains `base`.

Steps to reproduce :
```
.> pull https://github.com/unisonweb/base _base
.> test
```
Output:
```    
    Cached test results (`help testcache` to learn more)
    
    ◉ _base.releases._M1m.Abort.tests.ex1                                : Proved.
    ◉ _base.releases._M1m.Abort.tests.ex2                                : Proved.
    ◉ _base.releases._M1m.Abort.tests.ex3                                : Proved.
    ◉ _base.releases._M1m.Ask.tests.ex1                                  : Proved.
    ◉ _base.releases._M1m.Boolean.given.test                             : Proved.
    ◉ _base.releases._M1m.Boolean.implies.test                           : Proved.
    ◉ _base.releases._M1m.Boolean.xor.test                               : Proved.
    ◉ _base.releases._M1m.Char.ascii.tests.propUpperLower                : Passed 128 tests.
    ◉ _base.releases._M1m.Char.ascii.tests.toLower                       : Proved.
    ◉ _base.releases._M1m.Char.ascii.tests.toUpper                       : Proved.
    ◉ _base.releases._M1m.Either.mapLeft.tests.ex1                       : Proved.
    ◉ _base.releases._M1m.Either.mapRight.tests.ex1                      : Proved.
    ◉ _base.releases._M1m.Exception.tests.ex1                            : Proved.
    ◉ _base.releases._M1m.Exception.tests.ex2                            : Proved.
    ◉ _base.releases._M1m.Exception.tests.ex3                            : Proved.
    ◉ _base.releases._M1m.Int.decrement.test                             : Passed 100 tests.
    ◉ _base.releases._M1m.Int.inRange.test                               : Passed 100 tests.
    ◉ _base.releases._M1m.List.all.tests.deMorgan                        : Passed 100 tests.
    ◉ _base.releases._M1m.List.all.tests.homomorphism                    : Passed 100 tests.
    ◉ _base.releases._M1m.List.any.tests.deMorgan                        : Passed 100 tests.
    ◉ _base.releases._M1m.List.any.tests.homomorphism                    : Passed 100 tests.
[...]

Issue #1901, opened 2021-05-01
Errors in Map.fromList


asoltysik:
Working minimal case:
```
> Map.fromList [(1, "a"), (2, "b")]
```
Result:
```
    1 | > Map.fromList [(1, "a"), (2, "b")]
          ⧩
          Bin 2 1 "a" Tip (Bin 1 2 "b" Tip Tip)
```

Failing minimal case:
```
> Map.fromList [(1, "a"), (2, "b"), (3, "c")]
```
Result:
```
dumpData: bad closure: Foreign (Wrap ##Sequence _)
```

It doesn't seem to be exactly related to number of elements. For example this doesn't throw an error
```
> Map.fromList [ (774, "MACADDR"), (774, "MACADDR8"), (3614, "TSVECTOR"), (3615, "TSQUERY") ]
``` 
but it produces an incorrect result (the map contains only the last entry from the list).:
```
    1 | > Map.fromList [ (774, "MACADDR"), (774, "MACADDR8"), (3614, "TSVECTOR"), (3615, "TSQUERY") ]
          ⧩
          Bin 1 3615 (Cons "TSQUERY" ()) Tip Tip
```

However this one throws an error. Notice that the only thing that changed is that the first key became `773` instead of `774`:
```
> Map.fromList [ (773, "MACADDR"), (774, "MACADDR8"), (3614, "TSVECTOR"), (3615, "TSQUERY") ]
```

No idea whether the "bad closure" error and the incorrect result error are related.

**Unison version:** devel/M1l-1068-gb0dd6e549 (close to current HEAD of trunk)
**New runtime:** yes

pchiusano:
Thanks for reporting. It looks like Map.fromList does weird things with duplicate entries. That is just a bug in base. @runarorama can take a look.

The runtime error sounds like an unrelated decompilation bug - watch expressions get decompiled back to terms before being printed.

pchiusano:
Update on this, #1910 should fix the `bad closure` error. Still digging on the other issue. 

@dolio can you check this example when you fix the `||` compilation issue that we spotted?

pchiusano:
Fixed in trunk

PR #1910, opened 2021-05-05
Fix some variable capture problems

dolio:
This fixes some issues in the compiler for the new runtime that could cause variable confusion, leading to runtime errors about bad closures and the like. It fixes those aspects of #1901 and #1902.

Some Map tests still fail. I'm uncertain (yet) if this is a runtime bug or a Map bug.


pchiusano:
@dolio cool. Do you happen to have a smaller example that you could add as a regression test? If not I wouldn't worry about it, but I like to see regression tests for fixes. :)

pchiusano (approved):

dolio:
For one of the problems there's a much easier example I can add. That was probably causing the actual problem.

The general variable capture ones I'm not sure how to make an easy example.

pchiusano:
> For one of the problems there's a much easier example I can add. That was probably causing the actual problem.

K, I'd do that.

> The general variable capture ones I'm not sure how to make an easy example.

I wouldn't sweat it then.

PR #1911, opened 2021-05-06
Fix a boolean in the new runtime

dolio:
The ANF code had a the literal for `true` defined the same way as `false`. In that file it was only used in `||`, but I think it might also be used for some other builtins.

This does seem to fix the duplicate-key issue in #1901, although I'm unsure exactly which of its dependencies was broken. This also gets base's tests working.

I added a few truth table tests to a transcript.


pchiusano (approved):
👍

PR #1913, opened 2021-05-06
Delete unused code only used by old runtime

pchiusano:
The module `IR.hs` was dead code, as was a bunch of code in `ANF.hs` that isn't used by the new runtime.

Thanks to Michael Nygard in Slack for spotting this!


dolio (approved):
Looks good to me.

PR #1920, opened 2021-05-08
Move auth token from query string to resource path

### LINKED ISSUE #1914: Serve API and Codebase UI with the unique token as part of the url instead of query string
To ease the experience of running to UI with `ucm` we want to move the token from the query string to be part of the UI path:

`http://127.0.0.1:64739/api?ABC123` becomes `http://127.0.0.1:64739/ABC123/api`
`http://127.0.0.1:64739/ui?ABC123` becomes `http://127.0.0.1:64739/ABC123/ui`

(`ABC123` being a token example).

Note, a minor change in the UI is required for fetching assets via a relative path.

runarorama:
This changes the UCM server such that it serves the API and UI from `/token/api` and `/token/ui`, respectively, where `token` is the authentication token.

Fixes #1914


pchiusano (approved):
@runarorama cool. 

Question: how does one launch the server without the auth token? This will be needed for the version that's running on unison-lang.org.

PR #1925, opened 2021-05-10
Move `static` outside the auth token

runarorama:
This moves static content (like the UI javascript code and the favicon) outside of the auth protection, so it can be served without knowing what the token is (which was creating a chicken-and-egg situation).


pchiusano (commented):
LGTM but will let @hojberg review.

hojberg:
Works in my test with https://github.com/unisonweb/codebase-ui/pull/91 🎉

hojberg (approved):
🎉

Issue #1912, opened 2021-05-06
Additional Doc2 type features


runarorama:
Some features that I would like to have in Unison docs, in decreasing order of importance:

* **Checked Unison expressions** - Just like backtick snippets, but where the expression is typechecked. So kind of like `@eval` but just displaying the code instead of actually evaluating.
* **Get the type of a Unison expression** - Similar to the above, but instead of displaying the code, it displays the type of the expression. Maybe `@typeOf`.
* **Get the definition of a name** - Similar to `@source` but instead of showing the full source, it just expands the term to its definition. E.g. if I have `foo x = 10 + x`, then `@definition{foo}` expands to `x -> 10 + x`.

PR #1927, opened 2021-05-10
Add @typecheck block for including code snippets that are typechecked but not evaluated

pchiusano:
This addresses the first item of #1912. It adds a new block element type:

````
@typecheck ```
foo : Nat -> Nat
foo x = x + 1
```
````

Which just typechecks the above but doesn't evaluate it. Added a test to the `doc.md` transcript.

This syntax can be changed easily later if anyone has strong feelings about it.

I decided not to try either of the other items from #1912 for now.

### DIFF
--- a/parser-typechecker/src/Unison/TermParser.hs
+++ b/parser-typechecker/src/Unison/TermParser.hs
@@ -486,6 +486,9 @@ doc2Block =
         tm -> Term.apps' f [Term.nat (ann tm) 0, addDelay tm]
       "syntax.docTransclude" -> evalLike id
       "syntax.docEvalInline" -> evalLike addDelay
+      "syntax.docExampleBlock" -> do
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: parser-typechecker/src/Unison/TermParser.hs [hunk 1]
```diff
@@ -963,6 +966,11 @@ data BlockElement v
   | DestructuringBind (Ann, Term v Ann -> Term v Ann)
   | Action (Term v Ann)
 
+instance Show v => Show (BlockElement v) where
+  show (Binding ((pos,name), _)) = show ("binding: ", pos, name)
+  show (DestructuringBind (pos, _)) = show ("destructuring bind: ", pos)
+  show (Action tm) = show ("action: ", ann tm)
```
pchiusano, line 972: Added for debugging purposes. Think I'll leave it.

runarorama (approved):

Issue #1926, opened 2021-05-10
Database integrity error with watch expression cache


pchiusano:
```
CallStack (from HasCallStack):
  error, called at src/Unison/Codebase/SqliteCodebase.hs:261:81 in unison-parser-typechecker-0.0.0-2dM7tNDqeOvDQFHAl25PFL:Unison.Codebase.SqliteCodebase
  getCycleLen, called at src/Unison/Codebase/SqliteCodebase.hs:587:38 in unison-parser-typechecker-0.0.0-2dM7tNDqeOvDQFHAl25PFL:Unison.Codebase.SqliteCodebase
watch-integrity.md
```

pchiusano:
There's a regression test in https://github.com/unisonweb/unison/tree/fix/1926

aryairani:
This issue is similar to the one we worked around for patches, in bd949334a759d8a931bdbeed0dd0969ddc2b97b7.

In the regression test:
```unison
> 'sq

sq = 2934892384
```
The issue is that watch result term (`'sq`) contains a reference `sq`, which is not in the codebase.  This is expected, but to represent the reference in memory, we later ask the codebase to tell us the component cycle length of `sq`, which it doesn't know, which causes the crash.

As a fix, I guess I'm going to start off by crashing lazily instead of eagerly, and I think that will do the trick since nobody should be reading it anyway. 🤞

aryairani:
Crashing lazily was no help, so I tried returning a cycle length of `0`, thinking it wouldn't matter, but I got some output that I need some help understanding.

The transcript:
````
```unison
> 'sq

sq = 2934892384
```

```unison
> 'sq

sq = 2934892384
```
````
The debug output:
```
⚙️   Processing stanza 2 of 3.
getWatch "" #ptm7t
getWatch "test" #ptm7t
getWatch "" #ju93p
getWatch "test" #ju93p
⚙️   Processing stanza 3 of 3.
getWatch "" #ptm7t
getWatch "test" #ptm7t
getWatch "" #ju93p
unison: PE Lit
  AnnotatedText (fromList [Segment {segment = "Unknown term reference: #ptm7t.0c0", annotation = Nothing}])
[...]

PR #1928, opened 2021-05-11
drop Reference.Id cycle length from Eq/Ord instances

aryairani:
## Overview

This PR takes the cycle length field out of the `Eq` and `Ord` instances for `Reference.Id`, so that they can be found in a collection even if the cycle size isn't known at query time.

I still produce an `error` when looking up an unknown cycle length in the v2 codebase (though only if the value is bound and then forced), so that will be something to keep an eye on going forward.

## Test coverage

It includes the previously-failing transcript that inspired #1926.

### DIFF
--- a/unison-core/src/Unison/Reference.hs
+++ b/unison-core/src/Unison/Reference.hs
@@ -191,9 +191,6 @@ instance Hashable.Hashable Reference where
   tokens (Builtin txt) = [Hashable.Tag 0, Hashable.Text txt]
   tokens (DerivedId (Id h i n)) = [Hashable.Tag 1, Hashable.Bytes (H.toBytes h), Hashable.Nat i, Hashable.Nat n]
 
--- | A simpler version of `Id` that we'll use for some derived instances.
-data Id' = Id' H.Hash Pos deriving (Eq, Ord)
-toId' :: Id -> Id'
-toId' (Id h p _) = Id' h p
+-- | Two references mustn't differ in cycle length only.
 instance Eq Id where x == y = compare x y == EQ
 instance Ord Id where Id h i _ `compare` Id h2 i2 _  = compare h h2 <> compare i i2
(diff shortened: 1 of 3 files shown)


unison-src/transcripts/fix1926.md
```diff
@@ -0,0 +1,15 @@
+```ucm
+.> builtins.merge
```
aryairani, line 2: Sometimes I'll just use an `alias.type ##Nat Nat` instead of a full `builtins.merge`, but I didn't wanna modify it in this case.

unison-core/src/Unison/Reference.hs
```diff
@@ -190,3 +190,10 @@ instance Show Reference where show = SH.toString . SH.take 5 . toShortHash
 instance Hashable.Hashable Reference where
   tokens (Builtin txt) = [Hashable.Tag 0, Hashable.Text txt]
   tokens (DerivedId (Id h i n)) = [Hashable.Tag 1, Hashable.Bytes (H.toBytes h), Hashable.Nat i, Hashable.Nat n]
+
+-- | A simpler version of `Id` that we'll use for some derived instances.
+data Id' = Id' H.Hash Pos deriving (Eq, Ord)
+toId' :: Id -> Id'
+toId' (Id h p _) = Id' h p
+instance Eq Id where x == y = toId' x == toId' y
+instance Ord Id where x `compare` y = toId' x `compare` toId' y
```
pchiusano: I'd just write the instance out rather than introducing yet another thing ending in `'` which is only used for this one thing. :)

```suggestion
instance Eq Id where x == y = compare x y == EQ
instance Ord Id where Id h i _ `compare` Id h2 i2 _  = compare h h2 <> compare i i2
```

Also happens to be fewer lines. :)

FYI, the semigroup for `Ordering` does what you'd expect:

```Haskell
instance Semigroup Ordering where
    LT <> _ = LT
    EQ <> y = y
    GT <> _ = GT
```

pchiusano (approved):
Thanks for quick turnaround. Left one comment for you but other than that LGTM.

unison-core/src/Unison/Reference.hs
aryairani: Okay, cool thanks.

aryairani:
With this, we might be able to back out the previous workaround that added otherwise unnecessary dependencies for patches.

PR #1935, opened 2021-05-12
add gitsync22.*.lightweightPatch test

aryairani:
This adds a test that pushes and pulls a branch containing a patch but not explicitly containing the new or old definitions.  

This will be a sanity check once we quit explicitly syncing the old definitions along with a patch.  

I tried reverting the workaround that made old definitions full dependencies of patches (bd949334a759d8a931bdbeed0dd0969ddc2b97b7), hoping that #1928 would make it unnecessary, but it wasn't immediately the case.

### DIFF
--- a/parser-typechecker/tests/Unison/Test/GitSync.hs
+++ b/parser-typechecker/tests/Unison/Test/GitSync.hs
@@ -317,11 +317,14 @@ test = scope "gitsync22" . tests $
       .> builtins.merge
       ```
       ```unison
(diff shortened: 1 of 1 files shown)


pchiusano (approved):

aryairani:
The first CI run blew up with https://gist.github.com/aryairani/136313d116d74de438ba2aeca1013a8c but it seems unrelated to the PR.

PR #1945, opened 2021-05-13
Don't have two versions of Either.mapRight and List.map

### LINKED ISSUE #1944: Change implementation of `mapRight` and `List.map` in IOSource.hs to match what's in base
Right now there's versions of these functions in IOSource that don't match what's in base, leading to name conflicts if/when `builtin` gets merged into `base`, which is something we will be doing on each release.

The workaround in the meantime is to use `replace.term` to patch `builtin` before merging that into base.

—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)

pchiusano:
Fixes #1944 - the versions in `IOSource` now match what's in base, which should prevent future name conflicts when merging new builtins into base.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/IOSource.hs
+++ b/parser-typechecker/src/Unison/Runtime/IOSource.hs
@@ -570,12 +570,15 @@ unique[de2e0ee924578939213c950dfd8e0ba1047703ae] type ANSI.Color
   | BrightBlack | BrightRed | BrightGreen | BrightYellow | BrightBlue
   | BrightMagenta | BrightCyan | BrightWhite
 
-List.map : (a ->{g} b) ->{} [a] ->{g} [b]
-List.map f as =
-  go acc = cases
-    [] -> acc
-    h +: t -> go (acc :+ f h) t
-  go [] as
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -570,12 +570,15 @@ unique[de2e0ee924578939213c950dfd8e0ba1047703ae] type ANSI.Color
...
-    h +: t -> go (acc :+ f h) t
-  go [] as
+List.map : (a ->{e} b) -> [a] ->{e} [b]
+List.map f a =
+  go i as acc =
+    match List.at i as with
+      None   -> acc
+      Some a ->
+        use Nat +
+        go (i + 1) as (acc :+ f a)
```
pchiusano, line 580: Kinda interesting that `List.map` in base uses indexing rather than repeated uncons. The idea is to avoid performance hit of allocation and rebalancing while traversing, and my guess is that pays off for the right sequence type, but I wonder if it's true for Data.Sequence, the current implementation of lists.

Anyway, if/when List.map changes in base, we can swap this out.

runarorama (approved):

Issue #1915, opened 2021-05-07
v2 codebase loose ends megaticket


aryairani:
- [ ] ~use transactions~ investigated, but was kind of a pain.  possibly not too necessary given immutable nature of codebase. probably each ucm command should be a transaction, not each Codebase command, but it would mean saddling the v1 codebase with a fake transaction layer; which is perhaps fine if the v1 codebase is going away.
- [x] low quality error message when trying to load codebase that 
  - [x] doesn't exist
  - [x] has the wrong schema version
- [ ] ~back out the dependency workaround for patches~ didn't work as expected
- [ ] something for old names
- [ ] something to help squash namespaces to commits
  - [ ] ensure that we don't end up removing a commit someone else relies on for LCA/merge
- [ ] commit messages / metadata on causals
- [ ] utilize and debug diffing?
- [ ] detect external changes / root branch updates
- [x] move LCA computation to codebase
- [ ] change `Queries.schemaVersion` to use `EDB`
- [ ] use sql dependencies lookup in `ListDependenciesI` instead of loading the definitions
- [ ] use ShortByteString / ShortText in places?

PR #1947, opened 2021-05-13
Induce better error message when getting a codebase that doesn't exist

mitchellwrosen:
## Overview

Addresses part of #1915 

This PR makes a minimally-invasive improvement to the UX of running `unison` at the command line before initializing a codebase. (Previously, a SQLError would bubble up to the top level).

@aryairani and I decided this was ok for now, but noticed a couple weird things along the way that could use some attention, such as how the `Pretty` returned from this function is actually ignored, and replaced by a (nicer) `Pretty` that recommends running some command to initialize the codebase.

Some follow-up work would be good. I recommend going down the route of (whenever in IO, anyway) carving out an exception type for such cases where the application really has no options but to print a nice error message, and just catching these at the top-level. Then, intermediate helper functions need not do any rendering themselves, nor return `Either Pretty Thing` to their callers to indicate an error has occurred.

## Test coverage

No test coverage, but I did run `unison` at the command-line and observed a better error message than prior to this patch.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
@@ -158,7 +158,11 @@ getCodebaseOrError :: forall m. MonadIO m => CodebasePath -> m (Either Codebase1
 getCodebaseOrError dir = do
   prettyDir <- liftIO $ P.string <$> canonicalizePath dir
   let prettyError v = P.wrap $ "I don't know how to handle " <> P.shown v <> "in" <> P.backticked' prettyDir "."
-  fmap (Either.mapLeft prettyError) (sqliteCodebase dir)
+  doesFileExist (dir </> codebasePath) >>= \case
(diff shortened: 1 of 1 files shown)


pchiusano:
Can we see a screenshot of the after? :)

aryairani:
@pchiusano

M1m:
```
% ucm -codebase foo
No codebase exists in /Users/arya/unison-ucm-init/foo.
Run `ucm -codebase foo init` to create one, then try again!
```

trunk:
```
% stack build && stack exec unison -- -codebase foo
unison: SQLite3 returned ErrorCan'tOpen while attempting to perform open "foo/.unison/v2/unison.sqlite3": unable to open database file
```

PR:
```
% stack build && stack exec unison -- -codebase foo
No codebase exists in /Users/arya/unison-ucm-init/foo.
Run `unison -codebase foo init` to create one, then try again!
```

pchiusano (approved):

PR #1955, opened 2021-05-14
fix Sqlite.Queries.lca algorithm

aryairani:
Fixes two bugs, one that would fail if the one (but not both) of the arguments was the LCA; and one that would fail if the LCA was later in one history than the length of the other history. 😬


pchiusano:
Sweet!

PR #1957, opened 2021-05-14
Fix some runtime problems

### LINKED ISSUE #1940: Decompilation issue when rendering docs
Noticed a possible issue with doc rendering, where it produces

```
  cannot decompile an application to a local recusive binding
```

It's weird because I only see the error when calling the doc rendering in a slightly different way than usual, which I wouldn't think would matter at all.

There's a failing transcript in the [fix/1940](https://github.com/unisonweb/unison/tree/fix/1940) branch, see [this file](https://github.com/unisonweb/unison/blob/fix/1940/unison-src/transcripts/bug-strange-closure.md).

```
stack exec transcripts -- bug-strange-closure.md
```

Will run just the failing transcript.

dolio:
This fixes two runtime issues.

One fix retains decompilation information derived from files longer. Before it was being discarded, but loaded code was not, so there were situations where you could end up evaluating something but being unable to decompile it.

The other fixes the possibility of running off the stack when evaluating functions that do hundreds/thousands of lets.

Fixes #1940.


pchiusano (approved):

pchiusano:
Great!

PR #1977, opened 2021-05-18
Remove builtin.List.empty term

runarorama:
This deprecates `builtin.List.empty`, since it's defined in `base` as `base.List.empty = []`

### DIFF
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -426,7 +426,7 @@ builtinsSrc =
   , B "Bytes.fromBase64" $ bytes --> eithert text bytes
   , B "Bytes.fromBase64UrlUnpadded" $ bytes --> eithert text bytes
 
-  , B "List.empty" $ forall1 "a" list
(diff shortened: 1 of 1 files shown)


aryairani:
Question: Our current syntax provides a way to construct an empty list without using the `base` library, but will that be true for every syntax? And if not, should we not keep the builtin?

aryairani:
Hmm. I suppose every syntax should provide a way to construct every term, and `[]` is a term, or should it?  Like what would a Scala-like syntax do?

pchiusano:
Yeah, any future syntax should provide a way of constructing any valid AST, and the empty list is part of the current AST.

That said, this PR just removes that definition from the default builtin namespace. The builtin still exists, which is not really necessary but also not a big deal.

pchiusano (approved):

aryairani:
> That said, this PR just removes that definition from the default builtin namespace. The builtin still exists, which is not really necessary but also not a big deal.

It's necessary just as long as we wanna support programs that have already managed to reference it, though I think we're safe for now.

PR #1981, opened 2021-05-18
Avoid calling `error` in the new runtime

dolio:
Instead use custom defined exceptions that can be caught more reliably in the runtime interface and displayed without crashing.

Hopefully this makes compiler/runtime bugs somewhat less annoying when they occur.


pchiusano (approved):
Great!

PR #1983, opened 2021-05-19
fix missing-deps calculation for syncing branches that only exist in memory

aryairani:
All dependencies of memory-only branches were being treated as missing dependencies, causing an infinite loop during a push that merges.  This PR filters them and adds some more debug output.

I couldn't find a Github issue for the push hang @runarorama and I had been seeing, but if there is one, this closes it.

`termExists` and `declExists` in `SqliteCodebase` share an implementation, continuing with the assumption that we won't have hash collisions between terms and decls.

## Test coverage

I haven't added a minimized transcript to demonstrate the failure, but would if requested.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
@@ -567,6 +567,12 @@ sqliteCodebase root = do
             runDB conn . fmap isJust $
               Ops.primaryHashToMaybePatchObjectId (Cv.patchHash1to2 h)
 
+          -- well one or the other. :zany_face: the thinking being that they wouldn't hash-collide
+          termExists, declExists :: MonadIO m => Hash -> m Bool
+          termExists h = runDB conn . fmap isJust $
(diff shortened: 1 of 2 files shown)


pchiusano:
Nice!

pchiusano (approved):

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -869,7 +890,7 @@ emptySyncProgressState = SyncProgressState (Just mempty) (Right mempty) (Right m
 syncProgress :: MonadState SyncProgressState m => MonadIO m => Sync.Progress m Sync22.Entity
 syncProgress = Sync.Progress need done warn allDone
   where
-    quiet = False
+    quiet = True
```
pchiusano: Just checking - is this intentional?

pchiusano (approved):

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
aryairani: Unintentional—thanks.

PR #1994, opened 2021-05-20
Add `directoryContents` builtin

runarorama:
This adds a `directoryContents` builtin, implemented by Haskell's`System.Directory.getDirectoryContents`.


dolio (approved):
I didn't realize it would be this easy, but it seems legit.

pchiusano:
@runarorama You can add tests in `io.md`.

dolio:
Oh, there's one possible issue. It seems that the `BuiltinForeign` that is getting used here will be `FilePath`. It doesn't seem like that `FilePath` instance should even exist, since it builds a `Foreign` with the `FilePath` reference, but the latter seems to be a data type. And your declared type says it will return `Text` values, which it technically won't.

So I think the Haskell function should actually be something like `fmap pack <$> getDirectoryContents`, to use the `Text` instance.

It won't matter that the input is also a `FilePath`, because there is a `String` instance for `ForeignConvention` that will correctly unpack the `Text` (I think).

dolio (approved):
Okay, looks good now.

pchiusano:
I'd still like to see at least a quick test of it in `io.md` before merging. Could probably piggyback on this test, which creates a temp directory:

https://github.com/unisonweb/unison/blob/trunk/unison-src/transcripts/io.md#creatingdeletingrenaming-directories

unison-src/transcripts/io.md
```diff
@@ -191,3 +191,20 @@ testSystemTime _ =
...
+
+### Get directory contents
+
+```unison
+testDirContents : '{io2.IO} [Result]
+testDirContents _ = 
+  test = 'let
+    tempDir = newTempDir "dircontents"
+    c = reraise (directoryContents.impl tempDir)
+    check "directory size should be"  (size c == 2)
```
pchiusano, line 198: This testing is pretty thin - I'd create a file or two in the directory and check the name of the file coming back from getDirectory equals what was created. If you're just checking length then it's not testing that the Text values coming back from Haskell are encoded correctly as Unison text.

I did a double take that an empty directory has size 2 but I guess that's for current directory and parent? I feel like a version that filters these out would be more commonly used (think of like walking a directory tree) but that could also be done in pure Unison. Wdyt?

runarorama:
I think the testing is fine. The size wouldn't be exactly 2 unless it were working properly, and encoding/decoding names is "not done here". If it weren't working then no text operations would work at all.

pchiusano:
We have definitely both spent more time on these comments than it would have taken to just add the test 😀, but I am hoping to get on the same page, so let me give my thinking -

This code you wrote can typecheck returning stuff other than text, just depending on what `pack` function is in scope. (For instance, various `ByteString` types also have a `pack` function that accepts strings... without checking are you 100% certain you're not accidentally calling one of those here?)

```Haskell
  declareForeign "IO.directoryContents.impl.v3" boxToEFBox
    $ mkForeignIOF $ (fmap pack <$>) . getDirectoryContents
```

It's also possible that we'll be changing the Unison runtime representation of `Text` (I've got a branch where I started tinkering with this) and/or `List` and that after such a change the above code could still typecheck and be wrong. A test like I suggested would catch it.

Basically, the Haskell typechecker is just verifying that you're returning something that has a foreign instance, not that what you're returning is the type of thing expected by the particular Unison builtin you're implementing. It's like Haskell typechecks `f` but the full builtin is actually `g . f` and their composition isn't statically checked. So I would err on the side of basic tests that enforce that the right kind of thing is being returned.

pchiusano (approved):
@runarorama I went ahead and added the extra testing and synced with trunk since I'm assuming you're busy with the other codebase server backend changes. It'll merge after CI completes.

pchiusano:
CI is failing on linux on that test I added, I haven't gotten a chance to investigate. My test assumes the order is `[".", ".."]` so maybe that's too fragile.

pchiusano:
Tried checking for either order in the test, hopefully that does it.

PR #1996, opened 2021-05-21
Speed up `push` by using fast SQL-based `Branch.before` calculation

### LINKED ISSUE #1958: Push should sync to the staging codebase and do the merge there
By syncing to the staging codebase first, the staging codebase should have both branches locally and its LCA calculation will be done in codebase and be super fast.

Loose end after #1948

—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)

pchiusano:
This PR does a few things to close #1958 

* `push` no longer does a merge with the remote namespace - this isn't necessary since the thing you are pushing must be a fast forward or the `push` is rejected. So either it will be rejected or it will succeed with the result equaling the thing you're pushing. Basically, this is the observation that when `before a b`, `merge a b == b`. Just doing a `before` check rather than a full merge is a big deal since a cross-database merge isn't something we can do efficiently yet. 
* The `Branch.before` check that we do before `push` (to make sure that what you're pushing is a fast forward) is now done efficiently within the codebase.
* The two new tests demonstrate that ucm can [still] distinguish between a push to a codebase that is `before` us, vs. one that is not.

__Todo:__

- [x] During push, make sure to sync to staging codebase before calling before on staging codebase
- [x] @aryairani to audit and/or possibly improve the `SqlCodebase.beforeImpl` implementation and replace it with some SQL wizardry if the current LCA-based implementation is no good.
- [x] update `before` query to short-circuit per https://sqlite.org/forum/forumpost/ed2d66f197?t=h
- [x] verify that we have tests that will fail on non-FF push

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
@@ -180,7 +180,8 @@ codebase1' syncToDirectory branchCache fmtV@(S.Format getV putV) fmtA@(S.Format 
           (termReferentsByPrefix (getDecl getV getA) path)
           (pure 10)
           (branchHashesByPrefix path)
-          lca
+          (\_ _ -> pure Nothing) -- just use in memory Branch.lca
+          (\_ _ -> pure Nothing) -- just use in memory Branch.before
    in pure c
   where
     dependents :: Reference -> m (Set Reference.Id)
@@ -189,7 +190,6 @@ codebase1' syncToDirectory branchCache fmtV@(S.Format getV putV) fmtA@(S.Format 
     getTermsOfType r = listDirAsReferents (typeIndexDir path r)
     getTermsMentioningType :: Reference -> m (Set Referent.Id)
     getTermsMentioningType r = listDirAsReferents (typeMentionsIndexDir path r)
-    lca b1 b2 = Branch.lca b1 b2
   -- todo: revisit these
     listDirAsIds :: FilePath -> m (Set Reference.Id)
     listDirAsIds d = do
--- a/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/FileCodebase.hs
@@ -180,8 +180,8 @@ codebase1' syncToDirectory branchCache fmtV@(S.Format getV putV) fmtA@(S.Format 
           (termReferentsByPrefix (getDecl getV getA) path)
           (pure 10)
           (branchHashesByPrefix path)
-          (\_ _ -> pure Nothing) -- just use in memory Branch.lca
-          (\_ _ -> pure Nothing) -- just use in memory Branch.before
+          Nothing -- just use in memory Branch.lca
+          Nothing -- just use in memory Branch.before
    in pure c
   where
     dependents :: Reference -> m (Set Reference.Id)
--- a/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
@@ -701,25 +701,15 @@ sqliteCodebase root = do
             cs <- Ops.causalHashesByPrefix (Cv.sbh1to2 sh)
             pure $ Set.map (Causal.RawHash . Cv.hash2to1 . unCausalHash) cs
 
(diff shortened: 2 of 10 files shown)


parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -816,23 +818,23 @@ pushExhaustive = InputPattern
   (P.lines
     [ P.wrap $
       "The " <> makeExample' pushExhaustive <> "command can be used in place of"
-        <> makeExample' push <> "to repair remote namespaces"
+        <> makeExample' pushExhaustive <> "to repair remote namespaces"
```
aryairani: oops, used "in place of `push`"

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -816,23 +818,23 @@ pushExhaustive = InputPattern
...
     url : rest -> do
       (repo, sbh, path) <- parseUri "url" url
       when (isJust sbh)
         $ Left "Can't push to a particular remote namespace hash."
       p <- case rest of
         [] -> Right Path.relativeEmpty'
         [path] -> first fromString $ Path.parsePath' path
-        _ -> Left (I.help push)
-      Right $ Input.PushRemoteBranchI (Just (repo, path)) p SyncMode.Complete
+        _ -> Left (I.help pushExhaustive)
```
aryairani: This change seems right, though.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -292,13 +293,16 @@ pushGitRootBranch
...
   -> Branch m
   -> RemoteRepo
   -> SyncMode
   -> ExceptT GitError m ()
-pushGitRootBranch syncToDirectory cache branch repo syncMode = do
+pushGitRootBranch syncToDirectory cache onEmptyDest branch repo syncMode = do
   -- Pull the remote repo into a staging directory
   (remoteRoot, remotePath) <- viewRemoteBranch' cache (repo, Nothing, Path.empty)
+  when (remoteRoot == Branch.empty && onEmptyDest == AbortOnEmptyDest) $
+    throwError $ GitError.PushDestinationIsEmpty repo
```
aryairani: This is checking if the repo root is empty, but I think we instead want it to check the target sub-namespace before we get in here.

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -960,6 +966,16 @@ syncProgress = Sync.Progress need done warn allDone
       where
         v = const ()
 
+-- This might be handy later, if only as a starting point for something that ultimately returns a Connection, else delete
+-- getCodebaseForRepo :: forall m. MonadIO m => RemoteRepo -> m (Either GitError (m (), Codebase m Symbol Ann))
+-- getCodebaseForRepo repo = runExceptT @GitError @m do
```
pchiusano: Maybe delete this?

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -1005,47 +1021,91 @@ viewRemoteBranch' (repo, sbh, path) = runExceptT do
...
+    --   withStatus ("Staging codebase for upload to " ++ repoString ++ " ...") do
+    --     time "SqliteCodebase.pushGitRootBranch.stageAndPush.syncToDirectory" $ lift (syncToDirectory remotePath syncMode branch)
+    --     setRepoRoot remotePath (Branch.headHash branch)
+    --   -- push staging area to remote
+    --   withStatus ("Uploading to " ++ repoString ++ " ...") $
+    --     unlessM
+    --       ( push remotePath repo
+    --           `withIOError` (throwError . GitError.PushException repo . show)
+    --       )
+    --       (throwError $ GitError.PushNoOp repo)
```
pchiusano: Maybe delete

pchiusano (commented):
Left a couple comments

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
aryairani: deleted

aryairani (approved):

PR #1999, opened 2021-05-22
Add getLine I/O builtin

runarorama:
There's not really a good way to implement `getLine` (read a line of text from a file handle) without adding a new builtin. Added a builtin that is a foreign call out to `hGetLine`.


parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1408,9 +1409,15 @@ declareForeigns = do
   declareForeign "IO.setBuffering.impl.v3" set'buffering
     . mkForeignIOF $ uncurry hSetBuffering
 
-  declareForeign "IO.getBytes.impl.v3" boxNatToEFBox .  mkForeignIOF $ \(h,n) -> Bytes.fromArray <$> hGet h n
+  declareForeign "IO.getLine.impl.v3" boxToEFBox . mkForeignIOF
+    $ \h -> pack <$> hGetLine h
```
dolio: This probably doesn't make much of a difference, but there's an `hGetLine` in `Data.Text.IO`.

parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -531,6 +532,7 @@ ioBuiltins =
   , ("IO.createDirectory.impl.v3", text --> iof unit)
   , ("IO.removeDirectory.impl.v3", text --> iof unit)
   , ("IO.renameDirectory.impl.v3", text --> text --> iof unit)
+  , ("IO.directoryContents.impl.v3", text --> iof (list text))
```
dolio: Was directory contents stuff meant to be in this PR? I guess it'll probably go away once the other PR is merged?

pchiusano:
This is a weird operation to exist for any Handle since it would have to pick a text encoding out of a hat, which may not match the underlying source of bytes. I think the Haskell function does something random using the default locale but we don't want that here. Programs that use default locale are just bugs waiting to happen.

The encoding could be specified as an argument. Or maybe this operation could be specialized to stdin where using the default locale is fine? (Is it? @stew)

I think this one could wait for after M2.

pchiusano:
Did some more digging. I think a function with the proposed signature is fine assuming we have a way to set the encoding on a Handle and then whatever function we are using for getLine respects that encoding which has been set:

https://hackage.haskell.org/package/base-4.15.0.0/docs/System-IO.html#g:23

There's the question of what the encoding on a Handle should be if the user doesn't specify. We could default to the system locale or to utf8.

Standard in will have the encoding set to whatever.

parser-typechecker/src/Unison/Runtime/Builtin.hs
pchiusano: Yes that seems better / faster especially if you're piping a large file or something to standard in

And it looks like that uses the encoding set on the Handle: http://hackage.haskell.org/package/text-1.2.4.1/docs/Data-Text-IO.html#g:3

It says it uses the "default locale and/or the encoding on the Handle" which I think can only mean it uses what's on the Handle or otherwise falls back to the default locale.

pchiusano:
@runarorama I think this PR just needs a refresh against trunk, a unit test in `io.md` (maybe write a line to a temp file and read it back?), and [this comment](https://github.com/unisonweb/unison/pull/1999#discussion_r638038747) addressed.

Or @samgqroberts if you want to take a crack at updating sooner than Rúnar can get to it, then go for it (and let us know so we don't step on each other's toes 😀).

pchiusano:
Also, we can leave it out of scope for this PR to set the encoding on a `Handle`, but I think we should definitely support that at some point.

pchiusano (approved):
👍 

I had to do another round of transcript refreshing since there were some intervening updates to trunk.

PR #2002, opened 2021-05-24
Fix to `squash a b` to properly short circuit in the case where `b <= a`

### LINKED ISSUE #2000: Issue with squash
Two issues:

- [ ] The LCA calculation will gleefully return the empty branch as the LCA, which can be technically true but is almost never the LCA you want.
- [ ] `squash a b` where `a` is ahead of `b` needs to have some short circuiting logic that doesn't include `a` history. `merge` has such short circuiting.

pchiusano:
Fixes #2000 and adds a regression test.

- [x] Possibly add special case to LCA to avoid returning the empty branch unless there really are no other possible LCAs. We started on this but it made some unit tests fail and I'm not sure we had a non-hacky fix.
- [x] Add short circuiting of `squash`.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -397,10 +397,8 @@ merge'' _ mode b1 b2 | isEmpty b2 = case mode of
     pure $ cons (discardHistory0 (head b1)) b2
 merge'' lca mode (Branch x) (Branch y) =
   Branch <$> case mode of
-               RegularMerge -> Causal.threeWayMerge' lca' combine x y
-               SquashMerge  -> do
-                 traceM "Squashin' ..."
-                 Causal.squashMerge' lca' combine x y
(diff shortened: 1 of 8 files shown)


pchiusano:
@runarorama This PR should be rebased onto trunk, looks like it includes some of the new I/O primitives being proposed.

runarorama (approved):

pchiusano:
CI is failing due to fix2000.output.md and squash.output.md needing a refresh/ resolve conflict. Other than that LGTM.

I'll open a ticket to track updating the sql implementation.

pchiusano:
@runarorama - fyi, @aryairani and I had an idea for how we can leave the LCA calculation alone, and instead just say that:

     Branch.cons a Branch.empty = Branch.one a

This avoids making `Branch.empty` an ancestor of all branches - basically the first nonempty commit is used as the start of history. This might be a surgical change to `Branch` and feels less hacky than modifying `Causal`. We're exploring it now.

PR #2007, opened 2021-05-25
@signature is now @inlineSignature

pchiusano:
This is a minor papercut fix for the documentation format. 

Previously, `@signature{List.map}` would include a signature inline in a paragraph, and `@signatures{List.map, List.filter}` could be used to include 1 or more signatures as a block element. It's confusing to have singular vs plural decide block element vs inline element. @rlmark and I spotted this.

After this change it's `@inlineSignature{List.map}` for the inline version and `@signature{List.map}` (or `@signatures{List.map, List.filter}`) for the block version. I think this is a good default since signatures often don't fit nicely in the middle of a paragraph.

Note that `@signature` vs `@signatures` mean exactly the same thing. I didn't see any reason to disallow one or the other.

For testing, I just updated the syntax guide to use all of `@signature`, `@signatures`, and `@inlineSignature`.

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -375,12 +375,12 @@ lexemes' eof = P.optional space >> do
           _ <- lit "}"
           pure (join s)
         signature = wrap "syntax.docSignature" $ do
-          _ <- lit "@signatures" *> (lit " {" <|> lit "{") *> CP.space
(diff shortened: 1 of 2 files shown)


rlmark (approved):
LGTM!

Verification process: Pulled branch, built executable, observed both `@signature` and `@signatures` show up in docs as block signatures and `@inlineSignature` is rendered as inline.

PR #2008, opened 2021-05-25
Fix bug that would cause scopes to be improperly reported in EasyTest

### LINKED ISSUE #1860: fix off-by-one scope reporting for exceptions caught by EasyTest

pchiusano:
Fixes #1860 

I had to test this manually since there's not currently a combinator to check that a test fails with a particular scope. It would be pretty easy to add but I didn't have time now.


aryairani (approved):
Great, thanks.

Issue #2009, opened 2021-05-25
Round trip tests for new docs format


pchiusano:
There's currently a transcript that exercises parsing, pretty-printing, and display of docs, but it doesn't test that what the pretty-printer produces can be parsed again. 

We have some unit tests of this round trip property for other elements of Unison syntax. It's a little more involved to create this for the new doc type because of how it parses - it desugars into calls to `docWord`, `docParagraph`, and so on.

PR #2010, opened 2021-05-25
Fix to pretty-printing of 

pchiusano:
Oversight from #2007 - the parser was updated but the pretty-printer wasn't updated to match!

We could use round trip tests for the doc syntax - that would have caught this bug. I've created https://github.com/unisonweb/unison/issues/2009 to track that.


rlmark (approved):
👍

PR #2012, opened 2021-05-25
add `pragma optimize` on database close

aryairani:
## Overview

Sqlite's `pragma optimize` computes some metrics that the query optimizer had wished for earlier in the execution.

For long-lived connections (e.g. maybe the share server), it's recommended that you run `pragma optimize`
every so many hours or days; for short-lived connections (like desktop ucm processes) maybe on shutdown will suffice,
to benefit the user at the next startup.

I don't know if it will really help everything, because I think our database is pretty straightforward and I've already added all the indexes needed to run things optimally, but I saw in debugging (with the new `debugPragmaOptimize` setting) that the query optimizer had some questions about the contents of the `causal` and `causal_parents` tables, so this should allow it to get it.  It would be neat if sqlite would explain more about the optimizer's questions about these tables, but afaik it can't.

## Implementation notes

I created `Queries.shutdown :: DB m => m ()` and call it in place of `Sqlite.close` in various places.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/2012)

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
@@ -82,11 +82,12 @@ type EDB m = (DB m, Err m)
 
 type Err m = (MonadError Integrity m, HasCallStack)
 
-debugQuery, debugThread, debugConnection, debugFile :: Bool
+debugQuery, debugThread, debugConnection, debugFile, debugPragmaOptimize :: Bool
 debugQuery = False
 debugThread = False
 debugConnection = False
 debugFile = False
+debugPragmaOptimize = False
 
 alwaysTraceOnCrash :: Bool
 alwaysTraceOnCrash = True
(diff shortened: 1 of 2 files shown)


aryairani:
Weird failures in testing:

```
🦄  gitsync22.non-fastforward-push
gitsync22.non-fastforward-push FAILURE expected Left, got Right CallStack (from HasCallStack):
  crash, called at src/EasyTest.hs:83:24 in easytest-0.1-KCnEv2E8xRDLDbaYlI6kjn:EasyTest
  expectLeft, called at tests/Unison/Test/GitSync.hs:461:10 in main:Unison.Test.GitSync
```

and

```
gitsync22.sc.patching EXCEPTION!!!: SQLite3 returned ErrorIO while attempting to perform step: disk I/O error
```

on all CI images

aryairani:
Maybe we'll revisit in the future.

aryairani:
Also, I forgot to call `Sqlite.close` in `shutdown`

PR #2013, opened 2021-05-26
tag release whenever we merge to trunk

stew:
## Overview

Adds another GHA to upload and tag releases for macos and Linux

## Implementation notes

I am only building for macos-10, I was never able to get a working macos-11 build to work in my repo, no idea why.

This maintains two different tagged releases: latest-Linux and latest-macOS

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,8 +15,6 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
-          - ubuntu-18.04
-          # - ubuntu-16.04
           - macOS-11.0
           - macOS-10.15
     steps:
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -1,0 +1,68 @@
+name: "pre-release"
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    branches: [ trunk ]
+    types:
+      - completed
+jobs:
+  pre-release:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -15,8 +15,6 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
-          - ubuntu-18.04
```
pchiusano, line 18: Can you explain this one? Are Ubuntu 18 and 16 considered too old to support you think?

pchiusano (approved):
Left one comment. It'll be good if we can get those macOS 11 builds too, but we could still merge this and then figure that out later. CI is building macOS 11 so it seems like it ought to be possible.

So will this create and repeatedly update a "latest release" release? Like it won't keep old releases around indefinitely?

hojberg (approved):
🎉

.github/workflows/ci.yaml
stew: well, I just don't see the point since we are using the same toolchain for both. Do we ever see one fail and not the other for any meaningful reason? or are we just warming the earth with this one.

aryairani: It might be a libncurses version issue.  I guess we can fix it with package dependencies, but it some versions of ubuntu have libncurses6 and some have libncurses5 and when we link against one or the other and it's missing, that's when we get the libtinfo errors.  That said, I don't know that there's a difference between 18.x and 20.x for that.

PR #2017, opened 2021-05-26
Don't typecheck all scratch files on startup

pchiusano:
Before this PR, UCM would typecheck and run all `.u` files in the current directory on startup. This can be pretty spammy. Often you'll have more than one scratch file so you'd have to scroll to find what you want.

And even if you have just one scratch file, it's sort of bewildering to be shown all that on startup. What if you're about to switch to something totally different and want to do it in a new scratch file? "Oh, wait, what? Oh I guess that's the typechecking output of that random scratch file from last week? Gee thanks Unison."

You also miss the codebase ui url and friendly welcome message and ascii art. :)

Anyone have strong feelings to the contrary on this one?

Discussed in #2016

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Watch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Watch.hs
@@ -11,10 +11,6 @@ import           Control.Concurrent             ( forkIO
                                                 , threadDelay
                                                 , killThread
                                                 )
-import           UnliftIO.Directory             ( getModificationTime
-                                                , listDirectory
-                                                , doesPathExist
-                                                )
 import           UnliftIO.MVar                  ( newEmptyMVar, takeMVar
                                                 , tryTakeMVar, tryPutMVar, putMVar )
 import           UnliftIO.STM                   ( atomically )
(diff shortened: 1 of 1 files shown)


aryairani (approved):
👍

parser-typechecker/src/Unison/Codebase/Watch.hs
```diff
@@ -124,7 +114,7 @@ watchDirectory dir allow = do
       event@(file, _) <- watcher
       when (allow file) $
         STM.atomically $ TQueue.enqueue queue event
-  pending <- newIORef =<< existingFiles
+  pending <- newIORef []
```
pchiusano, line 117: It is also weird to me that this functionality was implemented here. I don't think a directory watcher should automatically trigger an event when you first start watching the directory - that feels like it belongs in another layer if that is the behavior being implemented. 

A directory watcher should just emit events when files are changed or touched anytime after the watching begins.

pchiusano:
I'll leave this open for a while just in case others want to weigh in.

PR #2030, opened 2021-05-27
Fix wrapper code for putBytes

### LINKED ISSUE #2026: "Cannot decompile Foreign" when running function with IO
The following code using IO prints out the error `Cannot decompile Foreign`

```unison
ex = unsafeRun '(printLine "hello world")

Exception.unsafeRun : '{Exception, g} a -> '{g} a
Exception.unsafeRun e _ = 
    h : Request {Exception} a -> a 
    h = cases 
        {Exception.raise fail -> _ } -> 
            bug fail 
        {a} -> a 
    handle !e with h 

```

In the UCM this is the command and error

```ucm
.base.io> run ex
hello world

  cannot decompile Foreign
```

Transcript below: 
[fix2026.md](https://github.com/unisonweb/unison/files/6549641/fix2026.md)

dolio:
The wrapper code used for `putBytes` was causing it to return the handle in successful cases instead of a unit value.

Should fix #2026. I wasn't able to pull base to check for certain, but I think I replicated the relevant definitions locally.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Builtin.hs
+++ b/parser-typechecker/src/Unison/Runtime/Builtin.hs
@@ -1410,7 +1410,7 @@ declareForeigns = do
 
   declareForeign "IO.getBytes.impl.v3" boxNatToEFBox .  mkForeignIOF $ \(h,n) -> Bytes.fromArray <$> hGet h n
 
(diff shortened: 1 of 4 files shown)


pchiusano:
@dolio great! can you just add the regression test (which has a full, self-contained example which doesn't require pulling base) to this PR?

It's linked in the original ticket https://github.com/unisonweb/unison/files/6549641/fix2026.md

pchiusano (approved):

PR #2033, opened 2021-05-27
Fix bad failure references in builtins

### LINKED ISSUE #2027: Ucm exits with "No builtin type called: io2.Failure" when calling new IO functions
When calling some of the new IO functions (I was experimenting with Sockets so the transcript contains that as an example) the following error will surface and the ucm will exit. 

```
No builtin type called: io2.Failure
  error, called at src/Unison/Runtime/IOSource.hs:63:16 in unison-parser-typechecker-0.0.0-6U6boimwb8GAC5qrhLfs8h:Unison.Runtime.IOSource
```

Attached is a transcript which should replicate the issue 
[fix2027.md](https://github.com/unisonweb/unison/files/6550160/fix2027.md)

dolio:
- When originally written, various builtins were referencing data types
  defined in IOSource. However, those no longer exist there.
- Instead, they are defined in Unison.Builtin.Decls
- This switches the builtins to the Decls values, some of which weren't
  exported yet.
- Also eliminates them from IOSource, since the ones there didn't
  actually work.
- Transcript testing the behavior added.

Fixes #2027, inasmuch as it now gives a proper failure condition.

### DIFF
--- a/parser-typechecker/src/Unison/Builtin/Decls.hs
+++ b/parser-typechecker/src/Unison/Builtin/Decls.hs
@@ -41,8 +41,11 @@ pairRef = lookupDeclRef "Tuple"
 optionalRef = lookupDeclRef "Optional"
 eitherRef = lookupDeclRef "Either"
 
-testResultRef, linkRef, docRef, ioErrorRef, stdHandleRef, failureRef, tlsSignedCertRef, tlsPrivateKeyRef :: Reference
+testResultRef, linkRef, docRef, ioErrorRef, stdHandleRef :: Reference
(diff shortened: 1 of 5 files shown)


pchiusano (approved):

dolio:
Hmm, the exact error message depends on OS, and that ends up in the transcript output.

pchiusano:
@dolio you can match on the failure and just strip out the message

```match f with Failure typ msg any -> Failure typ "oh noe!" any```

pchiusano:
Looks great, mergin'!

PR #2035, opened 2021-05-28
add `_edits` to `Branch0` hash calculation

aryairani:
Somehow we managed to get this far without branch hashes incorporating the patches / edits.

Changing this affects hash history and thus some tests that depend on hash values.  I deleted `Git.hs` which was a brittle test of v1 codebase git sync, because it isn't worth the effort to continue to update it with new hashes.


aryairani:
the `fix2000.md` transcript is failing because it looks up a particular hash that isn't there anymore.  I'm not sure what the intention was meant to be.
___
Checks that squash and merge do the same thing, with nontrivial history that
includes a merge conflict.

```unison
x.a.p = "af"
x.a.q = "ef"
```

```ucm

  change:
  
      x.a.p : Text
      x.a.q : Text

```
```ucm
.> add

    x.a.p : Text
    x.a.q : Text

.> fork x y

  Done.

.> fork x s

  Done.

.> fork x m

  Done.

.> delete y.a.p

  Name changes:
  
    Original    Changes
    1. m.a.p ┐  2. y.a.p (removed)
    3. s.a.p │  
    4. x.a.p │  
    5. y.a.p ┘  
  
```
```unison
y.a.p = "fij"
```

```ucm

  change:
  
      y.a.p : Text

```
```ucm
.> add

    y.a.p : Text

```
```unison
y.b.p = "wie"
```

```ucm

  change:
  
      y.b.p : Text

```
Merge back into the ancestor.

```ucm
.> add

    y.b.p : Text

.> merge y.b y.a

  Here's what's changed in y.a after the merge:
  
  New name conflicts:
  
    1. p#a3ef1630bu : Text
       ↓
    2. ┌ p#a3ef1630bu : Text
    3. └ p#fjqpdmdeqi : Text
  
  Tip: You can use `todo` to see if this generated any work to
       do in this namespace and `test` to run the tests. Or you
       can use `undo` or `reflog` to undo the results of this
       merge.

.> delete.term 1

  Resolved name conflicts:
  
    1. ┌ y.a.p#a3ef1630bu : Text
    2. └ y.a.p#fjqpdmdeqi : Text
       ↓
    3. y.a.p#fjqpdmdeqi : Text
  
.> history x.a

        message.
  
  ⊙ #4ehrt34sfm
  
    + Adds / updates:
    
      p q
  
  □ #sjg2v58vn2 (start of history)

.> history y.a

        message.
  
  ⊙ #5nb8t3khjo
  
    - Deletes:
    
      p
  
  This segment of history starts with a merge. Use
  `history #som3n4m3space` to view history starting from a given
  namespace hash.
  
  ⊙ #2v41qb0h90
  ⑃
  #0ilhqgrt9l
  #8c578gcant

.> history #0530pbcaqe

  😶
  
  I don't know of a namespace with that hash.

```

pchiusano (approved):
Wow good catch. I think the reason this one took so long to notice is you rarely just have a namespace with only a patch in it. And if you are using the default patch then an update will be updating that at the same time as another element of the Branch0.


I can take a look at updating that transcript, it's probably just too brittle.

aryairani:
I thought this would mean that `delete.term-replacement` / `delete.type-replacement` also wouldn’t work (with `consDistinct`), but I guess they do for some reason.

pchiusano:
Recording what we think was going on. 

The isEmpty check in updateChildren would delete the child if it was “empty”, where that was determined by looking at the hash.

```
updateChildren seg updatedChild =
  if isEmpty updatedChild
  then Map.delete seg
  else Map.insert seg updatedChild
```

Before the change
Screen Shot 2021-05-28 at 8 58 59 AM

when we went to add the patch to the namespace, it would create a causal of length 2 because it was using `stepDistinct`, and the `==` instance for `Branch0` was correct. Causals of length 2 have a different hash than a causal of length 1 even if the contents of the causal don’t change. So this would avoid the `isEmpty` check in `updateChildren` - the length 2 causal would be different than the empty branch.

After the change it would create a causal of length 1 upon adding just the patch. And now we were solely dependent on the `isEmpty` check being correct - and it wasn’t because it was relying on the incorrect hash which was the same as the empty branch. Thus this exposed the bug.

> I thought this would mean that delete.term-replacement / delete.type-replacement also wouldn’t work (with consDistinct), but I guess they do for some reason.

By the same reasoning, `consDistinct`, which is based on `==`, will still add a new element to the causal for these operations.

pchiusano:
Transcripts needed a refresh. Any transcript that shows namespace hashes would be affected by this, even if it does nothing with patches, since the namespace hash now correctly mixes in the hash of the patches. (It doesn't matter if the patches are empty, just mixing in another hash for the empty list will affect the overall hash)

PR #2036, opened 2021-05-28
Update mergify.yml to work with new build matrix

pchiusano:
CI isn't building Ubuntu 18 anymore, so Mergify shouldn't check for it.

I've already applied ready-to-merge to this PR so I think approving it will cause Mergify to merge it's own update to the Mergify config. 🤯

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,7 +2,7 @@ pull_request_rules:
   - name: automatic merge on CI success and review
     conditions:
       - status-success=ubuntu-20.04
-      - status-success=ubuntu-18.04
+      # - status-success=ubuntu-18.04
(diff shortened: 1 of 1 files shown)


nmcb (approved):
green is a go

pchiusano:
@nmcb Thanks. :) It looks like it's gotta be a committer though, which makes sense.

hojberg (approved):

pchiusano:
image

Okay looks like I gotta merge this one manually, but then after this it should work?

aryairani:
> I've already applied ready-to-merge to this PR so I think approving it will cause Mergify to merge it's own update to the Mergify config. 🤯

I wonder why it didn't.

PR #2037, opened 2021-05-28
Codebase server default ui path is relative to the UCM executable now

### LINKED ISSUE #1979: Improve running UI for developers of ucm
Running the UI from with `ucm` can be a little annoying as a developer of `ucm` if you're not also a developer of the UI as it involves running the UI dev server etc.

One possible solution could be to have `ucm` provide an error if there are no files in the ui folder it was looking at (either default location, set through args or through env vars) and tell the user that they can either start the UI dev server (with link to the ui repository instructions) or give the user a prompt to download the UI into a `ui` folder in the root of the `unison` project (where `ucm` by default looks for the UI), if the user then accepts this, `ucm` will download into that folder the `latest` tag from the Codebase UI repository and extract the `ucm.zip` into the `ui` folder and then on success provide the user with the URL to start the UI.

pchiusano:
Previously it was relative to the current directory, which was not very useful.

To help with my manual testing, I also added a script, `dev-ui-install.sh` which fetches the latest release of the codebase UI and puts it in the correct spot relative to the executable. After running this script, the URL that UCM prints on startup can be visited in the browser and it works. This script can be run as often as you like whenever new goodies land in the codebase UI and should streamline the dev workflow.

Both the `/ui` path and the executable path get symlinks resolved now as well. The thinking here is that you can now have a low tech install method - just unzip the release anywhere, and add a symlink to the executable in `/usr/local/bin` or wherever.

This just got manual testing.

### DIFF
--- a/development.markdown
+++ b/development.markdown
@@ -8,9 +8,13 @@ _Disclaimer_ If you have trouble getting started, please get in touch via [Slack
 
 To get cracking with Unison:
 
-* [Install `stack`](https://docs.haskellstack.org/en/stable/README/#how-to-install).
-* Build the project with `stack build`. This builds all executables.
-* After building, `stack exec unison` will fire up the codebase editor, create a codebase in the current directory, and watch for `.u` file changes.  If you want to run it in a different directory, just add `unison` to your `PATH`, after finding it with `find .stack-work -name unison -type f`.  (For me, this finds two, they both work, but have different contents.  ¯\\\_(ツ)\_/¯ )
+1. [Install `stack`](https://docs.haskellstack.org/en/stable/README/#how-to-install).
+2. Build the project with `stack build`. This builds all executables.
+3. (Optional) Run `./dev-ui-install.hs` to fetch the latest release of the codebase UI. If you don't care about running the codebase UI locally you can ignore this step.
+4. After building do `stack exec unison -- init` will initialize a codebase in your home directory (in `~/.unison`). This only needs to be done once.
+5. `stack exec unison` starts Unison and watches for `.u` file changes in the current directory. If you want to run it in a different directory, just add `unison` to your `PATH`, after finding it with `stack exec which unison`.
+
+On startup, Unison prints a url for the codebase UI. If you did step 3 above, then visiting that URL in a browser will give you a nice interface to your codebase.
 
 ## Running Tests
 
--- a/parser-typechecker/src/Unison/Server/CodebaseServer.hs
+++ b/parser-typechecker/src/Unison/Server/CodebaseServer.hs
@@ -87,9 +87,10 @@ import Servant.Server
     err404,
   )
 import Servant.Server.StaticFiles (serveDirectoryWebApp)
-import System.Directory (doesFileExist)
-import System.Environment (getArgs, lookupEnv)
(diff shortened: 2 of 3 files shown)


parser-typechecker/src/Unison/Server/CodebaseServer.hs
```diff
@@ -87,9 +87,10 @@ import Servant.Server
     err404,
   )
 import Servant.Server.StaticFiles (serveDirectoryWebApp)
-import System.Directory (doesFileExist)
-import System.Environment (getArgs, lookupEnv)
-import System.FilePath.Posix ((</>))
+import System.Directory (doesFileExist, canonicalizePath)
+import System.Environment (getArgs, lookupEnv, getExecutablePath)
+import System.FilePath ((</>))
```
pchiusano, line 92: @runarorama fyi, my understanding from reading [the package description](http://hackage.haskell.org/package/filepath-1.4.2.1) is that `System.FilePath` module directs to the correct version (Windows vs Posix depending on your environment) while explicitly importing `System.FilePath.Posix` assumes Posix file separators.

pchiusano:
Hmm the symlinks aren't getting resolved like I'd expect. I just tried:

```
ln $(stack exec which unison) /Users/pchiusano/.local/bin/unison
```

And then `unison` launches UCM but it can't find the UI.

EDIT, my bad, I forgot the `-s` flag to `ln`, so it was creating a hard link, which `canonicalizePath` doesn't do anything with. It works as expected with something like:

```
ln -s $(stack exec which unison) /Users/pchiusano/.local/bin/unison
```

hojberg:
Related to https://github.com/unisonweb/unison/issues/1979

aryairani:
@hojberg Would you say that it closes #1979, or what would be left?

development.markdown
```diff
@@ -8,9 +8,13 @@ _Disclaimer_ If you have trouble getting started, please get in touch via [Slack
 
 To get cracking with Unison:
 
-* [Install `stack`](https://docs.haskellstack.org/en/stable/README/#how-to-install).
-* Build the project with `stack build`. This builds all executables.
-* After building, `stack exec unison` will fire up the codebase editor, create a codebase in the current directory, and watch for `.u` file changes.  If you want to run it in a different directory, just add `unison` to your `PATH`, after finding it with `find .stack-work -name unison -type f`.  (For me, this finds two, they both work, but have different contents.  ¯\\\_(ツ)\_/¯ )
+1. [Install `stack`](https://docs.haskellstack.org/en/stable/README/#how-to-install).
+2. Build the project with `stack build`. This builds all executables.
+3. (Optional) Run `./dev-ui-install.hs` to fetch the latest release of the codebase UI. If you don't care about running the codebase UI locally you can ignore this step.
+4. After building do `stack exec unison -- init` will initialize a codebase in your home directory (in `~/.unison`). This only needs to be done once.
```
aryairani, line 14: When implementing #2040, we can remove the line about `init`.

aryairani (approved):

pchiusano:
Will leave this open for a day in case anyone else has comments.

hojberg:
@aryairani I think it'll be cool with a more integrated experience long term, but this seems fine for now.

Issue #2038, opened 2021-05-28
let `history` command show the contents of non-empty initial namespace


aryairani:
Transcript:
````
```unison:hide
x = 1
```

```ucm
.> add
```

```unison:hide
y = 2
```

```ucm
.> add
.> history
```
````

Output:
```unison
x = 1
```

```ucm
.> add

  ⍟ I've added these definitions:
  
    x : ##Nat

```
```unison
y = 2
```

```ucm
.> add

  ⍟ I've added these definitions:
  
    y : ##Nat

.> history

  Note: The most recent namespace hash is immediately below this
        message.
  
  ⊙ #is7tu6katt
  
    + Adds / updates:
    
      y
  
  □ #hkrqt3tm05 (start of history)

```

I'm thinking the `history` should say something about `x`; perhaps just a diff against `Branch.empty0` after `#hkrqt3`.

Issue #2031, opened 2021-05-27
Consider switching representation of `Causal`


pchiusano:
Suggested by @runarorama, proposal is switch to `Causal` having `Empty` and `Cons` rather than `One` and `Cons`:

```
data Causal hash a 
  = Empty 
  | Cons hash a (Causal hash a) 
  | Merge ...
```

We suspect this will clean up a number of things, for instance the LCA calculation: `Empty` doesn't count as an LCA.

Right now we have `One empty` for representing the empty namespace with no history, but then this can get chosen as an LCA, which is not great.

On the sql side, the only change seems like it would be making one of the columns nullable.

There are probably subtle implications with this change that aren't obvious right now, and those should be worked out.

PR #2039, opened 2021-05-28
replace "`One` isn't an LCA" algorithm with "`empty` isn't real history"

### LINKED ISSUE #2004: Update SQLiteCodebase implementation of LCA
The empty namespace can technically be the LCA due to how we represent history (all namespaces start out with an empty namespace at the start of their history), but it's basically never the LCA you want. Here's an example:

```
j1: ... - v1 - v2 - v3 - v4 - v4a - v5 - v6 - v7
                                              /
                                  <empty> - v5a

j2: ... - v1 - v2 - v3 - v4 - v4a - x1 - x2 - x3 
                             /    /     / 
                            /  <empty> <empty> 
                        big shallow wide tree
```

This could find `<empty>` as an LCA even though we want v4 as the LCA.

In this PR we avoid returning a Causal.One (a causal with no parents) as the LCA and keep looking.

https://github.com/unisonweb/unison/pull/2002

The sql implementation of LCA needs updating accordingly. I think it's as simple as restricting the query to namespaces with at least one parent.

aryairani:
## Overview

This PR reverts the LCA special case that was added in #2002, and instead says that
```
Branch.cons a Branch.empty = Branch.one a
```

It incorporates and closes #2015, and closes #2004.

## Implementation notes

Basically:
```haskell
Branch.step :: Applicative m => (Branch0 m -> Branch0 m) -> Branch m -> Branch m
Branch.step f = \case
  Branch (Causal.One _h e) | e == empty0 -> Branch (Causal.one (f empty0))
  b -> over history (Causal.stepDistinct f) b
```
with an equivalent change to `Branch.stepM`.

Probably the revert of #2002 should have been its own PR, and the initial change to LCA should have been its own PR that we could one-click revert.

## Test coverage

It adds [fix2004.md](https://github.com/unisonweb/unison/blob/60cf4c1f0010a7c13faf07679471c3da568872e4/unison-src/transcripts/fix2004.output.md) to verify that `<empty>` isn't selected as an LCA in common cases.

## Loose ends

* Implementing #2031 will probably involve undoing these changes to `step` and `stepM`.
* The `history` command currently assumes that the start of history doesn't contain anything worth displaying.  I opened #2038 to track this. Although this PR isn't the first to create counterexamples (builtins.merge will too), this means that a number of transcripts lose some of their output.


pchiusano (approved):
LGTM.

PR #2048, opened 2021-06-01
Simple install scripts

pchiusano:
This PR just adds two scripts, `linux-install.sh` and `mac-install.sh` which literally just downloads the release and unzips it to a directory, so it's almost the kind of thing you could just put in a box with a copy to clipboard button on the website. I hesitated to even do a PR for it, but I was thinking it's a good idea to have simple manual install instructions.

image

Here's the steps:

```
mkdir -p unisonlanguage 
curl -# -L https://github.com/unisonweb/unison/releases/download/latest-Linux/unison-Linux.tar.gz --output unisonlanguage/ucm.tar.gz 
tar -xzf unisonlanguage/ucm.tar.gz -C unisonlanguage
```

I didn't try to have a single script that detects the OS or which lets you pick which release you want to install. Maybe for future work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/2048)

### DIFF
--- a/linux-install.sh
+++ b/linux-install.sh
@@ -1,0 +1,12 @@
+mkdir -p unisonlanguage 
+curl -# -L https://github.com/unisonweb/unison/releases/download/latest-Linux/unison-Linux.tar.gz --output unisonlanguage/ucm.tar.gz 
+tar -xzf unisonlanguage/ucm.tar.gz -C unisonlanguage
+
+echo ""
+echo "All done!"
+echo ""
+echo "\x1b[1mRun ./unisonlanguage/ucm to start Unison.\x1b[0m" 
+echo ""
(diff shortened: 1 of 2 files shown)


aryairani (approved):
What is the goal with these?

- to document how one might grab the latest unison, e.g. from a website (which wouldn't include the `echo`s but it's fine to have them here)
- to give developers a quick way to download the latest binary (though they already have the source)
- as a substitute for package managers?

Do we want to test them in CI? :)  Maybe download ucm and pipe a transcript into it and see if it succeeds?

aryairani:
I think we should probably just get the packages going, or actually make this a one-liner on the website.

pchiusano:
I kinda like idea of integrating this into CI to make sure the one liner on the website actually works. I'll have a go at that. Going to convert back to a draft.

PR #2054, opened 2021-06-02
Add cache-control header to API Server

runarorama:
This adds HTTP headers to the API server as follows:

image


stew:
This seems like a huge diff for what I'm looking for but I love it nonetheless and I want it so bad thank you so much for doing it real fast

stew (approved):

PR #2064, opened 2021-06-02
Flush the WAL files on `push`

### LINKED ISSUE #2063: Flush write-ahead log before push and verify that no WAL files are committed
According to https://sqlite.org/tempfiles.html the WAL files aren’t necessarily flushed on close if there are other open connections to the database. Also see https://github.com/groue/GRDB.swift/issues/418

We are probably not closing the connection on `pull` and that should be tracked down, but we should make the `push` code robust to concurrent connections being open.

Also, we should ensure that `push` can’t commit WAL files or really any files other than the database files. 

```
$ git add v2/sqlite3.db
$ git commit -m blah
$ git status --porcelain # make sure this is empty
```
WIP in fix/wal-flush branch

Arya suggested enabling the WAL on `push` as well.

pchiusano:
This PR addresses #2063, ensuring that no stray WAL files get added to a commit, and (hopefully) forcing these files to get flushed before `push` (folded into the `unison.sqlite3` file then deleted). @aryairani spoon fed me the fix for this.

However, since we didn't have a reliable reproduction of the issue that resulted in those WAL files making it into a commit, it's unclear if the fix in this PR has actually worked. This PR should be understood as "here's a plausible fix, and a reliable failure detection if the issue is still present and ever arises again".

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
@@ -1100,16 +1100,16 @@ pushGitRootBranch srcConn branch repo = runExceptT @GitError do
       else Nothing
       where
         statusLines = Text.unpack <$> Text.lines status
-        okLine ('?':'?':(trimStart -> ".unison/v2/unison.sqlite3")) = True
-        okLine ('M':(trimStart -> ".unison/v2/unison.sqlite3")) = True
+        t = dropWhile Char.isSpace
+        okLine (t -> '?':'?':(t -> ".unison/v2/unison.sqlite3")) = True
+        okLine (t -> 'M':(t -> ".unison/v2/unison.sqlite3")) = True
         okLine line = isWalDelete line || isShmDelete line
-        isWalDelete ('D':(trimStart -> ".unison/v2/unison.sqlite3-wal")) = True
+        isWalDelete (t -> 'D':(t -> ".unison/v2/unison.sqlite3-wal")) = True
         isWalDelete _ = False
-        isShmDelete ('D':(trimStart -> ".unison/v2/unison.sqlite3-shm")) = True
+        isShmDelete (t -> 'D':(t -> ".unison/v2/unison.sqlite3-shm")) = True
         isShmDelete _ = False
         hasDeleteWal = any isWalDelete statusLines
         hasDeleteShm = any isShmDelete statusLines
-        trimStart = dropWhile Char.isSpace
 
     -- Commit our changes
     push :: CodebasePath -> RemoteRepo -> IO Bool -- withIOError needs IO
--- a/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
@@ -1115,6 +1115,10 @@ pushGitRootBranch srcConn branch repo = runExceptT @GitError do
     push :: CodebasePath -> RemoteRepo -> IO Bool -- withIOError needs IO
     push remotePath (GitRepo url gitbranch) = time "SqliteCodebase.pushGitRootBranch.push" $ do
       -- has anything changed?
+      -- note: -uall recursively shows status for all files in untracked directories
+      --   we want this so that we see
+      --     `??  .unison/v2/unison.sqlite3` and not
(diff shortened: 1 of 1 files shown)


parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -1069,27 +1073,76 @@ pushGitRootBranch srcConn branch repo = runExceptT @GitError do
...
+    --
+    -- Which will only happen if the write-ahead log hasn't been
+    -- fully folded into the unison.sqlite3 file.
+    --
+    -- Returns `Just (hasDeleteWal, hasDeleteShm)` on success,
+    -- `Nothing` otherwise. hasDeleteWal means there's the line:
+    --   D .unison/v2/unison.sqlite3-wal
+    -- and hasDeleteShm is `True` if there's the line:
+    --   D .unison/v2/unison.sqlite3-shm
+    --
```
pchiusano, line 1096: I made this more complicated than we originally discussed, to gracefully handle the case where a push cleans up / deletes old WAL files. The implementation of this is kinda fragile if the format of `git status --short` changes but I think it's okay.

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -1069,27 +1073,80 @@ pushGitRootBranch srcConn branch repo = runExceptT @GitError do
...
+                 <> "I was expecting only to see .unison/v2/unison.sqlite3 modified, but saw:\n\n"
+                 <> Text.unpack status <> "\n\n"
+                 <> "Please visit https://github.com/unisonweb/unison/issues/2063\n"
+                 <> "and add any more details about how you encountered this!\n"
+          Just (hasDeleteWal, hasDeleteShm) -> do
+            -- Only stage files we're expecting; don't `git add --all .`
+            -- which could accidentally commit some garbage
+            gitIn remotePath ["add", ".unison/v2/unison.sqlite3"]
+            when hasDeleteWal $ gitIn remotePath ["rm", ".unison/v2/unison.sqlite3-wal"]
+            when hasDeleteShm $ gitIn remotePath ["rm", ".unison/v2/unison.sqlite3-shm"]
```
pchiusano, line 1137: This is the only relevant diff. The rest is just needing to introduce an extra level of indentation for the pattern match. The logic in the `Just` case is otherwise the same as before.

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
pchiusano: Also somewhat reassuring is I had this function wrong in like every possible way (forgot to allow for the `unison.sqlite3` file to be created, forgot to allow the wal file to be deleted, forgot the `-uall` for pushing to an empty repo...) and got test failures or failures in some manual testing.

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -1069,27 +1073,80 @@ pushGitRootBranch srcConn branch repo = runExceptT @GitError do
       chId <- fromMaybe err <$> Q.loadCausalHashIdByCausalHash h2
       Q.setNamespaceRoot chId
 
+    -- This function makes sure that the result of git status is valid.
+    -- Valid lines are any of:
+    --
+    --   ?? .unison/v2/unison.sqlite3 (initial commit to an empty repo)
```
aryairani, line 1079: I think I had seen that there can be a space before the `M`; have you observed this? And will it still work if there is?

aryairani (approved):
I haven’t scrutinized every line but I left one question and it otherwise LGTM.

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
pchiusano: Yes, you're right, and no, it won't work. :| I have also observed this; don't totally get the logic it's using to pick spacing.

I've pushed a fix to trim whitespace at the start of the line.

aryairani: I choose to believe `t` is short for “thank goodness we’re going to get the release out the door”

PR #2069, opened 2021-06-03
Fix issue when syncing with >=2 namespaces only existing in memory

pchiusano:
Fixes a bug where sync would terminate prematurely after processing the first in-memory-only branch.


aryairani (approved):
LGTM

PR #2074, opened 2021-06-04
fix some benign mistakes in sync

aryairani:
See self-review.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
+++ b/parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
@@ -837,19 +837,19 @@ syncInternal progress srcConn destConn b = time "syncInternal" do
                       traceM $ "  decls: " ++ show ds
                       traceM $ "  edits: " ++ show es
                     (cs, es, ts, ds) <- lift $ runDB destConn do
-                      cs <- filterM (fmap not . runDB destConn . isCausalHash' . fst) branchDeps
-                      es <- filterM (fmap not . runDB destConn . patchExists') es
-                      ts <- filterM (fmap not . runDB destConn . termExists') ts
-                      ds <- filterM (fmap not . runDB destConn . declExists') ds
+                      cs <- filterM (fmap not . isCausalHash' . fst) branchDeps
+                      es <- filterM (fmap not . patchExists') es
+                      ts <- filterM (fmap not . termExists') ts
+                      ds <- filterM (fmap not . declExists') ds
                       pure (cs, es, ts, ds)
                     if null cs && null es && null ts && null ds
                       then do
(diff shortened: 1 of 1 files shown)


parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -837,19 +837,19 @@ syncInternal progress srcConn destConn b = time "syncInternal" do
...
                       traceM $ "  edits: " ++ show es
                     (cs, es, ts, ds) <- lift $ runDB destConn do
-                      cs <- filterM (fmap not . runDB destConn . isCausalHash' . fst) branchDeps
-                      es <- filterM (fmap not . runDB destConn . patchExists') es
-                      ts <- filterM (fmap not . runDB destConn . termExists') ts
-                      ds <- filterM (fmap not . runDB destConn . declExists') ds
+                      cs <- filterM (fmap not . isCausalHash' . fst) branchDeps
+                      es <- filterM (fmap not . patchExists') es
+                      ts <- filterM (fmap not . termExists') ts
+                      ds <- filterM (fmap not . declExists') ds
```
aryairani, line 843: These extra `runDB destConn`s are not needed given the outer one (which I suppose is ignored? one of them is ignored.)

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -837,19 +837,19 @@ syncInternal progress srcConn destConn b = time "syncInternal" do
...
                       pure (cs, es, ts, ds)
                     if null cs && null es && null ts && null ds
                       then do
                         lift . runDB destConn $ putBranch' b
                         processBranches @m sync progress rest
                       else do
-                        let bs = map (uncurry B) branchDeps
+                        let bs = map (uncurry B) cs
```
aryairani, line 850: Add the filtered set to the work queue, rather than the original.  I thought this might be responsible for some as-of-yet not observed infinite loops, and tried to create a failing regression test, but I think it is actually benign, in combination with the `null cs` check above it.

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -837,19 +837,19 @@ syncInternal progress srcConn destConn b = time "syncInternal" do
...
                             os = map O (es <> ts <> ds)
-                        processBranches @m sync progress (os ++ bs ++ B h mb : rest)
+                        processBranches @m sync progress (os ++ bs ++ b0 : rest)
```
aryairani, line 852: The parameters here aren't arbitrary, we're just postponing the current work item. (`b0` is bound during pattern matching.)

pchiusano (approved):

PR #2075, opened 2021-06-04
use `codebasePath` constant in SC.parseStatus instead of strings

aryairani:
I missed this during the first review. This should also help with Windows builds one day, too.

The autoformatter changed the whitespace, but it didn't seem important enough to change back.


pchiusano (approved):
Good idea

PR #2076, opened 2021-06-04
Series/m2

stew:
Merge the build updates to main

## Overview

We are trying to update our M2 release to contain images built for older ncurses5 (for ubuntu 18.04 users, for example). This update would instead of introducing yet another separate release for M2 which contains just a single tarball, this would create a single `release/M2` release containing all 3 tarballs (1 Mac, 1 Linux-ncurses5 and 1 Linux-ncurses6)

I'm requesting this merge to master, because the way these actions are setup, we have one action triggering another, and the subsequently triggered actions have to be on the default branch for that to work

## Implementation notes

This new method of creating releases first creates a draft release from the tag name, the job triggers the matrix builds which append to the draft release with their tarballs.

Then we would manually inspect the release, fix up the release notes and take away the `Draft:` flag

We could perhaps automate the taking away of the Draft flag. But I think we want to make sure the release notes look good. One way we could more reliably accomplish that would be to check in the current release notes in the commit we tag (perhaps in `docs/release_notes.md` or something, and then the job which creates the release can be sure to publish the right data

### DIFF
--- a/.github/workflows/upload-release-artifacts.yaml
+++ b/.github/workflows/upload-release-artifacts.yaml
@@ -1,9 +1,11 @@
-name: "release"
+name: "upload-release-artifacts"
 
 on:
-  push:
-    tags:
-      - "release/*"
+  workflow_run:
+    workflows: ["create-release"]
+    types:
+      - completed
+
 jobs:
   release:
     name: ${{ matrix.os }}
@@ -64,8 +66,11 @@ jobs:
         if: runner.os != 'macOS'
         run: echo "RELEASE_VERSION=release/${GITHUB_REF#refs/tags/release/*}-${{matrix.os}}" >> $GITHUB_ENV
 
-      - uses: "marvinpinto/action-automatic-releases@latest"
+      - name: "Upload ${{matrix.os}}"
+        uses: "actions/upload-artifact@v2"
         with:
+          path: "unison-${{runner.os}}.tar.gz"
+
           repo_token: "${{ secrets.GITHUB_TOKEN }}"
           automatic_release_tag: "${{ env.RELEASE_VERSION }}"
           prerelease: false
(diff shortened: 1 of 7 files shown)


pchiusano (approved):

.github/workflows/upload-release-artifacts.yaml
```diff
@@ -0,0 +1,74 @@
...
+      # One of the transcripts fails if the user's git name hasn't been set.
+      - name: set git user info
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "actions@github.com"
+
+      - name: remove ~/.stack/setup-exe-cache on macOS
+        if: runner.os == 'macOS'
+        run: rm -rf ~/.stack/setup-exe-cache
+
```
aryairani, line 48: If the runner isn't loading `~/.stack` from a cache, then this step can be skipped.

unison-core/package.yaml
```diff
@@ -46,4 +46,4 @@ flags:
 
 when:
   - condition: flag(optimized)
-    ghc-options: -funbox-strict-fields
+    ghc-options: -O2 -funbox-strict-fields
```
aryairani, line 49: I don't think `optimized` is actually being flagged for this package.

unison-core/unison-core1.cabal
```diff
@@ -98,5 +98,5 @@ library
     , util
     , vector
   if flag(optimized)
-    ghc-options: -funbox-strict-fields
+    ghc-options: -O2 -funbox-strict-fields
```
aryairani, line 101: Same here.

.github/workflows/upload-release-artifacts.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - name: build
+        run: stack --no-terminal build --flag unison-parser-typechecker:optimized
```
aryairani, line 50: I think the above is missing `-O2` / `optimized` on some of the other modules.  Should we use this the global build setting again?
```
run: stack --no-terminal build --ghc-options -O2
```
And maybe `-funbox-strict-fields` too; I don't know much about that one.

Issue #2068, opened 2021-06-03
Pushing to subpaths of length >= 1 crashes in SqliteCodebase


pchiusano:
As a simple example:

```
.> push https://github.com/pchiusano/myunisoncode:.some.subnamespace
```

As a workaround, you can push at the root or a path of length 1. For instance, instead of:

```
.mybase.List> push https://unisonweb/base:.trunk.List
```

You can do:

```
.mybase> push https://unisonweb/base:.trunk
```

PR #2077, opened 2021-06-04
wrapper type for DB connection provenance

aryairani:
`U.Codebase.Sqlite.Connection` replaces `Database.Sqlite.Simple.Connection` and adds a custom string identifier and the path to the database file, which are printed back to the console during debugging.

API functions that create codebases and database connections now allow/force you to provide a description for debugging.  API functions that create a codebase also allow/force you to provide a name that the codebase implementation may use for debug output.  (The V1 codebase implementation currently ignores the name.)

I used this to confirm that writes were going to the right databases when debugging #2068, though since that was _The Last Bug_, I'm not sure this is necessarily useful.

The PR also adds a centralized spot from which to close a v2 codebase db connections, which probably _is_ useful as I've implemented it twice now. (`Unison.Codebase.SqliteCodebase.shutdownConnection`).

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
@@ -23,6 +23,7 @@ import Control.Monad.Except (MonadError)
 import qualified Control.Monad.Except as Except
 import Control.Monad.IO.Class (MonadIO)
 import Control.Monad.Reader (MonadReader (ask))
+import qualified Control.Monad.Reader as Reader
 import Control.Monad.Trans (MonadIO (liftIO))
 import Control.Monad.Trans.Maybe (MaybeT (MaybeT), runMaybeT)
 import qualified Control.Monad.Writer as Writer
@@ -40,8 +41,7 @@ import Data.String (fromString)
 import Data.String.Here.Uninterpolated (here, hereFile)
 import Data.Text (Text)
 import Database.SQLite.Simple
-  ( Connection,
-    FromRow,
+  ( FromRow,
     Only (..),
     ToRow (..),
     (:.) (..),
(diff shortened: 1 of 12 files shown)


codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -85,11 +86,10 @@ type EDB m = (DB m, Err m)
 
 type Err m = (MonadError Integrity m, HasCallStack)
 
-debugQuery, debugThread, debugConnection, debugFile :: Bool
+debugQuery, debugThread, debugConnection :: Bool
 debugQuery = False
 debugThread = False
 debugConnection = False
-debugFile = False
```
aryairani, line 92: moves to Connection.hs

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
```diff
@@ -94,8 +95,8 @@ import UnliftIO.STM (atomically)
 
 init :: (MonadIO m, MonadCatch m) => Codebase.Init m Symbol Ann
 init = Codebase.Init
-  ((fmap . fmap) (pure (),) . openCodebase)
-  ((fmap . fmap) (pure (),) . createCodebase)
+  (const $ (fmap . fmap) (pure (),) . openCodebase)
+  (const $ (fmap . fmap) (pure (),) . createCodebase)
```
aryairani, line 99: Whee

parser-typechecker/src/Unison/Codebase/SqliteCodebase.hs
```diff
@@ -245,17 +252,23 @@ type TermBufferEntry = BufferEntry (Term Symbol Ann, Type Symbol Ann)
...
-  conn <- unsafeGetConnection root
+shutdownConnection :: MonadIO m => Connection -> m ()
+shutdownConnection conn = do
+  Monad.when debug $ traceM $ "shutdown connection " ++ show conn
+  liftIO $ Sqlite.close (Connection.underlying conn)
+
+sqliteCodebase :: MonadIO m => Codebase.DebugName -> CodebasePath -> m (Either SchemaVersion (m (), Codebase m Symbol Ann))
+sqliteCodebase debugName root = do
+  Monad.when debug $ traceM $ "sqliteCodebase " ++ debugName ++ " " ++ root
+  conn <- unsafeGetConnection debugName root
```
aryairani, line 271: Diff could try harder to match up definition names.

parser-typechecker/src/Unison/Codebase/FileCodebase.hs
pchiusano: Need moar fmap

pchiusano (approved):
LGTM

PR #2078, opened 2021-06-05
Don't print output for queries that don't have output

aryairani:
Reduces useless debug output.

Controversial: the substring match is pretty hacky, but it's okay.

Loose ends: change the merge target to `trunk` before merging, but after #2077 is merged.


pchiusano (approved):

aryairani:
EFF55974-CA26-4202-97E7-07ACC8ACC206

Magic! Maybe it has to do with deleting the branch on merge.

PR #2079, opened 2021-06-05
Disable pager inside emacs

jmorag:
## Overview

Disables paging when running ucm inside emacs to prevent repeated `WARNING: terminal is not fully functional`. This is in anticipation of completing the in progress ucm emacs integration I'm working on: 

Before
image
After
image

## Implementation notes

Modifies the `less` function to check the `INSIDE_EMACS` environment variable, and if so, output the string directly to STDOUT without going through the `less` pager. Behavior in a normal terminal is unchanged.

This could also be done by checking if the `TERM` environment variable is set to `dumb`, or using [hSupportsAnsi](https://hackage.haskell.org/package/ansi-terminal-0.11/docs/System-Console-ANSI.html#v:hSupportsANSI) to inspect the STDOUT handle. I'm not sure if those would cover any additional environments outside emacs, but I'm happy to change to one of those if you feel they'd be more idiomatic or robust.

## Test coverage

No additional tests, I'm not sure they'd be relevant for this change.

## Loose ends

It might be nice to disable ANSI colors as well for unsupported terminals. That's not necessary on emacs, though, and would probably require a bunch more invasive changes to the Pretty module, whereas this is a very minimal diff.

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -51,3 +51,4 @@ The format for this list: name, GitHub handle, and then optional blurb about wha
 * Sam Roberts (@samgqroberts)
 * Nigel Farrelly (@nini-faroux)
 * Johannes Huster (@JohannesHuster)
+* Joseph Morag (@jmorag)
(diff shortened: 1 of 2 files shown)


pchiusano (approved):
Lgtm

PR #2081, opened 2021-06-06
add parens around termLink and typeLink, according to ambient precedence

### LINKED ISSUE #2050: typeLink and termLink keywords should be treated as if they have same precedence as function application
`foo (typeLink Blah) 99` is rendered without the parens, which looks really weird even though it’s technically permitted by the parser. Similarly for `termLink`.

@atacratic any chance you have time to take a look at this one?

atacratic:
## Overview

Change the pretty-printer so that it produces, for example:

```haskell
  t : ()
  t =
    t1 = termLink foo
    t2 = id (termLink foo)
    t3 = termLink (+++)
    ty1 = typeLink List
    ty2 = id (typeLink List)
    ()
```
whereas previously it would have produced
```haskell
  t : ()
  t =
    t1 = termLink foo
    t2 = id termLink foo
    t3 = (termLink +++)
    ty1 = typeLink List
    ty2 = id typeLink List
    ()
```
fixes #2050 

## Implementation notes

Tweak to termPrinter.

None.

## Test coverage

Alas, none.  The termPrinter test harness can't furnish me with actual terms to get links to.  I have checked the example above as a one-off though. 

## Loose ends

None.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -122,6 +122,8 @@ data DocLiteralContext
 
      >=10
        10f 10x 10y ...
+       termLink t
+       typeLink t
 
      >=3
        x -> 2y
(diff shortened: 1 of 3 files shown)


pchiusano (approved):

pchiusano:
Thanks!

PR #2086, opened 2021-06-07
pass -f-terminfo to haskeline

### LINKED ISSUE #2085: Pass terminfo = disabled to haskeline when building
We use [haskeline](https://hackage.haskell.org/package/haskeline) for tab completion and line editing in UCM. It depends on terminfo at runtime by default, but can be passed a flag to disable that.

See https://hackage.haskell.org/package/haskeline

We should pass that flag to disable that dependency to make our linux binaries more portable (currently people get runtime errors if they've got a different version of libtinfo). According to the note here https://github.com/judah/haskeline:

> Runs on POSIX-compatible systems, using the terminfo library to support non-ANSI terminals.

But we expect an ANSI terminal anyway since we’re using ANSI codes for formatting. Also, realistically, we don’t care about supporting non ANSI terminals.

I’m not sure what the incantation is to pass a build flag to a dependency but this issue is just to do that.

aryairani:
Closes #2085

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,6 @@
-flags: {}
+flags:
+  haskeline: 
+    terminfo: false
 
 allow-newer: true # async package has needlessly strict upper bound
 


pchiusano (approved):
Great!

PR #2092, opened 2021-06-08
Delete questions.md

pchiusano:
It looks like this was just a file @aryairani was using to keep track of work on the codebase refactoring, but I don't think it belongs in the main repo. It's still there in the history and can be found via this PR if there's anything in there that's still relevant.


aryairani (approved):
Whoops.

PR #2093, opened 2021-06-08
Allow identifiers to start with `type` or `ability`

### LINKED ISSUE #2091: Any variable starting with the substring "type" is interpreted as not valid
[fix2091.md](https://github.com/unisonweb/unison/files/6611954/fix2091.md)
There are plenty of workarounds for this, (just don't start your variable with the word "type") but if you give a variable a name like "typesAndTermsDocumentation", Unison does not interpret your variable as valid. If you prefix your term like "myTypesAndTerms" everything works as expected. 

A transcript is attached to demonstrate the issue.

pchiusano:
Fixes #2091. 

The lexer was prematurely committing to the keyword branch even when it was followed by more identifier characters. Checking that the next character isn't an identifier character fixes it. 

I added a few more tests to the transcript that were also failing previously.

```Haskell
-- technically legal Unison syntax
-- this type is the same as `type Either a b = Left a | Right b`
type type! type_ ability_ = ability' type_ | type! type_
```

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -296,7 +296,7 @@ lexemes' eof = P.optional space >> do
      <|> token symbolyId <|> token blank <|> token wordyId
      <|> (asum . map token) [ semi, textual, backticks, hash ]
 
-  wordySep c = isSpace c || not (isAlphaNum c)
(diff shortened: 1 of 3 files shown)


rlmark (approved):
👍  LGTM 

Verification process: Pulled code, ran tests, ran against local codebase, prefixed terms now pass.

PR #2094, opened 2021-06-08
Simplify regression test to not pull all of base

pchiusano:
Minor cleanup. I noticed the regression test for #2027 was downloading all of base every time it was run. I just made the transcript self-contained so it includes exactly those parts of base that are needed to reproduce the original test case.


aryairani (approved):
Good catch.

PR #2099, opened 2021-06-08
add `debug.clear-cache` ucm command

aryairani:
Adds a new command that clears all (Regular & Test) watches from the codebase.

## Test coverage

There is a test that confirms that the new command clears the `RegularWatch` cache; it could be expanded to populate and clear the `TestWatch` cache too.

## Loose ends

#2100 In general, the forcing the result of the existing
```haskell
Codebase.watches :: Codebase m v a -> UF.WatchKind -> m [Reference.Id]
```
function may crash, because the codebase generally doesn't contain the watch expressions, and thus can't supply a reference cycle length.  Just don't look at the results too hard though, and you'll be fine.  Not sure how far removing the size from `Reference.showSuffix` would get us, though I suspect that `Show` instance is how the cycle lengths are recorded into the V1 codebase, for reconstructing the field when converting a `FilePath` to a `Reference`.

### DIFF
--- a/parser-typechecker/tests/Unison/Test/ClearCache.hs
+++ b/parser-typechecker/tests/Unison/Test/ClearCache.hs
@@ -12,7 +12,7 @@ import qualified Unison.Var as WatchKind
 
 test :: Test ()
 test = scope "clearWatchCache" $
-  for_ [minBound .. maxBound] \fmt -> scope (show fmt) do
+  for_ [minBound @Ucm.CodebaseFormat ..] \fmt -> scope (show fmt) do
     c <- io $ Ucm.initCodebase fmt
     let listWatches = io $ Ucm.lowLevel c \c ->
           Codebase.watches c WatchKind.RegularWatch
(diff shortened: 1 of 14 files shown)


pchiusano:
@aryairani Nice! we definitely want a version that clears the test cache though... I'd just have this command clear both caches and (later if we need) have separate commands for clearing one or the other cache.

aryairani:
@pchiusano The command clears all watches (Regular and Test), but the test of the command only tests that it clears Regular watches.

pchiusano:
Oh okay, gotcha.

aryairani:
@pchiusano I updated the description to clarify.

aryairani:
Err, I branched off the wrong branch and will rebase.

parser-typechecker/tests/Unison/Test/ClearCache.hs
```diff
@@ -0,0 +1,39 @@
...
+import Data.Foldable (for_)
+import Data.String.Here (i)
+import EasyTest
+import qualified Unison.Test.Ucm as Ucm
+import qualified Unison.Codebase as Codebase
+import qualified Unison.Var as WatchKind
+
+test :: Test ()
+test = scope "clearWatchCache" $
+  for_ [minBound .. maxBound] \fmt -> scope (show fmt) do
```
pchiusano: A comment here about what `minBound` and `maxBound` are might help. I did a double take when I first read this because I had in mind integers or something, but you're just running the test for the old and new codebase format.

pchiusano (approved):
LGTM! Left one minor comment.

PR #2105, opened 2021-06-10
new pr bundle process

aryairani:
Adds a new command, `pr.create2 <file> <base namespace> <topic namespace>` that creates a self-contained codebase + some metadata in `<file>.upr` (unison pull request) and then crashes due to missing `OutputMessages` work.

Here's an example of repackaging https://github.com/unisonweb/base/issues/12.
```
.> cd .prs

  ☝️  The namespace .prs is empty.

.prs> pull-request.load https://github.com/unisonweb/base:.trunk https://github.com/pchiusano/unisoncode:.prs.random2
  Importing downloaded files into local codebase...     
  Done syncing 4500 entities.
  Importing downloaded files into local codebase...           
  Done syncing 86 entities.
                                                   
  I checked out https://github.com/unisonweb/base:.trunk to base.
  I checked out https://github.com/pchiusano/unisoncode:.prs.random2 to head.
  
  The merged result is in merged.
  The (squashed) merged result is in squashed.
  Use `diff.namespace base merged` or `diff.namespace base squashed` to see what's been updated.
  Use `todo merged.patch merged` to see what work is remaining for the merge.
  Use `push https://github.com/unisonweb/base:.trunk merged` or
  `push https://github.com/unisonweb/base:.trunk squashed` to push the changes.

.prs> ls

  1. base/     (1914 definitions)
  2. head/     (1935 definitions)
  3. merged/   (1935 definitions)
  4. squashed/ (1935 definitions)

.prs> pr.create2 random2 base squashed

  Done syncing 15433 entities.
todo: print a nice diff
```
```
% ls -lh random2.upr
-rw-------  1 arya  staff   6.4M Jun  9 23:10 random2.upr
% unzip -p random2.upr pr.json       
{"version":1,"baseHash":"agacc40npn4o3ctgo89avotts70ircdri6c40d7mm0jmrf000vrqb81bf69o4utkqu633g672eqc24c9jg40t0r7vul1kekpa8sr8f8"}
```

The rest should be pretty self explanatory:
- [ ] add a `pr.load2` command that
  - [ ] unpacks `<file>.upr` to a temp dir
  - [ ] parses a version number out of `<tmp>/pr.json`
  - [ ] branch on `version: 1`
    - [ ] check that `baseHash` is known
    - [ ] merge the root of the codebase at `<tmp>` to wherever the user wants (can't use `Codebase.syncFromCodebase` directly, because it expects a branch loaded first for some reason)
  - [ ] deletes the temp dir
  - [ ] says something nice

### DIFF
--- a/parser-typechecker/package.yaml
+++ b/parser-typechecker/package.yaml
@@ -129,6 +129,7 @@ library:
     - unison-core
     - unison-core1
     - unison-util
+    - zip
 
 executables:
   unison:
--- a/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
@@ -193,8 +193,8 @@ data Command m i v a where
   SyncRemoteRootBranch ::
     RemoteRepo -> Branch m -> SyncMode -> Command m i v (Either GitError ())
 
-  SyncPrBundle ::
-    FilePath -> Branch.Hash -> Branch m -> Command m i v ()
+  CreatePrBundle ::
+    FilePath -> Branch m -> Branch.Hash -> Command m i v ()
 
   AppendToReflog :: Text -> Branch m -> Branch m -> Command m i v ()
 
@@ -275,7 +275,7 @@ commandName = \case
   ImportRemoteBranch{}        -> "ImportRemoteBranch"
   SyncLocalRootBranch{}       -> "SyncLocalRootBranch"
   SyncRemoteRootBranch{}      -> "SyncRemoteRootBranch"
-  SyncPrBundle{}              -> "SyncPrBundle"
+  CreatePrBundle{}            -> "CreatePrBundle"
   AppendToReflog{}            -> "AppendToReflog"
   LoadReflog                  -> "LoadReflog"
   LoadTerm{}                  -> "LoadTerm"
--- a/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Command.hs
@@ -193,6 +193,9 @@ data Command m i v a where
   SyncRemoteRootBranch ::
     RemoteRepo -> Branch m -> SyncMode -> Command m i v (Either GitError ())
 
+  SyncPrBundle ::
+    FilePath -> Branch.Hash -> Branch m -> Command m i v ()
+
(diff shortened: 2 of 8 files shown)


pchiusano:
Nice.

Can we swap the order of the arguments of pr.create2 so it's pr.create2 source destination, matching what we do for merge and squash and diff.namespace?

I know the reasoning behind the current order but I just don't think it's a good reason and I'd rather ucm be self-consistent.

JSON seems kinda overkill for the version/hash file but I guess it's fine.

aryairani:
> Nice.
> 
> Can we swap the order of the arguments of pr.create2 so it's pr.create2 source destination, matching what we do for merge and squash and diff.namespace?

Sure

> JSON seems kinda overkill for the version/hash file but I guess it's fine.

¯\\\_(ツ)_/¯

aryairani:
It doesn't look like there's any documented API or tool to attach files to Github issues, the way you can in the web browser.

The `gh` CLI gives a command to attach files to a gist (text files only), or to a tagged release, but not to an issue.

aryairani:
@pchiusano So we can hide files in github comments in 260k chunks, but the user has to download the whole thing every time they open the issue page on github.  Also we'd have to depend on the `gh` CLI or equivalent, when we already don't want to depend on `git` or even `less`.  

I think this PR will just quietly dump and load files, and we can build on that separately (beefing up the ucm output, chopping / encoding / uploading / searching / downloading / recombining / decoding / loading from github) if desired; but at a point, using github authentication against our own file service might be less work and less throwaway work.  

We should probably try to look even longer-term and only with that perspective, see if there are pieces we can carve out, without too much overhead, that would be useful in the short term too.  (e.g. Will dumping to files even be useful in the medium-/long-term?)

pchiusano:
Yeah, I would keep this PR limited to just creating and loading the PR objects. I think we'll get mileage out of that especially if the format can (eventually) be made to only include minimal history.

260kb doesn't seem sufficient and overall using GitHub issues is feeling pretty janky for stashing the data so I agree we should regroup on what overall experience we are going for in both short and long term. I might have some ideas on this.

aryairani:
The relational namespaces also come up here, where we wanted (but won't for this PR) to be able to quickly find which subnamespaces' history the base belongs to.  We'll be able to search up the tree from children, and forward in time from ancestors, whereas in V1 we could only go the down the tree and backwards in time.  (In V2 we can go forwards and backwards in time, but only down the tree, like before; not up.)

PR #2108, opened 2021-06-11
Preserve pattern row ordering during compilation

### LINKED ISSUE #2106: Bug in pattern matching with constants
```unison
type Tree = Branch Tree Tree | Leaf Nat

tst x = match x with
  Leaf n             -> 0
  Branch (Leaf 0) i -> 1
  Branch i (Leaf 0) -> 2
  Branch i1 i2       -> 3

> tst (Branch (Leaf 1) (Leaf 0))`
```
This returns 3, instead of the expected 2.
Possibly due to a reordering of the pattern matching cases.

dolio:
- `splitMatrixBuiltin` uses `fromListWith` which calls the merge
  function as `merge new old`, so `merge = (++)` was reversing the rows
  of the matrix.

fixes #2106


aryairani (approved):
@dolio Would you add a regression test for it somewhere?

PR #2115, opened 2021-06-15
By default build with optimization off in trunk

pchiusano:
I noticed compile times on trunk were super slow and then spotted that we're building with `-O2` by default on trunk. Guessing this happened during the shuffle of building M2 releases.

### DIFF
--- a/parser-typechecker/package.yaml
+++ b/parser-typechecker/package.yaml
@@ -21,7 +21,7 @@ ghc-options: -Wall -O0 -fno-warn-name-shadowing -fno-warn-missing-pattern-synony
 flags:
   optimized:
     manual: true
-    default: true
+    default: false
 
 when:
   - condition: flag(optimized)


stew (approved):

aryairani:
Disclaimer: despite my comments below, I didn't make this change in question. 😉 

I think the settings for the release build should be the default and we should use `stack build --fast` during development (or a shell alias to that effect), which sets `-O0` among possibly other things.  

I also think the package flags (e.g. `optimized`) are inconsistent (they have to be defined per-package, but aren't) and hard to use (they have to be enabled per-package, but often aren't), and that while they may make sense for package-specific functionality (e.g. "enable extra tracing" or "use experimental runtime" or "use tensorflow") we shouldn't be using them to switch between `-O2` and `-O0`.

pchiusano:
Good point about `--fast`. I'm sold. Will close.

aryairani:
To reduce confusion, I'm inclined to pull out the `optimized` flags as well, and switch all the defaults to `-O2`, which I or whoever can do in a separate PR sometime if there are no objections.  (Or maybe that is just something that can be set globally for our packages in the `stack.yaml` settings if it hasn't been already.)

pchiusano:
Sounds fine to me.

PR #2116, opened 2021-06-15
Add docs to `getDefinition` endpoint

### LINKED ISSUE #1470: let `view` show the type of a built-in term
```
.> view toText

  ⚠️
  
    toText

.> find toText

  1. _github.unisonweb.base.releases._M1j.Float.toText : Float -> Text
  2. _github.unisonweb.base.releases._M1j.Int.toText : Int -> Text
  3. _github.unisonweb.base.releases._M1j.Nat.toText : Nat -> Text
  

.> view 3

  -- base.Nat.toText is built-in.

.> 
```

I'd rather see:
```
.> view 3

  -- base.Nat.toText is built-in.
  -- base.Nat.toText : Nat -> Text

.>
```
or
```
.> view 3

  -- base.Nat.toText : Nat -> Text        (built-in)

.>
```
or something.  I put the output in comments so that it can be valid Unison without a definition.

As a side note, I think it would be cool if we could output a valid Unison definition (like below), and have it recognized as a no-op to the slurper. Wdyt?
```haskell
base.Nat.toText : Nat -> Text
base.Nat.toText = ##Nat.toText
```

### LINKED ISSUE #2112: Update the getDefinition endpoint to include docs for terms and types.
For the Codebase UI to render docs for a definition we want to include the AST of an associated doc.
Meaning if `List.map` has is both a term and a namespace and that namespace includes a `Doc` called `doc`, associate it with the `List.map` term.

Suggested (pseudo JSON):

```js
// GET /getDefinition

{
  typeDefinitions: [
      // A regular Type
    {
      typeNames: ["List"],
      bestTypeName: "List",
      termHash: "@@List",
      typeDoc: Maybe (docName, docHash, Doc), // Nothing when there is no List.doc
      typeDefinition: [/* Source of List */]
    },
  ],
  termDefinitions: [
    // A regular Term
    {
      termNames: ["List.map"],
      bestTermName: "List.map",
      termHash: "@@List.map",
      termDoc: Maybe (docName, docHash, Doc), // Nothing when there is no List.map.doc
      termDefinition: [/* Source of List.map */]
    },
    // A Doc Term
    {
      termNames: ["List.map.doc"],
      bestTermName: "List.map.doc"
      termHash: "@@List.map.doc",
      termDoc: Maybe (docName, docHash, Doc), // Never a Nothing
      termDefinition: [/* Source of List.map.doc */]
    }
  ]
}
```

Related https://github.com/unisonweb/codebase-ui/issues/77

pchiusano:
This PR adds doc rendering to the `getDefinition` endpoint. Ultimately it produces [this type](https://github.com/unisonweb/unison/blob/9a1d3669d6d3f207eaf16fd0403f66d69dfcb68a/parser-typechecker/src/Unison/Server/Doc.hs). 

@hojberg I was thinking you could create a branch in the codebase-ui project that works with this, and we can merge once we're satisfied everything is working.

I did verify that I can still run the codebase UI and do all the same things as before, so I guess that's a good sign. Actually I was a little surprised about this but I'm guessing that the JSON decoding done by the front end is just written in such a way that it ignores the extra info.

Also fixes #1470 

## Controversial/interesting decisions

If you `getDefinition` on a doc, the doc field will be the rendered form of the doc, and the source will be the doc source.

There can be multiple docs associated with a definition - all are returned.

It's just using the generic JSON-ification of that Haskell Doc type, so the JSON is not very space efficient - like the payload for List (which grabs List.doc) are somehow 167kb. This will probably need some improving.

## Loose ends

Tracked here -

- https://github.com/unisonweb/unison/issues/2234
- https://github.com/unisonweb/unison/issues/2235


parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -323,6 +329,11 @@ getTypeOfTerm c r = case r of
     pure $   fmap (const builtinAnnotation)
         <$> Map.lookup r Builtin.termRefTypes
 
+getTypeOfReferent :: (BuiltinAnnotation a, Var v, Monad m)
+                  => Codebase m v a -> Referent.Referent -> m (Maybe (Type v a))
```
pchiusano, line 333: Was shocked to find this function didn't exist / was inlined in random places.

parser-typechecker/src/Unison/Codebase/Editor/DisplayObject.hs
```diff
@@ -4,11 +4,17 @@ module Unison.Codebase.Editor.DisplayObject where
 
 import Unison.Prelude
 import Unison.ShortHash
+import Data.Bifunctor
 
-data DisplayObject a = BuiltinObject | MissingObject ShortHash | UserObject a
+data DisplayObject b a = BuiltinObject b | MissingObject ShortHash | UserObject a
```
pchiusano, line 9: Made `DisplayObject.BuiltinObject` take a `b`, which can be `()` or something more useful - for terms it is convenient for it to be a `Type` for the type of that builtin.

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
```diff
@@ -158,8 +158,8 @@ data Output v
   -- "display" definitions, possibly to a FilePath on disk (e.g. editing)
   | DisplayDefinitions (Maybe FilePath)
                        PPE.PrettyPrintEnvDecl
-                       (Map Reference (DisplayObject (Decl v Ann)))
-                       (Map Reference (DisplayObject (Term v Ann)))
+                       (Map Reference (DisplayObject () (Decl v Ann)))
+                       (Map Reference (DisplayObject (Type v Ann) (Term v Ann)))
```
pchiusano, line 162: This change allows `view` on a builtin to show its type signature.

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
```diff
@@ -181,8 +181,8 @@ data Output v
   | ConfiguredGitUrlParseError PushPull Path' Text String
   | ConfiguredGitUrlIncludesShortBranchHash PushPull RemoteRepo ShortBranchHash Path
   | DisplayLinks PPE.PrettyPrintEnvDecl Metadata.Metadata
-               (Map Reference (DisplayObject (Decl v Ann)))
-               (Map Reference (DisplayObject (Term v Ann)))
+               (Map Reference (DisplayObject () (Decl v Ann)))
+               (Map Reference (DisplayObject (Type v Ann) (Term v Ann)))
```
pchiusano, line 185: Similar treatment here.

parser-typechecker/src/Unison/CommandLine/Main.hs
```diff
@@ -157,10 +156,11 @@ main
   -> Path.Absolute
   -> (Config, IO ())
   -> [Either Event Input]
+  -> Runtime.Runtime Symbol
   -> Codebase IO Symbol Ann
   -> String
   -> IO ()
-main dir defaultBaseLib initialPath (config,cancelConfig) initialInputs codebase version = do
+main dir defaultBaseLib initialPath (config,cancelConfig) initialInputs runtime codebase version = do
```
pchiusano, line 163: A bit of replumbing: previously the `Runtime` was initialized locally in this function, but because the codebase server now needs the runtime (for rendering docs), it needed to be initialized before this function.

parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -103,6 +103,64 @@ prettyAnnotatedRef = typeNamed "Pretty.Annotated"
...
+pattern Doc2BulletedList ds <- Term.App' (Term.Constructor' Doc2Ref ((==) doc2BulletedListId -> True)) (Term.List' (toList -> ds))
+pattern Doc2Section title ds <- Term.Apps' (Term.Constructor' Doc2Ref ((==) doc2SectionId -> True)) [title, Term.List' (toList -> ds)]
+pattern Doc2NamedLink name dest <- Term.Apps' (Term.Constructor' Doc2Ref ((==) doc2NamedLinkId -> True)) [name, dest]
+pattern Doc2Image alt link caption <- Term.Apps' (Term.Constructor' Doc2Ref ((==) doc2ImageId -> True)) [alt, link, caption]
+pattern Doc2Special sf <- Term.App' (Term.Constructor' Doc2Ref ((==) doc2SpecialId -> True)) sf
+pattern Doc2Join ds <- Term.App' (Term.Constructor' Doc2Ref ((==) doc2JoinId -> True)) (Term.List' (toList -> ds))
+pattern Doc2UntitledSection ds <- Term.App' (Term.Constructor' Doc2Ref ((==) doc2UntitledSectionId -> True)) (Term.List' (toList -> ds))
+pattern Doc2Column ds <- Term.App' (Term.Constructor' Doc2Ref ((==) doc2ColumnId -> True)) (Term.List' (toList -> ds))
+pattern Doc2Group d <- Term.App' (Term.Constructor' Doc2Ref ((==) doc2GroupId -> True)) d
```
pchiusano, line 162: Very tedious and uninteresting code to allow for nice pattern matching on an evaluated `Doc` on the Haskell side.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -78,9 +82,11 @@ import Unison.Util.Pretty (Width)
 import qualified Unison.Util.Pretty as Pretty
 import qualified Unison.Util.Relation as R
 import qualified Unison.Util.Star3 as Star3
-import Unison.Util.SyntaxText (SyntaxText)
-import qualified Unison.Util.SyntaxText as SyntaxText
+import qualified Unison.Util.SyntaxText as UST
```
pchiusano, line 85: It was super confusing have two `SyntaxText` identifiers in this file, one of which was randomly imported unqualified. At one point my brain exploded. Now they are both imported qualified.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -522,21 +564,21 @@ prettyType width ppe =
...
   -> Maybe Width
   -> Suffixify
-  -> Codebase m v Ann
+  -> Rt.Runtime v
+  -> Codebase IO v Ann
   -> [HQ.HashQualified Name]
-  -> Backend m DefinitionDisplayResults
-prettyDefinitionsBySuffixes relativeTo root renderWidth suffixifyBindings codebase query
+  -> Backend IO DefinitionDisplayResults
+prettyDefinitionsBySuffixes relativeTo root renderWidth suffixifyBindings rt codebase query
```
pchiusano, line 580: This function has been modified to additionally return docs for each result.

parser-typechecker/src/Unison/Server/CodebaseServer.hs
```diff
@@ -166,12 +167,13 @@ serverAPI = Proxy
 
 app
   :: Var v
-  => Codebase IO v Ann
+  => Rt.Runtime v
+  -> Codebase IO v Ann
```
pchiusano, line 171: Just plumbing changes here to pass in the runtime.

parser-typechecker/src/Unison/Server/Endpoints/FuzzyFind.hs
```diff
@@ -179,29 +173,12 @@ serveFuzzyFind h codebase mayRoot relativePath limit typeWidth query =
...
-              $ FoundType (bestTypeName td) (typeDefinition td) namedType
-            )
-      )
-      refs
+          te <- Backend.typeListEntry codebase r n
+          let namedType = Backend.typeEntryToNamedType te
+          let typeName = Backend.bestNameForType @v ppe (mayDefault typeWidth) r
+          typeHeader <- Backend.typeDeclHeader codebase ppe r
+          let ft = FoundType typeName typeHeader namedType
+          pure (a, FoundTypeResult ft)
```
pchiusano, line 181: Previously this code was I think for convenience using `prettyDefinitionsBySuffixes`, but I gave it a more direct implementation. (Since we don't want to be fetching and rendering docs for every search result in the finder - that's too expensive)

parser-typechecker/src/Unison/Server/Types.hs
```diff
@@ -117,15 +111,17 @@ data TermDefinition = TermDefinition
   { termNames :: [HashQualifiedName]
   , bestTermName :: HashQualifiedName
   , defnTermTag :: Maybe TermTag
-  , termDefinition :: DisplayObject SyntaxText
+  , termDefinition :: DisplayObject SyntaxText SyntaxText
   , signature :: SyntaxText
+  , termDocs :: [(HashQualifiedName, UnisonHash, Doc)]
```
pchiusano, line 116: Here's the main change that will affect JSON served up.

parser-typechecker/src/Unison/Server/Types.hs
```diff
@@ -117,15 +111,17 @@ data TermDefinition = TermDefinition
...
   } deriving (Eq, Show, Generic)
 
 data TypeDefinition = TypeDefinition
   { typeNames :: [HashQualifiedName]
   , bestTypeName :: HashQualifiedName
   , defnTypeTag :: Maybe TypeTag
-  , typeDefinition :: DisplayObject SyntaxText
+  , typeDefinition :: DisplayObject SyntaxText SyntaxText
+  , typeDocs :: [(HashQualifiedName, UnisonHash, Doc)]
```
pchiusano, line 124: And this one will also affect the JSON.

parser-typechecker/src/Unison/Server/Types.hs
```diff
@@ -196,9 +192,14 @@ instance ToJSON TypeTag where
 
 deriving instance ToSchema TypeTag
 
-formatType :: Var v => PPE.PrettyPrintEnv -> Width -> Type v a -> SyntaxText
-formatType ppe w =
-  fmap Syntax.convertElement . render w . TypePrinter.pretty0 ppe mempty (-1)
```
pchiusano, line 201: Moved this into `Backend.hs`, I forget why though.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -51,9 +51,11 @@ pretty :: Var v => PrettyPrintEnv -> Term v a -> Pretty ColorText
 pretty env = PP.syntaxToColor . pretty0 env emptyAc . printAnnotate env
 
 prettyBlock :: Var v => Bool -> PrettyPrintEnv -> Term v a -> Pretty ColorText
-prettyBlock elideUnit env =
-  PP.syntaxToColor . pretty0 env (emptyBlockAc { elideUnit = elideUnit })
-                   . printAnnotate env
+prettyBlock elideUnit env = PP.syntaxToColor . prettyBlock' elideUnit env
+
+prettyBlock' :: Var v => Bool -> PrettyPrintEnv -> Term v a -> Pretty SyntaxText
```
pchiusano, line 56: Just exposing a version of this that gives back a `Pretty SyntaxText` - needed by doc rendering in codebase server.

parser-typechecker/src/Unison/TypePrinter.hs
```diff
@@ -143,7 +146,13 @@ prettySignatures'
   :: Var v => PrettyPrintEnv
   -> [(HashQualified Name, Type v a)]
   -> [Pretty ColorText]
-prettySignatures' env ts = map PP.syntaxToColor $ PP.align
+prettySignatures' env ts = map PP.syntaxToColor $ prettySignatures'' env ts
+
+prettySignatures''
+  :: Var v => PrettyPrintEnv
+  -> [(HashQualified Name, Type v a)]
+  -> [Pretty SyntaxText]
```
pchiusano, line 154: Same deal here - needed a version that returned `Pretty SyntaxText` for doc rendering in codebase server.

parser-typechecker/src/Unison/Util/SyntaxText.hs
```diff
@@ -57,7 +57,7 @@ data Element r = NumericLiteral
              | DocDelimiter
              -- the 'include' in @[include], etc
              | DocKeyword
-             deriving (Eq, Ord, Show, Generic, Functor)
+             deriving (Eq, Ord, Show, Functor)
```
pchiusano, line 60: Deleted this just to convince myself that this type wasn't being JSON-ified.

unison-core/src/Unison/HashQualified'.hs
```diff
@@ -19,7 +20,7 @@ import qualified Unison.ShortHash              as SH
 import qualified Unison.HashQualified          as HQ
 
 data HashQualified n = NameOnly n | HashQualified n ShortHash
-  deriving (Eq, Functor, Generic)
+  deriving (Eq, Functor, Generic, Foldable)
```
pchiusano, line 23: Came in handy at one point - the `toList` will give you just the name, ignoring the short hash if that exists.

unison-src/transcripts/suffixes.output.md
```diff
@@ -38,7 +38,7 @@ The `view` and `display` commands also benefit from this:
 ```ucm
 .> view List.drop
 
-  -- builtin.List.drop is built-in.
+  builtin builtin.List.drop : Nat -> [a] -> [a]
```
pchiusano, line 41: Shows the new behavior of `view` on a builtin term.

Re: #1470

pchiusano (commented):
Self review

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
aryairani: Kind of weird that it includes the type for the built-in but not for the user object.  I would suggest
```haskell
(Type v Ann, DisplayObject (Term v Ann))
``` 
instead?  Or better yet:
```haskell
Maybe (Type v Ann, Term v Ann, DefnType)
``` 
where 
```haskell
data DefnType = Builtin | UserDefined
```

In general (though this is not a particularly interesting example), I think we should create a new type when we want new semantics.

parser-typechecker/src/Unison/Server/Backend.hs
aryairani: Wait, what?

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
pchiusano: Hmm... While I didn't really consider shaking things up more here, fyi the type is already rolled into the term (as a type annotation) and in general the pretty-printer is already prepared to deal with type annotations on bindings elsewhere in the term (including at the top level), so I actually don't mind the current approach. 

There's also I don't think any real requirement that `UserDefined` terms have a type signature on them. We happen to infer the signature and add that in as an annotation if the user didn't provide one, but `x = 42` is a perfectly cromulent term definition IMO.

parser-typechecker/src/Unison/Server/Backend.hs
pchiusano: There's a `SyntaxText` defined in the codebase server package which has its own JSON-ification, and a different `SyntaxType` defined in `Unison.Util` and which is produced by the pretty-printer. I just made them both qualified so I could keep them straight - Haskell does a bad job with the error messages when one symbol is qualified and the other is not.

aryairani (approved):

hojberg:
Doing some integration with this on my branch right now, but so far so good 🎉

hojberg (approved):
I'm able to consume this on the frontend without any problems so far 🎉

hojberg:
@pchiusano Oh i'm actually seeing some strangeness in the Finder API. Let's connect. This otherwise is looking really good!

pchiusano:
Okie dokie, thanks for review. Simon and I sync'd up on the changes and have everything working end to end. Exciting!

Issue #2121, opened 2021-06-16
Error message for Boolean `&&` and `||` uses `and` and `or` instead of their actual names


runarorama:
This error message needs to be improved, as there's no operator called `and`:

![image](https://user-images.githubusercontent.com/130916/122247233-359f1880-ce95-11eb-9d90-fd8be70dece5.png)

PR #2125, opened 2021-06-16
Swap `and` and `or` with `&&` and `||` in boolean mismatch message

tapegram:
## Overview
For issue
https://github.com/unisonweb/unison/issues/2121

Before
Screen Shot 2021-06-16 at 6 35 24 PM

After
Screen Shot 2021-06-16 at 6 35 58 PM

## Implementation notes

Just updated some hard coded strings.

Hopefully, this is not controversial.

## Test coverage

This doesn't appear to be tested. Is this worth testing to prevent a regression and if so how do I go about it?


aryairani:
@tapegram This looks good!  We don't need a regression test for this.

Could you take a look at `CONTRIBUTORS.markdown`, make sure you're okay with the licensing terms, and add yourself to the end of the file as part of this PR?

aryairani (approved):
👍

PR #2132, opened 2021-06-18
Add cabal support via cabal.project

jneira:
## Overview

This is a development oriented change, to make easy start to contribute for cabal users.

## Implementation notes

Adding a `cabal.project` has been enough.

The main drawback is the project has to maintain two build config files.
One of the advantages of cabal is `allow-newer` is fine grained and you can set it per `package:dependency`.
The allow-newer i had to add does not correspond with the comment in the coarse grained `allow-newer` of stack: 
https://github.com/unisonweb/unison/blob/ef1b25af7b2fae38cd311f94b958781d045b021a/stack.yaml#L5

## Test coverage

I've tested the buid locally in windows and it has generated the executable: https://github.com/jneira/unison/releases/download/2d96be-windows/unison.exe
Ideally a job in ci using cabal should be added, to check it is in sync with stack.yaml

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -25,5 +25,41 @@ source-repository-package
   location: https://github.com/unisonweb/megaparsec.git
   tag: c4463124c578e8d1074c04518779b5ce5957af6b
 
+constraints: haskeline -terminfo
+
 allow-newer: 
   haskeline:base
+
+-- For now there is no way to apply ghc-options for all local packages
+-- See https://cabal.readthedocs.io/en/latest/cabal-project.html#package-configuration-options
+
+package easytest
+  ghc-options: -Wall -Werror -Wno-name-shadowing -Wno-type-defaults -Wno-missing-pattern-synonym-signatures -fprint-expanded-synonyms -fwrite-ide-info #-freverse-errors
+
+package parser-typechecker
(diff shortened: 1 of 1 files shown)


jneira:
I am afraid that cabal still does not have an equivalent for stack `$locals` so ghc-options has to be duplicated for every local package :worried: 
Other projects choose putting that options in each `.cabal` file, to make it available for every possible installation method (including nix). Usually a `common` stanza is used to apply them to all package components, to only have them once for each `.cabal` file.

pchiusano:
I noticed we already have this cabal.project that @dolio uses - https://github.com/unisonweb/unison/blob/trunk/contrib/cabal.project but is this more complete or better? If yes maybe a PR to that file.

jneira:
oh, I missed it completely, have you considered move it to top level? this way it would be more visible and usable, without having to set `--project-file` cli option (or move to top level)

pchiusano:
@jneira yes we have considered it, but for now, since stack is the official way of building the project and is what's used by CI etc, we moved it out of the root directory. 

It does get a mention in https://github.com/unisonweb/unison/blob/trunk/development.markdown#building-with-cabal but people may not read that... :)

jneira:
fair enough, thanks for pointing me in the right direction
will reopen the pr over the existing file

PR #2134, opened 2021-06-18
Update cabal.project

jneira:
## Overview

Update cabal.project taking as example stack.yaml.

I am afraid that cabal still does not have an equivalent for stack $locals so ghc-options has to be duplicated for every local package 😟
Other projects choose putting that options in each .cabal file, to make it available for every possible installation method (including nix). Usually a common stanza is used to apply them to all package components, to only have them once for each .cabal file.

## Test coverage

Tested building the project with cabal locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,8 @@ scratch.u
 .stack-work
 stack.yaml.lock
 
+# Cabal
+dist-newstyle
+
 # GHC
 *.hie
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -53,3 +53,4 @@ The format for this list: name, GitHub handle, and then optional blurb about wha
 * Johannes Huster (@JohannesHuster)
 * Joseph Morag (@jmorag)
 * Tavish Pegram (@tapegram)
+* Javier Neira (@jneira)
--- a/contrib/cabal.project
+++ b/contrib/cabal.project
@@ -1,10 +1,29 @@
-packages: unison-core/ parser-typechecker/ yaks/easytest/ yaks/haskeline/
+packages:
+    yaks/easytest
+    parser-typechecker
+    unison-core
+    codebase2/codebase
+    codebase2/codebase-sqlite
+    codebase2/codebase-sync
+    codebase2/core
+    codebase2/util
+    codebase2/util-serialization
+    codebase2/util-term
 
-allow-newer: base, directory, filepath
+source-repository-package
+  type: git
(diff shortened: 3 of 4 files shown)


aryairani:
Thanks @jneira — could you add yourself to CONTRIBUTORS.markdown (and agree to the MIT licensing terms in it) as part of this PR too?

aryairani (approved):
LGTM, but I'm not a cabal user.

jneira:
Testing this i am getting this error:
```
Configuring library for easytest-0.1..
<command line>: does not exist: #-freverse-errors
cabal.exe: Failed to build easytest-0.1 (which is required by exe:runtests
from easytest-0.1). The failure occurred during the configure step.
```
but i was able to build the project in #2132 🤔

jneira:
nvm i copied ghc-options from stack.yaml and `#` is a comment there

jneira:
@aryairani i would recommend to set up a `cabal build` in ci or this file will go out of sync very quickly.
If you agree i could try to add it to gha (in this or another pr, as you consider)

aryairani:
> @aryairani i would recommend to set up a `cabal build` in ci or this file will go out of sync very quickly.

@jneira Right, but the reason we haven't done that is because we aren't currently in a position to maintain both, and don't want to be forced to by CI.  We could have a separate repo (or possibly a third-party repo) that maintains the cabal build and keeps it up to date.

Or if it could be set up gha (in this or a separate pr) as a separate optional action that doesn't have to pass in order to merge a PR, but give some visibility to the status of the cabal build, which then others like yourself could use as a signal to come update the cabal file, that might be okay.

What do you think?

jneira:
> Or if it could be set up gha (in this or a separate pr) as a separate optional action that doesn't have to pass in order to merge a PR, but give some visibility to the status of the cabal build, which then others like yourself could use as a signal to come update the cabal file, that might be okay.

> What do you think?

That option seems a good compromise. The unique caveat is it would make ci slower without cache

aryairani:
> That option seems a good compromise. The unique caveat is it would make ci slower without cache

What do you mean?

jneira:
I mean that make a `cabal build` will make the ci slower when the cache used for the cabal build is invalidated.

aryairani:
> I mean that make a `cabal build` will make the ci slower when the cache used for the cabal build is invalidated.

We could not invalidate it :)

jneira:
well anyways I think we could merge this and add ci thereafter, if you agree

Issue #318, opened 2019-02-04
`replace` command


pchiusano:
Example:

    > replace foo bar

This has the same effect as creating a new version of `foo` in a new file then doing `update` (which creates an edit for `foo`), followed by `> rename foo bar`. In other words, it lets you replace one definition with another, manually, without needing to do an `update`. This is another way of resolving a name conflict. If you just do a `rename foo#sdf bar`, dependents of the `foo#sdf` definition are unaffected. If you do a `replace`, you are marking `foo#sdf` as obsolete and saying you want all dependents to use `bar` instead.

This might also just be useful if you realize that a definition you already have lying around is what you want to use in place of something else, like if you realize you've written an overly concrete / monomorphic function that turns out to already be implemented in more generic form.

Notes:

* Either of the two arguments may be hash-qualified
* Should do a `propagate` afterwards

hasheddan:
Hi @pchiusano! I recently discovered Unison and would love to start contributing. Would this be a good place to begin?

aryairani:
Hi @hasheddan — yes and no :) We are in the middle of a refactor (`wip/branchless` branch), so this feature should get implemented there.  It would be in `HandleInput.hs` and you could likely use the hopefully correctly implemented `UpdateI` and `RenameTermI` as guidance.  

However, since the refactoring is super incomplete, super doesn’t run, and is  largely a mess, I’d hate to subject you to it.  Nevertheless, this feature is a ways down on our to-do list, so if you are feeling masochistic you could definitely take a stab at it without stepping on anyone’s toes.

We’ll try to answer any questions we can.

runarorama:
Consider moving out of M1 to cut scope.

aryairani:
@hasheddan This is ready to be worked on now, if you're still open to it. 😅

pchiusano:
@aryairani how is this different than existing `replace.term` and `replace.type` commands?

aryairani:
@pchiusano Good point, it took me a while to decide the answer yesterday before not closing it, but then I didn't write it down.

I think `replace.term` and `replace.type` just need to be updated to take any `HashQualified` (right now they only accept `ShortHash`, and we could also have a generic `replace` command that figures out whether the args are both a term or both a type, and calls the respective underlying logic.

hasheddan:
@aryairani apologies for just circling back to this! Is this still something that needs to be worked on? If so, I would be happy to take it :) If not, I would be happy to work on anything else that is needing attention!

[...]

PR #2135, opened 2021-06-18
Create `replace` command instead of `replace.term` and `replace.type`

tapegram:
https://github.com/unisonweb/unison/issues/318
Paired with @bontaq 

## Overview

1) This PR creates a new command `ReplaceI` that combines the behaviors of `ReplaceTermI` and `ReplaceTypeI`. It also exposes this to the repl as `replace`.
2) `replace.term` and `replace.type` are removed.
3) Adds a new error message for type/term mismatches.

## Implementation notes

Basically just mashed the existing code together. Would appreciate suggestions on how to factor out these behaviors into smaller functions. Decided not to do that right now because so many dependencies are declared at the top of the function and I wasn't finding a clean way to pass them around (without big weird function types).

1) Ideally, this should be refactored into smaller reusable functions
2) Removed `replace.term` and `replace.type`

## Test coverage

Added a transcript test for happy path and failure cases.
Updated existing tests that depended on `replace.term` or `replace.type`.

## Loose ends

1) Returns an error in the big case block. I'm not sure what could actually be matched there -- but if you can think of something please let me know!

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -397,6 +397,10 @@ loop = do
           DeleteTypeI def -> "delete.type " <> hqs' def
           DeleteBranchI opath -> "delete.namespace " <> ops' opath
           DeletePatchI path -> "delete.patch " <> ps' path
+          ReplaceI src target p ->
+            "replace "      <> HQ.toText src <> " "
+                            <> HQ.toText target <> " "
+                            <> opatch p
           ReplaceTermI src target p ->
             "replace.term " <> HQ.toText src <> " "
                             <> HQ.toText target <> " "
@@ -1291,7 +1295,7 @@ loop = do
           BranchUtil.makeDeleteTermName (resolveSplit' (HQ'.toName <$> hq))
         go r = stepManyAt . fmap makeDelete . toList . Set.delete r $ conflicted
 
-      ReplaceTermI from to patchPath -> do
+      ReplaceI from to patchPath -> do
         let patchPath' = fromMaybe defaultPatchPath patchPath
         patch <- getPatchAt patchPath'
         QueryResult fromMisses' fromHits <- hqNameQuery [from]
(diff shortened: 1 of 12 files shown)


unison-src/transcripts/command-replace.output.md
```diff
@@ -0,0 +1,128 @@
...
+  type X = Two Nat Nat
+
+```
+Try with a type/term mismatch
+```ucm
+.scratch> replace.any X x
+
+  ⚠️
+  
+  The following names were not found in the codebase. Check your spelling.
```
runarorama: I'd want to improve this message. Instead of saying`X` isn't in the codebase, it should say something about `X` and `x` not being the same kind of thing.

parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -1014,6 +1014,9 @@ replaceEdit f s = self
       _ -> Left $ I.help self
     )
 
+replace :: InputPattern
+replace = replaceEdit Input.ReplaceI "any"
```
runarorama: I really want this command to be called `replace`. In that case we could just delete the `replace.term` and `replace.type` commands, or implement them as a call to `replace` that errors if the names given aren't the appropriate sort.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1291,6 +1295,92 @@ loop = do
           BranchUtil.makeDeleteTermName (resolveSplit' (HQ'.toName <$> hq))
         go r = stepManyAt . fmap makeDelete . toList . Set.delete r $ conflicted
 
+      ReplaceI from to patchPath -> do
```
runarorama, line 1290: This duplicates quite a lot of code that's already duplicated twice in this file. It would be best to move that code into a function that gets called from here with different arguments depending on whether the user typed `replace.term`, `replace.type`, or just `replace`.

Or better yet, I feel (not very strongly) that `replace.term` and `replace.type` should just go away.

runarorama (changes requested):

runarorama (approved):

runarorama:
This looks great, thank you! 🌈⭐

PR #2142, opened 2021-06-24
Fix misspelling of definition

hojberg:
## Overview

Very simple change: there was a few misspellings of definition with related to the docs help and conflicts.

## Loose Ends
~~I'm not sure which transcripts gets generated and how to deal with that? There are 2 more misspellings that I didn't fix since they were in an `*.output.*` transcript.~~ Edit: I ran transcripts locally with `stack exec transcripts` and that seems to have done the trick.


aryairani (approved):
Thanks.  Surprising that it showed up twice.

PR #2151, opened 2021-06-27
WIP: Walking through the Birthday Email Kata

tapegram:
This is a draft for some practical examples of writing Unison. This is a walkthrough of the Birthday Email kata.

TODO:
1) Break into smaller docs so the narrative can be more focused with links to deep dive "asides".
2) Include links to additional learning materials / talks like Dependency Rejection, Domain Modeling Made Functional, etc.
3) Improve the code to be more idiomatic?
4) Any other examples / decisions to include?

### DIFF
--- a/docs/birthday-kata.md
+++ b/docs/birthday-kata.md
@@ -61,7 +61,7 @@ unique type Date =
   , year: Nat
   }
 ```
-
+#### Decision 1 - Record Types
 Our employee is a record type consisting of an `id`, `name`, and a `birthday`. In Unison, the record type syntax is actually equivalent to a normal data type like
 ```unison
 type Employee = Employee EmployeeId Name Date
@@ -107,4 +107,179 @@ except the compiler will automatically generate a collection of functions (lens)
 
 `Name` and `Date` are record types as well.
 
+#### Decision 2 - Nominal vs Structural Typing
 Also note that we have opted in to using nominal typing (with the `unique` keyword) instead of structural typing. If we had left out `unique` the compiler would consider `EmployeeId`, `FirstName`, and `LastName` all to be the same type. For our current use case, we want to make sure that the compiler considers all of these different types, so we can prevent mistakes like accidentally passing in a `LastName` where an `EmployeeId` should go.
+
+#### Decision 3 - Avoiding Primitives
+We could have also modeled the Employee like this (or any other similar way)
+```unison
+unique type Employee  = 
+  { id: Text
+  , name: Name
+  , birthday: Date 
+  }
+
+unique type Name = 
+  { first: Text
+  , last: Text
+  }
+
+unique type Date = 
+  { day: Nat
+  , month: Nat
+  , year: Nat
+  }
+```
+
+When deciding whether to use primitives or to wrap them in more rich domain types we should consider some tradeoffs
+1) Using primitive types gives the compiler less context on what you are attempting to model. For example, using `Text` for the id, the first name, and the last name, means the compiler wouldn't catch an easy mistake like
+```unison
+foo = 
+    id = "12345"
+    firstName = "John"
+    lastName = "Lennon"
+    employee = Employee id (Name lastName firstName) (Date 1 1 1991)
+```
+(We've transposed the employees name!)
+Or even worse, maybe we accidentally swap the id and last name! However, the compiler would catch this problem.
+
+```unison
+foo = 
+    id = EmployeeId "12345"
+    firstName = FirstName "John"
+    lastName = LastName "Lennon"
+    employee = Employee id (Name lastName firstName) (Date 1 1 1991)
+    ()
+```
+```ucm
+  The 1st argument to `Name`
+  
+            has type:  LastName
+      but I expected:  FirstName
+  
+      9 | unique type LastName = LastName Text
+     10 | unique type Name = 
+     11 |   { first: FirstName
+      .
+     23 |     employee = Employee id (Name lastName firstName) (Date 1 1 1991)
+```
+Of course, you can still make a mistake like
+```unison
+firstName = LastName "Lennon"
+```
+But it's much harder to make this mistake. If we are ok with the extra types and the overhead of deconstructing them, we can catch many more typos/bugs at compile time, which means fewer tests are necessary to ensure correctness!
+
+2) In the future, Unison may introduce the ability to make to make data constructors private and/or refined types. Doing so would allow us to to avoid another huge source of bugs when using primitive types directly, which is that the set of possible values is much larger than what we are attempting to model. You can see this in our model for `Date`
+
+```unison
+unique type Date = 
+  { day: Nat
+  , month: Nat
+  , year: Nat
+  }
+```
+
+By modelling day/month/year all as `Nat` we have introduced the possibility of an illegal state into our system. For example, in this model, we could have illegal dates like `0/500/99999999999999`. In fact, lets try to restrict the possible values a bit, to reduce the surface area for future bugs. This means more tests and more runtime validation (and runtime validation failures).
+
+Lets try to contrain the possible values a bit.
+```unison
+unique type Date = 
+  { day: Day
+  , month: Month
+  , year: Year
+  }
+unique type Month 
+  = January
+  | February
+  | March
+  | April
+  | May
+  | June
+  | July
+  | August
+  | September
+  | October
+  | November
+  | December
+unique type Day = Day Nat
+unique type Year = Year Nat
+```
+
+It is really easy for us to just enumerate all of the possible months. Now we have eliminated the possiblity of an invalid month appearing in our system! Now we have to decide how far we want to go to make illegal state unrepresentable. Maybe `Nat` is ok for year if we don't care about `BCE` and we are ok with possibly talking about birthdays millions of years in the future (not ideal). But it would be nice to disallow days of months that don't exist, and this is where refined types, dependent types, or private constructors could be leveraged. Hopefully, we will have those one day!
+
+### Email
+Let's apply some similar decision making to get an Email.
+```unison
+unique type Email = 
+  { to: EmailAddress
+  , subject: Subject
+  , body: Body
+  }
+
+unique type EmailAddress = EmailAddress Text
+unique type Subject = Subject Text
+unique type Body = Body Text
+```
+
+It would be nice to add some constraints to the email address (like a regex to ensure only valid email addresses can exist) but this is pretty good so far!
+
+But maybe we can build a little bit toward the future here. The prompt told us the future will likely include sending birthday messages via other platforms, like SMS or robocalls.
+
+Maybe our model should actually be about a `Message` of which there is some relationship with an `Email`. One possible way to model this would be
+```unison
+unique type Message 
+  = Email EmailAddress Subject Body
+```
+which would give us the option of extending it in the future as needed
+```unison
+unique type Message 
+  = Email EmailAddress Subject Body
+  | SMS PhoneNumber Body
+  | TwitterDM TwitterHandle Body
+```
+
+### Summary
+There isn't one "best" way to model this, but this is one possible way:
+```unison
+unique type Employee = 
+  { id: EmployeeId
+  , name: Name
+  , birthday: Date 
+  }
+
+unique type EmployeeId = EmployeeId Text
+unique type FirstName = FirstName Text
+unique type LastName = LastName Text
+unique type Name = 
+  { first: FirstName
+  , last: LastName
+  }
+
+unique type Date = 
+  { day: Day
+  , month: Month
+  , year: Year
+  }
+unique type Month 
+  = January
+  | February
+  | March
+  | April
+  | May
+  | June
+  | July
+  | August
+  | September
+  | October
+  | November
+  | December
+unique type Day = Day Nat
+unique type Year = Year Nat
+
+unique type Message 
+  = Email EmailAddress Subject Body
+
+unique type EmailAddress = EmailAddress Text
+unique type Subject = Subject Text
+unique type Body = Body Text
+```
--- a/docs/birthday-kata.md
+++ b/docs/birthday-kata.md
@@ -283,3 +283,103 @@ unique type EmailAddress = EmailAddress Text
 unique type Subject = Subject Text
 unique type Body = Body Text
 ```
+
+## Abilities
+Now that we have our model, let's quickly talk about the effects/infrastructure we know we will need.
+
+One of the goals of this exercise (and generally working as a software engineer) is to decouple infrastructure from our business logic (which is often approximately the same thing as spearating pure and impure code). In OO we would use interfaces to define contracts and dependency inject in whatever implementations we need (be it postgres, aws, in memory, mocks, etc).
+
+In Unison, we leverage Abilities (also called Algebraic Effects) to accomplish something similar. See [the abilities documentation](ability-typechecking.markdown) for a more detailed overview of Abilities.
+
+In our case, let's limit ourselves to the following effects: sending a message, fetching employees from somewhere (a flatfile, a db, an api, anything), and fetching the current date. For the sake of simplicity let's ignore an exception throwing ability (and failures in general) or any other abilities that may make sense to add in a real production system.
+
+```unison
+ability MessageService where
+  send: Message ->{MessageService} Message
+
+ability EmployeeRepository where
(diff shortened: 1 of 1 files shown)


docs/birthday-kata.md
```diff
@@ -0,0 +1,385 @@
...
+
+Platform Goals:
+1) Testable; we should be able to test the internal application logic with no need to ever send a real email.
+2) Flexible: we anticipate that the data source in the future could change from a flat file to a relational database, or perhaps a web service. We also anticipate that the email service could soon be replaced with a service that sends greetings through Facebook or some other social network.
+3) Well-designed: separate clearly the business logic from the infrastructure.
+
+For further information about this kata please checkout the original post [here](http://matteo.vaccari.name/blog/archives/154).
+
+### Our Approach
+First, we will come up with a rich domain model and discuss different ways we can model the system to take advantage of the compiler nd what the tradeoffs are.
```
JoshNavi: typo nit: `compiler nd what` -> `compiler and what`

docs/birthday-kata.md
```diff
@@ -0,0 +1,385 @@
...
+  findById: EmployeeId -> {EmployeeRepository} {Employee}
+```
+2) have there behavior defined elsewhere (using handler syntax we will get to later)
+3) are composable (hence "algebraic effects")
+4) require that any calling function provide the ability in their type signature, which bubbles all the way to the top of the call stack (until it finds a handler).
+
+So let's scaffold our main function for this exercise and see this stuff in action.
+
+```unison
+sendBirthdayEmails : '[Message]
```
JoshNavi: do you mean to have `'[Message]` here or should it just be `[Message]` until the next example

docs/birthday-kata.md
```diff
@@ -0,0 +1,385 @@
...
+```
+Let's say we want a function we can call that will find all the employees whose birthday is today, create emails for them, send those emails, and then return those emails so we can see what was created and sent.
+
+As an aside, let's avoid modeling failures simply because it will clutter this exercise a bit. In real life, many of our effects would have type signatures with `Either` so we could model possible failures, but let's just naively pretend everything always works.
+
+So let's do some more scaffolding!
+```unison
+sendBirthdayEmails : '[Message]
+sendBirthdayEmails = 
+  today' = today
```
JoshNavi, line 351: does `today` automatically get called here? should it be `!today` since otherwise its a deferred computation by default?

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+
+So lets start with:
+```unison
+unique type Employee  = 
+  { id: EmployeeId
+  , name: Name
+  , birthday: Date 
+  }
+
+unique type EmployeeId = EmployeeId Text
```
bontaq, line 47: Might wanna put these above `Employee` so it's not defined after use even tho it works

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+      Name.first.set           : FirstName -> Name -> Name
+      Name.last                : Name -> LastName
+      Name.last.modify         : (LastName ->{g} LastName) -> Name ->{g} Name
+      Name.last.set            : LastName -> Name -> Name
+```
+
+`Name` and `Date` are record types as well.
+
+#### Decision 2 - Nominal vs Structural Typing
+Also note that we have opted in to using nominal typing (with the `unique` keyword) instead of structural typing. If we had left out `unique` the compiler would consider `EmployeeId`, `FirstName`, and `LastName` all to be the same type. For our current use case, we want to make sure that the compiler considers all of these different types, so we can prevent mistakes like accidentally passing in a `LastName` where an `EmployeeId` should go.
```
bontaq: > all to be the same type

Could be helpful to note which type it'd be to help understanding

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
+# Practical Example - Birthday Kata
```
rlmark: I think many folks may not know what the birthday Kata is off-hand. What if we call this Practical Example - Modeling a Birthday Message" or something like that, and then in the very next section, under "# The Kata" we can both link to the blog and foreground it with some context to help non DDD aficionados come up to speed.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+
+## The Kata
+[Source: Matteo Vaccari](http://matteo.vaccari.name/blog/archives/154)
+
+### Goals
+This kata is normally used to practice TDD, dependency injection, and ports and adapters architectures in an Object Oriented style. Instead, we will tackle the same problem in Unison and explore TDD, dependency rejection, and how pure FP languages naturally encourage ports and adapters like architectures (through seperation of pure and impure functions).
+
```
rlmark, line 8: We might need to add a high-level blurb about what the ports and adapters architecture idea is given that it's a central part of the stated goal.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+  , month: Nat
+  , year: Nat
+  }
+```
+#### Decision 1 - Record Types
+Our employee is a record type consisting of an `id`, `name`, and a `birthday`. In Unison, the record type syntax is actually equivalent to a normal data type like
+```unison
+type Employee = Employee EmployeeId Name Date
+```
+except the compiler will automatically generate a collection of functions (lens) that allow for easy access and updates of this immutable data structure, which will be a big help later. You can see this in the output in `ucm`:
```
rlmark: What if we omit the parenthetical about lenses here, in the spirit of "teach one thing" 😁?

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+```unison
+unique type Employee  = 
+  { id: EmployeeId
+  , name: Name
+  , birthday: Date 
+  , email: EmailAddress
+  }
+```
+
+It would be nice to add some constraints to the email address (like a regex to ensure only valid email addresses can exist) but this is pretty good so far!
```
rlmark: I think we could remove this sentence because in all likelihood we wouldn't use a regex to do validation of emails, and we probably couldn't solely use types to delimit the full email rfc. You've made the point really well earlier that types can make illegal states impossible so the details of how we'd validate emails can probably be hand-waved away for now 😄.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+  today' = today
+  employees = fetchAllByBirthday (Date.day today', Date.month today')
+  messages = map (Employee.toBirthdayMessage) employees
+  map (send) messages
+
+Employee.toBirthdayMessage : Employee -> Message
+Employee.toBirthdayMessage _ = todo "toBirthdayMessage"
+```
+
+And the compiler seems happy so far! Note that we just deferred the entire computation by putting the `'let` at the beginning.
```
bontaq, line 390: above I think you meant `sendBirthdayEmails : '{MessageService, EmployeeRepository, Calendar} [Message]`instead of `sendBirthdayEmails : '[Message]`?

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+Calendar.handler.mock : Date -> k -> k
+Calendar.handler.mock date k =
+  h : Request {Calendar} k -> k
+  h = cases
+    {today -> resume} -> handle resume date with h
+    { resume } -> resume
+  handle k with h
+
+EmployeeRepository.handler.mock : [Employee] -> k -> k
+EmployeeRepository.handler.mock employees k =
```
bontaq, line 439: I dig the style of naming for the handlers, ez to discover

rlmark (commented):
Hey @tapegram I just realized this PR is targeting the docs folder in /unisonweb/unison. Most of the documentation is hosted out of `https://github.com/unisonweb/unisonweb-org/tree/master/src/data/docs` at the moment so maybe once you're ready to move it out of the draft stage you could open the PR against the unisonweb-org repo for easier hosting? In that PR we can also talk about where to put this document in the nav tree / website / blog etc.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+unique type Subject = Subject Text
+unique type Body = Body Text
+```
+
+## Abilities
+Now that we have our model, let's quickly talk about the effects/infrastructure we know we will need.
+
+One of the goals of this exercise (and generally working as a software engineer) is to decouple infrastructure from our business logic (which is often approximately the same thing as spearating pure and impure code). In OO we would use interfaces to define contracts and dependency inject in whatever implementations we need (be it postgres, aws, in memory, mocks, etc).
+
+In Unison, we leverage Abilities (also called Algebraic Effects) to accomplish something similar. See [the abilities documentation](ability-typechecking.markdown) for a more detailed overview of Abilities.
```
rlmark, line 301: I think this link could point to the hosted abilities page, which is a little more fleshed out: https://www.unisonweb.org/docs/abilities/

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+  send: Message -> Message
+
+ability EmployeeRepository where
+  fetchAllByBirthday: (Day, Month) -> [Employee]
+
+ability Calendar where
+  today: Date
+```
+
+Abilities can sort of be thought of as providing interfaces, and interfaces in turn can be thought of as providing commands.
```
rlmark: I'm not sure what this means 😅. I'm probably missing some context on what "commands" means in this domain, so a note about that might be good for the general readership.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+
+One of the goals of this exercise (and generally working as a software engineer) is to decouple infrastructure from our business logic (which is often approximately the same thing as spearating pure and impure code). In OO we would use interfaces to define contracts and dependency inject in whatever implementations we need (be it postgres, aws, in memory, mocks, etc).
+
+In Unison, we leverage Abilities (also called Algebraic Effects) to accomplish something similar. See [the abilities documentation](ability-typechecking.markdown) for a more detailed overview of Abilities.
+
+In our case, let's limit ourselves to the following effects: sending a message, fetching employees from somewhere (a flatfile, a db, an api, anything), and fetching the current date. For the sake of simplicity let's ignore an exception throwing ability (and failures in general) or any other abilities that may make sense to add in a real production system.
+
+```unison
+ability MessageService where
+  send: Message -> Message
```
rlmark, line 307: While the abilities are inferred by the compiler to the right of these arrows, could we spell them out here for clarity? 
`send: Message -> {MessageService} Message` etc. 

A more general data modeling topic to ponder is if returning the `Message` is material to the `MessageService` ability, or if `send` should simply return unit `()` to represent the effect of a message having been sent.  I suspect that returning an actual message might aid in the TDD nature of the exercise, but I'd be happy to think through if there's a different way of capturing / validating these messages if you felt like the MessageService's send method is better modeled with `()`.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+```unison
+sendBirthdayEmails : '[Message]
+sendBirthdayEmails = 
+  today' = today
+  employees = fetchAllByBirthday (Date.day today', Date.month today')
+  messages = map (Employee.toBirthdayMessage) employees
+  map (send) messages
+
+Employee.toBirthdayMessage : Employee -> Message
+Employee.toBirthdayMessage _ = todo "
```
rlmark: closing quote here

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+            !sendBirthdayEmails 
+          with (Calendar.handler.mock (Date (Day 1) January (Year 1991)))
+        with (EmployeeRepository.handler.mock [])
+      with MessageService.handler.mock
+    expected = []
+    if expected === actual
+    then true
+    else bug (expected, actual)
+
+Calendar.handler.mock : Date -> k -> k
```
rlmark: Love the idea of mock handlers for testing. Could we add the abilities in these signatures to clarify for folks what they look like?

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+Well we can't defer this call any more, this is the boundary of the system! So it's time to implement handlers for our abilities!
+
+```unison
+test> sendBirthdayEmails.tests.noBirthdaysToday = 
+  check let
+    actual = 
+      handle 
+        handle 
+          handle 
+            !sendBirthdayEmails
```
rlmark, line 421: I'm gonna think on this structure overnight 🤔 but I think this is combining two concerns, it's creating a conglomerate mock handler for our core business logic, and also doing the test assertion. One thing we might do is separate the test assertion from the handler for better code re-use.  

```
use Universal 
testBusinessLogicHandler : '{MessageService, EmployeeRepository, Calendar} a -> a  
testBusinessLogicHandler myBusinessLogic = 
  handle 
    handle 
      handle 
        !myBusinessLogic 
      with (Calendar.handler.mock (Date (Day 1) January (Year 1991)))
    with (EmployeeRepository.handler.mock [])
  with MessageService.handler.mock

test> sendBirthdayEmails.tests.noBirthdaysToday = 
  check let
    actual = testBusinessLogicHandler sendBirthdayEmails
    expected = []
    if expected === actual
    then true
    else bug (expected, actual)
```

This might be better for future tests because you might imagine other tests using this test fixture, or a more general one which would let the caller of the mock handler provide expectations: `testBusinessLogicHandler2 : Date -> [Employee] -> '{MessageService, EmployeeRepository, Calendar} a -> a `

rlmark: Hah, silly me I should have read a bit farther down where you do exactly this.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+      handle resume birthdays with h
+    { resume } -> resume
+  handle k with h
+
+test> sendBirthdayEmails.tests.noBirthdaysToday = 
+check let
+  actual = 
+    handle 
+      !sendBirthdayEmails 
+    with sendBirthdayEmails.handlers.mock (Date (Day 1) January (Year 1991)) []
```
rlmark: Love this simplified handler! If you delay the computation `sendBirthdayEmails.handlers.mock : Date -> [Employee] -> '{Calendar, EmployeeRepository, MessageService} k -> k` you don't need to use the `handle` syntax in the actual test.  `actual = sendBirthdayEmails.handlers.mock (Date (Day 1) January (Year 1991)) [] sendBirthdayEmails` <-- this is more idiomatic for calling a handler.

docs/birthday-kata.md
```diff
@@ -0,0 +1,890 @@
...
+
+Note that abilities 
+1) can provide multiple interfaces if we wanted
+```unison
+ability EmployeeRepository where
+  fetchAllByBirthday: (Day, Month) -> [Employee]
+  fetchAll: [Employee]
+  findById: EmployeeId -> {Employee}
+```
+2) have there behavior defined elsewhere (using handler syntax we will get to later)
```
brianbollerman: slight typo. `there` should be `their`

docs/birthday-kata.md
tapegram: Done but it would also be cool to get refined/dependent/ liquid types ;) And also maybe access modifiers (if we are content with run time validation)

docs/birthday-kata.md
tapegram: Yeah I think you are right, it should probably return a `()`, or better yet `Either SendFailure ()`, though I think I might want to defer editing this to include chaining Eithers, though it would be interesting to see if Unison has a `do` notation.

I think I only wrote it this way out of laziness from an earlier implementation where I changed it to return Message so I didn't need to add an extra `tap` function (or whatever it is called when you call a function with some value and then ignore the result and return the original value).

docs/birthday-kata.md
tapegram: Yeah it might just be worth removing. I stole this from the Do Be Do Be Do paper.
```
2.4 Abilities Collect Interfaces; Interfaces Offer Commands
Abilities (Frank’s realisation of algebraic effects) are collections
of parameterised interfaces, each of which describes a choice of
commands (known elsewhere as operations [49]). Command types
may refer to the parameters of their interface but are not otherwise
polymorphic. Here are some simple interfaces
```

docs/birthday-kata.md
tapegram: I haven't updated this yet, but probably will on the next pass!

pchiusano:
Going to close this now that it's been published on the blog. Thanks @tapegram!

Issue #2159, opened 2021-06-30
duplicate/out of date file between unisonweb/unison and unisonweb/unisonweb-org


Carreau:
See slack's thread,  https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1625081221298300

https://github.com/unisonweb/unison/blob/4418667b79e1788cb9511c1dddc2239694bee1ec/docs/LanguageReference.md 
(last modif 2019)
https://github.com/unisonweb/unisonweb-org/blob/41963a806f293a79b4b838a12819aa10fbdb3a01/src/data/docs/language-reference/README.md 
(last modif 2020)

At least are duplicated and out-of-sync, almost everything in /doc/ is likely out of date or to be removed in favor of the version in unison/unisonweb-org

PR #2160, opened 2021-06-30
Remove old duplicate file now on unisonweb-org repo.

Carreau:
Addresses part of #2159

A newer more up-to-date version of this is on the website repo under
 /src/data/docs/language-reference/README.md

The other files will need to also be checked independently for wether
the contents need to be kept or not I'm not familiar enough to judge.


pchiusano (approved):
thanks!

PR #2170, opened 2021-07-02
Avoid building the complete FQN mapping when only a few keys are needed

pchiusano:
Noticed this while working on `Backend`. One of the things done when fetching a definition is to return all the fully-qualified names for that definition. Currently, that's done by building a `Map Reference (Set Text)` for ALL references in the `Branch`, even though we only ever do one (or a few) lookups in that `Map` for the references we're actually returning.

This tweak avoids building any map at all - it just does the relational lookup and filters on the way out to just include the FQNs.

Manually tested just by loading some definitions in the codebase UI.


runarorama (approved):
Nice!

pchiusano:
Not rigorously tested but it does seem like definition loading is noticeably snappier with this change.

PR #2182, opened 2021-07-08
Fix bad calling convention for `Value.load`

dolio:
In cases where a value had missing dependencies, `Value.load` was returning a list of raw references, rather than properly wrapping them as `Referent`. This subsequently caused problems that could result in a segfault if those erroneous values were used.


pchiusano:
Can you add a regression test?

pchiusano (approved):
LGTM!

stew (approved):

Issue #2189, opened 2021-07-09
Delete files containing namespace blocks.


philderbeast:
Please delete the 10 *.u files in this repo that still use namespace. These can be found with this [search](https://github.com/search?q=namespace+repo%3Aunisonweb%2Funison+extension%3Au&type=Code&ref=advsearch&l=&l=). It is a natural thing to want to write scratch files with namespaces and finding examples like these suggests it is still possible but namespace blocks have been removed (but may come back #449).

pchiusano:
@philderbeast PR welcome! Or we'll get to it sometime...

PR #2191, opened 2021-07-09
Delete 9 files with namespace blocks.

philderbeast:
See #2189. I deleted all *.u files with namespace blocks. I see that unison-src/Base.u doesn't use namespace blocks but may have in the past:

```
-- namespace Tuple where
-- namespace List where
-- namespace Search where
-- namespace Map where
```


pchiusano (approved):

pchiusano:
Thanks!

aryairani:
Wait, should we not update the files instead of deleting them?

pchiusano:
The random .u files in unison-src should probably just all be deleted or turned into things run by CI if we think they are covering something important (I don't think they are). But they are basically random text files that may or may not have once parsed and/or typechecked with some past version of Unison's syntax. :)

The poor error messages stuff should probably become a transcript where we can track regressions/improvements to the error reporting: https://github.com/unisonweb/unison/issues/2207

PR #2192, opened 2021-07-10
`{-#COMPLETE#-}` seems to have been fixed

aryairani:
Until recently we weren't able to make use of the `COMPLETE` pragma for `Reference` pattern synonyms; but now we can, apparently.

https://github.com/unisonweb/unison/blob/a166d718c01bec43bc2590b21a5daacab31f4613/unison-core/src/Unison/Reference.hs#L60-L61

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,9 +41,9 @@ jobs:
             unison-core/.stack-work
             yaks/easytest/.stack-work
             # Main cache key: commit hash. This should always result in a cache miss...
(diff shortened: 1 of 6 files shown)


aryairani:
Just kidding; why does it build on my machine but not in CI???

aryairani:
@mitchellwrosen FYI — In #1751 we weren't sure if situations could arise where caching `.stack-work` could cause an incorrect build, but I ran into one here. See the build failure on the first commit vs the second; I also needed a `stack clean` locally to build correctly.

mitchellwrosen (approved):

Issue #2158, opened 2021-06-30
Spurious ability check failure with new inference algorithm


pchiusano:
This looks like a regression, as it typechecks with M2 release:

```
ability Async t g where
  fork : '{Async t g, g} a -> t a
  await : t a -> a

Async.parMap : (a ->{Async t g, g} b) -> [a] ->{Async t g} [b]
Async.parMap f as = 
  tasks = List.map (a -> fork '(f a)) as 
  List.map await tasks
```

```
  The expression in red needs the {g12} ability, but this location only has access to these abilities: {Async t15 g12, g12}
  
      8 |   tasks = List.map (a -> fork '(f a)) as
                                         ^^^^^ 
```
If we trust the message, it seems like a simple fix - just allow the request if it’s a strict subset of what’s available?

One note is that in generating the above message, the context applied to all types involved, so the actual types during typechecking might not have been so simple.

pchiusano:
Also, the inferred type of `parMap` if I leave off the signature is a little weird too:

```Haskell
Async.parMap : (i ->{g} o) -> [i] ->{Async t {{}, g}} [o]
```

since `{{},g}` is equivalent to `{g}`, right?

```Haskell
-- ideally
Async.parMap : (i ->{Async t g, g} o) -> [i] ->{Async t g} [o]

-- not bad either, this is just cleaning up the stray `{}`
Async.parMap : (i ->{g} o) -> [i] ->{Async t g} [o]
```

PR #2197, opened 2021-07-12
Fix some ability checking regressions.

### LINKED ISSUE #2161: Typechecker regression inferring ability type parameters
This snippet typechecks with release/M2 but fails to typecheck with a build from the current trunk:

```
unique type Brother e = Larry e | Barry

ability KungFu e where
  chop : Brother e -> x

showTime: '{KungFu e} ()
showTime _ =
  match 1 with
    1 -> chop Barry

```
```
  The expression in red needs the {KungFu e11} ability, but this location only has access to the {KungFu e6} ability,
     10 |     1 -> chop Barry
```

### LINKED ISSUE #2167: Type inference regression for parameterized abilities
This doesn’t typecheck with trunk but did in M2g:

```Haskell
ability R t where
  die : () -> x
  near.impl : Nat -> Either () [Nat]

R.near n = match near.impl n with
  Left e -> die ()
  Right a -> a

R.near1 region loc = match R.near 42 with
  [loc] -> loc
  ls -> R.die ()
```

```
  I found a value  of type:  t1
  where I expected to find:  t
  
      3 | ability R t where
      4 |   die : () -> x
      5 |   near.impl : Nat -> Either () [Nat]
      6 | 
      7 | R.near n = match near.impl n with
      8 |   Left e -> die ()
      9 |   Right a -> a
  
    from right here:
  
      3 | ability R t where
```
The location is highlighting the entire body of `R.near` starting from the `match` keyword.

This one feels pretty big, it seems to prevent writing even basic combinators for any ability that has type parameters like `R` does.

dolio:
Fixes the failure part of #2158. I'm still not sure what is causing the weird inference part (It's some kind of context application thing, but my attempts to flatten that row have so far failed).

Fixes #2167.
Fixes #2161 (same as 2167 I think).

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -1699,9 +1699,21 @@ substAndDefaultWanted
 substAndDefaultWanted want ctx
   | want <- (fmap.fmap) (applyCtx ctx) want
   , want <- filter q want
-  = coalesceWanted want []
+  , repush <- filter keep ctx
+  = appendContext repush *> coalesceWanted want []
   where
-  p (Var v@TypeVar.Existential{}) = Just v
+  isExistential TypeVar.Existential{} = True
+  isExistential _ = False
+
+  -- get the free variables of things that aren't just variables
+  necessary (Type.Var' _) = mempty
+  necessary t = Set.filter isExistential $ Type.freeVars t
+
+  keeps = foldMap (necessary.snd) want
+  keep (Var v) = v `Set.member` keeps
+  keep _ = False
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
Great!

For the inference thing, it doesn't really impact anything, but you could add a call to `Type.cleanup` in the `Context.topLevelComponent` function - that should flatten out effect lists in top level type signatures that come out of typechecking.

runarorama:
Is this ready to merge?

pchiusano:
Ya, I merged but created #2208 to track cleaning up the inferred type to not include an empty nested `{}`.

PR #2226, opened 2021-07-15
Fix division and modulus for Nat

### LINKED ISSUE #2225: Division on `Nat` is wrong for large numbers
The behaviour of the built-in DIVN (and probably MODN too) is incorrect.

``` unison
> maxNat / 2
```

This results in `maxNat` again, which is obviously wrong.

runarorama:
Fixes #2225

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/MCode.hs
+++ b/parser-typechecker/src/Unison/Runtime/MCode.hs
@@ -308,6 +308,7 @@ data UPrim1
 data UPrim2
   -- integral
   = ADDI | SUBI | MULI | DIVI | MODI -- +,-,*,/,mod
+  | DIVN | MODN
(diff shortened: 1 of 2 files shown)


aryairani:
Maybe a new or updated transcript, to show them working correctly?

pchiusano:
@runarorama you can add a few quick tests in this file https://github.com/unisonweb/unison/blob/trunk/unison-src/transcripts-using-base/builtins.md#nat-functions should take you like 2 minutes.

dolio (approved):
Looks good to me.

pchiusano:
Dammit. I shoulda removed ready-to-merge - I'd probably avoid adding that label before PR review unless you really think a PR is trivial and there's no chance of any discussion about it.

@runarorama I created https://github.com/unisonweb/unison/issues/2232 to track adding those tests.

Issue #2245, opened 2021-07-16
add testing of `.unisonConfig` settings


aryairani:
I don't believe we currently have a test runner that accepts config file settings, meaning we have no tests of functionality that depends on config file settings.

Issue #2213, opened 2021-07-13
PR process revamp


aryairani:
__UPDATE:__ [latest design is here](https://github.com/unisonweb/unison/pull/2247#issuecomment-900409775)

---

No longer relevant - 

> There's some initial work in #2105, but we went down a rabbit hole looking at how we might embed PR bundles into github markdown, and never picked it back up again.

pchiusano:
The last idea on this that I recall is that PRs are encoded as namespaces with two subnamespaces, `base`, and `head`. An anonymous namespace is created for this when `pr.create` is made.

```
.> pr.create my_new_feature29 current_trunk

<maybe show diff here of what merged result will be>

Created PR bundle #a3048b120 and pushed to https://github.com/pchiusano/unisoncode.
Use:
  
    pr.load https://github.com/unisonweb/pchiusano:#a3048b120

To load this PR locally.
```

Then `pr.load` loads that namespace and can intelligently provide a list of targets for the merge, based on things which have the same base as a common ancestor.

So no weird bundle format needed, and this takes advantage of the efficient syncing of namespaces that already exists.

pchiusano:
Phase 1 of this should just be a timeboxed design jam / prototyping / implementation sketch of 2 days to decide on 1 or more possible approaches and better estimate level of effort.

aryairani:
Ok, I like this idea.

aryairani:
I closed #2105, but we might want to copy from its `InputPatterns.hs`.

aryairani:
We(Simon and I)'re thinking that we'll just be using normal branch comparison tools to inspect and compare contributor branches. We'll sort out how to compare against the PR base vs how to compare against the current `main` on share. Maybe this diff would only be on Share and not something we try to do in UCM; idk tbd.

PR #2246, opened 2021-07-16
finer-grained types for `RemoteRepo`

aryairani:
This PR is just a refactor, with no changes in functionality expected.

## Overview

Refines the model in [`RemoteRepo.hs`](https://github.com/unisonweb/unison/pull/2246/files#diff-79fe1181f9c8a3b44873791ea43e78ad64f11ee8b2238a42f3750187308c6f89) to distinguish between `Read`able namespace URIs we can pull from vs `Write`able ones we can push to, and then propagates those changes.

As a bonus, the PR eliminates a lot of dead/error code, even if it also adds code (more, distinct functions to deal with the new distinct types).

This yak came to a head while working on #2213 (pr revamp) which changes the argument types for `pr.create` and `pr.load`.

## Implementation notes

You can convert a `WriteRepo` to a `ReadRepo` which reads the root of the same repository, but you can't convert a `ReadRepo` to a `WriteRepo` (without throwing out information recklessly).

## Interesting/controversial decisions (not very interesting)

The configurator file `Path -> RemoteRepo` parser must find at most a `WriteRemotePath`, or it will complain.  

Why not the other way around, `ReadRemoteNamespace`?  It seems safe to rule out that a user wants to create a permanent URI association between a local path and an immutable remote one specified by hash, not path.  You can't push to a hash, and you shouldn't need to pull from one more than once.

I could change the type or constructor names.

## Test coverage

Existing tests seem enough for most of these changes, and I added #2245 to track that we don't have any testing for anything controlled by configurator afaik.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -16,7 +16,7 @@ import Unison.Codebase.Branch (Branch)
 import qualified Unison.Codebase.Branch as Branch
 import qualified Unison.Codebase.CodeLookup as CL
 import Unison.Codebase.Editor.Git (withStatus)
-import Unison.Codebase.Editor.RemoteRepo (RemoteNamespace, RemoteRepo)
+import Unison.Codebase.Editor.RemoteRepo (ReadRemoteNamespace, WriteRepo)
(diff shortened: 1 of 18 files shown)


pchiusano (approved):
👍

Issue #567, opened 2019-07-12
Do kind checking


aryairani:
```
foo : Optional   -- (Should be `Optional a`)
foo = None
```
yields
![image](https://user-images.githubusercontent.com/538571/61097604-f0ad1480-a429-11e9-8e36-26d102baa9bc.png)

pchiusano:
Phase 1 of this might be: come up with implementation plan and sketch it out, timeboxed.

pchiusano:
@dolio Just recording this here. 

For implementation, we realized that there's probably no need to change the codebase format or AST at all, at least for now where kinds can be fully inferred.

What we could do is have the typechecker do a kind inference pass up-front before starting typechecking. The typechecker should already have all data declarations used or referenced by the program in its environment (if not we could tweak things to ensure all transitively referenced decls are included in the environment). 

So it can just do the kind-inference on all those data decls, report errors, and then use the kind info when doing typechecking of the terms. It would be a change that's totally local to the typechecker - other than some new error messages.

ceedubs:
Some test cases that came up from [this discussion](https://unisonlanguage.slack.com/archives/CLKV43YE4/p1631197470159600). All of these _currently_ compile, but only a few of them should.

```
structural type Reader f a b = Reader (a -> {f} b)

-- should compile
x1: Reader Abort Text Nat
x1 = Reader (_ -> abort)

-- should compile and be the same as x1
x2 : Reader {Abort} Text Nat
x2 = Reader (_ -> abort)

-- should compile
x3 : Reader {Abort, .base.Exception Text} Text ('Nat)
x3 = Reader (_ -> _ -> abort)

-- shouldn't compile because `f` is expected to be an ability and `Int` is not
x4: Reader Int Text Nat
x4 = Reader (_ -> 3)

-- shouldn't compile because `Optional` is neither a rank-1 type nor an ability.
x5: Reader Optional Text Nat
x5 = Reader (_ -> 3)
```

tstat:
https://github.com/unisonweb/unison/pull/4331

PR #2259, opened 2021-07-22
Allow constructor replacement during update propagation

### LINKED ISSUE #2254: Constructor mapping done during propagation doesn’t work
In theory, there's supposed to be some heuristic where same-named constructors are associated with each other, but it doesn't actually work.

Did a little digging, and found a couple things:

* There's no logic for actually applying constructor mappings in `Term.updateDependencies`. The signature of that function needs to change to work with `Map Referent Referent` instead of `Map Reference Reference` 
* The constructor mapping isn't being properly initialized using the type replacements in the patch.

Fix is forthcoming.

pchiusano:
Fixes #2254 which should be a huge quality of life improvement. See [this transcript](https://github.com/unisonweb/unison/blob/fix/2254/unison-src/transcripts/fix2254.output.md)

This PR has gone through several phases... at one point I was convinced the approach wasn't going to work out but I came up with a nice fix.

## Overview

This PR allows constructor replacement to happen during update propagation, so the common use case of adding or removing a constructor from a data type, or adding a field to an existing constructor or record type is now fully automatic. 

This also lays the groundwork for adding constructor replacement directly to patches.

## Implementation notes

During propagation, we now maintain a `Map Referent Referent` rather than `Map Reference Reference`. And `Term.updateDependencies` now takes such a `Map`, and will rewrite pattern matches and constructor calls.

The map is initially populated by aligning same-named types and constructors, which works out nicely as long as the namespace being updated is a fork of a namespace that already exists. If updating a namespace that's not a fork, there are no old names for the type to use for establishing the mapping.

During propagation, you need to build up new constructor mappings as types are rewritten, and the approach used in this PR is robust and works great. See the comments in `Propagate.hs`. That module is still very frightening but I added more comments.

The existing logic doesn't check that type updates are well-kinded. It has actually been like this for a while, and I just left it that way. I'm okay with this for now until #567 - the programmer is responsible for knowing when it is safe to call `update` on a data type. You can also produce ill-kinded data types now, manually, via typing them in, so I'd say propagation should be allowed to do automatically whatever the programmer can currently do manually.

## Test coverage

I added a transcript to exercise this and caught lots of bugs this way, and also the existing transcripts caught some bugs.

I also did manual testing for a simple refactoring to the distributed library - I added some constructors to the `Location` type and it propagated automatically via multiple transitive dependents, including types and terms, as expected. (Before it totally did not work)

## Loose ends

- [x] gitsync22.fc.lightweightPatch test is currently failing, filed https://github.com/unisonweb/unison/issues/2265 to track figuring that out

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/Output.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Output.hs
@@ -287,7 +287,7 @@ isFailure o = case o of
   Typechecked{} -> False
   DisplayDefinitions _ _ m1 m2 -> null m1 && null m2
   DisplayRendered{} -> False
-  TodoOutput _ todo -> TO.todoScore todo /= 0 && not (TO.noConflicts todo)
+  TodoOutput _ todo -> TO.todoScore todo > 0 || not (TO.noConflicts todo)
   TestIncrementalOutputStart{} -> False
   TestIncrementalOutputEnd{} -> False
   TestResults _ _ _ _ _ fails -> not (null fails)
--- a/parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -Wno-partial-type-signatures #-}
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE PartialTypeSignatures #-}
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -144,8 +145,14 @@ propagate patch b = case validatePatch patch of
         (Set.fromList
           [ r | Referent.Ref r <- Set.toList $ Branch.deepReferents b ]
         )
-    initialDirty <-
-      R.dom <$> computeFrontier (eval . GetDependents) patch names0
+    initialDirty <- do
+      deps <- R.dom <$> computeFrontier (eval . GetDependents) patch names0
+      -- We consider edited types to be dirty in order to visit them and
+      -- establish constructor mappings. This can be removed once patches contain
+      -- constructor mappings. TODO
+      let editedTypes = R.dom (Patch._typeEdits patch)
+      pure $ deps <> editedTypes
+
     order <- sortDependentsGraph initialDirty entireBranch
     let
 
@@ -165,11 +172,8 @@ propagate patch b = case validatePatch patch of
           Reference.DerivedId _ -> go r todo
        where
         go r todo =
-          if Map.member r termEdits
-             || Map.member r typeEdits
-             || Set.member r seen
-          then
-            collectEdits es seen todo
+          if Map.member r termEdits || Set.member r seen -- or Map.member r typeEdits
+          then collectEdits es seen todo
           else
             do
               haveType <- eval $ IsType r
@@ -194,6 +198,7 @@ propagate patch b = case validatePatch patch of
                   collectEdits edits' seen' todo'
         doType :: Reference -> F m i v (Maybe (Edits v), Set Reference)
         doType r = do
+          traceM $ "Rewriting type: " <> show r
           componentMap <- unhashTypeComponent r
           let componentMap' =
                 over _2 (Decl.updateDependencies typeReplacements)
@@ -236,17 +241,17 @@ propagate patch b = case validatePatch patch of
             seen' = seen <> Set.fromList (view _1 . view _2 <$> joinedStuff)
             writeTypes =
               traverse_ (\(Reference.DerivedId id, tp) -> eval $ PutDecl id tp)
-            newCtorMappings = traceShowId $ generateConstructorMapping componentMap hashedComponents'
-            constructorMapping = constructorReplacements <> newCtorMappings
+            !newCtorMappings = traceShowId $ generateConstructorMapping componentMap hashedComponents'
+            constructorReplacements' = constructorReplacements <> newCtorMappings
           writeTypes $ Map.toList newNewTypes
           pure
             ( Just $ Edits termEdits
-                           (termReplacements <> newCtorMappings)
+                           (newCtorMappings <> termReplacements)
                            newTerms
                            typeEdits'
                            typeReplacements'
                            newTypes'
-                           constructorMapping
+                           constructorReplacements'
             , seen'
             )
         doTerm :: Reference -> F m i v (Maybe (Edits v), Set Reference)
--- a/parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
@@ -84,7 +84,7 @@ propagateAndApply rootNames patch branch = do
 -- This function produces constructor mappings for propagated type updates.
 --
 -- For instance in `type Foo = Blah Bar | Zoink Nat`, if `Bar` is updated
--- from `Blah#old` to `Blah#new`, `Foo` will be a "propagated update" and
+-- from `Bar#old` to `Bar#new`, `Foo` will be a "propagated update" and
 -- we want to map the `Foo#old.Blah` constructor to `Foo#new.Blah`.
 --
 -- The function works by aligning same-named types and same-named constructors,
@@ -125,12 +125,12 @@ propagateCtorMapping oldComponent newComponent = let
     ]
   in if debugMode then traceShow ("constructorMappings", r) r else r
 
--- TODO: This function will go away soon, once constructor mappings can be
+
+-- TODO: Use of this function will go away soon, once constructor mappings can be
 -- added directly to the patch.
 --
--- Given two data declaration components, creates a mapping from old data
--- constructors to new data constructors via the heuristic of looking
--- at constructors with matching names.
+-- Given a set of type replacements, this creates a mapping from the constructors
+-- of the old type(s) to the constructors of the new types.
 --
 -- Constructors for the same-unqualified-named type with a same-unqualified-name
 -- constructor are mapped to each other.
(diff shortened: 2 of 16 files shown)


parser-typechecker/src/Unison/Util/Star3.hs
```diff
@@ -182,17 +183,36 @@ deleteFact facts Star3{..} =
...
             => fact -> fact -> Star3 fact d1 d2 d3 -> Star3 fact d1 d2 d3
-replaceFact f f' Star3{..} =
-  let updateFact fact = 
-        if Set.member f fact
-        then (Set.insert f' . Set.delete f) fact
-        else fact
-  in Star3 (updateFact fact)
-        (R.replaceDom f f' d1)
-        (R.replaceDom f f' d2)
-        (R.replaceDom f f' d3)
```
pchiusano, line 195: This was unconditionally calling `replaceDom` on each dimension, even if the fact wasn't known. Now it's a noop if the replacement source isn't a fact.

parser-typechecker/src/Unison/Codebase/Editor/Output.hs
```diff
@@ -287,7 +287,7 @@ isFailure o = case o of
   Typechecked{} -> False
   DisplayDefinitions _ _ m1 m2 -> null m1 && null m2
   DisplayRendered{} -> False
-  TodoOutput _ todo -> TO.todoScore todo /= 0 && not (TO.noConflicts todo)
+  TodoOutput _ todo -> TO.todoScore todo > 0 || not (TO.noConflicts todo)
```
pchiusano, line 290: this was just wrong! should be `||` not `&&`. It's a failure if EITHER there's conflicts or there is more todo.

unison-src/transcripts/command-replace.output.md
```diff
@@ -55,9 +55,29 @@ Test that replace works with types
...
+  Edited Terms: #jk19sm5bf8 -> x
+  
+  Tip: To remove entries from a patch, use
+       delete.term-replacement or delete.type-replacement, as
+       appropriate.
+
 .scratch> view X
 
-  type X = Two Nat Nat
+  type X = One Nat Nat
```
pchiusano, line 80: It took me a while staring at this to understand that this is the correct output. I think this transcript example is kind of weird, in that it is replacing one type with another, but both types are under the current namespace, so they both get replaced with the target type, but with two different names for the constructors.

unison-src/transcripts/deleteReplacements.md
```diff
@@ -23,15 +23,15 @@ x = 2
 ```
 
 ```unison
-type Foo = Foo
+unique[a] type Foo = Foo
```
pchiusano, line 26: Added some of these deterministic unique types to avoid possible confusion with `()`.

pchiusano (commented):
self review

pchiusano:
@aryairani somehow this test, which checks that pushing a patch also pushes its dependencies, has started crashing -

```haskell
  pushPullTest "lightweightPatch" fmt
    (\repo -> [i|
      ```ucm
      .> builtins.merge
      ```
      ```unison
      type A = A Nat
      type B = B Int
      x = 3
      y = 4
      ```
      ```ucm
      .defns> add
      .patches> replace .defns.A .defns.B
      .patches> replace .defns.x .defns.y
      .patches> view.patch
      .patches> push ${repo}
      ```
    |])
    (\repo -> [i|
      ```ucm
      .> builtins.merge
      .> pull ${repo} patches
      .> view.patch patches.patch
      ```
    |])
```

With 

```
Missing type declaration #d97e0
  error, called at src/Unison/Codebase/Editor/Propagate.hs:424:27 in unison-parser-typechecker-0.0.0-60QGCyggypuLep7O0vZ7Pv:Unison.Codebase.Editor.Propagate
```

Is there a chance that my changes are exposing a pre-existing issue in the sync code? The old propagation code was sloppy about adding stuff to the star relations in the branch. I could vaguely see how it might create some invisible garbage that doesn't have a name, but that could still get sync'd? The new code is more precise about this and I'm somewhat more confident it isn't creating garbage in the star relations.

I know once you recycle it to disk and back that gets rid of some garbage, but if you sync the in memory branch, then what happens?

pchiusano:
@aryairani I was worried that maybe definitions created during propagation weren't being added to the codebase, but adding this line makes the test succeed, so it's definitely something with the sync algorithm. 

```
      .patches> alias.type .defns.A A
```

Kinda interesting that I don't need to alias the old term that's referenced in the patch, nor do I have to alias the new term or type that are referenced in the patch.

unison-src/transcripts/command-replace.output.md
aryairani: I don't get it.

unison-src/transcripts/deleteReplacements.md
aryairani: And then it needs a guid tag because the transcript references the type by hash below?

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -46,21 +50,19 @@ import qualified Unison.Util.Star3             as Star3
...
-  , termReplacements :: Map Reference Reference
+  , termReplacements :: Map Referent Referent
   , newTerms :: Map Reference (Term v Ann, Type v Ann)
   , typeEdits :: Map Reference TypeEdit
   , typeReplacements :: Map Reference Reference
   , newTypes :: Map Reference (Decl v Ann)
-  , constructorReplacements :: Map (Reference, Int, ConstructorType)
-                                   (Reference, Int, ConstructorType)
+  , constructorReplacements :: Map Referent Referent
   } deriving (Eq, Show)
```
aryairani, line 66: I'm so relieved that `Propagate.hs` has its own `Edits` representation and doesn't try to share `Patch` with everything else.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -132,10 +207,11 @@ generateConstructorMapping oldComponent newComponent = Map.fromList
 propagate
   :: forall m i v
    . (Applicative m, Var v)
-  => Patch
+  => Names0
+  -> Patch
   -> Branch0 m
   -> F m i v (Edits v)
-propagate patch b = case validatePatch patch of
+propagate rootNames patch b = case validatePatch patch of
```
aryairani: Could you elaborate a bit on how the `Names0` is used here, whether there are any constraints on its construction needed to ensure a useful result from this function, and maybe even consider use a simpler type like `Map` if `Names0` is overkill/immediate convenience.

I'm anticipating largely phasing `Names` out, because we're eventually going to stop trying to preload the global namespace into memory, and at that time, having as much info as possible about what data and functionality `propagate` really needs is going to be super helpful.

Is it really just a `Map (Reference, ConstructorId) NameSuffix` for the types that are... on one side of the patch or the other or both?

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -146,8 +222,41 @@ propagate patch b = case validatePatch patch of
...
+    let initialTypeReplacements = Map.mapMaybe TypeEdit.toReference initialTypeEdits
+    initialCtorMappings :: Map Referent Referent <- do
+      let mappings :: (Reference,Reference) -> _ (Map Referent Referent)
+          mappings (old,new) = do
+            old <- unhashTypeComponent old
+            new <- fmap (over _2 (either Decl.toDataDecl id)) <$> unhashTypeComponent new
+            pure $ genInitialCtorMapping @v (typeNamesMatch initialTypeReplacements)
+                                             ctorNamesMatch old new
+      Map.unions <$> traverse mappings (Map.toList initialTypeReplacements)
+
```
aryairani: Could you add a bit to the algorithmic explanation for this comment to incorporate this stuff, because I don't think it will be very obvious later.

aryairani (commented):
Left a partial review, sorry I wasn't able to dig into all of it yet!

unison-src/transcripts/command-replace.output.md
pchiusano: TL;DR: the old output was wrong / a consequence of constructor propagation not working. The new output is correct, but also, the transcript is kind of a weird example.

unison-src/transcripts/deleteReplacements.md
pchiusano: Yeah.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
pchiusano: Okay, I did some rejiggering of this. 

I left `Names0` alone as a dependency but pulled all usage of it out into a separate function with a big comment that it can be deleted once patches more natively support constructor mapping. 

I could also have passed in like, a pair of functions for looking up term and type names but that didn't seem worth it to me given that the usage of `Names0` only pertains to code that will hopefully be deleted soon anyway.

pchiusano: Also just added an explicit export list to `Propagate.hs` - it's literally one function, `propagateAndApply`.

pchiusano:
@aryairani okay, this is all set I think!

parser-typechecker/src/Unison/Util/Star3.hs
pchiusano: @aryairani I suspect but can't prove that the old behavior may have been responsible for garbage unused metadata mappings getting added to the namespaces.

aryairani: Hmm, I think the issue is elsewhere, but don't see any harm in this change.  `replaceDom` already looks like a no-op when `f` isn't a fact.

unison-src/transcripts/command-replace.output.md
aryairani: Doesn't it seem wrong or at least super confusing that the original constructor name (which doesn't appear in the replacement type) is reused, but given a new type? I'm imagining things like `None : a -> Maybe a`.  

Is this something we should track separately?

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -439,51 +483,61 @@ applyDeprecations patch = deleteDeprecatedTerms deprecatedTerms
 applyPropagate
   :: Var v => Applicative m => Patch -> Edits v -> F m i v (Branch0 m -> Branch0 m)
 applyPropagate patch Edits {..} = do
-  let termRefs = Map.mapMaybe TermEdit.toReference termEdits
-      typeRefs = Map.mapMaybe TypeEdit.toReference typeEdits
-      termTypes = Map.map (Type.toReference . snd) newTerms
+  -- let termRefs = Map.mapMaybe TermEdit.toReference termEdits
+  --     typeRefs = Map.mapMaybe TypeEdit.toReference typeEdits
```
aryairani: Kill these comments?

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -146,8 +169,26 @@ propagate patch b = case validatePatch patch of
         (Set.fromList
           [ r | Referent.Ref r <- Set.toList $ Branch.deepReferents b ]
         )
-    initialDirty <-
-      R.dom <$> computeFrontier (eval . GetDependents) patch names0
+      ns = traceShowId $ Branch.toNames0 b
```
aryairani: Leftover `traceShowId`

aryairani (approved):
I still don't really get the `type X = One Nat Nat` thing, but I don't want to hold things up.

I added a couple more comments on lines that looked like accidental checkins.

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
pchiusano: Looks like a comment on an old diff? traceShowId doesn't appear in latest version.

unison-src/transcripts/command-replace.output.md
pchiusano: I think it could be confusing but given that patches are nameless mappings it's the correct / only behavior. When you say `replace Type1 Type2`, you're mapping one hash to another (and by extension, to the constructors of that hash). Wherever the patch is applied, if names exist for the old hash it moves those names to the new hash. So this is just the same thing. Term replacement works the same way, incidentally.

So it is a feature of how patches are "nameless" and can be applied even to a namespace where the names are different than the namespace where the patch was prepared, and doesn't mess with the local names chosen within that namespace for the same definitions.

I do think sometimes you do want to update a thing, and also rename it (or perhaps suggest a renaming), but right now that's not something patches talk about directly. It could be interesting to make patches capable of doing more stuff, or I think a codebase ability in Unison would do it, so you could express more interesting codebase transformations.

aryairani: I get it for replacing types and I get  for replacing terms (though, maybe we should rethink this), but I don’t get it for replacing constructors; it still seems super wrong / random to me.  What if there’s more than one constructor? Is there any way to predict which old constructor name will correspond to which new constructor at the end?

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
aryairani: Oops, I tried to make sure I was looking at the latest but I guess I was not successful.

unison-src/transcripts/command-replace.output.md
pchiusano: > What if there’s more than one constructor? Is there any way to predict which old constructor name will correspond to which new constructor at the end?

Yes, it's just whatever the constructor mapping is that is established before propagation starts, or (when patches support this) whatever constructor mapping is in the patch. In this PR (and eventually the `update` command, when patches support this directly) the identity of the same named constructor is mapped to the identity of the new same named constructor, and then this nameless mapping can be applied to other namespaces, preserving whatever names exist for those constructors in those namespaces.

An example I was thinking about is like - you have two namespaces, one with the identifiers all in English, and another where the identifiers are in Spanish.

You can create a patch in the English namespace, then apply it to the Spanish namespace, mapping constructors that have different names but "the same role" in their respective aliases of the same type definition. So, it's not a random mapping that is like, positional or something.

Issue #2216, opened 2021-07-13
Codebase server API: `/list` should include a fully qualified name


hojberg:
When querying the `/list`API, there's a relative `namespaceListingName`, which is great, but it would be very helpful to additionally have a `namespaceListingFQN` field (perhaps based on `namespaceListingName` to pick a name if there are multiple). 

The UI currently walks the tree and constructs FQN based on `namespaceListingName` and this is a bit brittle and expects that the `/list` API is always queried from the root, which wont work for https://github.com/unisonweb/codebase-ui/issues/78

❗  Blocks: https://github.com/unisonweb/codebase-ui/issues/78

pchiusano:
Fixed by #2266

PR #2266, opened 2021-07-23
Add fully-qualified-name to namespace in list endpoint

runarorama:
This addresses #2216 by adding the qualified name of the namespace being listed. It does this by simply returning the query that was used.

Loose ends: this doesn't address the case when the namespace is listed using its hash. In that case, we don't know any of the namespace's names and finding them is currently very expensive. That will get solved by #2219


hojberg (approved):
Sweet 🎉

PR #2272, opened 2021-07-28
Fix #2261

### LINKED ISSUE #2261: Docs: `@source` and `@foldedSource` sometimes doesn't include term signatures
There seem to be an inconsistence between the rendered `SyntaxText` of `@source` and `@foldedSource`.

The API returns 2 fields (in a tuple), 1 for folded source and one for unfolded source, but in the attached screenshot, where `@source{List.map}` and `@source{example.example1}` are used, the JSON has different shapes. 

`@source{example.example1}` has the signature included in its "unfolded" field, `List.map` does not.

Resulting example (double signature for `example.example1`, single for `List.map`):

Screen Shot 2021-07-21 at 20 27 42

The shape of the "unfolded" field should be consistent across terms.

pchiusano:
After this PR, all term source embedded in docs will include a type signature, so the front-end should just toggle between the folded and unfolded source rather than treating those two docs as "summary" and "detail" and attempting to concatenate the two to produce the unfolded view. 

Manually tested.

Here's what it currently looks like:

Screen Shot 2021-07-27 at 11 45 48 PM

### DIFF
--- a/parser-typechecker/src/Unison/Server/Doc.hs
+++ b/parser-typechecker/src/Unison/Server/Doc.hs
@@ -262,9 +262,12 @@ renderDoc pped terms typeOf eval types tm = eval tm >>= \case
                   Just tm -> do
                     typ <- fromMaybe (Type.builtin() "unknown") <$> typeOf (Referent.Ref ref)
                     let name = PPE.termName ppe (Referent.Ref ref)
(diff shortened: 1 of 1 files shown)


pchiusano:
@hojberg crap I forgot that I was gonna have you shoulder surf for this fix. Next time!

hojberg (approved):

PR #2273, opened 2021-07-28
Assorted fixes

dolio:
This fixes some things I ran into when implementing Hindley-Milner in unison.

- MVar.tryRead was implemented incorrectly; it wasn't correctly decoding the `Optional` result
- Added a `Reference` to the `Unpack` instruction so that `dumpData` can report what closure type it expected
- Implemented universal equality separate from ordering, because `MVar` can't be ordered


pchiusano (approved):
Cool, though I'd add a regression test to mvar.md transcript. :)

pchiusano (approved):

PR #2278, opened 2021-07-30
WIP: add functions for converting Nats and Bytes with multiple word sizes

### LINKED ISSUE #2277: Binary encoding/decoding primitives
```haskell
Bytes.decodeNat64be : Bytes -> Optional (Nat, Bytes)
Bytes.decodeNat64le : Bytes -> Optional (Nat, Bytes)
Bytes.decodeNat32be : Bytes -> Optional (Nat, Bytes)
Bytes.decodeNat32le : Bytes -> Optional (Nat, Bytes)
Bytes.decodeNat16be : Bytes -> Optional (Nat, Bytes)
Bytes.decodeNat16le : Bytes -> Optional (Nat, Bytes)

Bytes.encodeNat64be : Nat -> Bytes
Bytes.encodeNat32be : Nat -> Bytes
Bytes.encodeNat16be : Nat -> Bytes
Bytes.encodeNat64le : Nat -> Bytes
Bytes.encodeNat32le : Nat -> Bytes
Bytes.encodeNat16le : Nat -> Bytes

-- bitwise casting from Nat <-> Float
Float.toRepresentation : Float -> Nat
Float.fromRepresentation : Nat -> Float
Int.toRepresentation : Int -> Nat
```

Design with @dolio and @stew, this should be sufficient primitives for fast binary decoding stuff.

stew:
## Overview

Fixes: #2277 

## Implementation notes

I added encode and decode functions to Bytes.hs and calling them as foreign functions

I would love for someone that knows what they are actually doing to check my encode functions on Bytes, and see if what I'm doing with a mutable block is an efficient / correct way of creating Bytes

## Test coverage

TODO


parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1750,6 +1750,23 @@ declareForeigns = do
   declareForeign "Bytes.fromBase64" boxToEBoxBox . mkForeign $ pure . Bytes.fromBase64
   declareForeign "Bytes.fromBase64UrlUnpadded" boxDirect . mkForeign $ pure . Bytes.fromBase64UrlUnpadded
 
+  declareForeign "Bytes.decodeNat64be" boxToMaybeBox . mkForeign $ pure . (\n -> let bs = Bytes.decodeNat64be n in
+                                                                           traceShow bs bs)
+  declareForeign "Bytes.decodeNat64le" boxToMaybeBox . mkForeign $ pure . Bytes.decodeNat64le
```
dolio: `boxToMaybeBox` is not going to work here.

Unison tuples are like cons lists `(a,b) = Cons a (Cons b ())`. And the way that the foreign operation is going to pass things back is like this:

  - Tag word indicating `Nothing` vs `Just`
  - In the latter case, word for the `Nat` on the unboxed stack and one closure for the `Bytes` on the boxed stack

So the wrapper code needs to read that stack convention and build the Unison values up.

parser-typechecker/src/Unison/Util/Bytes.hs
```diff
@@ -86,6 +89,159 @@ at i bs = case Unison.Util.Bytes.drop i bs of
...
+decodeNat16le :: Bytes -> Maybe (Word64, Bytes)
+decodeNat16le bs = do
+  b1 <- at 0 bs
+  b2 <- at 1 bs
+
+  let b = (shiftL (fromIntegral b2) 8) .|. (fromIntegral b1)
+  return (b, (Unison.Util.Bytes.drop 2 bs))
+
+encodeNat64be :: Word64 -> Bytes
+encodeNat64be n =
```
dolio: I think instead of implementing this manually, it'd be better to use on of the binary encoding libraries we already depend on. E.G. we use bytes which has a bunch of endian-specific functions:

    https://hackage.haskell.org/package/bytes-0.17.1/docs/Data-Bytes-Put.html

And can spit out a `ByteString` afterward.

dolio (commented):
So, something occurred to me. Tagging @pchiusano too.

If encoding is single values, like you encode one `Nat` and then concatenate, that's going to lead to pretty fragmented `Bytes` values, isn't it? The chunks will be extremely small, or if they aren't, it will be a bunch of expensive (due to copying) ByteString concatenation.

It's probably better to have something like Haskell's binary where you 'write' to an intermediate type and then finalize to get a `Bytes` with reasonable chunk sizes.

pchiusano:
> If encoding is single values, like you encode one Nat and then concatenate, that's going to lead to pretty fragmented Bytes values, isn't it? The chunks will be extremely small, or if they aren't, it will be a bunch of expensive (due to copying) ByteString concatenation.

So, the idea is that you can just call `Bytes.flatten` to convert a fragmented `Bytes` to a single-chunk one, and that function is quite efficient and does everything in one pass. You can even control the chunk size, like if you want 1024-sized chunks, you can `take 1024`, flatten, and concatenate with the remaining recursively chunked `Bytes`.

Basically, `Bytes` can function as a buffer (unlike `ByteString`). Maybe the constant factors aren't quite as good as a separate `Builder` type, but OTOH having a separate `Builder` type leads to both more plumbing code and also inefficiencies with converting back and forth at various API boundaries.

Also on that note, I wonder if going through `Put`/`Get` is overkill and possibly less efficient. Or maybe it gets optimized and is fine. But more problematic with those libraries is they work only with `ByteString`, which uses pinned memory (and this is never getting fixed it seems, due to `ByteString` also being used for FFI interchange), so to avoid needless copying we'd need to have `Bytes` also use `ByteString` for its chunk type. I think using pinned memory is the wrong choice and don't want to change `Bytes` to use that.

That said, the implementation of the decoders could be much more efficient @stew to avoid going through `Maybe` for all those bytes getting pulled out. Will leave a note on how to improve that.

dolio:
I see. I guess flattening the bytes afterwards works fine.

The stuff Stew wrote is also producing `ByteString`, though, so then that needs to be implemented using something else if you don't want to use `ByteString`.

pchiusano:
> The stuff Stew wrote is also producing ByteString, though, so then that needs to be implemented using something else if you don't want to use ByteString.

Ah yeah, good point.

parser-typechecker/src/Unison/Util/Bytes.hs
stew: @dolio so I think we are doing much better now in this respect?

pchiusano:
Great! Seems like this is pretty much there.

What's going on with the CI though?

parser-typechecker/src/Unison/Util/Bytes.hs
dolio: Yeah, that looks better. I just default to preferring something in a more widely used library if we're already using it. That way there's hopefully less question whether some fiddly detail is mistaken (because more people would have noticed by now).

pchiusano (approved):
Great!

PR #2283, opened 2021-08-02
Codebase Server: Correctly parse rootBranch param

hojberg:
## Overview
Fix a bug where the Codebase Server wasn't stripping out the `#` symbol
of the `rootBranch` param.

## Implementation notes
We were relying on `Servant` (via `FromHttpApiData`) and `Text` for `ShortBranchHash` when parsing the query param. Instead, add a manual instance of `parseUrlPiece` (used for both URI parts and Query String params) for `FromHttpApiData`.


pchiusano (approved):
We paired on this.

PR #2284, opened 2021-08-02
Type checker improvements to avoid ambiguous ability variable collections

dolio:
This PR contains some modifications to the type checker to try to avoid getting into situations where there are two possible existential ability variables to unify with in a row. When that happens, subtype checks start failing, because it isn't clear which existential variable should 'pick up the slack', and the choice matters.

The (new) type checker already avoided inferring types like `forall e. A ->{e} B`, because the `e` is unnecessary (this is the type of a 'pure' function), and the unneccessary variable leads to the above problem. However, many builtins would effectively have such types, and functions in base that were checked before the modified type checker could have such types as well. It's also possible for users to annotate types this way.

So, this patch tries to prune off such variables immediately, so that they cannot percolate through and cause problems. There are still situations that can cause the multiple-variable problem, but this common source should be eliminated at least.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker.hs
+++ b/parser-typechecker/src/Unison/Typechecker.hs
@@ -16,7 +16,6 @@ import           Control.Monad.State        (State, StateT, execState, get,
                                              modify)
 import           Control.Monad.Writer
 import qualified Data.Map                   as Map
-import qualified Data.Set                   as Set
 import qualified Data.Sequence.NonEmpty     as NESeq (toSeq)
 import qualified Data.Text                  as Text
 import qualified Unison.ABT                 as ABT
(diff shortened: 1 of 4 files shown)


pchiusano (approved):
Great! Looking forward to testing this out.

pchiusano:
@dolio Question: the `TypePrinter.hs` currently has a call to `Type.removePureEffects` 

```Haskell
pretty0 n im p tp = prettyRaw n im p (cleanup (removePureEffects tp))
```

But I think that could be replaced with a call to your function that strips covariant ability variables.

dolio:
It could probably work, but it'd have a slight mismatch I think. My function needs you to tell it which variables to consider dropping, because it only removes things you're about to generalize over (or in the case of this PR, ones that just got instantiated from generalization).

PR #2286, opened 2021-08-03
Make U.R.ANF.Serialize independent of V1

dolio:
This should make it so that removing the V1 codebase isn't blocked on serialization of code for runtime values.


pchiusano (approved):

aryairani:
Thanks @dolio!

PR #2290, opened 2021-08-04
Support suffix-based names lookup directly, and pick more reasonable default names when multiple exist

### LINKED ISSUE #2133: Pretty-printer should prefer names whose FQN has fewest number of segments
Right now, we don't really have any way of prioritizing what names are chosen by the pretty-printer. The current algorithm is deterministic but basically never what you want / it's just accidental implementation-defined (see example below). We've talked about having some separate name prioritization scheme and maybe that's a good idea but feels maybe overkill and it would need more design. 

This is a simple proposal that can be implemented in a few lines of code today. It also wouldn't preclude having something fancier for prioritizing names later.

The thing I want is for:

````
zoink = 230498234
ooga = zoink + 1209
````
```
.> alias.term zoink zoink.aliases.aaaa
.> view ooga
```

To use the name `zoink` (whose FQN has fewer segments) rather than `aaaa` (whose FQN has more segments but whose unique suffix comes alphabetically first). Currently it picks `aaaa` because the unique name has 1 segment and it comes first alphabetically.

The proposed algorithm is to have the pretty-printer prefer names whose fully-qualified name has the fewest number of name segments (like `foo.bar.baz` has 3 segments). If there's multiple names with the the same number of segments, it can pick alphabetically.

Rationale for this change is to just have a predictable and simple algorithm for picking names which is easy to think about and lets users prioritize names. Another way of looking at it is that we're wasting an opportunity with the current algorithm, which is pretty random and not useful IMO. Just having something that's non-random and useful is a step up.

For instance, with this change, we can start including aliases for a definition `foo` under `foo.aliases`. It also works when someone includes a copy of `base` in `external` of their library and maybe changes some of the names - these won’t be chosen either (though archived namespaces will probably help with this also - I think `_external` should be archived).

My view is that “pick which comes first alphabetically” should only be done if there is no other information to make a decision on.

One way of doing this with our current name printing is tweaking `PrettyPrintEnv.fromNames`. It already picks the shortest name (in terms of number of segments), it can be modified to take the shortest name (according to the number of segments of the longest name each name is a suffix of, or 1 if the name is not a suffix of any name other than itself).

So in the above example, `aaaa` will have a length of 3,  because `aaaa` is a suffix of the name `zoink.aliases.aaaa` which has length 3. Whereas `zoink` will have a length of 1 because it is not a suffix of any other name.

Aside: If the representation of names were more like `(Path, UniqueSuffix)` then this would be simpler - it would just be sorting based on the combined segment length of Path + Suffix. Right now we’re throwing away information by just associating this bag of names with a hash, forgetting that some of those names were formed by taking a suffix of an existing name.

cc @runarorama @hojberg @aryairani

### PR DESCRIPTION
**Pull request by pchiusano**
This PR simplifies and improves the efficiency of suffix-based name resolution. It also improves definition loading speed, since the definition can just directly use the names index which is cached in the `Branch` itself.

It also fixes https://github.com/unisonweb/unison/issues/2133 and paves the way for doing something fancier to prioritize chosen names, since the names index is left alone rather than "baking in" the suffixes.

The new name selection algorithm used by the pretty-printer is explained [in this transcript](https://github.com/unisonweb/unison/blob/d67b99095a559a141bf37965f93037bb198fdddc/unison-src/transcripts/name-selection.output.md).

## Overview

Currently, resolving a suffix like `List.map` involves first transforming the names index via a process called "suffixification", where all the suffixes of a FQN like `base.List.map` which still refer to the same definition are also added as aliases to the names index. This is inefficient but also loses track of what the original names were. It's also more complicated because we now have two different kinds of names indexes, one which has been suffixed, and the other which hasn't and we have to keep track of which to use where.

This PR alters the names index to directly support suffixed lookup.

## Implementation notes

This PR changes the `Ord Name` instance to order from last segment to first. So `base.List.map` will get compared as `[map,List,base]`. As a result, all things that end in `map` will be adjacent to each other in the index, and can be located in logarithmic time. The main new function on `Relation` is `searchDom` (and `searchRan`). I added some unit tests for these.

This change caught a bunch of places where names were being sorted for display using the `Ord Name` even though that isn't really correct and we should be using some unicode collation algorithm like https://www.rfc-editor.org/rfc/rfc5051.txt.

I was debating how to catch all the places where we were sorting `HashQualified` by their bytes rather than alphabetically, and went with introducing a new class `Alphabetical`, and having the `HashQualified` types require that for the name parameter. Thus, `HashQualified` by default sorts alphabetically, which is what you want for display purposes and it's also a perfectly fine total order. 

## Test coverage

The existing transcripts caught a lot of bugs. I added some unit tests and caught some bugs that way, and I also added a transcript testing the name selection behavior. I feel pretty good about the coverage.

## Loose ends

- [x] https://github.com/unisonweb/unison/pull/2290#discussion_r682264346
- [x] Add a transcript for testing the name selection algorithm in the pretty-printer, just showing that it's no longer arbitrary and now prefers FQNs with fewer segments.

### CODE REVIEW

File: parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs [hunk 1]
```diff
@@ -283,7 +283,7 @@ loop = do
             ParseErrors text [ err | Result.Parsing err <- toList notes ]
           Just (Left errNames) -> do
             ns <- makeShadowedPrintNamesFromHQ hqs errNames
-            ppe <- prettyPrintEnv (Names3.suffixify ns)
+            ppe <- suffixifiedPPE ns
```
pchiusano, line 287: Rather than creating a `PrettyPrintEnv` from a suffixified `Names`, you instead can just directly create a suffixified `PrettyPrintEnv` from a regular `Names`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -636,9 +636,7 @@ loop = do
           case uf of
             Nothing -> do
               let parseNames0 = (`Names3.Names` mempty) basicPrettyPrintNames0
-                  -- use suffixed names for resolving the argument to display
-                  parseNames = Names3.suffixify parseNames0
-                  results = Names3.lookupHQTerm hq parseNames
+                  results = Names3.lookupHQTerm hq parseNames0
```
pchiusano, line 641: `lookupHQTerm` does suffix-based lookup now, so no need to suffixify the `Names` passed in.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1004,44 +1002,18 @@ loop = do
...
+            types = Names3.lookupHQType thing ns
+            printNames = Names basicPrettyPrintNames0 mempty
+            terms' :: Set (Referent, Set (HQ'.HashQualified Name))
+            terms' = Set.map go terms where
+              go r = (r, Names3.termName hqLength r printNames)
             types' :: Set (Reference, Set (HQ'.HashQualified Name))
-            types' = (`Set.map` Names.typeReferences filtered) $
-                        \r -> (r, Names3.typeName hqLength r printNames)
+            types' = Set.map go types where
+              go r = (r, Names3.typeName hqLength r printNames)
```
pchiusano, line 1017: Simplified this implementation to use `lookupHQTerm/Type`.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1004,44 +1002,18 @@ loop = do
...
---          where
---            results p namePortion = let
---              name = Path.toName . Path.unprefix currentPath' . Path.snoc' p
---                   $ namePortion
---              ns = prettyPrintNames0
---              terms = [ (r, Names.namesForReferent ns r)
---                      | r <- toList $ Names.termsNamed ns name ]
---              types = [ (r, Names.namesForReference ns r)
---                      | r <- toList $ Names.typesNamed ns name ]
---              in (terms, types)
```
pchiusano, line 1046: Deleted some commented out code that's no longer relevant.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1544,7 +1516,7 @@ loop = do
         names <- makePrintNamesFromLabeled' $
           LD.referents testTerms <>
           LD.referents [ DD.okConstructorReferent, DD.failConstructorReferent ]
-        ppe <- prettyPrintEnv names
+        ppe <- fqnPPE names
```
pchiusano, line 1521: This is a spot where we want to see fully qualified names - the test result report, we want the test names to be their full path.

parser-typechecker/src/Unison/FileParser.hs
```diff
@@ -52,9 +52,9 @@ file = do
   --
   -- There's some more complicated logic below to have suffix-based name resolution
   -- make use of _terms_ from the local file.
-  local (\e -> e { names = Names.push (Names.suffixify0 locals) namesStart }) $ do
+  local (\e -> e { names = Names.push locals namesStart }) $ do
     names <- asks names
-    stanzas0 <- local (\e -> e { names = names }) $ sepBy semi stanza
```
pchiusano, line 57: This `local` call wasn't doing anything! Since it is just setting the `names` it just `ask`d.

parser-typechecker/src/Unison/FileParser.hs
```diff
@@ -52,9 +52,9 @@ file = do
   --
   -- There's some more complicated logic below to have suffix-based name resolution
   -- make use of _terms_ from the local file.
-  local (\e -> e { names = Names.push (Names.suffixify0 locals) namesStart }) $ do
+  local (\e -> e { names = Names.push locals namesStart }) $ do
```
pchiusano, line 55: `Names.push` just takes a regular `Names0` now, not a suffixed one.

parser-typechecker/src/Unison/FileParser.hs
```diff
@@ -68,7 +68,9 @@ file = do
     let (terms, watches) = (reverse termsr, reverse watchesr)
     -- suffixified local term bindings shadow any same-named thing from the outer codebase scope
     -- example: `foo.bar` in local file scope will shadow `foo.bar` and `bar` in codebase scope
-    let (curNames, resolveLocals) = (Names.deleteTerms0 locals (Names.currentNames names), resolveLocals)
+    let (curNames, resolveLocals) =
+          ( Names.shadowSuffixedTerms0 locals (Names.currentNames names)
+          , resolveLocals )
```
pchiusano, line 73: Renamed `Names3.deleteTerms0` to `shadowSuffixedTerms0`.

This name resolution stuff was very subtle so I tweaked some stuff like this to improve clarity a bit.

parser-typechecker/src/Unison/Parser.hs
```diff
@@ -244,8 +244,7 @@ run' p s name env =
             then L.lexer name (trace (L.debugLex''' "lexer receives" s) s)
             else L.lexer name s
       pTraced = traceRemainingTokens "parser receives" *> p
-      env' = env { names = Names.suffixify (names env) }
-  in runParserT pTraced name (Input lex) env'
+  in runParserT pTraced name (Input lex) env
```
pchiusano, line 247: Parser no longer needs a suffixed names since the names index can directly resolve unique suffixes.

parser-typechecker/src/Unison/PrettyPrintEnv.hs
```diff
@@ -35,12 +34,14 @@ instance Show PrettyPrintEnv where
 
 fromNames :: Int -> Names -> PrettyPrintEnv
 fromNames len names = PrettyPrintEnv terms' types' where
-  terms' r = shortestName . Set.map HQ'.toHQ $ Names.termName len r names
-  types' r = shortestName . Set.map HQ'.toHQ $ Names.typeName len r names
+  terms' r = shortestName . Set.map Name.convert $ Names.termName len r names
+  types' r = shortestName . Set.map Name.convert $ Names.typeName len r names
```
pchiusano, line 38: Gratuitous change, but trying to move away from the special purpose conversion functions I can never remember. Just use `Name.convert` or `Name.parse`, and add a type annotation at the call site if necessary for clarity.

parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -303,7 +303,7 @@ unique[b28d929d0a73d2c18eac86341a3bb9399f8550c11b5f35eabb2751e6803ccc20] type
   IsPropagated = IsPropagated
 
 d1 Doc.++ d2 =
-  use Doc
+  use Doc2
```
pchiusano, line 306: The original `use Doc` had no effect, although no `use` is needed here since there is only one `Join` constructor in the file.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -444,15 +444,14 @@ searchBranchExact len names queries =
...
   :: Monad m
-  => Bool
-  -> Maybe Path
+  => Maybe Path
   -> Branch m
   -> Codebase m v Ann
   -> [HQ.HashQualified Name]
   -> m QueryResult
-hqNameQuery' doSuffixify relativeTo root codebase hqs = do
+hqNameQuery relativeTo root codebase hqs = do
```
pchiusano, line 454: Removed this `doSuffixify` parameter, but @runarorama I don't totally understand what its purpose was - was it just an efficiency thing, to avoid doing suffixification in some cases? 

It doesn't seem harmful to do lookup by suffix.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -496,24 +494,6 @@ hqNameQuery' doSuffixify relativeTo root codebase hqs = do
...
-hqNameQuery = hqNameQuery' False
-
-hqNameQuerySuffixify
-  :: Monad m
-  => Maybe Path
-  -> Branch m
-  -> Codebase m v Ann
-  -> [HQ.HashQualified Name]
-  -> m QueryResult
-hqNameQuerySuffixify = hqNameQuery' True
```
pchiusano, line 515: Consolidated these into one `hqNameQuery` function.

unison-core/src/Unison/Names3.hs
```diff
@@ -91,6 +78,17 @@ push n1 ns = Names (unionLeft0 n1 cur) (oldNames ns <> shadowed) where
...
+  -- of that name [[foo.bar.baz], [bar.baz], [baz]]. Any suffix which uniquely
+  -- refers to a single definition is added as an alias
+  suffixify0 :: Names0 -> Names0
+  suffixify0 ns = ns <> suffixNs
+    where
+    suffixNs = names0 (R.fromList uniqueTerms) (R.fromList uniqueTypes)
+    terms = List.multimap [ (n,ref) | (n0,ref) <- R.toList (terms0 ns), n <- Name.suffixes n0 ]
+    types = List.multimap [ (n,ref) | (n0,ref) <- R.toList (types0 ns), n <- Name.suffixes n0 ]
+    uniqueTerms = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList terms ]
+    uniqueTypes = [ (n,ref) | (n, nubOrd -> [ref]) <- Map.toList types ]
```
pchiusano, line 100: This function `push` was the one spot where we still use suffixification, so it's in a where clause of that function now. I think the names index could be more like a list of indexes which are consulted in turn (kind of like old names, but that also controls whether the identifier is hash qualified or not), and then this function `push` could have a direct implementation without doing suffixification.

unison-core/src/Unison/Type.hs
```diff
@@ -68,9 +67,10 @@ bindNames
   -> Names.Names0
   -> Type v a
   -> Names.ResolutionResult v a (Type v a)
-bindNames keepFree ns t = let
+bindNames keepFree ns0 t = let
+  ns = Names.Names ns0 mempty
   fvs = ABT.freeVarOccurrences keepFree t
-  rs = [(v, a, R.lookupDom (Name.fromVar v) (Names.types0 ns)) | (v,a) <- fvs ]
+  rs = [(v, a, Names.lookupHQType (Name.convert $ Name.fromVar v) ns) | (v,a) <- fvs ]
```
pchiusano, line 73: Generally `lookupDom` on a names relation is going to be wrong and will only find exact matches. I caught a bunch of places that were doing that and relying implicitly on the names being suffixified.

unison-src/transcripts-using-base/fix2027.output.md
```diff
@@ -90,10 +90,7 @@ myServer = unsafeRun! '(hello "127.0.0.1" "0")
   I've encountered a call to builtin.bug with the following
   value:
   
-    builtin.io2.Failure.Failure
-      (typeLink builtin.io2.IOFailure)
-      "problem"
-      !builtin.Any.Any
+    Failure (typeLink IOFailure) "problem" !Any
```
pchiusano, line 93: @dolio bonus! this got fixed for free

unison-src/transcripts-using-base/hashing.output.md
```diff
@@ -104,15 +104,15 @@ And here's the full API:
...
-  9.  hash : HashAlgorithm -> a -> Bytes
+  1.  hash : HashAlgorithm -> a -> Bytes
+  2.  builtin type HashAlgorithm
+  3.  HashAlgorithm.Blake2b_256 : HashAlgorithm
+  4.  HashAlgorithm.Blake2b_512 : HashAlgorithm
+  5.  HashAlgorithm.Blake2s_256 : HashAlgorithm
+  6.  HashAlgorithm.Sha2_256 : HashAlgorithm
+  7.  HashAlgorithm.Sha2_512 : HashAlgorithm
+  8.  HashAlgorithm.Sha3_256 : HashAlgorithm
+  9.  HashAlgorithm.Sha3_512 : HashAlgorithm
```
pchiusano, line 127: Just different sorting here.

unison-src/transcripts-using-base/utf8.output.md
```diff
@@ -5,8 +5,8 @@ Unison has function for converting between `Text` and a UTF-8 `Bytes` encoding o
 ```ucm
 .> find Utf8
 
-  1. Text.fromUtf8 : Bytes ->{Exception} Text
-  2. builtin.Text.toUtf8 : Text -> Bytes
+  1. builtin.Text.toUtf8 : Text -> Bytes
+  2. Text.fromUtf8 : Bytes ->{Exception} Text
```
pchiusano, line 9: Just different sorting. Honestly not sure what the old sorting was here... was it reversed or what?

unison-src/transcripts/ambiguous-metadata.output.md
```diff
@@ -32,7 +32,7 @@ x = 1
   there are multiple matches:
   
     foo.doc
-    boo.doc#tj3gfqdnje
+    doc#tj3gfqdnje
```
pchiusano, line 35: Not sure why the old code added the extra segment here, or if we want to preserve that behavior... it seems fine?

unison-src/transcripts/builtins.md
```diff
@@ -69,7 +76,7 @@ test> Int.tests.conversions =
         toText +10 == "10",
         toText -1039 == "-1039",
         fromText "+0" == Some +0,
-        fromText "a8f9djasdlfkj" == None,
+        Int.fromText "a8f9djasdlfkj" == None,
```
pchiusano: This is a regression with wildcard `use` clauses, looks like. Will investigate.

pchiusano (commented):
Self review

unison-src/transcripts/builtins.md
pchiusano: Fixed and re-added the test.

pchiusano:
Okay, addressed all outstanding todos that I know about, so I think this is ready for review!

pchiusano:
Also @hojberg and @runarorama good news: this improves the fuzzy-find performance, it seems. It appears that a chunk of the runtime was just producing suffixed names for picking term names and rendering their type signatures:

https://user-images.githubusercontent.com/11074/128216558-b6442eca-29cf-4c38-a080-67de8d21f7b6.mov

I don't quite understand why sometimes it's returning results almost instantly and other times not.

unison-src/transcripts-using-base/utf8.output.md
aryairani: Weird, idk

parser-typechecker/src/Unison/PrettyPrintEnv.hs
aryairani: I don’t think I’ve learned this yet and may need reminders in the future!

unison-core/src/Unison/Name.hs
```diff
@@ -120,6 +134,13 @@ stripNamePrefix prefix name =
   where
   mid = if toText prefix == "." then "" else "."
 
+-- suffixFrom Int builtin.Int.+ ==> Int.+
+-- suffixFrom Int Int.negate    ==> Int.negate
+suffixFrom :: Name -> Name -> Maybe Name
```
aryairani, line 144: What’s a concrete example of when to use this?

unison-core/src/Unison/Name.hs
```diff
@@ -178,9 +199,73 @@ fromSegment = unsafeFromText . NameSegment.toText
 segments :: Name -> [NameSegment]
 segments (Name n) = NameSegment <$> segments' n
 
+reverseSegments :: Name -> [NameSegment]
+reverseSegments (Name n) = NameSegment <$> NameSegment.reverseSegments' n
+
 countSegments :: Name -> Int
 countSegments n = length (segments n)
 
+-- The `Ord` instance for `Name` considers the segments of the name
```
aryairani, line 213: Should we use a newtype around Name for this Ord logic that’s kind of surprising?  Like, we have an application that’s more efficient with this new Ord, but wouldn’t we quickly find a second application that this Ord is not good for?

unison-core/src/Unison/Name.hs
```diff
@@ -190,6 +275,8 @@ class Parse a b where
 instance Convert Name Text where convert = toText
 instance Convert Name [NameSegment] where convert = segments
 instance Convert NameSegment Name where convert = fromSegment
+instance Convert [NameSegment] Name where
```
aryairani, line 283: God, should we have a Convert from String to Text and back? Is there already one?

unison-core/src/Unison/Names3.hs
aryairani: Go on? 
Although the wrong lines seem to be selected here somehow

unison-core/src/Unison/Names3.hs
```diff
@@ -192,6 +216,30 @@ termName length r Names{..} =
...
+  ( suffixedName termName (Names.terms . currentNames) HQ'.fromNamedReferent
+  , suffixedName typeName (Names.types . currentNames) HQ'.fromNamedReference )
+  where
+  suffixedName fallback getRel hq' length r ns@(getRel -> rel) =
+    if R.memberRan r rel
+    then go $ toList (R.lookupRan r rel)
+    else sort $ map Name.convert $ Set.toList (fallback length r ns)
+    where
+      -- Orders names, using these criteria, in this order:
+      -- 1. NameOnly comes before HashQualfied,
```
aryairani: typo HashQualfied

unison-core/src/Unison/Names3.hs
```diff
@@ -251,11 +299,24 @@ expandWildcardImport prefix ns =
...
+    suffix <- Name.stripNamePrefix prefix rem
+    -- suffix = negate
+    pure (suffix, full)
+
+-- Deletes from the `n0 : Names0` any definitions whose names
+-- share a suffix with a name in `ns`. Does so using logarithmic
+-- time lookups, traversing only `ns`.
+--
+-- See usage in `FileParser` for handling precendence of symbol
+-- resolution where local names are preferred to codebase names.
```
aryairani, line 325: nice

aryairani (approved):
Awesome! I didn’t try to verify any asymptotics though.

unison-core/src/Unison/Name.hs
pchiusano: Yeah, I considered that, but then, I think that the purpose of an `Ord` instance is to be computationally useful. The current byte ordering isn't useful for any purpose we have right now - not for indexing, and not for display-level alphabetical sorting (byte ordering is wrong for that as well). So I don't think there's a good argument for it being the default. The current ordering is just the first ordering we happened to think of given the representation as a newtype around `Text`. :)

IMO we have been too influenced by that rather arbitrary representation choice - if we have thought of suffix-based resolution and search as being important things for names, we might have represented it as a snoc list of segments, and then that representation's "natural" ordering would actually coincide with what this PR is doing. I think it could be cool to try swapping out the representation (in a separate PR), but I didn't try for that here.

That said, we can definitely revisit what the default is if/when we have other competing orderings for `Name` that are useful! :)

unison-core/src/Unison/Name.hs
pchiusano: Ha... maybe. I haven't added one. I have been pretty conservative about adding these.

unison-core/src/Unison/Names3.hs
pchiusano: Fixed.

unison-core/src/Unison/Name.hs
pchiusano: Added comment on this - it's currently used as an implementation detail of `expandWildcardImports`, but it seemed general enough to include here.

unison-core/src/Unison/Name.hs
pchiusano: Maybe another POV here is that what an `Ord` instance buys you over just a hashing function is the ability to do meaningful range queries, and the current `Ord` isn't letting us do any range query we care about - it lets us do prefix matching based on the byte order of the name, which is not very useful.

PR #2292, opened 2021-08-04
Case insensitive substring matching used as pre-filter to fzf

pchiusano:
Right now, the fuzzy find implementation can take several seconds to return results (though sometimes it is fast...) This PR adds a case-insensitive substring match as a pre-filter before consulting the fuzzy-find implementation. This returns fewer results but returns instantly, as it uses `Text.isInfixOf` which is highly optimized.

The stuff that matches the filter is still passed to fuzzy find to deal with alignment and ranking.

I'm not sure if this PR should be merged, but including some videos for comparison. 

The hope is that the pre-filter could be removed later once there's a more optimized implementation. 

One thing that seems like low-hanging fruit @runarorama is that the implementation currently converts all the names to strings before passing that to the FZF implementation. Then I'm guessing that FZF copies that into some unboxed arrays before doing its thing. Could it instead go directly from `Text` into the unboxed arrays used by the implementation?

__Before:__

https://user-images.githubusercontent.com/11074/128216558-b6442eca-29cf-4c38-a080-67de8d21f7b6.mov

__After:__

https://user-images.githubusercontent.com/11074/128217397-a53164f6-fd89-4218-b824-7257d2821405.mov

## Other ideas

I was thinking - is there some other _conservative_ pre-filter we could do? Like if `fast q name` returns `False`, then there is no way `q` can be a subsequence of `name`?

One really simple one - does `name` contain all the characters of `q`?

Are there others?

### DIFF
--- a/unison-core/src/Unison/Names2.hs
+++ b/unison-core/src/Unison/Names2.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE PatternSynonyms     #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ViewPatterns        #-}
 
 module Unison.Names2
   ( Names0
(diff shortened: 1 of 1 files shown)


aryairani:
@hojberg If this speeds things up like crazy, should we back out the UI workarounds like debouncing and min. query length?

hojberg:
@aryairani I suspect that on Share we'd continue to need those.

hojberg (approved):

PR #2299, opened 2021-08-06
Disallow floating point patterns

dolio:
This PR addresses #2914

It disallows floating point pattern matching, based on the rationale that exact equality is poorly behaved for floating point numbers. The error message if you write a floating point pattern suggests testing that the scrutinized value is within a certain acceptable bound of the target.


pchiusano (approved):
LGTM

pchiusano:
@dolio I can't find the issue associated with this, #2914 is not a thing yet :)

pchiusano:
You can close the ticket once you find it though.

Issue #2215, opened 2021-07-13
The codebase server `/list` API returns the wrong hash when queried with a name


hojberg:
Querying a namespace with the list API and a FQN returns the codebase hash instead of the hash of the queried namespace (the children are correct):

`GET http://127.0.0.1:8080/asdf/api/list`

returns 

```js
{
  "namespaceListingName": ".",
  "namespaceListingHash": "#qm8rg9p4ba0hd8haabpqnjdhr2b0td0565os0a3m02gru41vvvr0allb9i14hsna2pta95u9niqbdkabfbtqf37ocr96vh6fe0nobso",
  "namespaceListingChildren": [ /* ... abbreviated ...  */ ]
}
```

`GET http://127.0.0.1:8080/asdf/api/list?namespace=.base.List`

returns

```js
{
  "namespaceListingName": ".base.List",
  "namespaceListingHash": "#qm8rg9p4ba0hd8haabpqnjdhr2b0td0565os0a3m02gru41vvvr0allb9i14hsna2pta95u9niqbdkabfbtqf37ocr96vh6fe0nobso",
  "namespaceListingChildren": [ /* ... abbreviated ...  */ ]
}
```

Notice that the `namespaceListingHash` are the same in both examples (it's the codebase hash)

hojberg:
The RICE score is based on the `/list` API being used over https://github.com/unisonweb/unison/issues/2172 in the initial version of https://github.com/unisonweb/codebase-ui/issues/78

PR #2301, opened 2021-08-06
List Endpoint: Add rootBranch & relativeTo params

### LINKED ISSUE #2275: Codebase Server API: `/list` should support a `rootBranch` and a `relativeTo`
The `/list` API should support the `rootBranch` and `relativeTo` params like the `/find` and `/getDefinition` API.

hojberg:
## Overview
Add 2 new query params to the namespace listing `/list` API:
  * `rootBranch` - a hash of the root, typically the codebase hash
  * `relativeTo` - a namespace fqn that the listing should be relative
    to.

This enables URLs that are focused into a specific namespace. For
instance if we were focused on `base.List`:

Fetch all entries in `base.List`:

  `/list?rootBranch=#mycodebasehash&relativeTo=base.List`

Fetch all `Nonempty` entries in `base.List`:

  `/list?rootBranch=#mycodebasehash&relativeTo=base.List&namespace=Nonempty`

Fetch all `Nonempty.append` entries in `base.List`:
  `/list?rootBranch=#mycodebasehash&relativeTo=base.List&namespace=Nonempty.append`

This also fixes a minor bug where the NamespaceListing were including
the root hash as the `namespaceListingHash`. This is now correctly the
hash of the returned NamespaceListing.

## Notes
* Paired with @aryairani for most of this.
* Tested with the Codebase UI
* Closes: https://github.com/unisonweb/unison/issues/2275 and https://github.com/unisonweb/unison/issues/2215

### DIFF
--- a/parser-typechecker/src/Unison/Server/Endpoints/ListNamespace.hs
+++ b/parser-typechecker/src/Unison/Server/Endpoints/ListNamespace.hs
@@ -15,8 +15,6 @@ import Data.OpenApi (ToSchema)
 import qualified Data.Text as Text
 import Servant
   ( QueryParam,
-    ServerError (errBody),
-    err400,
     throwError,
     (:>),
   )
@@ -34,9 +32,8 @@ import qualified Unison.Codebase.Branch as Branch
 import qualified Unison.Codebase.Causal as Causal
 import qualified Unison.Codebase.Path as Path
 import qualified Unison.Codebase.ShortBranchHash as SBH
+import Unison.Codebase.ShortBranchHash (ShortBranchHash)
 import qualified Unison.Hash as Hash
-import qualified Unison.HashQualified as HQ
-import qualified Unison.Name as Name
 import qualified Unison.NameSegment as NameSegment
 import Unison.Parser (Ann)
 import Unison.Prelude
@@ -44,7 +41,6 @@ import qualified Unison.PrettyPrintEnv as PPE
 import qualified Unison.Server.Backend as Backend
 import Unison.Server.Errors
   ( backendError,
-    badHQN,
     badNamespace,
     rootBranchError,
   )
@@ -52,6 +48,7 @@ import Unison.Server.Types
   ( APIGet,
     APIHeaders,
     HashQualifiedName,
+    NamespaceFQN,
     NamedTerm (..),
     NamedType (..),
     Size,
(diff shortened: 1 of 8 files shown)


parser-typechecker/src/Unison/Server/Endpoints/ListNamespace.hs
```diff
@@ -80,16 +80,14 @@ instance ToSample NamespaceListing where
     [ ( "When no value is provided for `namespace`, the root namespace `.` is "
         <> "listed by default"
       , NamespaceListing
-        (Just ".")
-        (Just ".")
+        "."
```
hojberg, line 83: I removed the maybe constraint. It seems like that is appropriate, but i'm not sure?

parser-typechecker/src/Unison/Server/Types.hs
```diff
@@ -54,6 +54,8 @@ type APIGet c = Get '[JSON] (APIHeaders c)
 
 type HashQualifiedName = Text
 
+type NamespaceFQN = Text
```
hojberg, line 57: Used this over `HashQualifiedName` for namespace params. In the future, I bet we could get these parsed by Servant.

parser-typechecker/src/Unison/Server/Errors.hs
```diff
@@ -64,8 +67,14 @@ noSuchNamespace :: HashQualifiedName -> ServerError
...
 
+couldntLoadBranch :: Branch.Hash -> ServerError
+couldntLoadBranch h =
+  err404 { errBody = "The namespace "
+    <> munge (Text.toStrict . Text.pack $ show h)
+    <> " exists but couldn't be loaded." }
+
 ambiguousNamespace :: HashQualifiedName -> Set HashQualifiedName -> ServerError
-ambiguousNamespace name namespaces = err400
+ambiguousNamespace name namespaces = err409
```
hojberg, line 77: Slight HTTP Status code update.

parser-typechecker/src/Unison/Server/Endpoints/ListNamespace.hs
hojberg: We also weren't using both `namespaceListingName` and `namespaceListingFqn`. Removed the former for simplicity.

aryairani: I guess a "name"space listing would have to have a name, not just `Maybe` have a name, so it makes sense to me.

aryairani (approved):
My questions are: a) does it work for you? and b) does it need tests?

hojberg:
@aryairani I tested it against the codebase ui and its working great 🎉 

I'm not sure how best to add test coverage for this :/

aryairani:
@hojberg I opened a separate ticket for tests in #2304.

PR #2303, opened 2021-08-07
Add single argument form of diff.namespace

### LINKED ISSUE #1531: simplify diff.namespace with current namespace as "after"
## motivation

Almost every time that I run `diff.namespace before after` I want to use the current namespace as `after`. I think that it would be nice to have a concise way to do this.

possibly related: #1515 

## proposal 1: default `after` to current namespace

This would be similar to `git` diffs. If you run `.prs._myPR> diff.namespace ._base.trunk` it would be equivalent to `.prs._myPR> diff.namespace ._base.trunk .prs._myPR`.

### pros

* familiar to people with git experience
* probably usually the thing that you want
* aligns with `merge` command behavior

### cons

* probably causes a mess if it's _not_ what you want
* default arguments can be confusing and error-prone (if I only provide one arg, is it `before` or `after`?)
* git maybe isn't the best reference for how to make an intuitive UI
* doesn't generalize beyond `diff.namespace` (see proposal 2)

## proposal 2: keyword for current namespace

If there were a shortcut for the current namespace, one could run `prs._myPR> diff.namespace ._base.trunk <shortcut>`.

Since `ucm` is a command-line tool, and `cd .prs._myPR` can be used to switch to the `.prs._myPR` namespace, someone with experience in Unix/Linux could reasonably expect `.` to be a reference to the current namespace, and that `prs._myPR> diff.namespace ._base.trunk .` would do what they wanted. Unfortunately, in Unison `.` refers to the root namespace.

### pros

* could be generally useful for other commands, such as `pull`
* avoids confusion with default arguments

### cons

* Finding the right shortcut could be tricky. It should be short but not conflict with `.` or namespace names that might reasonably be used.
* magic nature of this shortcut could throw off beginners

sullyj3:
## Overview

You can now do

`.> diff.namespace before`

to diff `before` with the current namespace.

Closes #1531

## Implementation notes

Modifies the appropriate `InputPattern`

None, fairly straightforward change

## Test coverage

Added a line to `diff.md` to check for same output as two argument form.

## Loose ends

none.

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
+++ b/parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
@@ -926,11 +926,15 @@ diffNamespace :: InputPattern
 diffNamespace = InputPattern
   "diff.namespace"
   []
-  [(Required, pathArg), (Required, pathArg)]
+  [(Required, pathArg), (Optional, pathArg)]
   (P.column2
     [ ( "`diff.namespace before after`"
       , P.wrap
         "shows how the namespace `after` differs from the namespace `before`"
+      )
+    , ( "`diff.namespace before`"
+      , P.wrap
+        "shows how the namespace `before` differs from the current namespace"
       )
     ]
   )
@@ -939,6 +943,9 @@ diffNamespace = InputPattern
       before <- Path.parsePath' before
       after <- Path.parsePath' after
       pure $ Input.DiffNamespaceI before after
+    [before] -> first fromString $ do
+      before <- Path.parsePath' before
+      pure $ Input.DiffNamespaceI before Path.currentPath
     _ -> Left $ I.help diffNamespace
   )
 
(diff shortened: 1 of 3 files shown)


parser-typechecker/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -926,19 +926,26 @@ diffNamespace :: InputPattern
...
-  [(Required, pathArg), (Required, pathArg)]
+  [(Required, pathArg), (Optional, pathArg)]
   (P.column2
     [ ( "`diff.namespace before after`"
       , P.wrap
         "shows how the namespace `after` differs from the namespace `before`"
       )
+    , ( "`diff.namespace before`"
+      , P.wrap
+        "shows how the namespace `before` differs from the current namespace"
```
aryairani: for symmetry with the previous example, let's say "shows how the current namespace differs from the namespace `before`"

sullyj3: You're right, that makes more sense

sullyj3:
Ok I'm happy with this, let me know if I need to add any more tests/other forms of testing.

pchiusano:
Great, thanks so much for knocking this out. I have definitely missed this.

pchiusano (approved):

sullyj3:
No problem ☺️

PR #2323, opened 2021-08-14
gzip and zlib compression builtins

pchiusano:
Adds a few builtins:

```Haskell
Bytes.zlib.compress : Bytes -> Bytes
Bytes.zlib.decompress : Bytes -> Either Text Bytes
Bytes.gzip.compress : Bytes -> Bytes
Bytes.gzip.decompress : Bytes -> Either Text Bytes
```

Request for comment - 

Things I wondered about:

* Should we expose more compression options?
* The error handling, considered just returning an `Optional Bytes`
* Okay to depend on https://hackage.haskell.org/package/zlib which binds to the C library. Will this generally be available everywhere?
* Streaming... but I lean toward this being out of scope, possibly addressed with separate builtins


aryairani:
> * Should we expose more compression options?

Someone probably will want to eventually. What's the immediate use case that motivated these?  I think the answer depends on that.

> * The error handling, considered just returning an `Optional Bytes`

We should have a planning session to nail down how we want to error handling across all builtins & base, and then come back and revise whatever needs revising.  We could do that before or after merging this.

> * Okay to depend on https://hackage.haskell.org/package/zlib which binds to the C library. Will this generally be available everywhere?

I would be impressed if we found a relevant platform without a zlib.

> * Streaming... but I lean toward this being out of scope, possibly addressed with separate builtins

Yeah, definitely definitely need streaming versions if it's meant for non-toy stuff; even for implementing general-purpose HTTP.  If we need these to get some simple early product out the door though, it seems fine and good! 

A future non-streaming version could be implemented in terms of the streaming versions; but I am guessing we don't have a great system for expressing builtins with more interesting types like we'd want for that.

parser-typechecker/src/Unison/Util/Bytes.hs
```diff
@@ -35,12 +38,27 @@ empty = Bytes mempty
...
+
+gzipCompress :: Bytes -> Bytes
+gzipCompress = fromLazyByteString . GZip.compress . toLazyByteString
+
+gzipDecompress :: Bytes -> Bytes
+gzipDecompress = fromLazyByteString . GZip.decompress . toLazyByteString
+
+zlibDecompress :: Bytes -> Bytes
+zlibDecompress = fromLazyByteString . Zlib.decompress . toLazyByteString
+
```
aryairani, line 52: Consider putting these into `Unison.Util.Bytes.Compress` or something, though I don't feel super strongly in this case.

aryairani (approved):

pchiusano:
Motivating use case - wanted to produce a tiny self contained compiled bytecode file (I think in lieu of standalone binary), which you can do pretty easily in pure Unison by serializing a term and all its transitive dependencies. But then I'd like to compress this since there's a lot of duplicate hashes and it's also just generally highly compressible I suspect.

I also suspect we will want to tweak the value serialization to not emit a hash more than once, so then there are no duplicate hashes to begin with. Like it emits a hash once, and then all subsequent references for that hash just backreference its earlier definition via a byte offset or something. Even with that though, it'll be highly compressible. 

For the streaming thing, I'd be tempted to do

```Haskell
builtin type Compressor
Compressor.gzip : Settings -> Compressor
Compressor.feed : Compressor -> Bytes -> (Compressor, Bytes)
Compressor.done : Compressor -> Bytes
```

(similar for decompression) 

But `Compressor` isn't necessarily a thing that can be serialized so I think this might be too simple, and you want `Compressor` to live in some `ST`-like ability.

This also came up for the hashing functions and I bailed on trying to come up with a streaming version for those.

cc @dolio

aryairani:
> Motivating use case - wanted to produce a tiny self contained compiled bytecode file (I think in lieu of standalone binary), which you can do pretty easily in pure Unison by serializing a term and all its transitive dependencies. But then I'd like to compress this since there's a lot of duplicate hashes and it's also just generally highly compressible I suspect.

Ah, sure. So if it's basically just for us, then we don't need to expose more compression options.

> But `Compressor` isn't necessarily a thing that can be serialized 

Why not?

> This also came up for the hashing functions and I bailed on trying to come up with a streaming version for those.

I'm no expert on streaming hashing, but isn't it pretty common?  I guess I haven't seen the non-streaming hashing functions.

pchiusano:
I'm not sure what I want to do with this PR yet. Since these are pure functions, I want to make sure they're well-specified in case we end up moving to a different runtime than the Haskell one.

pchiusano:
Okay, decided to just leave these as is for now, under the assumption that we can always later add streaming support and/or more options as separate builtins.

Issue #2140, opened 2021-06-21
something fishy with find by type / type-mention


aryairani:
`find : Nonempty a -> a` doesn't seem to find `Nonempty.last` (or anything else).
`Nonempty.last` is `#4s2e6fpcpb`

In my v1 codebase, I see these hashes in the type-mentions-index for `Nonempty.last`:
`% find ~/.unison/v1/type-mentions-index/ | grep 4s2e6fpcpb`
```
#gku1kqleg9n619vfvu1iuig4fe3f937o3nbaa0dq0man4tjhg07idfadst0e5pufnnbh5ckun4rhl0f3ob2fnodimgm4sel20oau2qo
#9f58iojnfqnkf14ifh966q052vv9u8f3ov58icl53qre0jtn2m05fdfhasnjllvv7qv599ctbt188uc2b7d8bp4em1ds4t9lhvnkuso
#kh9ar9iiaa3sn503l5rqmo7josbjvpe9qfh2nn5qpqdtg0df9un3bn59dkhdfkhphcddril2b0d4g4gjs8eqipejrmco16dj4ta2vog
#mlcfeh2294igm228dlq3u9bn0i0c42q8fc700f6c57tpqntn7omoml9liepcvkq3v9v79n824d9t0rp63otl6f86purvlv08gu56ih0
#94a25njontg9htj2q8elv0jajublc92q5c6t3mrj0p17if0j5lfc6nnp7jgr4km0vjm0v2rucb63dvjnpeltj53jhs4m63mpnqp42mo
```

In my v2 codebase I see these hashes in the type-mentions-index for `Nonempty.last`:
```sql
select out.base32 
from find_type_mentions_index, hash_object, hash, hash as out 
where hash.base32 like '4s2e6fpcpb%' 
  and hash.id = hash_id 
  and object_id=term_referent_object_id 
  and type_reference_hash_id = out.id;
```

```
4ej7f14gj2761smju0pdk9b1v1aflunmjtej8ek562rrd9dlkvr9ak2q0p28tf2m37003jq9a2sffdavpgcv9uiinncdruf7qps2l3g
6hvd1hivr57ok2ftti8carcai9eag2dunrb3r5k84nef2e6vcjjosnqdsogqfjhebh8hbc7tsmq4jc2gmi0bl1q0ke06cud2e8utkbo
[...]

PR #2324, opened 2021-08-14
Start of fix for confused find-by-type (issue #2140)

puffnfresh:
There's inconsistency between hashing for V1 and V2 versions of the
codebase. The incoming search reference is constructed from the V1
version type, while the V2 types are indexed. This means we couldn't
find results for things like:

    find : [a] -> Optional a

This commit fixes the above but it's both messy and probably not
correct. It's also missing changes in various other places which have
this same issue (e.g. the "docs" command)

## Overview

## Implementation notes

## Test coverage

## Loose ends

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -23,7 +23,9 @@ import Unison.Codebase.Editor.RemoteRepo (ReadRemoteNamespace, WriteRepo)
 import Unison.Codebase.GitError (GitError)
 import Unison.Codebase.Patch (Patch)
 import qualified Unison.Codebase.Reflog as Reflog
+import qualified U.Codebase.Reference as V2
 import Unison.Codebase.ShortBranchHash (ShortBranchHash)
+import qualified Unison.Codebase.SqliteCodebase.Conversions as Cv
 import Unison.Codebase.SyncMode (SyncMode)
 import Unison.DataDeclaration (Decl)
 import qualified Unison.DataDeclaration as DD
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -365,13 +368,12 @@ dependents c r
     . Set.map Reference.DerivedId
   <$> dependentsImpl c r
 
-termsOfType :: (Var v, Functor m) => Codebase m v a -> Type v a -> m (Set Referent.Referent)
-termsOfType c ty =
+termsOfType :: (Var v, Monad m) => Codebase m v a -> Type v a -> m (Set Referent.Referent)
+termsOfType c ty = do
+  r <- Cv.reference2to1 (const (pure 0)) (TypeUtil.toReference (Cv.type1to2' Cv.reference1to2 ty))
```
puffnfresh, line 373: What's the size function here? Is it important to look up an actual size?

parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -23,7 +23,9 @@ import Unison.Codebase.Editor.RemoteRepo (ReadRemoteNamespace, WriteRepo)
 import Unison.Codebase.GitError (GitError)
 import Unison.Codebase.Patch (Patch)
 import qualified Unison.Codebase.Reflog as Reflog
+import qualified U.Codebase.Reference as V2
 import Unison.Codebase.ShortBranchHash (ShortBranchHash)
+import qualified Unison.Codebase.SqliteCodebase.Conversions as Cv
```
puffnfresh, line 28: Bad idea to depend on the Sqlite conversion module. Seems like it was designed to mostly go one way.

I kinda do like the idea of converting Codebase to require V2 types but it's a big change.

aryairani:
A more comprehensive refactor + fix is in the works!

PR #2330, opened 2021-08-16
Improve the error message for pattern match failures

dolio:
Used to just tell you that a pattern match failure happened somewhere. Now it says something like this:

```
  I've encountered a pattern match failure in function `f` while
  scrutinizing:

    None

  This happens when calling a function that doesn't handle all possible inputs
  I'm sorry this message doesn't have more detail about the location of the
  failure. My makers plan to fix this in a future release. 😢
```

The `f` is gotten from the `PrettyPrintEnv` for the top-level reference being compiled.


pchiusano:
Excellent! A nice quality of life improvement.

pchiusano (approved):

Issue #2150, opened 2021-06-25
Feature Proposal: A `ui` command in `ucm`


hojberg:
It would be really cool to have a `ui` command in `ucm`. Here's how I imagine it working:

* `.> ui`, opens the UI in the browser if there's a UI folder that `ucm` knows about (set through `--ui` or `UCM_WEB_UI`).
* `.> ui install` to install and perhaps also update? (could be a separate command) the UI. Which version of the UI that gets downloaded could be specified and/or determined by a mechanism to tag which version of ucm the ui can run with (per @aryairani's suggestion).
* `.> ui` without a UI folder present, `ucm` would tell the user (instead of attempting to open the URL) and ask if the user would want to download the UI, allowing them to choose between `latest` and the current release (`M2g` for instance).
* `.> ui List.map` would work similar to `display List.map` except that it would open `List.map` in the UI—in the future this could open directly in an Electron app which would allow us to support multiple calls to `ui <definition>` to open in the same window instead of in a new browser tab every time.
* `.> ui List.map List.filter` would open both `List.map` and `List.filter` in the UI.
* `.> ui .base` would open the UI zoomed into the on the `base` namespace (relative to)

aryairani:
wrt downloading the UI, could we have a way of keeping track of the newest ui that is supported by the user's version of ucm?  (this involves tweaking our git tags to produce a `git describe` version number that is meaningful in sequential, which is possible and common, though not obvious)

And if we have a way to automatically download the ui, we should have a way of automatically updating the ui to the latest supported.

hojberg:
Right! `.> ui update` and `.> ui install` and `.> ui install unstable` for instance. I like it—added it to the list.

ceedubs:
I imagine that this would come for "free", but just to throw it out there:

It would be nice if `ui 3` worked like `view 3` does (the 3rd result from the previous command).

hojberg:
@ceedubs nice!

hojberg:
Closing for now.

PR #2331, opened 2021-08-18
Add a ucm `ui` command to open the Codebase UI

hojberg:
## Overview
Add a command, `ui`, that opens the Codebase UI with the current
codebase server URL.

Part of https://github.com/unisonweb/unison/issues/2150

It's likely that there could be a better separation of concerns here. The way the URL is moved throughout the various functions is a bit annoying. Any suggestions?

## Test coverage
Not sure how we normally do tests for commands, or even how to best test something that has an external side effect like this?

## Loose ends
I definitely want to support giving this command arguments for opening a definition or namespace, but plan to tackle that in a subsequent effort.

### DIFF
--- a/parser-typechecker/src/Unison/Server/CodebaseServer.hs
+++ b/parser-typechecker/src/Unison/Server/CodebaseServer.hs
@@ -10,7 +10,9 @@ module Unison.Server.CodebaseServer where
 
 import Control.Concurrent (newEmptyMVar, putMVar, readMVar)
 import Control.Concurrent.Async (race)
+import Data.ByteString.Char8 (unpack)
 import Control.Exception (ErrorCall (..), throwIO)
+import qualified Network.URI.Encode as URI
 import Control.Lens ((&), (.~))
 import Data.Aeson ()
 import qualified Data.ByteString as Strict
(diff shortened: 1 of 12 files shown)


parser-typechecker/src/Unison/Server/CodebaseServer.hs
```diff
@@ -113,6 +115,24 @@ type AuthedServerAPI = ("static" :> Raw) :<|> (Capture "token" Text :> ServerAPI
 instance ToSample Char where
   toSamples _ = singleSample 'x'
 
+-- BaseUrl and helpers
+
+data BaseUrl = BaseUrl { urlHost :: String
+                       , urlToken :: Strict.ByteString
+                       , urlPort :: Port
+                       }
```
aryairani: We want to start shifting to using `ormolu` for formatting, specifically for new or changed code.  Can you set it up in your editor and "format region" here, and possibly other modified sections (specifically, the `ui` `InputPattern`).  I wouldn't absolutely hold up the PR for it, but we do want to start switching and gotta start somewhere.  

OTOH, I also wouldn't reformat the entire file on a feature PR because it will make it hard to review the diff, unless it's a newly created file where there's nothing to diff against.

aryairani (approved):

parser-typechecker/src/Unison/Server/CodebaseServer.hs
hojberg: Reformatted with Ormolu.

aryairani: It looks like we missed the `ui` `InputPattern` but we'll get it eventually.

hojberg: @aryairani whats the impact of missing that?

aryairani: @hojberg I meant it was just another region of code to auto-format, the impact is low.

PR #2336, opened 2021-08-18
add toRepresentation / fromRepresentation for Float Int

### LINKED ISSUE #2313: Bitwise casting from/to Int, Nat, Float
We could use these
```
-- bitwise casting from Nat <-> Float <-> Int
Float.toRepresentation : Float -> Nat
Float.fromRepresentation : Nat -> Float
Int.toRepresentation : Int -> Nat
```

stew:
The new functions do bitwise coersion to and from Nat

fixes: #2313


dolio (approved):
This looks good to me.

pchiusano:
@stew I think you just need to `stack exec transcripts` locally to refresh. Some of the transcript output depends on the set of builtins.

Issue #2242, opened 2021-07-16
Always show ability variables if they appear in invariant or contravariant position


pchiusano:
Example:
``` 
f : ‘a -> Nat
f x = 42
```
Unison will infer that as `forall e a . ‘{e} a -> Nat` but then elide the `’{e}` from the signature on display.

I’m imagining writing:

`removeCovariantAbilities : Type v a -> Type v a` can strip out ability variables in covariant position, and then the term printer should call this instead of whatever it’s doing now.

runarorama:
Is this resolved? Please close if so.

pchiusano:
Ya

PR #2343, opened 2021-08-19
Only elide covariant variables in `removePureEffects`

dolio:
Addresses #2242


pchiusano:
Cool. Transcript test? :)

pchiusano:
Or I guess just a transcript that will show the output, which we can inspect manually.

pchiusano (approved):

Issue #2341, opened 2021-08-19
Need either `unsafeCoerce` or some more restricted version


pchiusano:
This has come up several times. Currently it’s blocking work on a local interpreter of `Remote`.

dolio:
This is easy to do, but what version do you actually need? The version with type:

    forall a b. a -> b

will cause subtle problems unless it's actually the case that `a` and `b` are identical. If the desire is for reinterpreting bits, those need to be written for specific types to actually work correctly.

pchiusano:
Right. The version I need for now is just:

```Haskell
unsafeCoerceAbilities : forall a g g2 b . (a ->{g} b) -> (a ->{g2} b)
```

Which doesn't seem too unsafe... at worst it leads to uncaught requests. Maybe we add this for now and consider other kinds of coercing for later? WDYT.

dolio:
That can only be the general one in behavior, I think. I just needed to know which one to implement, and there should probably be caveats documented that using it for `Nat -> Int` isn't actually going to work correctly, because you'll get an `Int` that still has information specifying that it's a `Nat`.

I guess you could restrict the type to be somewhat less unsafe, but people might eventually want the general one for something.

pchiusano:
I think it's likely someone will want `unsafeCoerce` at some point, but I'd be inclined to hold off on adding the fully general version for now.

dolio:
Okay.

PR #2346, opened 2021-08-20
Add unsafe.coerceAbilities builtin

dolio:
Addresses #2341 

I think this is going to need a merge back from base because of the negative effect variable thing not being merged before I forked, but we'll see.


pchiusano:
Cool. Yeah, you can just merge in trunk and then refresh the transcripts to resolve the conflicts. I can see that the output type is wrongly eliding some ability variables in your branch.

pchiusano (approved):

Issue #2340, opened 2021-08-19
`Scope` ability and local mutable references


pchiusano:
Allow local mutation in pure code, with an `ST`-like interface to prevent it from being observed outside some scope.

Needs some more design, @dolio we should jam on this a bit, but a while ago I started sketching this, and it comes out quite nice using abilities -

https://github.com/unisonweb/unison/issues/798#issuecomment-702487151

We’d need `Scope.Ref` (aka `STRef`) as well.

This is useful for the distributed API, for local pure testing, but it’s also just generally useful for other stuff.

pchiusano:
Latest design: 

```Haskell
builtin unique ability Scope t 
builtin type Ref g a = ...
Ref.get : Ref g a ->{g} a
Ref.set : a -> Ref g a ->{g} ()
Scope.ref : a -> Ref {Scope t} a
IO.ref : a -> Ref {IO} a
```

pchiusano:
Fixed by #2351

PR #2351, opened 2021-08-24
Mutable references and locally scoped mutation

dolio:
This PR adds mutable references and local scoping of them.

There is a `Ref` type that is tagged with the type of its contents and the ability it works in. It has `read` and `write` functions.

There is a `Scope s` ability for local mutation that is externally pure.

There is a `Scope.ref` function that yields a new locally scoped reference and an `IO.ref` function that yields a reference usable with the `IO` ability.

Addresses #2340. I made the API more consistent with e.g. the `TVar` API, though (naming and argument order wise).


pchiusano (approved):
LGTM! Thanks for quick turnaround. 

I'm not sure what's going on with CI but maybe merge with trunk and re-push. #2349 just upgraded LTS version we are using.

pchiusano:
@dolio Looks like transcripts need another refresh. Like maybe you ran them at one point but then made some more changes.

Issue #2251, opened 2021-07-20
Make unique types the default or make structural/unique required


pchiusano:
It's very common to leave off `unique` for types that really should be. Suggest just making `structural` a keyword and have `type Foo` mean `unique type Foo` and `structural type Foo` mean the structural type. 

Kinda goes along with #2196

pchiusano:
Steps:

- [x] Change in `Lexer.hs` to add `structural` as a keyword, audit of the lexer to see if that's going to mess anything up.
- [x] `Parser.hs`, add a new error type for missing keyword
- [x] `FileParser.hs`, modify that to look for `unique` keyword
- [x] `PrintError.hs`, modify to render the new parse error
- [x] `DeclPrinter.hs`, modify to insert the keyword as needed
- [x] Update all the transcripts! Idea:
  - [x] replace `unique type` with `basldklfjsdflkjsdf93`
  - [x] replace `type` with `structural type`
  - [x] replace `basldklfjsdflkjsdf93` with `unique type`

hojberg:
This was fixed in #2251 🎉

PR #2352, opened 2021-08-24
Adds Structural Keyword #2251

rlmark:
## Overview

When defining types and abilities the user will now have to specify unique vs structural as a keyword - written in tandem with @hojberg. 🥳 
 
Addresses #2251 

### Potential risks 

If a user has defined a term or namespace called 'structural' they will need to update that before being able to use this change. Existing structural/unique types will not be impacted but will render with the structural keyword. 

## Implementation notes

We used the existing unique keyword code patterns and added 'structural' as a keyword. We removed the optional nature of this keyword. Changes to transcripts, tests, and .u files in the haskell codebase are numerous. 

n/a

## Test coverage

We have included a transcript which covers this and have also manually tested round-trip experience. 

## Loose ends

We discovered a `demo` folder that included a bunch of unison files that should probably be deleted.


pchiusano:
@rlmark cool. Can you link to the transcript that exercises this specifically? Having a hard time locating it given the number of transcripts that were updated. :) 

I'd like to take a look at the error message if you forget to supply the modifier. If you don't have an example of that, you can tag a transcript block as `unison:error` when a parse or type error is expected.

rlmark:
> @rlmark cool. Can you link to the transcript that exercises this specifically? Having a hard time locating it given the number of transcripts that were updated. :)
> 
> I'd like to take a look at the error message if you forget to supply the modifier. If you don't have an example of that, you can tag a transcript block as `unison:error` when a parse or type error is expected.

@pchiusano, the diff is: 
https://github.com/unisonweb/unison/blob/b2cd7153d341606121447310482623357be5797a/unison-src/transcripts/type-modifier-required.md

and the file is called `unison-src/transcripts/type-modifier-required.md`

pchiusano (approved):
Great!

PR #2356, opened 2021-08-26
Call `minimize'` underneath each destructuring bind

### LINKED ISSUE #2344: confusion in code using a mix of abilities
Tried to minimize this as much as possible. I'm suprised that `nate` needs to be a tuple, but it I can't seem to eliminate that. In the following code, `grumpy` does compile, but `sneezy` and `doc` both complain about the ultimate term in the expression

```
unique ability Nate where
  nate: (Boolean, Nat)
  antiNate: ()

grumpy: (Nat -> {d}()) -> '{d, Nate}()
grumpy dee _ =
  antiNate
  dee 1
  (_, _) = nate
  ()

sneezy: (Nat -> {d}()) -> '{d, Nate}()
sneezy dee _ =
  (_, _) = nate
  antiNate
  dee 1

doc: (Nat -> {d}()) -> '{d, Nate}()
doc dee _ =
  (_, _) = nate
  dee 1
  antiNate
```

The error for `sneezy`:
```
  The expression in red needs the {d36} ability, but this location does not have access to any abilities.
  
     17 |   dee 1
```

The error for `doc`:
```
  The expression in red needs the {Nate} ability, but this location does not have access to any abilities.
  
     17 |   antiNate

```

dolio:
Destructuring binds do not actually produce/add to `let rec` terms from the parser, so they were causing the final `minimize'` call to only operate on a subset of bindings including a destructuring bind. If the destructuring bind was first, that subset was empty. This would cause the other bindings to be unnecessarily recursive, and have unexpected ability limitations (and possibly other issues that went unnoticed).

Fixes #2344

### DIFF
--- a/parser-typechecker/src/Unison/TermParser.hs
+++ b/parser-typechecker/src/Unison/TermParser.hs
@@ -11,6 +11,7 @@ module Unison.TermParser where
 import Unison.Prelude
 
 import           Control.Monad.Reader (asks, local)
+import           Data.Foldable (foldrM)
 import           Prelude hiding (and, or, seq)
 import           Unison.Name (Name)
 import           Unison.Names3 (Names)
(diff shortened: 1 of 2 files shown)


pchiusano (approved):
Ah. That makes sense.

Issue #2355, opened 2021-08-26
Typechecker hang in this example


pchiusano:
Here’s a self-contained example that hangs the typechecker. Adding a type signature on the inner function `go` causes it to typecheck fine.

```
ability A t g where 
  fork : '{g, A t g} a -> t a
  await : t a -> a
  empty! : t a
  put : a -> t a -> ()

example : '{A t {}} Nat
example = 'let
  r = A.empty!
  -- adding this signature makes it not loop
  -- go : Nat ->{A t {}} x
  go u = 
    t = A.fork '(go (u + 1))
    A.await t
  
  go 0
  t2 = A.fork '(A.put 10 r)
  A.await r
```

dolio:
Somehow the typechecker is deciding that the `go u = A.fork '(go (u+1))` sort of thing means that we should do a refinement `e = A t e`, and there's no occurs check on that part. Not sure exactly why yet. I think it might have to do with the stand alone `g` in `fork` getting associated with effect inference variable for `go`.

PR #2364, opened 2021-08-30
Avoid a loop caused by inferring cyclic abilities

dolio:
Gives an error with a new error message in that case.

Addresses #2355


pchiusano:
LGTM, you just need to `stack exec transcripts -- fix2355.md` again, looks like you forgot to refresh after tweaking the error message - https://github.com/unisonweb/unison/pull/2364/checks?check_run_id=3466549405

pchiusano (approved):

Issue #2348, opened 2021-08-23
UCM onboarding improvements


pchiusano:
- [x] UCM should just go ahead and create a codebase in `~/` if none exists, rather than requiring this `init` step. Can instead add a `--no-init` flag or something to disable this functionality.
- [x] UCM should just go ahead and pull the correct version of base on codebase initialization. Can add a `--no-base` flag or something to disable this functionality.
- [x] Make author/license setup painless, this probably needs design

stew:
perhaps include #2326 with this effort

rlmark:
Proposal for another task/story: 
When running `init` a user most likely wants to start programming in Unison immediately, currently `init` drops the user back into the terminal to start the codebase in a second command. Would it be an onboarding improvement to change the behavior of `init` to open the UCM? Would that be a non-trivial amount of work to implement?

rlmark:
I'm gonna close this for now! Hopefully onboarding is a bit easier for new codebases.

PR #2363, opened 2021-08-30
remove FileCodebase and some weeds

aryairani:
In addition to removing v1 `FileCodebase` support, this PR decouples many module dependencies that I felt were making refactors difficult.

Apart from a couple of tweaked output messages, there shouldn't be any functionality changes here.

This will merge-conflict pretty hard with #2365 unfortunately, but I'm happy to help with reconciling them.

**Things the PR did**
- pulled these into their own modules:
    - Codebase (types) into `Codebase.Type`; `Codebase` module re-exports them
    - `CodebasePath`
    - `Codebase.Branch.Merge`, 
    - `Codebase.Branch.Names` (Branch shouldn't depend on `Names`)
    - `Codebase.BuiltinAnnotation`
    - `Codebase.Causal.FoldHistory`
    - `Codebase.CodeLookup.Util` (remove `CodeLookup` dependency on `UnisonFile`)
    - `Codebase.DataDeclaration.ConstructorId` trying to use this alias in relevant places
    - `Codebase.Init`
    - `CodebasePath`, the `FilePath` alias.
    - `Lexer.Pos` (because `AnnotatedText`, `Range`, `Parser.Ann` shouldn't depend on the whole lexer)
    - `Names.ResolutionResult` (`Type` shouldn't depend on `Names`)
    - `PrettyPrintEnv.FQN` (contains `elideFQN` logic)
    - `PrettyPrintEnv.Names` (pulled out references to `Names` from `PPE`)
    - `PPE.Util` not 100% sure what's happening here
    - `PrettyPrintEnvDecl` pull this data type and supporting functions into separate module
    - `Path.Parse` (`Path` shouldn't depend on the lexer)
    - `Path.Convert`, `Path.Parse` into `Unison.Util.Convert`
    - `Parser.Ann` (`Codebase`, etc. etc. shouldn't depend on `Parser`)
    - `Reference.Util` (`Reference` shouldn't depend on `ABT`)
    - `Referent'` (`SyntaxText`/`ColorText` shouldn't depend on `Reference`)
        - `Referent` module re-exports stuff from `Referent'`, hard-coded to `Reference`.
    - removed `SyntaxText.SyntaxText`
    - `TermEdit.Typing` (because `TermEdit` shouldn't depend on the full typechecker)
    - `UnisonFile` / `TypecheckedUnisonFile` (types into `UnisonFile.Type`)
    - `UnisonFile.Env`
    - `UnisonFile.Error`
    - `UnisonFile.Names`
[...]

PR #2365, opened 2021-08-30
Inits codebase if default does not exist

rlmark:
Addresses #2348 
Tagging @hojberg - coauthor

## Overview

When a user downloads the UCM and no default codebase is found, it should automatically create the codebase, as opposed to prompting the user to initialize it. 

This PR also redirects the user of a v1 codebase to use an older version of UCM to upgrade (as part of the deprecation of v1) 

## Test coverage

If this direction is ok, we'll add some tests based on the `Unison.Test.Ucm`


parser-typechecker/src/Unison/Codebase/Init.hs
```diff
@@ -31,10 +32,44 @@ data Init m v a = Init
...
+                  pure (CreatedCodebase resolvedDir cb)
+            )
+        Just specifiedDir -> do
+          ifM (FCC.codebaseExists specifiedDir)
+            (pure (Error specifiedDir FoundV1Codebase))
+            (pure (Error specifiedDir NoCodebaseFoundAtSpecifiedDir))
+
 createCodebase :: MonadIO m => Init m v a -> DebugName -> CodebasePath -> m (Either Pretty (m (), Codebase m v a))
-createCodebase debugName cbInit path = do
+createCodebase cbInit debugName path = do
```
hojberg, line 92: The types of all of these lined up, but the variable names were inverted

pchiusano:
Seems good to me - I'd just add some tests like you say. 

Also we just merged https://github.com/unisonweb/unison/pull/2363 which likely conflicts with some changes you made in `Main` but @aryairani can help you with resolving those.

@rlmark Also don't forget to add yourself to CONTRIBUTORS.markdown :)

pchiusano (approved):
Great!

PR #2367, opened 2021-08-31
update feature/compress

aryairani:
So dumb, I merged in trunk as well, so now it looks like a billion changes. 😞


aryairani:
I think this is obsolete.

Issue #2172, opened 2021-07-02
Codebase Server API: `/namespace` — namespace details


hojberg:
To support nice URLs in the Codebase UI—both local and on Share—we need to be able to ask the codebase server the hash of a namespace name as well the hash of the whole codebase when given a special argument.

Additionally we want to support readme's and metadata features for rendering them nicely in the UI.

Completing this ticket will support and unblock these UI tickets:
  * https://github.com/unisonweb/codebase-ui/issues/78
  * https://github.com/unisonweb/unison-local-ui/issues/3
  * https://github.com/unisonweb/codebase-ui/issues/153
  * https://github.com/unisonweb/codebase-ui/issues/144

### Suggested API

* Namespace details: `/namespace?namespace=.base`
* Namespace details: `/namespace?namespace=#asdiojw213dq2`—querying details by hash should also be possible
* Codebase details: `/namespace`—without a namespace, it returns namespace details for the whole codebase
  * Alternatively this could be  `/namespace?namespace=.`
 
### Suggested Response Schema

```elm
-- Main response

type NamespaceResponse = {
  name: String, -- ".base",
  hash: String, -- "#asdiojw213dq2",
  readme: Maybe Doc, -- Same structure as definition docs, a tree of doc elements, could be a nullable field instead of maybe since this is JSON
  metadata: {
    numAuthors: Nat,
    numCopyrightHolders: Nat
    numLicenses: Nat
    createdAt: String, -- ISO 8601 formatted date time 
    updatedAt: String -- ISO 8601 formatted date time
  }
}
```

Alternatively, instead of counts of metadata (counts were used since the UI shows a list of counts in the sidebar before the user clicks through), it could be a list of hashes that could be queried separately.

https://github.com/unisonweb/unison/issues/2172 was factored out of this, but could easily be merged back into here — the major consideration was about how long it takes to compute all metadata from a whole namespace subtree.

❗This is dependent on https://github.com/unisonweb/unison/issues/2205

hojberg:
After talking to @runarorama we realized we can get the core of this data from the `/list` endpoint already and we can use this for now. We'll move the metadata requesting to 2 separate endpoints: 1 for metadata about the codebase (like default author etc) and 1 for metadata about a namespace that is an aggregate of all linked objects within.

Here's the 2 tickets: 

https://github.com/unisonweb/unison/issues/2186
https://github.com/unisonweb/unison/issues/2185

hojberg:
[...]

PR #2368, opened 2021-08-31
Add new server endpoint: /namespaces/:fqn

hojberg:
## Overview
Add a new server endpoint for fetching the details of a namespace:
`/namespaces/:namespaceFQN`, can be used like so `/namespaces/base.List`.

This is served by the `NamespaceDetails` API which looks for a readme
(any capitalization) in the namespace and renders it.

This address part of https://github.com/unisonweb/unison/issues/2172.

## Implementation notes
The readme is looked up shallowly in the namespace branch and if found, is rendered.

Additionally I cleaned up a few things with regards to naming of the
previous `ListNamespace` module to match what it was returning. Down the
line the 2 APIs might be merged into a single `Namespaces` API.

`Backend.renderDoc` returns a `List` and i'm not sure why. @pchiusano do you know the details of that?

## Test coverage
Tested the API and its giving the right results. We still need a better story for testing server endpoints.

## Loose ends
This is kinda slow? I'm not sure how best to debug that.

I think there's a separate effort related to working through these endpoints to make them more RESTful now that rootBranch is almost always used. I could imagine a set of endpoints like so (rough):

* `/:rootBranch/definitions/find/:query`
* `/:rootBranch/definitions/:definitionFQN`
* `/:rootBranch/namespaces/list`
* `/:rootBranch/namespaces/:namespaceFQN`

### DIFF
--- a/parser-typechecker/src/Unison/Server/Backend.hs
+++ b/parser-typechecker/src/Unison/Server/Backend.hs
@@ -258,6 +258,50 @@ findShallow codebase path' = do
     Nothing -> pure []
     Just b  -> findShallowInBranch codebase b
 
+findShallowReadmeInBranch ::
+  (Monad m, Var v) =>
+  Codebase m v Ann ->
+  Branch m ->
+  Backend m (Maybe (TermEntry v Ann))
+findShallowReadmeInBranch codebase branch =
+  let find :: [ShallowListEntry v Ann] -> Maybe (TermEntry v Ann)
+      find entries =
+        case entries of
+          [] -> Nothing
+          e : rest ->
+            case e of
+              ShallowTermEntry termEntry@(TermEntry _ segment _ _) ->
+                if (Text.toUpper . HQ'.toText $ segment) == "README"
+                  then Just termEntry
+                  else find rest
+              _ ->
+                find rest
+   in fmap find (findShallowInBranch codebase branch)
+
+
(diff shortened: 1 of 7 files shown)


pchiusano (changes requested):
Cool!

On the implementation, I would not go through `[ShallowListEntry]` if all you're doing is locating a term called `README`. That's going to do a bunch of extra work and might be contributing to slowness (the other thing that could be slow is doc rendering - which you can test by making the README in question a very small document).

I think that helper `renderDoc` returns a list because it's possible for a single definition to have multiple docs, like you could have two terms called `List.map.doc`.

I'll try to suggest some changes on the diff to improve the implementation of finding the README term.

pchiusano:
@hojberg check https://github.com/unisonweb/unison/pull/2368/commits/fbb62e3b6a060afe581538e34dd54b26fcdaa54c which I just pushed to this branch. This was my attempt at simplifying and to avoid needing to construct all these term results, and also avoiding a linear scan. 

I wasn't sure how to test, so can you give it another go? One possibly controversial decision is that I just looked for one of `README`, `Readme`, `ReadMe`, or `readme`. Avoids a linear scan, and any other capitalization doesn't seem like a thing anyone would ever intentionally type... :)

pchiusano (approved):

hojberg:
@pchiusano Nice! thanks, that works great :D

PR #2369, opened 2021-08-31
Improve testing of link/unlink

pchiusano:
Adds a check for displaying metadata by type, and a check for unlink


aryairani (approved):
LGTM

PR #2371, opened 2021-08-31
Make transcript tests depend on building unison exe

dolio:
HPack seems to be using the deprecated way of specifying this in cabal. Hopefully that doesn't cause a problem.


pchiusano (approved):
Oh nice, so this will prevent you from accidentally doing `stack exec transcripts` using a stale exe?

dolio:
Yeah, exactly. The `unison` executable is declared as a build tool dependency of `transcripts`, so it should be built and up to date whenever you run the transcripts.

PR #2374, opened 2021-09-01
remove cycle length

aryairani:
This PR removes the cycle length (`Pos`) from `Reference`s everywhere.

To support that, we add a couple of Codebase operations on whole cycles instead of individual definitions:
```haskell
getTermComponentWithTypes :: Hash -> m (Maybe [(Term v a , Type v a)])
getDeclComponent :: Hash -> m (Maybe [Decl v a])
dependentsofComponentImpl :: Hash -> m (Set Reference.Id)
```
and rejiggers some `SqliteCodebase` implementation to use them in places where the cycle length was needed previously to come up with a whole component.

The PR adds to earlier work pulling hashing code into a single module, removing Hashable instances for `Codebase.Patch`, `Codebase.Branch`, etc.

It pulls the main types out of `Codebase.Branch` into `Codebase.Branch.Type`, leaving just helpers; `Branch.Raw` and `Branch.BranchDiff` are in their own modules now too.

`unhashComponent` now deals in `Reference.Id`s rather than `Reference`s

**Todo:**
- [ ] Figure out why `suffixes.output.md` changed.

**Loose ends:**
- [ ] Should we calling `anyHashToMaybeObjectId` everywhere instead of `primaryHashToMaybeObjectId`? We think probably yes but should convince ourselves.


unison-src/transcripts/link.output.md
```diff
@@ -101,8 +101,8 @@ We can look at the links we have:
 ```ucm
 .> links coolFunction
 
-  1. alice                : Author
-  2. coolFunction.license : License
+  1. coolFunction.license : License
+  2. alice                : Author
   3. coolFunction.doc     : Doc
```
aryairani, line 106: This looks super confusing, but these links are sorted accoring to `Ord (Type Symbol)`!

unison-src/transcripts/suffixes.output.md
```diff
@@ -38,7 +38,7 @@ The `view` and `display` commands also benefit from this:
 ```ucm
 .> view List.drop
 
-  builtin builtin.List.drop : Nat -> [a] -> [a]
+  builtin builtin.List.drop : builtin.Nat -> [a] -> [a]
```
aryairani, line 41: @pchiusano Any idea how `Nat` could suddenly end up with a longer suffix? (but not in the `find` output)

unison-src/transcripts/fix2053.output.md
```diff
@@ -4,10 +4,10 @@
   go f i as acc =
     _pattern = List.at i as
     match _pattern with
-      None           -> acc
       Some _pattern1 ->
         use Nat +
         go f (i + 1) as (acc :+ f _pattern)
+      None           -> acc
```
aryairani, line 10: @pchiusano @dolio Any idea why the match cases are shown in a different order? Will this break some programs, or did some completeness analysis happen to prove this transformation is safe?

unison-src/transcripts/suffixes.output.md
pchiusano: This could happen if a fully-qualified PPE is being used here for some reason - are all type names being shown fully qualified?

It's also possible `Nat` is no longer unique in this transcript's codebase. You could dump out the names to see what's in scope.

unison-src/transcripts/fix2053.output.md
pchiusano: I'm guessing it's ordering non-overlapping matches by their constructor id, which is fine (non-overlapping patterns can be reordered relative to each other without affecting semantics), though I think it should try to preserve source order. But I think this diff is not a problem.

aryairani: Ok, sounds good.

unison-src/transcripts/suffixes.output.md
aryairani: In the previous and subsequent stanzas, it shows up at `Nat`.  I'm not aware of anything in the branch that would've changed the PPE, but I guess I can dump it out eventually.

pchiusano: Weird. Is it something specifically with printing out builtin types signatures?

parser-typechecker/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -134,8 +134,11 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
     SyncRemoteRootBranch repo branch syncMode ->
       lift $ Codebase.pushGitRootBranch codebase branch repo syncMode
     LoadTerm r -> lift $ Codebase.getTerm codebase r
-    LoadType r -> lift $ Codebase.getTypeDeclaration codebase r
+    -- LoadTermComponentWithType :: H.Hash -> Command m i v (Maybe [(Term v Ann, Type v Ann)])
```
aryairani:
```suggestion
```

parser-typechecker/src/Unison/Codebase/Branch/Type.hs
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+makeLenses ''Branch
+makeLensesFor [("_edits", "edits")] ''Branch0
+
+instance Eq (Branch0 m) where
+  a == b = _terms a == _terms b
+    && _types a == _types b
+    && _children a == _children b
+    && (fmap fst . _edits) a == (fmap fst . _edits) b
```
aryairani:
```suggestion
    && (fmap fst . _edits) a == (fmap fst . _edits) b

```

parser-typechecker/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -634,9 +634,9 @@ computeFrontier
    . Monad m
   => (Reference -> m (Set Reference)) -- eg Codebase.dependents codebase
   -> Patch
-  -> Names0
+  -> (Reference -> Bool)
```
aryairani, line 637: `computeFrontier` takes a predicate, not a `Names0` (in general trying to minimize use of `Names0` to where it's exactly needed)

unison-src/transcripts/suffixes.output.md
aryairani: @ChrisPenner /cc @mitchellwrosen We wanna understand/fix this change before merging.

PR #2375, opened 2021-09-01
`cd ..` is an alias for new `up` command, which moves up a level

pchiusano:
If you're already at the root, has no effect.

Screen Shot 2021-08-31 at 10 16 52 PM


hojberg (approved):
🎉

PR #2377, opened 2021-09-01
Add missing Eq case for Pattern.Char

### LINKED ISSUE #2345: cannot view some terms in my codebase
I know of at least two terms in my codebase which I'm not able to view or edit. If I try to view or edit them my ucm process starts chewing through memory. I've let it run for 30 minutes and ~60Gb of ram without gettng a response. I've gone through the process of squashing this codebase compeltely, and no matter what, for at least these two terms, I don't seem to be able to get ucm to show me the body of these functions

stew:
fixes: #2345

## Overview

A missing Eq case for Pattern.Char caused an infinite loop when pretty printing Pattern matches on char literals such as:

```
foo = match ?0 with 
  ?0 -> ()

```

### DIFF
--- a/unison-core/src/Unison/Pattern.hs
+++ b/unison-core/src/Unison/Pattern.hs
@@ -125,6 +125,7 @@ instance H.Hashable (Pattern p) where
 instance Eq (Pattern loc) where
   Unbound _ == Unbound _ = True
   Var _ == Var _ = True
+  Char _ c == Char _ d = c == d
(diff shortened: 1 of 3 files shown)


stew:
big thanks to @dolio for being the one to actually find the bug :)

pchiusano:
Awesome. Can you add a regression test please?

pchiusano (approved):

PR #2380, opened 2021-09-02
Fix some order dependence in `pruneAbilities`

### LINKED ISSUE #2378: Spurious ability check failure in this example
```  
unique ability C c where 
  new : c a
  receive : c a -> a
  send : a -> c a -> ()

unique ability A t g where 
  fork : '{A t g, g, Exception} a -> t a
  await : t a -> a

unique ability Ex where raise : () -> x

Ex.catch : '{Ex, g} a ->{g} Either () a
Ex.catch _ = todo "Exception.catch"

C.pure.run : (forall c . '{C c, g} r) ->{Ex, g} r
C.pure.run _ = todo "C.pure.run"

A.pure.run : (forall t . '{A t g, g} a) ->{Ex,g} a
A.pure.run _ = todo "A.pure.run"

x : '{} (Either () Nat)
x _ = Ex.catch 'let
  C.pure.run 'let
  A.pure.run 'let
    c = C.new
    x = A.fork 'let
      a = receive c 
      a + 10 
    y = A.fork 'let
      send 0 c
      ()
    A.await x
```
The error is:

```  
The expression in red needs the {C c137} ability, but this location only has access to these abilities: {A t69 {𝕖132, 𝕖133, 𝕖132, C c137}, 𝕖132, 𝕖133, 𝕖132, C c137}
  
     25 |     c = C.new
  
```

`C c137` is right there in the available list, so not sure how it’s not finding it. 

If I had to guess, it’s attempting to unify the `C c137` with one of those existentials. If that’s the case, perhaps it should probably just look through the fully substituted ambient and check for an exact match on the head (in this case `C`)

dolio:
Previously the results could depend on the order that wanted abilities were tested against provided abilities. The new approach runs until a fixed point is reached, allowing some unifications on concrete abilities to better inform the overall process.

Fixes #2378

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -2220,6 +2220,37 @@ expandWanted
   . (traverse.traverse) applyM
 
 
+pruneConcrete
+  :: Var v
+  => Ord loc
+  => (Maybe (Term v loc) -> Type v loc -> M v loc ())
+  -> Wanted v loc
+  -> Wanted v loc
+  -> [Type v loc]
+  -> M v loc (Wanted v loc)
+pruneConcrete _ acc [] _ = pure (reverse acc)
+pruneConcrete missing acc ((loc, w):ws) have
+  | Just v <- find (headMatch w) have = do
+    subtype v w `orElse` missing loc w
+    ws <- expandWanted ws
+    have <- expandAbilities have
+    pruneConcrete missing acc ws have
+  | otherwise = pruneConcrete missing ((loc,w):acc) ws have
+
+pruneVariables
+  :: Var v
+  => Ord loc
+  => Wanted v loc
+  -> Wanted v loc
+  -> M v loc (Wanted v loc)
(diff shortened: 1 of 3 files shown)


pchiusano (approved):
👍

pchiusano:
CI failing with

```
/home/runner/work/unison/unison/parser-typechecker/src/Unison/Typechecker/Context.hs:60:1: error: [-Wunused-imports, -Werror=unused-imports]
    The import of ‘Data.Ord’ is redundant
      except perhaps to import instances from ‘Data.Ord’
    To import instances alone, use: import Data.Ord()
   |
60 | import           Data.Ord                       ( comparing )
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/home/runner/work/unison/unison/parser-typechecker/src/Unison/Typechecker/Context.hs:2279:3: error: [-Wunused-local-binds, -Werror=unused-local-binds]
    Defined but not used: ‘isVar’
     |
2279 |   isVar (Type.Var' _) = True
```

Issue #2381, opened 2021-09-03
TDNR doesn’t work for higher-rank functions


pchiusano:
So if I have

```
Scope.run : (forall s . ‘{Scope s} a) -> a
Blah.run  : (forall s . ‘{Blah s} a) -> a
```
And then I write `run` somewhere, it won’t be able to resolve to `Scope.run` or `Blah.run`.

The workaround is to refer to `Scope.run` or `Blah.run` explicitly.

dolio:
Is there an example for where it doesn't work?

My guess would be that this is akin to expecting higher-rank types to be inferred automatically, because a higher rank type needs to be inferred for the higher rank candidates to be subtypes. That is not a tractable expectation.

pchiusano:
> My guess would be that this is akin to expecting higher-rank types to be inferred automatically, because a higher rank type needs to be inferred for the higher rank candidates to be subtypes. That is not a tractable expectation.

Yeah, I think it's that.

I could imagine TDNR working differently though and doing checking rather than inference. The way it works right now is something like - it infers the type of the symbol from context, then sees if that matches the types of symbols with a matching suffix. That approach obviously won't find higher-rank stuff since that will never be inferred.

Does not seem high priority though.

dolio:
Yeah, I'm not sure how to fix it other than just trying to type check with all possibilities.

Issue #448, opened 2019-04-09
Make TDNR more predictable


pchiusano:
I'd like TDNR to be more predictable. Lately I've been avoiding it since it gives odd results unless the rest of the code also typechecks (often not the case).

Something that might be relevant is that the way we currently select possibilities is by putting a blank in the program and then running typechecking to learn the type of that blank. If the program has type errors elsewhere, it's possible that inserting that blank may cause the program to typecheck, but with a weird type for the blank.

I'd probably start by collecting a number of examples of behavior we don't like for TDNR, then see if we can come up with something that's highly predictable. 

Or it might be that TDNR is fine and just needs better docs and a few bugfixes.

Related issues:

#377 #376 #370

pchiusano:
Some options on the table: 

1. rethink it, come up with a more predictable algorithm
2. disable it entirely
3. leave as is, with disclaimers

francisdb:
https://acronyms.thefreedictionary.com/TDNR ?

aryairani:
@francisdb IMO, The Dark Night Rises is predictable enough; the one we want to improve is Type-Directed Name Resolution.

pchiusano:
Current algorithm for TDNR:

1. For each TDNR slot, replace it with a blank, which is a fresh existential, then run typechecking and learn the type of that blank. Based on that type, substitute in the unique matching term, if that exists, otherwise it's a type error.
2. Repeat for the next TDNR slot, until all TDNR slots have been filled. Note that we do N typechecking passes, one for each TNDR slot.

In a program with type errors elsewhere and multiple TNDR slots, sticking a blank with a fresh existential type for each slot is way underconstrained and generates unpredictable results. The blanks can cover up the type errors and cause them to resurface elsewhere (or prevent TDNR elsewhere), and when multiple TNDR slots interact (think `foo (x + y)`, it's also unpredictable).

### Alternate algorithm

Here's a single pass algorithm that I think would be predictable:

1. To synthesize a function application, `f x y z`, where `f` is a TDNR slot, we first synthesize `x`, `y`, and `z`. Then we see if any of the possibilities for `f` can be applied to these types. If yes, and it's unique, do it. If no, type error.
2. To check a TDNR slot, `f`, against a type, `t`, we do the same - see if any of the possibilities for `f` are a subtype of `t` and if it's got a unique solution, use it.

[...]

PR #2382, opened 2021-09-03
Fix to subtle name resolution bug that results in using TDNR unexpectedly

### LINKED ISSUE #2353: Why is TDNR getting used in this example?
Here’s a self-contained example:

```
use builtin Scope
unique ability Async t g where async : Nat
unique ability Exception where raise : Nat -> x

pure.run : a -> (forall t . '{Async t g} a) ->{Exception, g} a
pure.run a0 a = 
  a' : forall s . '{Scope s, Exception, g} a
  a' = 'a0 -- typechecks
  Scope.run a'
```
This yields:
```
  I'm not sure what Scope.run means at line 10, columns 3-12
  
     10 |   Scope.run a'
  
  Whatever it is, it has a type that conforms to (Unit ->{g, 𝕖, Exception, builtin.Scope s} a) -> a.
  I found a term in scope with a matching name but a different type. If this is what you meant, try using the fully qualified name and I might be able to give you a more illuminating error message: 
  
    - .builtin.Scope.run : (Unit ->{builtin.Scope s, g} r) ->{g} r
```
Changing this to `##Scope.run a’` causes it to work, so it seems like some sort of name resolution failure - basically, it isn’t resolving `Scope.run` to a definition and is instead resorting to TDNR to figure out what definition it is.

Scope.run is a unique suffix in my codebase (according to `find Scope.run`, which only returns 1 result), so I don’t know what’s going on.

I kinda want to dig into this, since it could result in a lot of weird typechecking issues if unique suffixes aren’t being resolved properly.

pchiusano:
Fixes #2353 and adds a regression test.

The bug was resulting in use of the TDNR code path for more cases than expected. I'm still not totally sure what the implications are of the old behavior, since the TDNR code path has a short circuit if a definition has a unique suffix and the right type.

This fix avoids TDNR for the use case that was given and all the other transcripts are unchanged, including [this one](https://github.com/unisonweb/unison/blob/trunk/unison-src/transcripts/fix1578.output.md) which checks various subtle cases of name resolution.

Filed a separate bug for https://github.com/unisonweb/unison/issues/2381

### DIFF
--- a/parser-typechecker/src/Unison/FileParser.hs
+++ b/parser-typechecker/src/Unison/FileParser.hs
@@ -76,7 +76,7 @@ file = do
     -- suffixified local term bindings shadow any same-named thing from the outer codebase scope
     -- example: `foo.bar` in local file scope will shadow `foo.bar` and `bar` in codebase scope
     let (curNames, resolveLocals) =
-          ( Names.shadowSuffixedTerms0 locals (Names.currentNames names)
+          ( Names.shadowTerms0 locals (Names.currentNames names)
           , resolveLocals )
           where
           -- All locally declared term variables, running example:
--- a/parser-typechecker/src/Unison/FileParser.hs
+++ b/parser-typechecker/src/Unison/FileParser.hs
@@ -1,5 +1,6 @@
 {-# Language DeriveTraversable #-}
 {-# Language OverloadedStrings #-}
+{-# Language ViewPatterns #-}
 
 module Unison.FileParser where
 
@@ -8,7 +9,9 @@ import Unison.Prelude
 import qualified Unison.ABT as ABT
 import Control.Lens
 import           Control.Monad.Reader (local, asks)
+import Data.List.Extra (nubOrd)
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 import           Prelude hiding (readFile)
 import qualified Text.Megaparsec as P
 import           Unison.DataDeclaration (DataDeclaration, EffectDeclaration)
@@ -92,8 +95,9 @@ file = do
           -- suffix, but `bob` is. `foo.alice` and `bob.alice` are both unique suffixes but
           -- they map to themselves, so we ignore them. In our example, we'll just be left with
           --   [(bob, Term.var() zonk.bob)]
-          replacements = [ (Name.toVar n, Term.var() v') | (n,[v']) <- Map.toList varsBySuffix
-                                                         , Name.toVar n /= v' ]
+          replacements = [ (Name.toVar n, Term.var() v')
+                         | (n, nubOrd -> [v']) <- Map.toList varsBySuffix
+                         , Name.toVar n /= v' ]
           locals = Map.keys varsBySuffix
           -- This will perform the actual variable replacements for suffixes
           -- that uniquely identify definitions in the file. It will avoid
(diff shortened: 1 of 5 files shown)


parser-typechecker/src/Unison/FileParser.hs
```diff
@@ -92,19 +95,22 @@ file = do
           -- suffix, but `bob` is. `foo.alice` and `bob.alice` are both unique suffixes but
           -- they map to themselves, so we ignore them. In our example, we'll just be left with
           --   [(bob, Term.var() zonk.bob)]
-          replacements = [ (Name.toVar n, Term.var() v') | (n,[v']) <- Map.toList varsBySuffix
-                                                         , Name.toVar n /= v' ]
+          replacements = [ (Name.toVar n, Term.var() v')
+                         | (n, nubOrd -> [v']) <- Map.toList varsBySuffix
+                         , Name.toVar n /= v' ]
```
pchiusano, line 100: Missing `nubOrd` would cause this logic to not fire for any name with multiple segments. This was wrong.

unison-core/src/Unison/Names3.hs
```diff
@@ -311,13 +311,13 @@ expandWildcardImport prefix ns =
...
 -- resolution where local names are preferred to codebase names.
-shadowSuffixedTerms0 :: [Name] -> Names0 -> Names0
-shadowSuffixedTerms0 ns n0 = names0 terms' (types0 n0)
+shadowTerms0 :: [Name] -> Names0 -> Names0
+shadowTerms0 ns n0 = names0 terms' (types0 n0)
   where
-  shadowedBy name = Name.searchBySuffix name (terms0 n0)
-  terms' = R.subtractRan (foldMap shadowedBy ns) (terms0 n0)
+  terms' = foldl' go (terms0 n0) ns
+  go ts name = R.deleteDom name ts
```
pchiusano, line 323: This function no longer needs to do any fancy suffix-based stuff, so I renamed and simplified it. It wasn't being called anywhere else besides the name resolution code.

pchiusano:
It's also likely this speeds up typechecking since the TDNR codepath is kinda slow, even when there's unique matches.

ceedubs:
I'm not well-versed enough in the compiler to be able to offer a meaningful review, but thanks, for fixing this, @pchiusano. I ran into this over the weekend and was stumped for quite a while. I was working with some funky types, so I thought that I just didn't have the types lining up properly.

aryairani:
I'm not really sure how to evaluate this either, but I think it demonstrates the need for a boring comprehensive set of transcripts that fully spec out the behavior of TDNR. (Can be a separate effort.)  LMK if you want to walk through the PR together or if you want me to rubber-stamp it; and I'll leave it for @runarorama to take a look at in the mean time.

pchiusano:
@aryairani I'd just do a rubber stamp. If we do a walkthrough it'll soon be forgotten. :) I think we should task the next person who has to do surgery on this code to produce a writeup of how name resolution works. :)

I kind of want to just rethink TDNR and don't really want to expend effort trying to document exactly how it works right now. It's unpredictable and often not useful if your program has any type errors. See old discussion in https://github.com/unisonweb/unison/issues/448. I don't think addressing that is top priority right now, but if we were going to exhaustively document something, I'd want it to be a more predictable algorithm than what's there now.

aryairani (approved):
Makes sense.

runarorama (approved):

ceedubs:
I can confirm that this solves the nasty issue that I was running into where I got a cryptic error if I named my function `fold` but it was fine if I called it `fold2`. 🥳

PR #2384, opened 2021-09-07
update to lts-18.9 (published Sept 6), ghc 8.10.7

aryairani:
Should be safe to merge once HLS is released for ghc 8.10.7 (https://github.com/haskell/haskell-language-server/releases/).


hojberg:
Note we should wait with upgrading till https://github.com/haskell/vscode-haskell has a release that supports `8.10.7` as well.

hojberg (approved):
🎉

PR #2387, opened 2021-09-08
add builtin Term.Link.toText

stew:
wanted this builtin for log messages which are helping us debug our
implementation of Remote


pchiusano (approved):
Maybe just refresh the transcripts?

PR #2388, opened 2021-09-08
Pattern matching on constructors with polymorphic fields now works properly

pchiusano:
Worked on this with @dolio. Previously, the typechecker was ungeneralizing the types of fields at every level of pattern matching. This led to some spurious type errors when dealing with constructor fields that contained polymorphic functions. Example is like:

```Haskell
unique type Functor f = Functor (forall a b . (a -> b) -> f a -> f b)
```

Instead, we now just ungeneralize the root scrutinee, which has the effect of preserving polymorphic field types properly even when pattern matching on them.

[The transcript that we put together](https://github.com/unisonweb/unison/blob/d1a7692c40163f4459d543cb223d81e6ef85c87b/unison-src/transcripts/higher-rank.output.md) exercises the cases that didn't previously work, and I added a few other examples I thought of for additional testing.


dolio (approved):
Still looks good to me. Pretty straight forward change.

PR #2390, opened 2021-09-09
Add parens to destructuring binds

### LINKED ISSUE #1772: Destructuring bind requires parens (unless there's a guard)
There's a transcript in the repo (unison-src/transcripts/destructuring-binds.md), showing this example, which works fine.

```unison
ex3 =
  Some x | x > 10 = Some 19
  x + 1
```

```ucm

  change:
  
      ex3 : Nat

```
But if I delete the guard then it doesn't work.

```unison
ex3' =
  Some x = Some 19
  x + 1
```

```ucm

  I'm not sure what x means at line 3, columns 3-4
  
      3 |   x + 1
  
  Whatever it is, it has a type that conforms to builtin.Nat.
  

```
You need parentheses to make it happy, which is non-obvious.

```unison
ex3' =
  (Some x) = Some 19
  x + 1

```

```ucm

  change:
  
      ex3' : Nat

```
Would be better if the parens weren't needed?  Either way it's tricky that it depends on the presence of a guard.

(trunk@b586abb)

[destructure.md.txt](https://github.com/unisonweb/unison/files/5719578/destructure.md.txt)
[destructure.output.md.txt](https://github.com/unisonweb/unison/files/5719579/destructure.output.md.txt)

### LINKED ISSUE #2337: Pretty Printer: missing parens in sub expression pattern match.
pattern matching like so:

```hs
myFunction arg =
  (MyType someValue) = arg
  -- ...
```

compiles and works great, but is pretty printed like so:

```hs
myFunction arg =
  MyType someValue = arg
  -- ...
```

which does not compile.

runarorama:
Fixes #2337 and #1772 using the simplest thing that could possibly work.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -298,7 +298,7 @@ pretty0
           -- know bc.)  So we'll fail to take advantage of any opportunity
           -- this let block provides to add a use statement.  Not so bad.
           (fmt S.ControlKeyword "let") `PP.hang` x
-      lhs = PP.group (fst (prettyPattern n (ac 0 Block im doc) (-1) vs pat))
+      lhs = PP.group (fst (prettyPattern n (ac 0 Block im doc) 10 vs pat))
          <> printGuard guard
       printGuard Nothing = mempty
       printGuard (Just g') = let (_,g) = ABT.unabs g' in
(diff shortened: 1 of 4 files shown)


pchiusano:
Cool! CI is failing with: 

```
  860 passed
    "termprinter.match x with Optional.Some (Optional.Some _) -> ()"
```

I'm guessing this renders with extra parens and you can just update the test? You might have to refresh other transcripts too.

Might be cool for later to special case eliding the parens if the pattern is a tuple or list pattern, but not a big deal.

pchiusano (approved):

PR #2393, opened 2021-09-09
Change precedence of list patterns

### LINKED ISSUE #2224: Pretty printing multiple list head matches removes parens
The pretty printer removes parens on list head matches when they are chained:

```
match xs with
  x +: (x' +: rest) -> ...
```

gets pretty printed as

```
match xs with
  x +: x' +: rest -> ...
```

which does not type check

runarorama:
This changes precedence of list operator patterns so a pattern will always have parens around it unless it is the outermost pattern.

Fixes #2224


pchiusano (approved):
👍

PR #2396, opened 2021-09-09
Add parens around effectful arrow special case.

### LINKED ISSUE #2392: Type application doesn’t insert needed parens
```Haskell
unique ability Zonk where zonk : Nat
unique type Foo x y = 

foo : Nat -> Foo ('{Zonk} a) ('{Zonk} b) -> Nat
foo n _ = n
```
The signature prints as:
```
foo : Nat -> Foo '{Zonk} a '{Zonk} b -> Nat
```
Which does not parse.

Thanks to @ceedubs for reporting

runarorama:
Just adds what I think was simply a missing set of parens.

Fixes #2392 

## Controversial decision:

Previously, the expression `Pair a '({e} b)` was pretty-printing as `Pair a '{e} b`, but this would not parse.
Now it roundtrips perfectly as `Pair a '({e} b)`, and I removed the pending test for `Pair a '{e} b`. I don't think that syntax should be valid anyway (and it hasn't been valid in the past), but if we want to add support for that later, we can.

### DIFF
--- a/parser-typechecker/src/Unison/TypePrinter.hs
+++ b/parser-typechecker/src/Unison/TypePrinter.hs
@@ -104,8 +104,9 @@ prettyRaw n im p tp = go n im p tp
            in (fmt S.TypeOperator "∀ " <> vformatted <> fmt S.TypeOperator ".")
               `PP.hang` go n im (-1) body
     t@(Arrow' _ _) -> case t of
-      EffectfulArrows' (Ref' DD.UnitRef) rest -> arrows True True rest
-      EffectfulArrows' fst rest ->
+      EffectfulArrows' (Ref' DD.UnitRef) rest ->
+        PP.parenthesizeIf (p >= 0) $ arrows True True rest
+      EffectfulArrows' fst rest -> PP.parenthesizeIf (p >= 0) $
         case fst of
           Var' v | Var.name v == "()"
             -> fmt S.DelayForceChar "'" <> arrows False True rest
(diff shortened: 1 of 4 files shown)


pchiusano:
@runarorama cool, something not quite right though, `stack exec tests -- typeprinter` has a bunch of round-tripping failures that I definitely think should still pass unchanged.

```
    "typeprinter.'a -> b"
    "typeprinter.((a -> b) -> c) -> d"
    "typeprinter.(a -> b) -> c"
    "typeprinter.(a -> b) -> c -> d"
    "typeprinter.Pair (a -> b) (c -> d)"
    "typeprinter.Pair a ('{e} b)"
    "typeprinter.Pair a b ->{e1, e2} Pair a b ->{} Pair (a -> b) d -> Pair c d"
    "typeprinter.a -> 'b -> c"
    "typeprinter.a -> 'b ->{f} c"
    "typeprinter.a -> '{e} () ->{f} b"
    "typeprinter.a -> '{e} b -> c"
    "typeprinter.a -> '{e} b ->{f} c"
    "typeprinter.a -> (b -> c) -> d"
```

Maybe an off-by-1 problem with the precedence levels?

pchiusano:
Unrelated: I think `undo` will do it instead of the reflog stuff, it's only when there's multiple steps (like adding metadata) that `undo` wouldn't be enough or you'd need to do it multiple times.

runarorama:
On it, nothing to see here yet.

pchiusano:
My bad I missed that it was a draft. Taking a look now.

pchiusano (approved):

pchiusano:
Sweet!

PR #2397, opened 2021-09-09
Base library download on codebase init

rlmark:
coauthored @hojberg 

## Overview

This change will download the base library upon initialization of a new codebase. Addresses part of issue 
https://github.com/unisonweb/unison/issues/2348 

## Implementation notes

We used the existing Input data type and prepended a command to the startup of the ucm. 

We moved the Welcome process into a separate module with the intention to augment it for a full interactive onboarding workflow. 
We fixed a bug in the getDefaultBase function which referenced git tags as `devel` to `latest` - it seemed to us that `devel` hadn't been used as a tag prefix in a long while. (Let us know if this was the wrong call)

## Test coverage

We have use-it tested this code, but are unsure how to test the main function and would need existing testing fixtures to help us. 

## Loose ends

We did not implement the `--no-base` flag at this time. 
In the future, we think the base download output should be smaller, and before the "get started" part of the welcome message.


pchiusano:
Cool, can we see screenshot or paste of what the startup process looks like now?

rlmark:
> Cool, can we see screenshot or paste of what the startup process looks like now?

Screen Shot 2021-09-09 at 2 03 30 PM

This is what the startup screen looks like after a fresh codebase init.

rlmark:
Perhaps less desirable is the next screen, where you have to exit out of less: 
Screen Shot 2021-09-09 at 2 06 44 PM

aryairani:
@rlmark Absent `--no-base`, is there a way (e.g. by supplying `--codebase`) to skip downloading `base`?

hojberg:
@aryairani with this, only a non-empty codebase will skip the download. Our intent was to implement `--no-base` in a follow up if thats ok.

hojberg:
As noted in the loose ends, we're pretty keen on getting the base download output a lot smaller for this initial download.
The thinking is that it's part of an onboarding flow that goes between the welcome banner and the "Get started" section.

We had played around with an output like so (rough draft):

```
  Now starting the Unison Codebase Manager (UCM)...


   _____     _
  |  |  |___|_|___ ___ ___
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|

  👋 Welcome to Unison!

  You are running version: latest-370-g63bc75ea3' (built on 2021-09-07)

  ──

  ☕ First things first

  It looks like this is the first time you started UCM, I went ahead and created a 
  codebase for you in `~/hojberg/.unison`, next, to really get you started writing 
  Unison code, let me download the base library for you:

    🕐 Downloading `releases._M2h` of the base library into `.base`, this might take a minute...

  ✅ Success! The base library is the Unison standard library that includes core types and
  functions to write Unison code. Visit https://docs.unison-lang.org/base to learn more.

  To assign authorship to your code I need a little information about you. 

    Ready? `Yes`/`skip` > Yes
    Whats your preferred name? > Simon Højberg

  😎 Thank you `Simon`, you are good to go to start writing Unison code. Have fun!
  
  ──

  Get started:

    📖 Type help to list all commands, or help <cmd> to view help for one command
    🎨 Type ui to open the Codebase UI in your default browser
    📚 Read the official docs at https://unisonweb.org/docs
    🌏 Visit Unison Share at https://share.unison-lang.org to discover libraries
    👀 I'm watching for changes to .u files under ~/code/unison/unison
```

This would be where prompts for name etc would go as well (not fully explored in the above draft).


Anyway, we were going to look at this in a follow as well.

hojberg:
(accidentally closed the pr)

parser-typechecker/src/Unison/Codebase/Editor/VersionParser.hs
```diff
@@ -12,12 +12,12 @@ import qualified Unison.Codebase.Path as Path
...
---  "devel/*" -> "trunk"
+--  "latest-*" -> "trunk"
 defaultBaseLib :: Parsec Void Text ReadRemoteNamespace
-defaultBaseLib = fmap makeNS $ devel <|> release
+defaultBaseLib = fmap makeNS $ latest <|> release
   where
-  devel, release, version :: Parsec Void Text Text
-  devel = "devel/" *> many anyChar *> eof $> "trunk"
+  latest, release, version :: Parsec Void Text Text
+  latest = "latest-" *> many anyChar *> eof $> "trunk"
```
pchiusano, line 20: Okay, so this will use a released version of base if you're using a named release, and otherwise will pull trunk.

pchiusano (approved):
Looks great. I'd add the `--no-base` flag, but that could be a follow up.

If you're interested limiting the output of that initial pull, we could add `pull.quiet` command, which just produces `✅ Finished.` output. It could be implemented by adding an extra argument to the existing `PullRemoteBranchI` constructor, with verbosity info. And then dispatch on this in `HandleInput`. This would be a good way to learn about adding new commands to `ucm`.

pchiusano:
https://github.com/unisonweb/unison/issues/2398 to track the `--no-base` loose end

parser-typechecker/src/Unison/Codebase/Editor/VersionParser.hs
rlmark: > Looks great. I'd add the `--no-base` flag, but that could be a follow up.
> 
> If you're interested limiting the output of that initial pull, we could add `pull.quiet` command, which just produces `✅ Finished.` output. It could be implemented by adding an extra argument to the existing `PullRemoteBranchI` constructor, with verbosity info. And then dispatch on this in `HandleInput`. This would be a good way to learn about adding new commands to `ucm`.

Yeah, this sounds like something we'd be interested in!

hojberg: Yep. This was already being used with regards to the "pull base" hint, but had a bug with `devel`

aryairani:
Wait, for codebase development I definitely need a way to create a new codebase without having to wait to pull base, and a way to create a codebase without an internet connection. 

There might be implications for the transcript runner too; I'm not sure?

pchiusano:
@aryairani it's cool, we will still do it! https://github.com/unisonweb/unison/issues/2398#issue-992664792

Transcript runner is unaffected, the change only affects UCM, and also does not affect the `ucm transcript` subcommand.

PR #2400, opened 2021-09-10
add a Any.unsafeExtract builtin

stew:
This provides an unsafe way to extract a value that was previously wrapped in an Any.


dolio:
I think it might be a little better to not use the FFI wrapper. The wrapped function is just acting as a case statement, right?

pchiusano:
~~@dolio do you want to hold this up until @stew can address, and if so I think you gotta spell out what you mean or work with him on fixing.~~

Nevermind, I see @stew has already done it!

pchiusano (approved):

PR #2405, opened 2021-09-14
removes init command, replaces with --codebase-create flag

rlmark:
Coauthored with @hojberg 

## Overview

General onboarding improvement: https://github.com/unisonweb/unison/issues/2348 
We removed the separate `init` command, in favor of a `--codebase-create` flag when running the ucm. The user no longer needs a two step process to create and start the ucm. 

The new --help text describes this change: 

```
Use unison [command] --help to show help for a command.

Usage: unison [--codebase codebase/path | --codebase-create codebase/path] 
              [COMMAND | [--token STRING] [--host STRING] [--port NUMBER] [--ui DIR]]

Available options:
  --codebase codebase/path The path to an existing codebase
  --codebase-create codebase/path
                           The path to a new or existing codebase (one will be created if there
                           isn't one)
```

## Implementation notes

Flag was added to the UCM lunch command. 
We contemplated having a --codebase flag and an additional --create flag but having two separate flags which each took the codebase path as their argument ended up making the combination a little simpler. 

## Test coverage

We have added tests to reflect this change

### DIFF
--- a/parser-typechecker/tests/Unison/Test/CodebaseInit.hs
+++ b/parser-typechecker/tests/Unison/Test/CodebaseInit.hs
@@ -7,6 +7,10 @@ module Unison.Test.CodebaseInit where
 import EasyTest
 import qualified Unison.Codebase.Init as CI
 import Unison.Codebase.Init
+    ( CodebaseInitOptions(..)
+    , Init(..) 
+    , SpecifiedCodebase(..)
+    )
 import qualified System.IO.Temp as Temp
 
 -- keep it off for CI, since the random temp dirs it generates show up in the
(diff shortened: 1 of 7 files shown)


pchiusano (approved):
Great!

parser-typechecker/tests/Unison/Test/CodebaseInit.hs
```diff
@@ -37,18 +41,35 @@ test = scope "Codebase.Init" $ tests
...
           _ -> expect False
     ] 
+  , scope "*with* a --codebase-create flag" $ tests
+    [  scope "a v2 codebase should be created when one does not exist at the Specified dir" do
+        tmp <- io (Temp.getCanonicalTemporaryDirectory >>= flip Temp.createTempDirectory "ucm-test")
+        cbInit <- io initMockWithoutCodebase
+        res <- io (CI.openOrCreateCodebase cbInit "ucm-test" (Specified (CreateWhenMissing tmp)))
+        case res of 
+          CI.CreatedCodebase _ _ -> expect True
+          _ -> expect False
```
aryairani, line 63: For future reference you can use `ok` in place of `expect True`, and `fail "msg"` or `crash "msg"` instead of `expect False`.

parser-typechecker/unison/ArgParse.hs
```diff
@@ -138,7 +144,8 @@ versionCommand = command "version" (info versionParser (fullDesc <> progDesc "Pr
 initCommand :: Mod CommandFields Command
 initCommand = command "init" (info initParser (progDesc initHelp))
   where
-    initHelp = "Initialise a unison codebase"
+    initHelp = 
+        "This command is has been removed. Use --codebase-create instead to create a codebase in the specified directory when starting the UCM."
```
aryairani, line 148: 👏

aryairani (commented):
Great!

aryairani (approved):

Issue #2398, opened 2021-09-09
Add `--no-base` flag


pchiusano:
~~I don’t think this one is particularly urgent since the autoinitialization only affects totally empty codebases.~~

Disregard, @aryairani thinks this one is pretty important to support local testing. So I'd try to knock it out. :)

aryairani:
Wait, 100% of codebases start out totally empty.

pchiusano:
Ha, I just meant... it will only be run once for a codebase, then never again. It also isn't run for transcripts.

It's probably more a feature that we might use sometimes when developing Unison, like for testing purposes. I'm not sure many end-users would make use of it.

pchiusano:
Fixed by #2414

PR #2414, opened 2021-09-16
Adds the --no-base flag option for initializing new codebase

rlmark:
Coauthored with @hojberg 

## Overview

when running the `ucm` command, if you are creating a new codebase, you can supply a flag --no-base to avoid the base download. By default, new codebases download base. 

See available options from help text: 

```
Available options:
  --codebase codebase/path The path to an existing codebase
  --codebase-create codebase/path
                           The path to a new or existing codebase (one will be created if there
                           isn't one)
  --token STRING           API auth token
  --host STRING            Codebase server host
  --port NUMBER            Codebase server port
  --ui DIR                 Path to codebase ui root
  --no-base                if set, a new codebase will be created without downloading the base
                           library, otherwise the new codebase will download base
  -h,--help                Show this help text
```

Addresses https://github.com/unisonweb/unison/issues/2398 

## Implementation notes

Added flag option to argParse. 
Updated Welcome module with types! 

## Test coverage

We have use-it tested this by building the executable, but were not sure how to test the Command line Main changes. We are open to suggestions!


pchiusano (approved):
Great!

@aryairani can rest easy now. :)

PR #2415, opened 2021-09-17
Allow blocks to appear as elements in lists and tuples

### LINKED ISSUE #2258: Parser rejects if-then-expression if it's a direct element of a list
```
/Users/acarrasco/workspace-ac/unison-scratch-v2/scratch.u:1:30:
  unexpected ]
      1 | > [if true then "x" else "y"]
```

this works fine:
```
1 | > [(if true then "x" else "y")]
          ⧩
          ["x"]
```

Unfortunately, it gets rendered by the pretty-printer in the correct, but un-parsable version:
```
.> view x

  x : [Text]
  x = [if true then "x" else "y"]
```

version: release/M2g

runarorama:
Fixes #2258.

This changes and unifies the way the parser handles brackets—namely parentheses (`()`) and square brackets (`[]`)—and commas (`,`) inside them. A bracket now opens a block, and the corresponding closing bracket closes it. Any comma between brackets also closes any inner blocks that were opened inside the brackets.

As a result, blocks can now appear as elements in lists and tuples, so this is valid syntax:

```
[if true then 0 else 1]
```

as is

```
(if true then 0 else 1, "hi")
```

This incidentally also adds support for redundant commas in tuples, like `(,"one",,"two",)`, making that syntax consistent with lists.

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -917,8 +917,8 @@ lexemes' eof = P.optional space >> do
           where quote s = "'" <> s <> "'"
         Just (block, n) -> do
           S.put (env { layout = drop (n-1) (layout env) })
-          let opens = [Token (Reserved close) pos1 pos2]
-          pure $ replicate (n-1) (Token Close pos1 pos2) ++ opens
+          let delims = [Token (Reserved close) pos1 pos2]
+          pure $ replicate (n-1) (Token Close pos1 pos2) ++ delims
 
     close' :: Maybe String -> [String] -> P String -> P [Token Lexeme]
     close' reopenBlockname open closeP = do
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -916,7 +916,7 @@ lexemes' eof = P.optional space >> do
         Nothing -> err pos1 (UnexpectedDelimiter (quote close))
           where quote s = "'" <> s <> "'"
         Just (block, n) -> do
-          S.put (env { layout = drop n (layout env), opening = Just block })
+          S.put (env { layout = drop (n-1) (layout env) })
           let opens = [Token (Reserved close) pos1 pos2]
           pure $ replicate (n-1) (Token Close pos1 pos2) ++ opens
 
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -920,7 +920,7 @@ lexemes' eof = P.optional space >> do
       case findClose open (layout env) of
         Nothing -> err pos1 (UnexpectedDelimiter (quote close))
           where quote s = "'" <> s <> "'"
-        Just (block, n) -> do
+        Just (_, n) -> do
           S.put (env { layout = drop (n-1) (layout env) })
           let delims = [Token (Reserved close) pos1 pos2]
           pure $ replicate (n-1) (Token Close pos1 pos2) ++ delims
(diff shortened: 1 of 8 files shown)


parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -882,6 +908,18 @@ lexemes' eof = P.optional space >> do
...
+    blockDelimiter open closeP = do
+      (pos1, close, pos2) <- positioned $ closeP
+      env                 <- S.get
+      case findClose open (layout env) of
+        Nothing -> err pos1 (UnexpectedDelimiter (quote close))
+          where quote s = "'" <> s <> "'"
+        Just (block, n) -> do
+          S.put (env { layout = drop n (layout env), opening = Just block })
+          let opens = [Token (Reserved close) pos1 pos2]
+          pure $ replicate (n-1) (Token Close pos1 pos2) ++ opens
```
runarorama: The `n-1` here is key. Close as many blocks as necessary to get us back to the bracket block.

parser-typechecker/src/Unison/TypeParser.hs
```diff
@@ -84,9 +84,9 @@ effectList = do
 
 sequenceTyp :: Var v => TypeP v
 sequenceTyp = do
-  open <- reserved "["
+  open <- openBlockWith "["
```
runarorama, line 87: Maybe it's a bit weird that blocks can appear in list types, but there aren't any block type elements at this level anyway.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -854,7 +872,15 @@ lexemes' eof = P.optional space >> do
         inLayout <- S.gets inLayout
         when (not inLayout) $ void $ P.lookAhead (CP.satisfy (/= '}'))
         pure l
+    matchWithBlocks = ["match-with", "cases"]
     parens = open "(" <|> close ["("] (lit ")")
+    brackets = open "[" <|> close ["["] (lit "]")
+    commaSeparator = do
+      env <- S.get
+      case topBlockName (layout env) of
+        Just match | not (match `elem` matchWithBlocks || match == "{") ->
```
pchiusano: Suggestion: pull this logic out into a function: `allowCommaToCloseIn match` and comment its purpose.

parser-typechecker/src/Unison/Lexer.hs
```diff
@@ -882,6 +908,18 @@ lexemes' eof = P.optional space >> do
...
 
+    blockDelimiter :: [String] -> P String -> P [Token Lexeme]
+    blockDelimiter open closeP = do
+      (pos1, close, pos2) <- positioned $ closeP
+      env                 <- S.get
+      case findClose open (layout env) of
+        Nothing -> err pos1 (UnexpectedDelimiter (quote close))
+          where quote s = "'" <> s <> "'"
+        Just (block, n) -> do
+          S.put (env { layout = drop n (layout env), opening = Just block })
```
pchiusano: Did you try just -  

```suggestion
          S.put (env { layout = drop (n-1) (layout env) })
```

The existing line is I think popping that block from the layout stack, then later re-adding it back (with a newly updated starting column), which seems goofy and is probably subtly wrong for error reporting.

parser-typechecker/src/Unison/Lexer.hs
pchiusano: Minor suggestion: renaming the `opens` list to `closes` (or maybe `delims`). I did a double-take when reading it the first time.

```suggestion
          let delims = [Token (Reserved close) pos1 pos2]
          pure $ replicate (n-1) (Token Close pos1 pos2) ++ delims
```

pchiusano (approved):
Cool. I left a few suggestions that I think might be good improvements in the `Lexer`. Have a look. I also added one additional test case, directly to the transcript.

I removed ready-to-merge so you can review my comments and address and merge at your leisure.

parser-typechecker/src/Unison/Lexer.hs
pchiusano: I think I understand this better... it's really that there is a set of blocks where `,` is not treated specially and it just emits a `Reserved ","`. There are currently only three, `cases`, `match-with`, and `{`.

PR #2418, opened 2021-09-18
New version of systemTime that returns an Int representing microseconds

### LINKED ISSUE #2359: `systemTime` is returning seconds when it should be returning nanoseconds
On trunk (prob other versions too):

```
timeApp: '{IO, Exception} ()
timeApp _ = 
  match !systemTime with
    EpochTime t -> printLine (toText t)
```
```
.base.io> run timeApp
1630089157
```
the seconds are coming from the underlying built-in `systemTime.impl`

alvaroc1:
Resolves #2359

Tests pass and have used the new function successfully (i hacked the previous version for the test), but will require an update of base.

Base refers to `systemTime.impl` which is a built-in, but i'm not sure how that connects to `systemTime.impl.v3`, i'm guessing something else needs to be wired (?)


atacratic:
Hooray! ☺️

But looks like impl just multiplies by a million, and can't actually work out the time to precision better than a second?

alvaroc1:
> Hooray! ☺️
> 
> But looks like impl just multiplies by a million, and can't actually work out the time to precision better than a second?

My very limited understanding is that `getPOSIXTime` produces something equivalent to a Real (?) or a Float representing seconds including fractions of a second. Multiplying it by a million turns it to microseconds and something later turns it into Int.

My animation code depends on it having microsecond resolution, and seems to work fine at least when rendering 30 fps, which makes me think the code is correct. But i can't claim to understand all the details :/

atacratic:
Oh magic, OK, nice one, sorry for misunderstanding!

Cheers for doing this 👍

parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -560,6 +560,7 @@ ioBuiltins =
   , ("IO.putBytes.impl.v3", handle --> bytes --> iof unit)
   , ("IO.getLine.impl.v1", handle --> iof text)
   , ("IO.systemTime.impl.v3", unit --> iof nat)
+  , ("IO.systemTime2.impl.v1", unit --> iof int)
```
pchiusano: How about we call this `IO.systemTimeMicroseconds.v1`? I think the return result can just be `io int` too, since this function can't fail. You can then simplify the calling convention.

We use the `.impl` convention only when the builtin has to return an `Either`, so that the nicely named version returns `{Exception,IO} a`. 

The `.v1` will be stripped out of the default name that shows up when you `builtins.merge`, and we'll add an alias for all new builtins in base when we do the next release.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -992,6 +992,19 @@ outIoFailNat stack1 stack2 stack3 fail nat result =
...
+  [ (0, ([BX, BX],)
+        . TAbss [stack1, stack2]
+        . TLetD fail BX (TCon Ty.failureRef 0 [stack1, stack2])
+        $ TCon eitherReference 0 [fail])
+  , (1, ([UN],)
+        . TAbs stack3
+        . TLetD int BX (TCon Ty.intRef 0 [stack3])
+        $ TCon eitherReference 1 [int])
+  ]
+
```
pchiusano: Maybe rename this to `unitDirectInt` and simplify it.

@dolio do you know what this should be for a function `() -> IO Int`?
```

pchiusano (changes requested):
Great! Left a couple comments though.

pchiusano:
Also anytime you update the builtins, run `stack exec transcripts` after. CI checks that transcript output hasn't changed, and the set of builtins affects some transcript output.

parser-typechecker/src/Unison/Runtime/Builtin.hs
dolio: I think in the naming scheme `Direct` means that it just returns a boxed value from the operation, but this is wrapping an unboxed int. Also this is only the part of the code that deals with the result, not the whole function.

I think this bit would just disappear, and the boxing would just be inlined into a renamed `unitToInt` `ForeignOp` or something.

pchiusano: @dolio can you give `unitToInt :: ForeignOp` ?

dolio: Instead of `outIoFailInt`, just use `TCon Ty.intRef 0 [result]`, I think.

alvaroc1: Done

parser-typechecker/src/Unison/Builtin.hs
alvaroc1: Done

pchiusano (approved):
Great!

PR #2420, opened 2021-09-19
Remove trace from fieldnames

hojberg:
## Overview
The trace snug through to the merge. Remove it, and avoid random print
statements when running UCM.

cc @alvaroc1


pchiusano (approved):
Ha, whoops!

PR #2421, opened 2021-09-19
Fixed built-in IO ability to be reported as Ability on Server

### LINKED ISSUE #2325: IO builtin reported by API server as a Data type but should be an Ability type
image

The API tags it as a `Data`, but should be `Ability`:

```
{"termDefinitions":{},"typeDefinitions":{"##IO":{"typeNames":[".base.IO",".builtin.io2.IO",".contrib.stew.logging.builtin.io2.IO"],"bestTypeName":"base.IO",

"defnTypeTag":"Data"

,"typeDefinition":{"tag":"BuiltinObject","contents":[{"annotation":null,"segment":"base.IO"}]},"typeDocs":[]}},"missingDefinitions":[]}
```

alvaroc1:
Fixes #2325 

Not sure if I am doing this right, but it seems to work:
http://127.0.0.1:50120/RVo7z7c4KYLtSIobQ5n6QozbtpVIoGMU/api/getDefinition?names=base.IO&rootBranch=%23jev7fp8580qth12tmf68f221ir5a3ahbvvl54d4f200q3nk1dln0dnhla4f72clu3fm0dni1ubj9ossgucq7mqvmnciepemm2kopuhg
```
{"termDefinitions":{},"typeDefinitions":{"##IO":{"typeNames":[".base.IO"],"bestTypeName":"IO","defnTypeTag":"Ability","typeDefinition":{"tag":"BuiltinObject","contents":[{"annotation":null,"segment":"IO"}]},"typeDocs":[]}},"missingDefinitions":[]}
```

### DIFF
--- a/parser-typechecker/src/Unison/Server/Backend.hs
+++ b/parser-typechecker/src/Unison/Server/Backend.hs
@@ -308,6 +308,10 @@ termListEntry codebase b0 r n = do
       tag = if isDoc then Just Doc else if isTest then Just Test else Nothing
   pure $ TermEntry r n ot tag
 
+
+builtInIoAbility :: Reference
+builtInIoAbility = Reference.unsafeFromText "##IO"
+
 typeListEntry
   :: Monad m
   => Var v
@@ -323,7 +327,8 @@ typeListEntry codebase r n = do
       pure $ case decl of
         Just (Left _) -> Ability
         _             -> Data
-    _ -> pure Data
+    -- IO is the only built-in ability
+    _ -> pure (if r == builtInIoAbility then Ability else Data)
   pure $ TypeEntry r n tag
 
 typeDeclHeader
--- a/parser-typechecker/src/Unison/Server/Backend.hs
+++ b/parser-typechecker/src/Unison/Server/Backend.hs
@@ -323,7 +323,6 @@ typeListEntry codebase r n = do
       pure $ case decl of
         Just (Left _) -> Ability
         _             -> Data
-    -- IO is the only built-in ability
     _ -> pure (if Set.member r Type.builtinAbilities then Ability else Data)
   pure $ TypeEntry r n tag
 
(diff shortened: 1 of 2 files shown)


parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -323,7 +327,8 @@ typeListEntry codebase r n = do
       pure $ case decl of
         Just (Left _) -> Ability
         _             -> Data
-    _ -> pure Data
+    -- IO is the only built-in ability
+    _ -> pure (if r == builtInIoAbility then Ability else Data)
```
pchiusano:
```suggestion
    _ -> pure (if r == Type.builtinIORef then Ability else Data)
```

Might need to `import qualified Unison.Type as Type` if that doesn't already exist in this module.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -308,6 +308,10 @@ termListEntry codebase b0 r n = do
       tag = if isDoc then Just Doc else if isTest then Just Test else Nothing
   pure $ TermEntry r n ot tag
 
+
+builtInIoAbility :: Reference
+builtInIoAbility = Reference.unsafeFromText "##IO"
+
```
pchiusano:
```suggestion
```

Can get rid of this if using the constant from `Type.builtinIORef`

pchiusano (changes requested):
Great! Left one thing for you to address - what you have will work but there's already a pre-existing constant you can use.

pchiusano (approved):
Great - 

One thing I just thought of is that `STM` is another builtin ability, defined in the same spot as `builtinIORef` (it's `Type.stmRef`), so if you wanted to include that here as well then that would be welcome!

Suggestion is have `Type.builtinAbilities : Set Reference`, and then do `Set.member` to check if it's in that set.

I'll leave this PR open for a day in case you'd like to add that otherwise I'll create a separate ticket to track that one.

pchiusano:
Looks good, merging!

PR #2427, opened 2021-09-22
Fix emphasis roundtrip in docs

### LINKED ISSUE #2408: Doc pretty print round trip issue with emphasize and underscores.
To emphasize parts of a doc the user can use `**my text**` or `__my text__`. Both of these will be pretty printed (when edited) as `__my text__`.  This works fine, but if there is an underscore in the part being emphasized, the parser breaks:

`**MY_TEXT**` is pretty printed as `__MY_TEXT__` which does not parse.

@pchiusano I'm not sure if I should file this as a pretty printing issue or a parsing issue. Thoughts?

runarorama:
Fixes #2408

This changes the lexer and pretty-printer for doc literals slightly. Underscores are now allowed inside emphasis elements surrounded by `__`, as long as there are strictly fewer `_` inside than around.

When pretty-printing, we print as many `_` as necessary so that it's one more than the longest run of `_` inside the element. 

Same for `**` and `~~` and `''`.

This also changes the meaning of `*` and `_` so that now any number of `*` means "bold" and any number of `_` means "italic".

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -1342,7 +1342,16 @@ prettyDoc2 ppe ac tm = case tm of
       (toDocItalic ppe -> Just d) ->
         PP.group $ "*" <> rec d <> "*"
       (toDocBold ppe -> Just d) ->
-        PP.group $ "__" <> rec d <> "__"
+        let inner = rec d
+            numUnderscores =
+              case
+                  filter (\s -> take 2 s == "__")
+                    $ group (PP.toPlainUnbroken $ PP.syntaxToColor inner)
+                of
+                  [] -> 2
+                  x  -> 1 + (maximum $ map length x)
+            underscores = replicate numUnderscores '_'
+         in PP.group $ PP.string underscores <> inner <> PP.string underscores
       (toDocStrikethrough ppe -> Just d) ->
         PP.group $ "~~" <> rec d <> "~~"
       (toDocGroup ppe -> Just d) ->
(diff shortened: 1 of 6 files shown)


parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -1342,7 +1342,16 @@ prettyDoc2 ppe ac tm = case tm of
...
+              case
+                  filter (\s -> take 2 s == "__")
+                    $ group (PP.toPlainUnbroken $ PP.syntaxToColor inner)
+                of
+                  [] -> 2
+                  x  -> 1 + (maximum $ map length x)
+            underscores = replicate numUnderscores '_'
+         in PP.group $ PP.string underscores <> inner <> PP.string underscores
       (toDocStrikethrough ppe -> Just d) ->
         PP.group $ "~~" <> rec d <> "~~"
```
pchiusano: Nice. Do you want to give the same treatment to strikethrough? It'll have the same bug when pretty-printing.

Not sure what to do about italics but we could consider that out of scope for this PR. One idea I had was use 1 or more `*` for italics, and 1 or more `_` for bold. That way you can apply the same trick for italics, bold, and strikethrough.

pchiusano (approved):
Great! Left one comment for you.

parser-typechecker/src/Unison/TermPrinter.hs
runarorama: I kind of like that solution, although for some reason I want it to be the other way around (`_` for emphasis, `*` for bold).

pchiusano: Sure

PR #2428, opened 2021-09-22
Add better diagnostics when compilation fails with missing data case

pchiusano:
I've been tracking down a compilation failure with "missing data case". Still haven't managed to reproduce it more simply, but this PR adds slightly better diagnostics by telling you what `Reference` is involved in the failure.


dolio (approved):
Looks fine to me. Pretty straight forward.

PR #2432, opened 2021-09-23
Add parens around let blocks, fix delayed effect parser

### LINKED ISSUE #1778: Round trip fails for parenthesised let block + operator
`ucm version: release/M1m`

Place this code in scratch.u:
```
x = '(let
  abort
  0) |> Abort.toOptional
```

Then
```
>. update
>. edit x
```
prepends to scratch.u:
```
x : Optional Nat
x =
  'let
    abort
    0 |> toOptional
```
which is different and doesn't typecheck.

runarorama:
Fixes #1778 

This adds parens around delayed et blocks that appear as arguments to operators.

```
x = ('let
  abort
  0) |> Abort.toOptional

y = (let
        x = 10
        x + 2) |> id
```

Also fixes the type parser so that it allows the following:

```
x -> '{Abort} Optional a 
```

instead of requiring

```
x -> '{Abort} (Optional a)
```


parser-typechecker/src/Unison/Parser.hs
```diff
@@ -53,6 +54,7 @@ import qualified Unison.Hashable as Hashable
 import Unison.Referent (Referent)
 import Unison.Reference (Reference)
 import Unison.Parser.Ann (Ann(..))
+import Text.Megaparsec.Error (ShowErrorComponent)
```
runarorama, line 57: This is added purely for debugging. No harm in having this.

parser-typechecker/src/Unison/TermPrinter.hs
```diff
@@ -368,7 +368,7 @@ pretty0
            -> ([Pretty SyntaxText] -> Pretty SyntaxText)
            -> Pretty SyntaxText
   printLet elideUnit sc bs e im uses =
-    paren ((sc /= Block) && p >= 12)
+    paren ((sc /= Block) && p >= 3)
```
runarorama: This will cause parens to print around a `let` block if it appears as an argument to a function.

parser-typechecker/src/Unison/TypeParser.hs
```diff
@@ -72,7 +72,7 @@ type2 = do
 effect :: Var v => TypeP v
 effect = do
  es <- effectList
- t <- valueTypeLeaf
+ t <- type2
```
runarorama, line 75: This allows `'{Foo} Optional a` instead of requiring `'{Foo} (Optional a)`

parser-typechecker/src/Unison/TypePrinter.hs
```diff
@@ -109,37 +109,37 @@ prettyRaw n im p tp = go n im p tp
...
   arrows delay first ((mes, arg) : rest) =
-    arrow delay first mes
-      <> (  parenNoGroup (delay && (not $ null rest))
-         $  go n im 0 arg
-         <> arrows False False rest
-         )
+    arrow delay first mes <> parenNoGroup
+      (delay && not (null rest))
+      (go n im 0 arg <> arrows False False rest)
+
```
runarorama, line 134: These changes are just for readability to help me understand this code. Small improvement.

pchiusano (approved):
👍

CI is failing - looks like unit tests need updating.

unison-src/transcripts-round-trip/main.output.md
```diff
@@ -34,15 +34,15 @@ x = 1 + 1
   most recent, along with the command that got us there. Try:
   
     `fork 2 .old`             
-    `fork #bt17giel42 .old`   to make an old namespace
+    `fork #m41m2ql36i .old`   to make an old namespace
```
aryairani, line 37: Do we expect changed hashes here?

runarorama: I don't know why this changes. The transcript is starting up with a different root hash than before for some reason.

runarorama: This happens in trunk as well, so it's unrelated to this change.

aryairani: Oh weird, ok.

pchiusano:
@runarorama adding the round-trip tests to `stack exec transcripts` will cause CI to fail every time.

There's a separate CI target for the round trip tests, which is run automatically by CI (see [this commit](https://github.com/unisonweb/unison/pull/2385/commits/2ab39a2ac4786e99c4fc3a7bfbf4513353b2c612)). You might want to define an alias for the command it's running.

The reason for the separation is that CI checks that the working directory is clean after running `stack exec transcripts` (to catch transcript changes). The working dir won't be clean after the round trip tests are dumping definitions to a scratch file. We also can't check in the scratch file produced by round-tripping since running the round trip test will push definitions at the top of the scratch file (i.e., it's not idempotent in how it updates the scratch file, nor could it be).

If you don't like that, another option might be to have CI somehow delete the scratch file after running the transcript... we just don't want to be deleting files called `scratch.u` in the developer's working directory. :)

Issue #2416, opened 2021-09-17
upgrade codebase hashing


aryairani:
It is time for our first codebase db migration! 🙃 

- [x] pull out v1 hashing module
- [x] pull out v2 hashing module (probably just a clone of v1)
- [x] fix tests that broke with `TypecheckedUnisonFile` changes
- [ ] create migration harness (check schema version, do things & update version)
- [ ] converting objects
  - [ ] updating objects in place? or create new objects and clean up db after?
  - [ ] update decls and cache v1->v2 constructor mapping
  - [ ] update terms, using constructor mapping
  - [ ] update patches, using constructor mapping
  - [ ] update namespaces, using constructor mapping (namespaces and causals still share a hash)
- [ ] update hard-coded hashes and constructors in runtime
- [x] update hashes in `AuthorInfo.hs`
- [ ] might as well fix #2332 too (`find_type_mentions_index` has redundant rows)

Issue #2373, opened 2021-09-01
standardize representation for hashing (decouple from codebase representation)


aryairani:
We have hashes computed in different places — sometimes in the codebase implementation and sometimes outside of it. We should pick one, IMO passed in, with its own independent representation that's locked to "hashing" version.  (related to #2140)

pchiusano:
Discussed this, decided the following:

* We're going to amend the codebase interface so the codebase is not responsible for doing ANY hashing. Currently, the codebase is doing some type hashing in `putTerm` to populate the type search index. It should not be responsible for that. Not sure what exactly the new signature for `putTerm` should be but it seems doable. Suggestion: `putTermImpl` is the codebase interface function, and then `Codebase.putTerm` is a helper function that calls `putTermImpl` with the type search hashing info.
  * Any implementation of hashing can just be removed from the codebase package.
* The hashes will be different after pulling out the cycle length, which is fine. Will just use the old hashes as IDs until someone changes their implementation.
  * May result in some spurious updates even if a definition is unchanged, but this is pretty rare (if you are `edit`-ing a definition it's probably to... actually make changes to it)
  * We should bump the hash version number used for storing new hashes in the database.
  * We have this other planned hash update https://github.com/unisonweb/unison/issues/2276#issuecomment-889317543 so we should roll both these hash updates into the same release

aryairani:
> The hashes will be different after pulling out the cycle length

I think for this to work without the constructor problems we discussed on Friday, and without rehashing the whole codebase (I don't think it's necessary or desirable yet), the hashes *won't*(/can't) be different.

The hashing code can just convert definitions to a v1 `Hashable` data type (identical to the one we've been using thus far), and produce an unchanged result.  i.e. To avoid footguns, we'll be very selective about what types get `Hashable` instances, and never alter those types or instances except in a provably backwards-compatible way.

The hash scheme only has to change if the v1 `Hashable` data type doesn't capture all the data we want in the hash (e.g. #2276), which doesn't have to be coupled to this or #2140.

PR #2433, opened 2021-09-23
extract hashing modules

aryairani:
- This PR moves the `Term`, `Type`, and `Decl` hashing code into a separate `Unison.Hashing.V2` module, complete with their own copy of `Reference` and all the dependents in between.
- This is a prerequisite of #2416, #2373, #2100, and maybe some others. This PR also incidentally adds a `Unison.Hashing.V1` module which is not used yet, but is a prerequisite of those too and I didn't seem worth deleting for the PR.
- Because hashing is all done on the data types in the `Unison.Hashing.V2` module, you should `import Unison.Hashing.V2.Convert as Hashing` to access the hashing functions there that convert from the default "memory" versions of `Term`, `Type`, and `Decl`
- In consolidating the hashing code, we changed the contract of `TypecheckedUnisonFile.hashTermsId`. This now includes all watches, and adds an extra tuple field that can be used to check the WatchKind, if any, associated with each term.  This is a potential source of gotchas for some older code that assumed that "regular" watches were not included here.
- rewrote `UnisonFile.typecheckedUnisonFile` smart constructor to be less bonkers


pchiusano (approved):

parser-typechecker/src/Unison/Codebase/Editor/AuthorInfo.hs
```diff
@@ -2,58 +2,68 @@
...
+                  (a, Term.text a t)
+                ]
+       in AuthorInfo
+            (guidRef, guidTerm, guidType)
+            (authorRef, authorTerm, authorType)
+            (chRef, chTerm, chType)
+    hashAndWrangle v tm =
+      Foldable.toList $
+        H.hashTermComponents
+          (Map.fromList [(Var.named v, tm)])
```
aryairani, line 62: This is the only real change in this file, I think.  `hashComponents` -> `H.hashTermComponents`.

Issue #2175, opened 2021-07-02
Codebase Server API: Definition summaries


hojberg:
We want to support a notion of a definition summary and source summary which is more compact response than the `/getDefintion` result.

This ticket will support and unblock these UI tickets:
* https://github.com/unisonweb/codebase-ui/issues/112
* https://github.com/unisonweb/codebase-ui/issues/157
* https://github.com/unisonweb/ui-core/issues/22

## Suggested response schema

A `DefinitionSourceSummary` is defined is defined as such:

* Types: up until the `=`: `type List a`
* Abilities: up until `where`: `ability IO`
* Terms: the term signature: `Char -> Text -> [Text]` without the name of the term (example is `Text.split`).

We should support these for all definitions, including builtins (which might be a challenge for types, though we can at least do `type List`)

`DefinitionSourceSummary` will be used in several endpoints:

* The `sourceSummary` field replaces the `termType` and `typeDef` fields of the `/find` endpoint,
* The `sourceSummary` field replaces the `signature` field of `/getDefinition` for terms
* The `sourceSummary` field is added to `/getDefinition` for types
* A new `/definitionSummary` endpoint which includes a the same response shape as `/find`, but for a single definition which can be queried by hash or name—used for hovering references.

hojberg:
To the last point about the new `/definitionSummary` endpoint:

There's an other option to allow hovers on references in source to get the type of them in a tooltip and that is to embed that in the original source. The benefits of this is that we don't have to do an ad-hoc Ajax request on hover and also this opens up the door for a more uniform representation of hovers—we'd eventually want to allow hover on non-references as well such that you can get the type of a sub expression (@rlmark brought up this kind of UX after our last Wednesday hack day)

The major downsides to embedding summaries is performance and download size. Perhaps there's an in-between that allows for both where we can do ad-hoc queries for types of sub-expressions in some way?

cc @pchiusano @runarorama

PR #2434, opened 2021-09-23
Add TermSummary endpoint

hojberg:
## Overview
Add a new endpoint: `/definitions/terms/by_name/:fqn/summary` that will
return a term if one is found in this shape:

```js
{
  "fqn" : "base.List.map",
  "hash" : "#asdf1234",
  "summary" : {
    "tag": "UserObject",
    "contents": [/* ... */ ]
  },
  "tag" : "Plain"
}
```

This will help support hover operations in the Codebase UI and other
similar UIs like the recent VS Code extension work by Tom Sherman.

One part of https://github.com/unisonweb/unison/issues/2175

### URL Design
The URL scheme is designed such that it is as RESTful as possible.
The main outlier in that is the `by_name` prefix. This is done to
ensure sub resources doesn't clash with fqns:

`/definitions/terms/:fqn` with a term name like `histories` would prevent `/definitions/terms/histories` to be a specific sub resource

### Various

This adds a few helper functions to Backend for things like getting the tag of a term (`Doc` vs `Test` vs `Plain`) and also makes termTag a non-maybe field on all definition entities by adding the `Plain` variant.

I only found out about `TermEntry` and `TypeEntry` 90% of the way through this and I think I should likely use that as an intermediate before constructing a `TermSummary`. What do you think @runarorama ?

## Loose ends

* There's another half of this, notably a `TypeSummary` endpoint, that still needs to be done.
* There's quite a few minor helper functions we use on every endpoint that would be nice to put somewhere, like `inBackend`, but not sure where they belong?

### DIFF
--- a/parser-typechecker/src/Unison/Server/Backend.hs
+++ b/parser-typechecker/src/Unison/Server/Backend.hs
@@ -113,6 +113,7 @@ listEntryName = \case
 
 data BackendError
   = NoSuchNamespace Path.Absolute
+  | BadNamespacePath String String
   | BadRootBranch Codebase.GetRootBranchError
   | CouldntExpandBranchHash ShortBranchHash
   | AmbiguousBranchHash ShortBranchHash (Set ShortBranchHash)
@@ -188,7 +189,7 @@ data TermEntry v a = TermEntry
   { termEntryReferent :: Referent,
     termEntryName :: HQ'.HQSegment,
     termEntryType :: Maybe (Type v a),
-    termEntryTag :: Maybe TermTag
+    termEntryTag :: TermTag
   }
   deriving (Eq, Ord, Show, Generic)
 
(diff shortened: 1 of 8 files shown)


parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -286,44 +287,69 @@ findShallowReadmeInBranchAndRender width runtime codebase branch =
...
-  ot <- lift $ loadReferentType codebase r
+-- Indicate what kind of term it is; Doc, Test, or Plain
+getTermTag ::
+  Monad m =>
+  Var v =>
+  Branch0 m ->
+  Referent ->
+  Maybe (Type v Ann) ->
+  Backend m TermTag
+getTermTag b0 r sig = do
```
hojberg, line 298: Pulled out from `termListEntry`

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -286,44 +287,69 @@ findShallowReadmeInBranchAndRender width runtime codebase branch =
...
-typeListEntry codebase r n = do
   -- The tag indicates whether the type is a data declaration or an ability.
-  tag <- case Reference.toId r of
+getTypeTag ::
+  Monad m =>
+  Var v =>
+  Codebase m v Ann ->
+  Reference ->
+  Backend m TypeTag
+getTypeTag codebase r =
```
hojberg, line 335: pulled out from `typeListEntry` (not used yet though, since this PR doesn't add `TypeSummary`).

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -286,44 +287,69 @@ findShallowReadmeInBranchAndRender width runtime codebase branch =
...
+
+typeListEntry ::
+  Monad m =>
+  Var v =>
+  Codebase m v Ann ->
+  Reference ->
+  HQ'.HQSegment ->
+  Backend m TypeEntry
+typeListEntry codebase r n = do
+  tag <- getTypeTag codebase r
```
hojberg, line 352: Diff here is formatting

parser-typechecker/src/Unison/Server/Endpoints/Definitions/Terms/TermSummary.hs
```diff
@@ -0,0 +1,140 @@
...
+    tag :: TermTag
+  }
+  deriving (Generic, Show)
+
+instance ToJSON TermSummary where
+  toEncoding = genericToEncoding defaultOptions
+
+deriving instance ToSchema TermSummary
+
+-- TODO: Should this be using Backend.TermEntry as a way to get to TermSummary?
```
hojberg, line 80: @runarorama thoughts on this?

parser-typechecker/src/Unison/Server/Endpoints/Definitions/Terms/TermSummary.hs
```diff
@@ -0,0 +1,140 @@
...
+  Maybe ShortBranchHash ->
+  Maybe NamespaceFQN ->
+  Maybe Width ->
+  Handler (APIHeaders TermSummary)
+serve tryAuth codebase rawTermName mayRoot namespaceName mayWidth =
+  let inBackend a = do
+        ea <- liftIO $ runExceptT a
+        errFromEither backendError ea
+
+      errFromEither f = either (throwError . f) pure
```
hojberg, line 95: These helpers are used in almost all endpoints.

parser-typechecker/src/Unison/Server/Endpoints/Definitions/Terms/TermSummary.hs
```diff
@@ -0,0 +1,140 @@
...
+            Nothing ->
+              throwError (Backend.MissingSignatureForTerm termReference)
+            Just typeSig -> do
+              hashLength <- liftIO $ Codebase.hashLength codebase
+              let ppe = Backend.basicSuffixifiedNames hashLength branch Path.empty
+              let formattedTermSig = Backend.formatSuffixedType (PrettyPrintEnvDecl ppe ppe) width typeSig
+              let summary = mkSummary termReference formattedTermSig
+              tag <- Backend.getTermTag branch0 termReferent sig
+
+              pure $ TermSummary rawTermName hash summary tag
```
hojberg, line 138: I feel like this is a lot of dense code to achieve this and could use some pointers to clean up this whole function.

parser-typechecker/src/Unison/Server/Endpoints/Definitions/Terms/TermSummary.hs
```diff
@@ -0,0 +1,140 @@
...
+            Nothing ->
+              Backend.resolveRootBranchHash mayRoot codebase
+            Just n -> do
+              namespacePath <- fqnToNamespacePath n
+              root <- Backend.resolveRootBranchHash mayRoot codebase
+              pure $ Branch.getAt' namespacePath root
+
+          let branch0 = Branch.head branch
+          let rel = Branch.deepTerms branch0
+          -- TODO: Map exception of missing element in set to 404 - noSuchDefinition
```
hojberg, line 123: I'm not sure how to do this.

parser-typechecker/src/Unison/Server/Endpoints/Definitions/Terms/TermSummary.hs
```diff
@@ -0,0 +1,140 @@
...
+        pure (Path.fromPath' path')
+
+      width = mayDefaultWidth mayWidth
+
+      mkSummary reference termSig =
+        if Reference.isBuiltin reference
+          then BuiltinObject termSig
+          else UserObject termSig
+   in do
+        -- TODO: Should servant do this Name.unsafeFromString?
```
hojberg, line 109: This feels like a guarantee that'd be nice to do on the edge in Servant param parsing.

parser-typechecker/src/Unison/Server/Errors.hs
```diff
@@ -57,16 +59,30 @@ rootBranchError rbe = err500
...
+          <> mungeString err
+    }
+
 noSuchNamespace :: HashQualifiedName -> ServerError
 noSuchNamespace namespace =
   err404 { errBody = "The namespace " <> munge namespace <> " does not exist." }
 
+noSuchDefinition :: HashQualifiedName -> ServerError
+noSuchDefinition definition  =
+  err404 { errBody = "The definition " <> munge definition  <> " does not exist." }
```
hojberg, line 74: This isn't used yet, but should be the response if we can't find the definition.

parser-typechecker/src/Unison/Server/Types.hs
```diff
@@ -140,7 +140,7 @@ data DefinitionDisplayResults =
     , missingDefinitions :: [HashQualifiedName]
     } deriving (Eq, Show, Generic)
 
-data TermTag = Doc | Test
+data TermTag = Doc | Test | Plain
```
hojberg, line 143: This removed the maybe from a few types

unison-core/src/Unison/Reference.hs
```diff
@@ -180,6 +181,10 @@ groupByComponent refs = done $ foldl' insert Map.empty refs
       Map.unionWith (<>) m (Map.fromList [(Left r, [(k,r)])])
     done m = sortOn snd <$> toList m
 
+isBuiltin :: Reference -> Bool
+isBuiltin (Builtin _) = True
+isBuiltin _ = False
```
hojberg, line 186: There was a `isBuiltinType` somewhere else, but it didn't quite fit the need. Perhaps there is something somewhere that is better to use?

aryairani:
Bump! Is this still desired? What needs to happen to get it merged?

hojberg:
Let's close this for now. We want something for hover abilities in the UI, where the signature of functions are shown, but I've been considering having that hover ability take a modifier key to show the full definition, and with that in mind the current endpoints would work ok.

That said, I do think this PR introduces an API convention thats worth following going forward.

PR #2436, opened 2021-09-24
Numbered list backend doc rendering fix

### LINKED ISSUE #229: Numbered lists in ui show "Sos" errors
Docs seem to have trouble rendering numbered lists. 

```
testDoc = {{

# Test Num list 

1. A
2. B 
3. C 

}}
```

is valid in the UCM, but I've attached a screenshot showing what it looks like in the UI. (Also I'm running the UI which ships with the recent m2h release, so it's possible more features have been added since then, in which case, disregard!)
 
Screen Shot 2021-09-24 at 12 42 32 PM

pchiusano:
I think this fixes https://github.com/unisonweb/codebase-ui/issues/229 but can someone test it out? I have something messed up with my UI install where I'm not able to actually open any definitions. I think `./dev-ui-install.sh` might be busted somehow since that's usually how I refresh the UI code.


pchiusano:
@rlmark you can approve and merge if it's working for you - I'll figure out what's going on with my setup later. :)

rlmark (approved):

PR #2443, opened 2021-09-28
Welcome code revamp which includes authorship

rlmark:
Coauthored with @hojberg 

## Overview

* Onboarding workflow is now expressed in terms of discrete states, with the goal of making it more intelligible / extendable. 
* The welcome message now suggests to the user that they do the authorship step if it is a new codebase. 
* Slight change in ordering of welcome printout for ucm. 

New welcome screen for fresh codebase: 
Screen Shot 2021-09-28 at 4 12 51 PM

When you've already onboarded: 
Screen Shot 2021-09-28 at 4 13 39 PM

"Closes" https://github.com/unisonweb/unison/issues/2348

The authorship workflow as originally intended (with a prompt to the user) is non-trivial to add, so this is a bit of a stop-gap, but I do think it significantly improves the onboarding workflow. 

For reference, the workflow we were working towards is in a draft here: https://gist.github.com/hojberg/f1803050b59c007d3dbaf3bdf8ecf324

## Implementation notes

I added Input/Output types to allow the messages in the Welcome process to have a specific order. This also means Welcome uses the existing mechanisms for communicating with the user as opposed to doing raw IO (sorry Simon 😅) for printing to the console.

## Test coverage

I've use-it tested this change a lot. (I could be convinced to write a test of the Welcome module itself.)


rlmark:
Actually, closing this PR, I think there was something we needed to do with the  `newCodebasePath :: Maybe FilePath` that I've forgotten about in the churn. One sec.

PR #2446, opened 2021-09-30
Standalone Binaries

dolio:
This implements support for a sort of standalone binary executable.

The `compile.output` command will dump a binary `.uc` file with enough state to execute a single `'{IO, Exception} ()` value in the codebase.

Then you can subsequently run it with the exact same `ucm` binary by passing the `run.compiled` option. The `ucm` binary and the compiled output file should be bundled together and considered the whole 'compiled binary'.


dolio:
@stew If you have suggestions about how to pack everything into one file, let me know. I looked for how to read one section out of an ELF, but it doesn't seem like there's anything tailor made for that. It seems like the closest would be to shell out to something like `objcopy ... <myself> /dev/stdout` or something, but that seems a little janky.

There's a Haskell library for parsing ELF, but I think it'd give a bytestring, instead of the handle that the library I'm using expects. Maybe if I looked into the GHC internals there'd be something that would accept a bytestring's contents. Not sure.

pchiusano:
@dolio sweet! Can you work on resolving conflicts vs trunk?

pchiusano (approved):
Sweet! Can't wait to try this out. 

One thing I'd like to see. It looks like the file format is literally just a dump of whatever the compact library produces which I think means not great error messages if you load an incorrect binary. 

Can you add a header at the start of the file format which includes the version number (from `Version.hs` and also `os` and `arch` from https://hackage.haskell.org/package/base-4.15.0.0/docs/System-Info.html). 

Then on load, check that the header is the same and if it's not, complain and give the two headers:

```
I can't run this compiled program since it works with a different version of Unison 
than the one you're running.
   
   Program version   M2g (built on 2021-08-16) for darwin x86_64
   Your version      M2h (built on 2021-09-07) for linux x86_64

The program was compiled from hash #xyz9384. If you have that hash in your codebase,
you can do `.> compile.output #xyz9384 outputname.uc` to produce a new compiled
program that matches your version of Unison.
```

That info I suggested for the version identifier probably isn't enough in all cases... (you'd want to like, hash the executable itself and make sure it matches or something?), but it will catch a lot of the obvious ones so I think it's valuable.

dolio:
Yeah, I'll see if I can make that happen.

pchiusano (approved):

PR #2447, opened 2021-09-30
Welcome onboarding state machine authorship

rlmark:
Coauthored with @hojberg 

## Overview

* Onboarding workflow is now expressed in terms of discrete states, with the goal of making it more intelligible / extendable. 
* The welcome message now suggests to the user that they do the authorship step if it is a new codebase. 
* Slight change in ordering of welcome printout for ucm. 

New welcome screen for fresh codebase: 
Screen Shot 2021-09-28 at 4 12 51 PM

When you've already onboarded: 
Screen Shot 2021-09-28 at 4 13 39 PM

"Closes" https://github.com/unisonweb/unison/issues/2348

The authorship workflow as originally intended (with a prompt to the user) is non-trivial to add, so this is a bit of a stop-gap, but I do think it significantly improves the onboarding workflow. 

For reference, the workflow we were working towards is in a draft here: https://gist.github.com/hojberg/f1803050b59c007d3dbaf3bdf8ecf324

## Implementation notes

I added Input/Output types to allow the messages in the Welcome process to have a specific order. This also means Welcome uses the existing mechanisms for communicating with the user as opposed to doing raw IO (sorry Simon 😅) for printing to the console.

The previous pr had a missing (rare) corner case where a user could create a codebase with the --no-codebase flag, leave the onboarding workflow without adding anything, then start their codebase again, after which the ucm would download base. This version does not do that. 

## Test coverage

I've use-it tested this change a lot. (I could be convinced to write a test of the Welcome module itself.)


pchiusano (approved):
Great! I'll leave this open a day or two in case @hojberg wants to also take a look.

parser-typechecker/src/Unison/CommandLine/Welcome.hs
```diff
@@ -110,3 +126,51 @@ asciiartUnison =
     <> P.cyan "|___|"
     <> P.purple "_|_|"
 
+
+downloading :: Path -> P.Pretty P.ColorText
+downloading path =
+  P.lines
+    [ P.group (P.wrap "🐣 Since this is a fresh codebase, let me download the base library for you." <> P.newline ),
```
hojberg, line 133: 🐣 hah! love it 🎉

hojberg (approved):
:tada: Nice! Having this return inputs for the messages makes a lot of sense in lieu of prompts (maybe a Prompt input down the line 🤔)

PR #2448, opened 2021-10-01
Avoid linebreak between equals sign and `'let`

### LINKED ISSUE #1536: Consider printing 'let on the same line
I.e. print:
```
wat = 'let
  bla
```
instead of
```
wat =
  'let
    bla
```
This had at least 4 upthumbs [on Slack](https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1588879127258700).

runarorama:
Fixes #1536 by adding a "soft hang" element.


parser-typechecker/src/Unison/Util/Pretty.hs
```diff
@@ -668,6 +670,20 @@ string = fromString
...
+  -> Pretty s
+softHang from = softHang' from "  "
+
+softHang' :: (LL.ListLike s Char, IsString s)
+  => Pretty s
+  -> Pretty s
+  -> Pretty s
+  -> Pretty s
+softHang' from by p = group $
+  (from <> " " <> group p) `orElse` (from <> "\n" <> group (indent by p))
```
pchiusano: Maybe a comment on what these do and how they differ from `hang` would be good, along with an example or two.

pchiusano (approved):
Great! I feel like `softHang` could get used in a bunch of places, basically anytime a keyword follows an `=` or `->`.

PR #2450, opened 2021-10-04
Update release-steps.md

pchiusano:
Add an extra step to mention updating the slack post


aryairani (approved):

aryairani:
Over 4 hours of cloud compute is going into making sure this doesn't break the build.  We gotta figure out how to cut that down. 😅

PR #2455, opened 2021-10-05
document and improve performance of innerJoinRanMultimaps

mitchellwrosen:
## Overview

This PR improves the performance of `innerJoinRanMultimaps` (and thus `innerJoinDomMultimaps`) 

I also added a bit of documentation along the way.

## Implementation notes

Uses `Map.intersectionWith`, which has better asymptotics than the hand-rolled equivalent we had before.

## Test coverage

No additional tests, but some could be added.

## Loose ends

I noticed there's also a `U.Util.Relation` module, but I'm not sure how it relates to this one. Should both be receiving the same patches?

### DIFF
--- a/unison-core/src/Unison/Util/Relation.hs
+++ b/unison-core/src/Unison/Util/Relation.hs
@@ -112,23 +112,11 @@ outerJoinRanMultimaps :: (Ord a, Ord b, Ord c)
                       -> Map c (Set a, Set b)
 outerJoinRanMultimaps a b = outerJoinDomMultimaps (swap a) (swap b)
 
--- | @innerJoinDomMultimaps xs ys@ returns the "inner join" of the domains of @xs@ and @ys@.
+-- | @innerJoinDomMultimaps xs ys@ returns the "inner join" of the domains of @xs@ and @ys@, which has intersection-like
+-- semantics:
 --
--- You can think of this function as having the following, higher-level type:
---
--- @
--- innertJoinDomMultimaps
---   :: (Ord a, Ord b, Ord c)
---   => Relation a b
---   => Relation a c
---   => Relation a (Either b c)
--- @
---
--- with the invariant on the output relation that no @a@ will be related to /only/ @Left b@s or @Right c@s; rather, each
--- @a@ will be related to at least one @Left b@ and at least one @Right c@.
---
--- However, this function returns a lower-level @Map a (Set b, Set c)@ instead, for performance reasons, because the
--- calling code currently does not have a need for the higher-level relation.
+-- * @a@s that do not exist in both @xs@ and @ys@ are dropped.
+-- * The @a@s that remain are therefore associated with non-empty sets of @b@s and @c@s.
 --
 -- /O(a2 * log(a1/a2 + 1)), a1 <= a2/, where /a1/ and /a2/ are the numbers of elements in each relation's domain.
 innerJoinDomMultimaps ::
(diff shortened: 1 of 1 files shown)


aryairani:
Okay great. The two modules are basically identical (diff to confirm) and both used, and probably should be consolidated into one module in its own package.  I think initially it wasn’t clear which package it should belong to, so we ended up with two copies.

We could even fork https://github.com/haskell-works/relation or wherever we got it from, but it doesn’t have to be in this PR.

aryairani (approved):

unison-core/src/Unison/Util/Relation.hs
```diff
@@ -112,15 +112,36 @@ outerJoinRanMultimaps :: (Ord a, Ord b, Ord c)
...
+--   => Relation a b
+--   => Relation a c
+--   => Relation a (Either b c)
+-- @
+--
+-- with the invariant on the output relation that no @a@ will be related to /only/ @Left b@s or @Right c@s; rather, each
+-- @a@ will be related to at least one @Left b@ and at least one @Right c@.
+--
+-- However, this function returns a lower-level @Map a (Set b, Set c)@ instead, for performance reasons, because the
+-- calling code currently does not have a need for the higher-level relation.
```
aryairani: I found the explanation kind of confusing. I think the actual type alone is pretty clear though.

mitchellwrosen: Any suggestions for change? :)

(Or were you saying to nix the comment entirely?)

mitchellwrosen:
> Okay great. The two modules are basically identical (diff to confirm) and both used, and probably should be consolidated into one module in its own package. I think initially it wasn’t clear which package it should belong to, so we ended up with two copies.

Sounds good! I can take care of that while I'm in here.

unison-core/src/Unison/Util/Relation.hs
aryairani: Ha, fair. I guess I would nix it entirely (keeping the asymptotics), but I will defer the final decision to you if you find the text clear and helpful.

mitchellwrosen: I don't think the writing is particularly clear or anything, so I'd like to improve it if possible.

But I do want to point out the type doesn't tell the full story - it has the same type as the "outer join" variant(s) above, where one (but not both) of the returned sets may be empty. (c.f. `intersection` / `union` have the same type)

aryairani: Oh, I see, good point.  You could say: 
> Unlike the "outer join" variant(s) above, where one (though not both) of the returned sets may be empty, the sets from "inner join" variants are always non-empty.

or

> For these "inner join" functions, some keys of the resulting map may be missing but all values are nonempty, in contrast to the "outer join" variant(s) above, for which all keys are present but some values may be empty.

Or you could literally use `Data.Set.NonEmpty` in the type (helps for inner join, though doesn't completely solve for outer join).

mitchellwrosen: Nice, I didn't know we were using `Data.Set.NonEmpty`. That sounds like an improvement.

But in the meantime, how does this look?

```
-- | @innerJoinDomMultimaps xs ys@ returns the "inner join" of the domains of @xs@ and @ys@, which has intersection-like
-- semantics:
--
-- * @a@s that do not exist in both @xs@ and @ys@ are dropped .
-- * The @a@s that remain are therefore associated with non-empty sets of @b@s and @c@s.
--
-- /O(a2 * log(a1/a2 + 1)), a1 <= a2/, where /a1/ and /a2/ are the numbers of elements in each relation's domain.
```

aryairani: Looks great!

ChrisPenner (commented):
Yup, really clean approach 👍🏼

PR #2456, opened 2021-10-05
Simplify the codebase server token with NanoID

### LINKED ISSUE #2426: Unison-server should either not use special characters in IDs, or accept their un-escaped versions
There's an annoying issue in my [devcontainer pr](https://github.com/unisonweb/vscode-devcontainer/pull/4/commits/24fb31997eca0472a09c8dca3d11d182c336130c#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R26) caused by what seems to be a [vscode bug](https://github.com/microsoft/vscode/issues/133551). It would be resolved if the IDs generated didn't include special characters that had to be escaped. (I wonder if they're encoded in base64, and you just need to use a url-safe encoding instead?)

hojberg:
## Overview
Fixes https://github.com/unisonweb/unison/issues/2426

Change the generated codebase server token to be a NanoID of 6 characters.

NanoID is a quick way to generate ids in a similar fashion to UUIDs with a [similar collision probability](https://github.com/ai/nanoid#comparison-with-uuid). NanoID also has a lot of implementations and seems quite popular. Read more here: https://zelark.github.io/nano-id-cc/

## Implementation notes
A length of 6 was chosen to have a small enough chance of collision (~2 days needed, in order to have a 1% probability of at least one collision according to their website) and minimal impact on how URLs look. NanoIDs also don't include a lot of special characters (just `_` and `-`), making them great for URLs.

We could likely make this as small as 4 or 5 characters, given the use case of the codebase server token.

The [NanoID library](https://hackage.haskell.org/package/NanoID) wasn't in the most recent Stack so and is thus included in the stack.yml file. There was another nanoid implementation, but it didn't exist on Hackage at all.

I explored [hashids](http://hashids.org) as well, but having to use a salt and a number felt annoying.

## Test coverage
Tested with `stack exec unison headless`.

## Edit

Updated to reflect decision to use 16 characters (which at 1000 request/second would take ~1000 years to collide)


pchiusano:
The token is a security feature to prevent another user on the same machine from being able to view your codebase. If we still care about that, which I think we do, then reducing the length to 6 basically nullifies that security feature - you could check 1000 tokens per second and have a 1% chance of finding the token in 37 seconds. :)

If you think the url is prettier with this implementation, that's fine, but I'm kind of hesitant to reduce the length.

One thing you could do is in the event of an incorrect token, take 1 second to respond with a failure, or even double the response time after each failure, first 1 second, then 2, 4, 8... then you can probably reduce the length without impacting security. Like at 10 characters, one trial per second, it would take 5 years to reach 1%. I can't imagine keeping UCM open for 5 years without restarting. :)

Actually, I'm not sure that does it since an attacker could treat a delayed reply as a failure, and do a much faster rate that way.

pchiusano:
What about just a `--no-token` flag for people running on a single-user machine?

aryairani:
I think it can be hard to determine what is a true "single-user" machine.  But for our DOS users, that could work :)

@stew WDYT?

pchiusano:
We discussed bumping the length to 16 but maybe in a subsequent PR flip the default so that by default no token gets used, but one could be used via `ucm --with-token` or some such. Anyone have strong opinions on that?

hojberg:
@pchiusano updated to reflect the 16 character length.

pchiusano (approved):

PR #2458, opened 2021-10-05
Improve the performance of `add` with default metadata

mitchellwrosen:
## Overview

This PR improves the performance of `add` in the case that there is default metadata to add.

## Implementation notes

Previously, if there was any default metadata to add to each of the new types/terms, `add` would:

- For each different stringy name of metadata (e.g. ".metadata.authors.mitchell"), 
  - Resolve it to an unambiguous term, and look up that term's type.
  - If that worked, add the metadata type/value to the branch for each new type/term
- Diff the current branch with the one prior to (possibly) adding any metadata
- If anything changed, report "I added some metadata" to the terminal

Since we are (currently) just reporting whether we add any metadata at all, this algorithm could be improved by eliding the full branch diff, which is slow. Instead, we (almost) know that by calling the "add metadata" function at all, we have metadata to add to each type/term, and we can just report "I added some metadata" anyway.

Small change in behavior: 

Previously, default metadata would be added to a type/term on a best-effort basis. For example, if I have two default metadata names "A" and "B", but only "A" names a real term whereas "B" does not, then newly-added types/terms would still be linked to metadata "A".

Now, if any of the default metadata names do not refer to a term, no metadata is added to a term.

This change was only made because it made the implementation simpler, and didn't seem significantly or semantically worse than the existing mechanism. However, if we actually intended for "partial" metadata to be added to new types/terms in the way described above, the implementation can change.

## Test coverage

No new automated tests, but I tested this manually, and I'm sure the existing tests cover any potential bugs.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
@@ -241,18 +241,6 @@ loop = do
       getHQ'Terms p = BranchUtil.getTerm (resolveSplit' p) root0
       getHQ'Types :: Path.HQSplit' -> Set Reference
       getHQ'Types p = BranchUtil.getType (resolveSplit' p) root0
-      getHQTerms :: HQ.HashQualified Name -> Action' m v (Set Referent)
-      getHQTerms hq = case hq of
-        HQ.NameOnly n -> let
-          -- absolute-ify the name, then lookup in deepTerms of root
-          path :: Path.Path'
-          path = Path.fromName' n
-          Path.Absolute absPath = resolveToAbsolute path
-          in pure $ R.lookupRan (Path.toName absPath) (Branch.deepTerms root0)
-        HQ.HashOnly sh -> hashOnly sh
-        HQ.HashQualified _ sh -> hashOnly sh
-        where
-        hashOnly sh = eval $ TermReferentsByShortHash sh
 
       basicPrettyPrintNames0 =
         Backend.basicPrettyPrintNames0 root' (Path.unabsolute currentPath')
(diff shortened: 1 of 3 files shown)


parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -241,18 +241,6 @@ loop = do
       getHQ'Terms p = BranchUtil.getTerm (resolveSplit' p) root0
       getHQ'Types :: Path.HQSplit' -> Set Reference
       getHQ'Types p = BranchUtil.getType (resolveSplit' p) root0
-      getHQTerms :: HQ.HashQualified Name -> Action' m v (Set Referent)
```
mitchellwrosen, line 244: This was moved to the top-level in an effort to start breaking up this enormous function :)

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -560,61 +548,57 @@ loop = do
         -- `mdValues` is (names of the) metadata to pass to `op`
         -- `op` is the operation to add/remove/alter metadata mappings.
         --   e.g. `Metadata.insert` is passed to add metadata links.
-        manageLinks :: Bool
```
mitchellwrosen, line 563: The type was formatted, but not changed.

parser-typechecker/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -560,61 +548,57 @@ loop = do
...
-                 respond DefaultMetadataNotification
+          runExceptT (for mdValues \val -> ExceptT (getMetadataFromName val)) >>= \case
+            Left output -> respond output
+            Right metadata -> do
+              before <- Branch.head <$> use root
+              traverse_ go metadata
+              if silent
+                then respond DefaultMetadataNotification
+                else do
+                  after <- Branch.head <$> use root
```
mitchellwrosen, line 571: No change in behavior in the "else" branch here. The key difference is that we only perform a branch diff here, but not in the "then" branch.

unison-core/src/Unison/Util/Relation.hs
```diff
@@ -112,15 +112,25 @@ outerJoinRanMultimaps :: (Ord a, Ord b, Ord c)
                       -> Map c (Set a, Set b)
 outerJoinRanMultimaps a b = outerJoinDomMultimaps (swap a) (swap b)
 
-innerJoinDomMultimaps :: (Ord a, Ord b, Ord c)
```
mitchellwrosen, line 115: Oops, I branched off of another PR that hasn't merged yet. Will remove this from the diff after that lands.

unison-src/transcripts/ambiguous-metadata.output.md
```diff
@@ -37,6 +37,4 @@ x = 1
   Tip: Try again and supply one of the above definitions
        explicitly.
 
-  I didn't make any changes.
```
mitchellwrosen, line 40: This transcript is kinda better now, per the "slight change in behavior" mentioned in the PR description. Previously, we would fail to resolve the metadata name to a single value, but then proceed anyway with a branch diff, which is empty, hence the `I didn't make any changes.` output due to a `respond NoOp`.

But now, we don't even bother to perform the branch diff, because it is not the case that each of the metadata names (in this case, only one) resolve to a single reference.

aryairani (approved):
LGTM.

pchiusano:
Nice!!

PR #2459, opened 2021-10-05
Updates version-parser for release tags

rlmark:
Changes to the version-parser to support base-downloads when given official release versions. 
With author: @ChrisPenner🚀  ⭐ 

## Overview

Stops including the date built in the base version passed to base download. 

## Implementation notes

Parser changes - returning a tuple instead of a concatenated string with date.

## Test coverage

Tests were updated/added


pchiusano (approved):
👍

PR #2468, opened 2021-10-06
Improve performance of `R4` -> `R3` conversion

mitchellwrosen:
## Overview

In at least one place in the code, we were converting from a `Relation4 a b c d` to a `Relation3 a b d` by round-tripping through a list.

There is a simple and asymptotically more efficient way to do this projection, which I've implemented in this PR. 

In my little test, which just calls `link foo bar` six times in a transcript, I've observed the following performance difference:

|  | Old | New | Improvement |
| --- | --- | --- | --- |
| Memory | 51% of branch diff | 1% of branch diff | 51x |
| Time | 50% of branch diff | 2% of branch diff | 25x |

The numbers here are the percentage of time/memory consumed by the branch diff calculation that this particular `R4 -> R3` accounts for.

## Test coverage

I've added tests for this function that assert the implementation matches the version that round trips through a list.

## Loose ends

~Depends on #2462; should not be merged until after.~


.github/workflows/ci.yaml
```diff
@@ -85,3 +85,5 @@ jobs:
           x=`git status --porcelain -uno` bash -c 'if [[ -n $x ]]; then echo "$x" && false; fi'
       - name: prettyprint-round-trip
         run: stack --no-terminal exec unison transcript unison-src/transcripts-round-trip/main.md
+      - name: other test suites
```
mitchellwrosen, line 88: `unison-util-relation` has its own test suite, run with `stack test [unison-util-relation]`. Is that ok?

lib/unison-util-relation/src/Unison/Util/Relation.hs
```diff
@@ -136,6 +137,16 @@ data Relation a b = Relation
...
 
+-- | Construct a relation from a mapping from the domain and range mappings.
+--
+-- /Precondition/: the multimaps together form a valid relation; i.e. if @x@ is related to @y@ in one map then @y@ is
+-- related to @x@ in the other.
+--
+-- /O(1)/.
+unsafeFromMultimaps :: Map a (Set b) -> Map b (Set a) -> Relation a b
+unsafeFromMultimaps domain range =
+  Relation {domain, range}
```
aryairani, line 148: Is this better than `Relation domain range` w/o `NamedFieldPuns`?

aryairani (approved):
I left a question about the language extension but otherwise LGTM.

lib/unison-util-relation/src/Unison/Util/Relation.hs
mitchellwrosen: I wouldn't say it's that much better. Just a somewhat common way of constructing a record because the order of the fields is irrelevant.

pchiusano (approved):
Wowser, nice!!

pchiusano (approved):

lib/unison-util-relation/src/Unison/Util/Relation.hs
```diff
@@ -385,10 +396,14 @@ manyRan :: Ord b => b -> Relation a b -> Bool
 manyRan b = (> 1) . S.size . lookupRan b
 
 -- | Returns the domain in the relation, as a Set, in its entirety.
+--
+-- /O(a)/.
```
pchiusano, line 400: I was curious about this. If you check https://hackage.haskell.org/package/containers-0.6.5.1/docs/src/Data.Map.Internal.html#keysSet you'll notice it does no rebalancing - `Map` and `Set` have identical structure, `Map` just has the extra field in `Bin` for the value and `keySet` throws it away. 

If `Set` were lazy, the call to `dom` would be O(1) and then it will incrementally copy over just the parts of the set you end up using. For any subtree of the set you access, it would be an extra O(1) factor to do the lazy copying over from the original map.

But it looks like `Set` is strict (probably for space leak reasons), so nevermind. :)

mitchellwrosen: Hmm... wouldn't it be O(a) regardless?

aryairani: Laziness is weird. Seems like it could be O(1) now and O(a) later, or O(a) now. Is that what you mean by "regardless" @mitchellwrosen? In any case, it's nice that it's not O(a log a).

mitchellwrosen: Yeah, I meant regardless of if that Set has bang-patterns on its constructors, because if it doesn't, I think the operation overall has enqueued O(a) worth of work, whether or not it's actually forced eventually :)

Analogously I think you would say List.map is O(n) in the length of the list

pchiusano: If it were lazy, I believe it would be O(1) amortized. The operation itself takes O(1) time and for any computation you do on the returned set, its asymptotic complexity is the same as if the set were already forced - there's just an extra constant factor. Constant factors are a thing but they don't impact big-O.

This is because it takes O(1) to look at a subtree even in a fully forced Set - so you can do a constant amount of forcing work and by the banker's method still end up with a non-negative balance.

https://www.cs.cornell.edu/courses/cs3110/2020fa/textbook/eff/amortized_methods.html

mitchellwrosen: Thanks for the link, I have some reading to do :)

PR #2470, opened 2021-10-06
Improve performance of Relation.difference, Relation3.difference

mitchellwrosen:
## Overview

This PR improves the performance of `Relation.difference` and `Relation3.difference`, which are both used in calculating a branch diff.

Performance numbers (relative to #2468, not trunk):

|  | Old | New | Improvement |
| --- | --- | --- | --- |
| Memory | 93% of diff algorithm | 52% of diff algorithm | 1.8x |
| Time | 92% of diff algorithm | 59% of diff algorithm | 1.6x |

## Test coverage

No additional tests.

## Loose ends

Depends on #2468


aryairani (approved):
Did a quick pair-review with Mitchell and convinced that it's right without unit tests.

ChrisPenner (approved):
These changes look great 👍🏼 

I notice that [here](https://github.com/unisonweb/unison/blob/903ba3a7ea7f391250b3143b0a580109fc82cf0d/parser-typechecker/src/Unison/Codebase/Branch.hs#L588-L592) we compute both sides of the diff, e.g. both (a - b) AND (b - a); I'm wondering if we can save even more time by doing it in a single pass?

Might be able to leverage [semialign](https://hackage.haskell.org/package/semialign-1.2) for this.

pchiusano:
Cool. You look into the single pass both sides of difference in a later PR maybe, since just this PR on its own is a clear win.

pchiusano (approved):

mitchellwrosen:
I thought about it for a bit, but couldn't figure out how to more efficiently compute A-B and B-A than what's already there. 🤔

Can you say more?

PR #2494, opened 2021-10-13
Implement by-hand serialization for standalone binary files

dolio:
Switched out the compact region serialization for a hand written serializer on the same structure, since the compact region version doesn't seem to work on some platforms.


cli/package.yaml
```diff
@@ -28,6 +28,7 @@ executables:
     ghc-options: -threaded -rtsopts -with-rtsopts=-I0 -optP-Wno-nonportable-include-path
     dependencies:
       - base
+      - bytestring
       - compact
```
pchiusano: Should we remove `compact` as a dependency now? Is it still being used anywhere?

cli/unison-cli.cabal
```diff
@@ -84,6 +84,7 @@ executable unison
   ghc-options: -threaded -rtsopts -with-rtsopts=-I0 -optP-Wno-nonportable-include-path
   build-depends:
       base
+    , bytestring
     , compact
```
pchiusano: Also here, do we still need `compact` here?

cli/unison/Main.hs
```diff
@@ -127,22 +127,29 @@ main = do
...
-             Right (getCompact -> (w, sto)) -> RTI.runStandalone sto w
+           | otherwise -> RTI.runStandalone sto w
            where
-           vmatch = v == Version.gitDescribeWithDate
-                 && o == os
-                 && a == arch
-           ws s = P.wrap (P.string s)
+           vmatch = v == Text.pack Version.gitDescribeWithDate
+                 && o == Text.pack os
+                 && a == Text.pack arch
```
pchiusano: It seems like we can get rid of the `os` and `arch` checks now, right? Only the Unison version matters? 

This is nice as the compiled files are platform independent.

Also, is it true to say that if no new builtins are added between Unison versions, you can run a compiled file from an old version? I don't think we should try to track that, but just wondering.

pchiusano (approved):
Cool! 

I left a couple comments for you re: removing the compact dependency which I think is unused now? Also it seems like we could relax the version check some since these files are platform and architecture independent, right?

When will a change actually invalidate the files? Is it just when new builtins get added? I think it is fine for the check to be conservative but I'm just wondering.

It would be nice to get some testing of this in CI. It could be a later PR though.

cli/unison/Main.hs
dolio: Hmm, yeah, that's probably correct.

pchiusano:
@dolio what do you want to do about testing? Can you think of an easy way to do it? I'd be okay with punting for now (if so please file a separate issue to track) but would be more comfortable if we had at least 1 end-to-end test of the functionality.

dolio:
I don't know how to test this easily. Testing just the serialization would require implementing random generators for multiple different code representation types that don't have them. Maybe it'd be easier to somehow call the binaries somewhere, but I'm not sure how to put that would require being able to run it with a codebase that has something appropriate, and I'm also not very knowledgeable about how to get the CI to do that sort of stuff, so I'd have to learn that.

pchiusano:
Okay cool, thanks for creating issue to track #2497 

Going to merge!

PR #2500, opened 2021-10-14
Add some special cases for universal compare

### LINKED ISSUE #2083: Universal.< gives the wrong answer for large Nat values
The first of these is wrong:

```
    2 | > (Universal.<) 9223372036854775808 0
          ⧩
          true
  
    3 | > (Nat.lt) 9223372036854775808 0
          ⧩
          false
```
At a guess, `Universal.<` seems to be comparing Nats as Ints.  9223372036854775808 (2^63) is the smallest Nat that doesn't fit in an Int - interpreted as an Int, that bit pattern means -9223372036854775808, which would fit with the behaviour above.

Nats go up to 2^64-1, so it's the upper half of the range that has the problem.  Presumably the other comparison operators are affected too.

This causes a failure in a property test for me, and the test used to work.  But adding the `--old-runtime` flag (is that a thing?) and retrying (with cleaned cache) still shows the failure, so I don't understand what's made the change.  

I can work around by switching to `Nat.lt` (which is probably better anyway - I didn't realise I was using universal comparison til I dug into this...)

This is on the M2 release candidate from 4-Jun.

(Code reading, I got as far as `universalCompare` in Stack.hs and then gave up.  But I do see that the `Closure` type in that file is using `Int` not `Nat`...)

tagging @dolio

### LINKED ISSUE #2257: Universal.compare of negative floats is backwards
I don't expect `-1.0 < -2.0` to be true.

```
# scratch.u
> 0.0 < 0.0
> 0.0 < 1.0
> 1.0 < 2.0
> +0.0 < +1.0
> +1.0 < +2.0
> -1.0 < 0.0
> -2.0 < -1.0
> -1.0 < -2.0

# ucm


    1 | > 0.0 < 0.0
          ⧩
          false

    2 | > 0.0 < 1.0
          ⧩
          true

    3 | > 1.0 < 2.0
          ⧩
          true

    4 | > +0.0 < +1.0
          ⧩
          true

    5 | > +1.0 < +2.0
          ⧩
          true

    6 | > -1.0 < 0.0
          ⧩
          true

    7 | > -2.0 < -1.0
          ⧩
          false

    8 | > -1.0 < -2.0
          ⧩
          true
```

dolio:
Handle `Float` and `Nat` specifically, since comparing them as if they were `Int` doesn't work correctly.

Addresses #2257


parser-typechecker/src/Unison/Runtime/Stack.hs
```diff
@@ -193,6 +193,21 @@ universalEq frn = eqc False
...
+  fi = indexByteArray ba 0
+  fj = indexByteArray ba 1
+
+compareAsNat :: Int -> Int -> Ordering
+compareAsNat i j = compare ni nj
+  where
+  ba = byteArrayFromList [i,j]
+  ni, nj :: Word
+  ni = indexByteArray ba 0
+  nj = indexByteArray ba 1
```
pchiusano: Is there no more direct way to do this, like casting from `Int` to `Float` / `Nat` ?

parser-typechecker/src/Unison/Runtime/Stack.hs
```diff
@@ -203,6 +218,9 @@ universalCompare frn = cmpc False
   where
   cmpl cm l r
     = compare (length l) (length r) <> fold (zipWith cm l r)
+  cmpc _ (DataC rf1 0 [i] []) (DataC rf2 0 [j] [])
+    | rf1 == Ty.floatRef && rf2 == Ty.floatRef = compareAsFloat i j
+    | rf1 == Ty.natRef && rf2 == Ty.natRef = compareAsNat i j
```
pchiusano: Bummer that we're having to compare full `Reference` values here, rather than a int tag. It slows all other comparisons down and when there's a hit it has to look at all 512 bits of the hash to prove that. 

Do we not currently have an integer tag for the type of boxed values?

pchiusano: One thought I had if we're missing an int tag for boxed types - just comparing the first K bytes of the hash is sufficient to establish equality for this purpose. Like K = 10 is  2^80 = 1,208,925,819,614,629,174,706,176... how many types would we need to hit a 1% chance of collision?

pchiusano (approved):

parser-typechecker/src/Unison/Runtime/Stack.hs
dolio: Hmm, I think I can do better. `DataC` isn't the most direct way to access these things.

Just casting `Int` to `Double` is not guaranteed to work, though.

dolio: Oh, but actually, the way these are going to be represented in practice is no better. We'll get `Int` values, because they will be specifically built with constructors for 1 unboxed value.

parser-typechecker/src/Unison/Runtime/Stack.hs
dolio: I don't really understand what you're asking. The reference checks already only occur when both arguments have exactly 1 unboxed field, and have tag 0. So, only stuff like `Int` `Nat` and `Float`, basically, if that's what you're worried about.

pchiusano: Ah, yeah, I was thinking it slowed everything down. Also ignore what I said about hashes, since these are builtin types.

When that pattern is matched, though it will end up doing a text comparison of the text `##Nat` or `##Float` (because `Ty.natRef` is `Builtin "##Nat"`) for every universal comparison of those kinds of boxed values. Which is a bit silly. 

Do you not have an integer tag for the _type_ of the thing? (I don't mean the constructor id tag, which is 0 in this case)

parser-typechecker/src/Unison/Runtime/Stack.hs
pchiusano: I mean the whole byte array business. Why not just use:

https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Float.html#v:int2Double 

To go from `Int` to `Double` safely. And then not sure how to go from `Int -> Word` bitwise... will `fromIntegral` do it? Or does that truncate at 0?

dolio: `int2Double` is like `1 -> 1.0`. `fromIntegral` should work for words, though, yeah.

parser-typechecker/src/Unison/Runtime/Stack.hs
dolio: At some point, I was packing some type information into the data tags. But I think that got eliminated when the entire reference was put in there.

dolio: I can try to add the type information back in. I think maybe the reason I got rid of it was that it was more difficult to handle for serialization. I think it's possible, though. Just a bit tricky.

pchiusano: Okay, cool. I'd say don't sweat it for now. We aren't even sure it's much slower.

Something we could also consider for later is just not having `Text` in the `Builtin` side of a reference - use an enum which we can convert to and from text. This could speed up the comparisons maybe but also give us coverage checking in places where we're inspecting builtins.

parser-typechecker/src/Unison/Runtime/Stack.hs
pchiusano: Ah, gotcha. Is there really no more direct way to bitwise cast from `Int <-> Double`? There's gotta be.

I just thought of something else, which is that I want `Universal.compare` to be a total order for floats, not the IEEE partial order thing. If you want the partial order you can always use `Float.<` etc. Basically keep the Universal functions as structural/lexicographical.

I think you just have to flip the sign bit then compare the ints?

PR #2533, opened 2021-10-21
Avoid adding indentation to raw code blocks in doc on roundtrip

### LINKED ISSUE #2271: `raw` multiline codeblocks in the docs indent themselves weirdly
Upon saving a Doc with a `raw` codeblock element, the element will indent the code successively. 

```
type Book = Book [Text]

flipToPage: Nat -> Book -> Optional Text
flipToPage desiredPage book = 
    match book with 
        Book pages -> List.at desiredPage 
        _ -> None  
```
becomes: 
``` raw
         type Book = Book [Text]
                
                        flipToPage: Nat -> Book -> Optional Text
                        flipToPage desiredPage book = 
                          match book with 
                            Book pages -> List.at desiredPage 
                            _ -> None  
```

Odd indentation is present upon viewing source in the console, and editing code once saved into the ucm.

runarorama:
Fixes #2271 

## Overview

This makes it so that the lexer detects the indentation level of the code fence, and it strips whitespace from the front of all lines inside the fence, according to that indentation level. This is instead of what it was doing before, which is trimming all whitespace from the _first line_ of the block, and otherwise leaving indentation alone.

Handling of tabs is sort of ad-hoc. It's not well defined what it means for indentation if the fence is indented by spaces but the contents are indented by tabs. In that case, we leave the indentation inside alone.

## Test coverage

A transcript is provided.

## Loose ends

We're not sure if other types of code blocks need the same treatment. We think not.


pchiusano (approved):
Great! Looks like the transcripts just need a refresh.

PR #2535, opened 2021-10-22
fix getCurrentDirectory

iamevn:
Trying to use `base.io.getCurrentDirectory` results in the following before this change:

```
.> view getCurrentDirectory

  base.io.getCurrentDirectory : '{IO, Exception} FilePath
  base.io.getCurrentDirectory = compose (compose FilePath reraise) getCurrentDirectory.impl

.> view printCurrentDir

  printCurrentDir : '{IO, Exception} ()
  printCurrentDir = compose (compose printLine FilePath.toText) getCurrentDirectory

.> run printCurrentDir

  applying non-function: Foreign (Wrap ##Text "/home/evan/unison")
```

After this change:
```
.> run printCurrentDir
/home/evan/unison
```

Not sure what tests to write, could I get some pointers for what would be most useful here?

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Builtin.hs
+++ b/parser-typechecker/src/Unison/Runtime/Builtin.hs
@@ -1596,7 +1596,7 @@ declareForeigns = do
        temp <- getTemporaryDirectory
        createTempDirectory temp prefix
 
-  declareForeign "IO.getCurrentDirectory.impl.v3" direct
+  declareForeign "IO.getCurrentDirectory.impl.v3" unitToEFBox
     . mkForeignIOF $ \() -> getCurrentDirectory
 
   declareForeign "IO.setCurrentDirectory.impl.v3" boxToEF0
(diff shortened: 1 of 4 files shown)


pchiusano:
@iamevn great! Could you add yourself to CONTRIBUTORS and also add a test to the `unison-src/transcripts/io.md` file?

iamevn:
Done! I also added a test for the `getTempDirectory` builtin as a sanity check. `testGetTempDirectory` passes against trunk and this branch, `testGetCurrentDirectory` fails against trunk and passes on this branch.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1596,7 +1596,7 @@ declareForeigns = do
        temp <- getTemporaryDirectory
        createTempDirectory temp prefix
 
-  declareForeign "IO.getCurrentDirectory.impl.v3" direct
+  declareForeign "IO.getCurrentDirectory.impl.v3" unitToEFBox
```
iamevn, line 1599: Should this version number be bumped up?

pchiusano: Nope, not in this case. 

We bump the version number if the old version did something reasonable and the new version does something different and/or has a different type signature. Bugfixes that everyone would want can just be made in place without a version bump so that anyone using the old version (I'd assume no one in this particular case :) gets upgraded to the new one.

pchiusano (approved):
Great, thanks! Will merge after CI passes.

PR #2538, opened 2021-10-22
Fix duplicate find_type(_mentions)_index entries

### LINKED ISSUE #2332: find_type_mentions_index table has redundant rows

mitchellwrosen:
## Overview

This PR fixes #2332.

We had duplicate rows in the `find_type_index` and `find_type_mentions_index` due to code that looked like:

```
for each member of a component,
  do some per-member stuff
  copy over the whole component's index rows
```

which was de-dented to become

```
for each member of a component,
  do some per-member stuff
copy over the whole component's index rows
```

Additionally, to guard against syncing from old databases that have this property, we now `nubOrd` the list index rows before copying over. As mentioned in a comment in the source, this could also be accomplished by putting a unique index on the index tables for the whole 6-tuple, if we want to do that some day.

## Test coverage

No automated tests were added. I did manually confirm this PR passes the Stew Test from #2332:

```
sqlite> select count(*) from find_type_index;
19206
sqlite> select count(*) from (select distinct type_reference_builtin,type_reference_hash_id,type_reference_component_index, term_referent_object_id,term_referent_component_index ,term_referent_constructor_index from find_type_index);
19206
sqlite> select count(*) from find_type_mentions_index;
112326
sqlite> select count(*) from (select distinct type_reference_builtin,type_reference_hash_id,type_reference_component_index, term_referent_object_id,term_referent_component_index ,term_referent_constructor_index from find_type_mentions_index);
112326
```

Note we have two pairs of equal numbers. Previously, the `count(*)` was far greater than the distinct `count(*)` on this database I'm testing against.

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Sync22.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Sync22.hs
@@ -9,27 +9,19 @@
 
 module U.Codebase.Sqlite.Sync22 where
 
-import Control.Monad (when)
 import Control.Monad.Except (ExceptT, MonadError (throwError))
 import qualified Control.Monad.Except as Except
-import Control.Monad.Extra (ifM)
-import Control.Monad.RWS (MonadIO, MonadReader, lift)
+import Control.Monad.RWS (MonadReader)
 import Control.Monad.Reader (ReaderT)
 import qualified Control.Monad.Reader as Reader
 import Control.Monad.Validate (ValidateT, runValidateT)
 import qualified Control.Monad.Validate as Validate
 import Data.Bifunctor (bimap)
 import Data.Bitraversable (bitraverse)
-import Data.ByteString (ByteString)
 import Data.Bytes.Get (getWord8, runGetS)
 import Data.Bytes.Put (putWord8, runPutS)
-import Data.Foldable (for_, toList, traverse_)
-import Data.Functor ((<&>))
 import Data.List.Extra (nubOrd)
-import Data.Maybe (catMaybes, fromMaybe)
-import Data.Set (Set)
 import qualified Data.Set as Set
-import Debug.Trace (traceM, trace)
 import qualified U.Codebase.Reference as Reference
 import qualified U.Codebase.Sqlite.Branch.Format as BL
 import U.Codebase.Sqlite.Connection (Connection)
@@ -48,6 +40,7 @@ import qualified U.Codebase.Sync as Sync
 import qualified U.Codebase.WatchKind as WK
 import U.Util.Cache (Cache)
 import qualified U.Util.Cache as Cache
+import Unison.Prelude
 
 data Entity
   = O ObjectId
@@ -170,16 +163,13 @@ trySync tCache hCache oCache cCache = \case
             lift do
               -- copy reference-specific stuff
               for_ [0 .. length localIds - 1] \(fromIntegral -> idx) -> do
+                let ref = Reference.Id oId idx
+                    refH = Reference.Id hId idx
+                    ref' = Reference.Id oId' idx
                 -- sync watch results
                 for_ [WK.TestWatch] \wk ->
-                  syncWatch wk (Reference.Id hId idx)
-                -- sync dependencies index
-                let ref = Reference.Id oId idx
-                    ref' = Reference.Id oId' idx
-                let fromJust' = fromMaybe (error "missing objects should've been caught by `foldLocalIds` above")
-                runSrc (Q.getDependenciesForDependent ref)
-                  >>= traverse (fmap fromJust' . isSyncedObjectReference)
-                  >>= runDest . traverse_ (flip Q.addToDependentsIndex ref')
+                  syncWatch wk refH
+                syncDependenciesIndex ref ref'
               syncTypeIndex oId oId'
               syncTypeMentionsIndex oId oId'
             pure oId'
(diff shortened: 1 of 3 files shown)


pchiusano (approved):
Wow, good catch.

It might be nice to have a "rebuild index" command at some point. 

Would one way to repair an existing codebase is just to push its root to an empty codebase? And then the `nubOrd` will take care of deduping, and then going forward `ucm` won't produce duplicates.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Sync22.hs
```diff
@@ -167,26 +167,21 @@ trySync tCache hCache oCache cCache = \case
...
+                -- sync watch results
+                for_ [WK.TestWatch] \wk ->
+                  syncWatch wk (Reference.Id hId idx)
+                -- sync dependencies index
+                let ref = Reference.Id oId idx
+                    ref' = Reference.Id oId' idx
+                let fromJust' = fromMaybe (error "missing objects should've been caught by `foldLocalIds` above")
+                runSrc (Q.getDependenciesForDependent ref)
+                  >>= traverse (fmap fromJust' . isSyncedObjectReference)
+                  >>= runDest . traverse_ (flip Q.addToDependentsIndex ref')
```
aryairani: We should split this and the other spot into its own `syncDependenciesIndex` function too, to match. Can be here or a separate PR after.

aryairani (approved):
Awesome! Go ahead and merge when ready.

aryairani:
@mitchellwrosen So is it correct that a codebase with existing bloat can be fixed by pushing the whole thing somewhere and the pulling to a new codebase?  /cc @stew 

This workaround is also a temporary one, because all codebases will have their indexes fixed during the codebase schema migration.

mitchellwrosen:
@pchiusano / @aryairani Yeah, that'd work (at least in my head)

aryairani:
Funny, I totally missed that Paul had already asked the same question.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Sync22.hs
mitchellwrosen: I added this in https://github.com/unisonweb/unison/pull/2538/commits/0f7b20d921a28b0b829ecc436819bf04fc30c9ed, and pulled in `unison-prelude` as a dependency of `codebase-sqlite`, which I think is ok, because it should be at the root of everything and not contain anything parser-typechecker specific.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Sync22.hs
```diff
@@ -346,6 +339,13 @@ trySync tCache hCache oCache cCache = \case
     isSyncedObjectReferenceId (Reference.Id oId idx) =
       isSyncedObject oId <&> fmap (\oId' -> Reference.Id oId' idx)
 
+    -- Assert that a reference's component is already synced, and return the corresponding reference.
+    expectSyncedObjectReference :: Sqlite.Reference -> m Sqlite.Reference
+    expectSyncedObjectReference ref = undefined
+      isSyncedObjectReference ref <&> \case
+        Nothing -> error (reportBug "E452280" ("unsynced object reference " ++ show ref))
+        Just ref' -> ref'
+
```
aryairani: I'm surprised no existing push-pull tests failed on this `undefined`, what am I missing?

aryairani (changes requested):

codebase2/codebase-sqlite/U/Codebase/Sqlite/Sync22.hs
aryairani: Disregard; they did fail on this `undefined`.

aryairani (approved):

ChrisPenner (approved):
Thanks for doing this; I think this provides us with at least two object lessons:

1. Aggressively pull out helpers so that the control-flow and nesting at the top level becomes clear (e.g. operate at a single layer of abstraction at a time)
2. Add as many unique indexes as we can, even if we _think_ we don't need them

aryairani:
> 2. Add as many unique indexes as we can, even if we _think_ we don't need them

I'd push back on this somewhat — I suspect I left this out on purpose to reduce the size of the index on disk (and I'd want to do a space comparison before adding it back in), though in that case I ought to have left a comment with some concrete numbers.

Or, I should have added the unique constraint during development & testing, and then dropped it before releasing it to the public, which would have caught this bug too.

PR #2542, opened 2021-10-24
update error message to match cli flag: "-codebase" -> "--codebase"

iamevn:
When an transcript errors, an error message is printed that ends in:
```
Run `ucm -codebase /tmp/transcript-18a0ab44dc0f8f92` to do more work with it.
```

Trying to run that command results in:
```
Invalid option `-codebase'
```

This PR changes "-codebase" -> "--codebase" in this error message, in comments, and in transcripts. I couldn't find any other flags that need a similar update.


pchiusano (approved):
Thanks!

PR #2548, opened 2021-10-26
delete OldName

mitchellwrosen:
This PR deletes `OldName`, which was left around just long enough to see the regression test suite pass in CI before merging #2488


aryairani (approved):

aryairani:
I already restarted CI once; it doesn't seem to be able to unarchive `stack` on the macOS-10.15 build.  I can unarchive it fine locally.  Not sure what to do about this besides dropping macOS-10.15 from the CI builds.

mitchellwrosen:
That's weird. :sob:

PR #2549, opened 2021-10-27
explicit export lists for `U.Codebase.Sqlite.{Operations,Queries}.hs`

aryairani:
Lots of exports.  One deletion.

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
@@ -10,7 +10,75 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE ViewPatterns #-}
 
-module U.Codebase.Sqlite.Operations where
+module U.Codebase.Sqlite.Operations (
+  -- *
+  dataVersion,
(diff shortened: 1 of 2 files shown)


mitchellwrosen (approved):

codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
```diff
@@ -10,7 +10,85 @@
...
+  getCycleLen,
+
+  -- * Error types
+  Error(..),
+  DecodeError(..),
+
+  -- ** Constraint kinds
+  EDB,
+
+  -- * somewhat unexpectedly unused definitions
```
pchiusano, line 81: wait, these are unused but they are in the export list? Or should that say "used definitions"

mitchellwrosen: Yeah, unused but exported to avoid warnings

aryairani: @pchiusano Yeah, didn’t realize they were present but unused until now, but not quite sure that they should be deleted.

PR #2552, opened 2021-10-27
TermPrinter: Use Referent for terms

### LINKED ISSUE #2154: References to terms in SyntaxText show up in the UI as TypeReference
I opened up `List.filter` then clicked through to `List.foldLeft` and got this:

Screen Shot 2021-06-10 at 9 16 05 AM

Notice the URL.

Interesting is that if I refresh the page, the definition still loads but has the wrong url.

hojberg:
## Overview
Fix a bug which manifested itself in the SyntaxText of the Codebase
server where references to terms, where using TypeReference instead of
TermReference. 

Fixes: https://github.com/unisonweb/unison/issues/2154

## Implementation notes
This was due to using a Reference instead of a Referent.

## Loose ends
@pchiusano you'd mentioned on https://github.com/unisonweb/unison/issues/2154 to change the constructors to be more explicit, but I wasn't sure exactly which constructors might change.

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/DisplayValues.hs
+++ b/parser-typechecker/src/Unison/CommandLine/DisplayValues.hs
@@ -316,10 +316,10 @@ displayDoc pped terms typeOf evaluated types = go
 
 termName :: PPE.PrettyPrintEnv -> Referent -> Pretty
 termName ppe r = P.syntaxToColor $
-  NP.styleHashQualified'' (NP.fmt $ S.Referent r) name
+  NP.styleHashQualified'' (NP.fmt $ S.TermReference r) name
   where name = PPE.termName ppe r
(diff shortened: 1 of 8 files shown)


pchiusano (approved):
@hojberg Cool. 

I was just saying that I think the name of that `SyntaxText` constructor is confusing. It's just called `Reference` when it really means "type reference". I'd just rename it to `TypeReference`. The confusing name is probably what led to this problem in the first place. :) 

Fixing would entail editing `Unison.Util.SyntaxText` to rename that constructor, and then fix any compile errors. Should be quick but not a big deal if you'd rather not futz with it now or think it belongs as a separate PR.

hojberg:
@pchiusano updated with new constructor names 🎉

PR #2553, opened 2021-10-27
`@signature{blah}` now correctly renders `blah` as a clickable link

pchiusano:
Previously this wasn't hyperlinked (see for example https://share.unison-lang.org/@mb2bmao6r5c17uu14ulp8n4ui0aoe1eaquhh59ro64oouiujaica1oh583okjdjvujogn6jjbfh4t7r1ubi6h43pl06quiajgj57g10/terms/unison/async/Async/doc) and try clicking on the signatures at the top.

`@signature` gets used frequently when writing overview docs that give the signatures of key functions in the library, and those definitions should be clickable.

`@source{blah}` should probably get the same treatment but I'm leaving that for a separate PR.

__Old n' busted__

https://user-images.githubusercontent.com/11074/139127473-583ca267-7149-4aa9-984a-2f54f3b3b714.mov

__After this PR__

https://user-images.githubusercontent.com/11074/139126888-5537fd23-96c1-4408-b531-5602baa74771.mov

### DIFF
--- a/parser-typechecker/src/Unison/CommandLine/DisplayValues.hs
+++ b/parser-typechecker/src/Unison/CommandLine/DisplayValues.hs
@@ -234,7 +234,7 @@ displayPretty pped terms typeOf eval types tm = go tm
     Just typ -> pure . P.group $
       TypePrinter.prettySignatures
         (PPE.suffixifiedPPE pped)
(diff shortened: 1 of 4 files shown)


hojberg (approved):

hojberg:
Nice. I think this fixes https://github.com/unisonweb/codebase-ui/issues/188

PR #2554, opened 2021-10-27
Doc.AsHtml: Correctly extract href & alt-text

hojberg:
## Overview
`Image` often (always?) have nested href and alt texts. Support 1 level
deeper.

## Notes
Perhaps we should have a `toText` function function of doc, that recursively unwraps and joins?


pchiusano (approved):
This is okay but you should probably flatten out the paragraph recursively and concatenate all the `Word` inside with no spaces for the href

For alt text I'd do the same but with spaces between words.

hojberg:
Improved the toText bit.

PR #2559, opened 2021-10-28
Run integration tests in CI

pchiusano:
This has CI run the integration tests added in https://github.com/unisonweb/unison/pull/2543 by @satotake. Thanks @satotake. 

The tests take 130 seconds to run on my machine. I'm guessing that could be sped up, either by trimming the transcripts to rely on less code, or having multiple tests share a codebase rather than recreating it every time.

Regardless, I think this is a good thing to add to CI.


aryairani (approved):

pchiusano:
I think that Mergify won't automatically merge changes to CI.

aryairani:
Ok.

satotake:
Thanks for this.

> The tests take 130 seconds to run on my machine.

Probably, `--codebase-create` is the culprit.
So we have to manage it by skipping test or improving execution itself as you mentioned.
I will try it though the latter may be hard for me.
But before that, I would like to measure and display the durations.

PR #2561, opened 2021-10-28
Bugfix: `Name.cons` is not monotonic wrt. Name ordering

mitchellwrosen:
In #2488 I erroneously replaced a few `Map.mapKeys` with `Map.mapKeysMonotonic` where I thought was appropriate, but it turns out `Name.cons` is not monotonic with respect to name ordering.

Here is the counterexample that @aryairani and I chased down:

- Name segment = "test"
- Name 1 = "Test"
- Name 2 = "internals.v1.Test.Test"

For `Name.cons "test"` be monotonic with respect to ordering, it must be the case that

- if `x < y`,
- then `Name.cons "test" x < Name.cons "test" y` 

However, this isn't the case.

- "Test" < "internals.v1.Test.Test"
- "test.Test" > "test.internals.v1.Test.Test"

Or, written out in reverse-segment order (which is the way names compare),

- ["Test"] < ["Test", "Test", "v1", "internals"]
- ["Test", "test"] > ["Test", "Test", "v1", "internals", "test"]

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -232,40 +232,38 @@ branch0 terms types children edits =
       where
         go :: (NameSegment, Branch m) -> Relation Referent Name
         go (n, b) =
-          R.mapRanMonotonic (Name.cons n) (deepTerms $ head b)
+          R.mapRan (Name.cons n) (deepTerms $ head b)
     deepTypes' :: Relation Reference Name
     deepTypes' =
       R.mapRanMonotonic Name.fromSegment (Star3.d1 types) <> foldMap go children'
       where
         go :: (NameSegment, Branch m) -> Relation Reference Name
         go (n, b) =
-          R.mapRanMonotonic (Name.cons n) (deepTypes $ head b)
+          R.mapRan (Name.cons n) (deepTypes $ head b)
     deepTermMetadata' :: Metadata.R4 Referent Name
     deepTermMetadata' =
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
Regression test might be good. I think a transcript might work. But also maybe more trouble than it's worth. I'd be fine merging without it.

mitchellwrosen:
Ok. I think a unit test would be easy enough to add.

ChrisPenner (approved):

aryairani (approved):

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -279,7 +277,7 @@ deepEdits' = go id where
   -- can change this to an actual prefix once Name is a [NameSegment]
   go :: (Name -> Name) -> Branch0 m -> Map Name (EditHash, m Patch)
   go addPrefix Branch0{_children, _edits} =
-    Map.mapKeysMonotonic (addPrefix . Name.fromSegment) _edits
```
aryairani, line 282: What's `addPrefix` vs `Name.cons`; the same?

aryairani: Oh, it's passed in.

mitchellwrosen:
I added a regression test and confirmed that it fails if we use `mapKeysMonotonic` to construct (e.g.) `deepTypes`

https://github.com/unisonweb/unison/pull/2561/commits/3fb585e85dfd11fae50d5f55c7f649c75860db83

Issue #2424, opened 2021-09-21
Spurious ability check failure


pchiusano:
```
structural ability Split m where
  skip! : x
  both : m -> a -> a -> a

structural type Split.Step m a r
  = Empty 
  | One a
  | Two m r r 

Split.un : '{Split m,g} a ->{Split m,g} (Step m a ('{Split m,g} a))
Split.un s = 
  go : Request {Split m} a ->{Split m,g} (Step m a ('{Split m,g} a))
  go = cases
    { a } -> One a
    { skip! -> _ } -> Empty
    { both m l r -> k } -> Two m '(k l) '(k r) 

  handle !s with go
```

Yields:

```
  The expression in red
  
                needs the abilities: {g9}
    but was assumed to only require: {Split m12}
  
  This is likely a result of using an un-annotated function as an argument with concrete abilities. Try adding an annotation to the function definition whose body is red.
  
     18 |   go = cases
     19 |     { a } -> One a
     20 |     { skip! -> _ } -> Empty
     21 |     { both m l r -> k } -> Two m '(k l) '(k r) 
```

Weirdly, putting an annotation on the `’(k r)` causes it to typecheck.

```
    { both m l r -> k } -> Two m ('(k l) : '{Split m,g} a) '(k r) 
```

PR #2562, opened 2021-10-28
Fix various ability checking issues

### LINKED ISSUE #2350: Spurious ability check failure in this example
```
unique ability Storage d g where
  save.impl : a ->{Storage d g} ('{g} (d a))

save : a ->{Storage d g, g} (d a)
save a = !(save.impl a)
```
This yields:
```
The expression in red needs the {Storage d9 {}} ability, but this location only has access to these abilities: {Storage d9 g12, g12}
  
      5 | save a = !(save.impl a)
                     ^^^^^^^^^^^
```
This is kinda blocking some work on the distributed API. I could work around for now, but hoping it’s something simple that’s easy to fix.

I don’t understand where is it getting the `Storage d9 {}` requirement.

I feel like the type signature should be getting pushed down to that use site, so when that `Storage.save.impl` operation is called, it uses the exact instantiation of `Storage` which is ambiently available.

### LINKED ISSUE #2474: Spurious ability check failure in this example
Might be another instance of the same issue, but check this out - 

```
structural ability Stream a where 
  emit : a -> ()

Stream.uncons : '{Stream a, g} r ->{g} Either r (a, '{Stream a, g} r)
Stream.uncons s = 
  go : Request {Stream a,g} r -> Either r (a, '{Stream a,g} r)
  go = cases 
    { r } -> Left r
    { Stream.emit a -> tl } -> Right (a, tl : '{Stream a,g} r)
  handle !s with go
```
```
  The expression in red
  
                needs the abilities: {g38}
    but was assumed to only require: {Stream a35}
  
  This is likely a result of using an un-annotated function as an argument with concrete abilities. Try adding an annotation to the function definition whose body is red.
  
     15 |   go = cases 
     16 |     { r } -> Left r
     17 |     { Stream.emit a -> tl } -> Right (a, tl : '{Stream a,g} r)
  
```
Removing the annotation on `go` causes it to work. 

The annotation on `tl` is needed, without that I get the same error.

dolio:
This PR includes work for a couple fixes.

- A dedicated equality procedure is implemented so that invariant types are less prone to causing issues with ability defaulting
- Handler typing was incorrect in a couple ways, causing it to have overly specific/wrong types
- There was a corner case where ability variable defaulting happened in a situation it shouldn't.

Test cases for associated issues. Fixes #2350 #2474 and #2424. I'm pretty sure the latter two were both related to the wrong handler typing.

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -595,6 +595,9 @@ renderTypeError e env src = case e of
   simplePath' :: C.PathElement v loc -> Pretty ColorText
   simplePath' = \case
     C.InSynthesize e -> "InSynthesize e=" <> renderTerm env e
+    C.InEquate t1 t2 ->
+      "InEquate t1=" <> renderType' env t1 <>
+      ", t2=" <> renderType' env t2
     C.InSubtype t1 t2 ->
       "InSubtype t1=" <> renderType' env t1 <> ", t2=" <> renderType' env t2
     C.InCheck e t ->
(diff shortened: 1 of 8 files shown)


pchiusano:
Nice!

pchiusano (approved):
Looking forward to trying this out!

PR #2568, opened 2021-10-29
Render doc html files

hojberg:
## Overview
Add a new ucm command to render docs in a namespace to html files

Authored with @ChrisPenner 

## Implementation notes
* Add a `Backend` (down the line, perhaps this should live somewhere else?) function to render all docs in a branch to html files
* Add a new ucm command `docs.to-html` that takes a namespace path and a directory path and calls the `Backend` function to render docs to html files.

## Test coverage
Since there's a side-effect of output files, i'm not sure how we can best test this?

## Loose ends
There's a not so great crash that happens when we try and write /to a file thats not within the cwd.
We should perhaps catch Lucid's error and translate it somehow?

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
+++ b/parser-typechecker/src/Unison/Codebase/Editor/Input.hs
@@ -108,50 +108,51 @@ data Input
     | ReplaceI (HQ.HashQualified Name) (HQ.HashQualified Name) (Maybe PatchPath)
     | RemoveTermReplacementI (HQ.HashQualified Name) (Maybe PatchPath)
     | RemoveTypeReplacementI (HQ.HashQualified Name) (Maybe PatchPath)
-  | UndoI
-  -- First `Maybe Int` is cap on number of results, if any
-  -- Second `Maybe Int` is cap on diff elements shown, if any
-  | HistoryI (Maybe Int) (Maybe Int) BranchId
-  -- execute an IO thunk
-  | ExecuteI String
-  -- execute an IO [Result]
-  | IOTestI (HQ.HashQualified Name)
-  -- make a standalone binary file
-  | MakeStandaloneI String (HQ.HashQualified Name)
-  | TestI Bool Bool -- TestI showSuccesses showFailures
-  -- metadata
-  -- `link metadata definitions` (adds metadata to all of `definitions`)
-  | LinkI (HQ.HashQualified Name) [Path.HQSplit']
-  -- `unlink metadata definitions` (removes metadata from all of `definitions`)
-  | UnlinkI (HQ.HashQualified Name) [Path.HQSplit']
-  -- links from <type>
-  | LinksI Path.HQSplit' (Maybe String)
-  | CreateAuthorI NameSegment {- identifier -} Text {- name -}
-    -- Display provided definitions. If list is empty, prompt a fuzzy search.
-  | DisplayI OutputLocation [HQ.HashQualified Name]
-    -- Display docs for provided terms. If list is empty, prompt a fuzzy search.
-  | DocsI [Path.HQSplit']
-  -- other
-  | SearchByNameI Bool Bool [String] -- SearchByName isVerbose showAll query
-  | FindShallowI Path'
-  | FindPatchI
-    -- Show provided definitions. If list is empty, prompt a fuzzy search.
-  | ShowDefinitionI OutputLocation [HQ.HashQualified Name]
-  | ShowDefinitionByPrefixI OutputLocation [HQ.HashQualified Name]
-  | ShowReflogI
-  | UpdateBuiltinsI
-  | MergeBuiltinsI
-  | MergeIOBuiltinsI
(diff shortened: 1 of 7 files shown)


parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -782,7 +800,7 @@ renderDoc ::
   Rt.Runtime v ->
   Codebase IO v Ann ->
   Reference ->
-  Backend IO [(HashQualifiedName, UnisonHash, Doc.Doc)]
+  IO [(HashQualifiedName, UnisonHash, Doc.Doc)]
```
hojberg, line 803: This didn't use `Backend` at all. Removing it made it cleaner to deal with from `Command`.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -294,6 +298,18 @@ findShallowReadmeInBranchAndRender width runtime codebase printNames namespaceBr
         hqLen <- liftIO $ Codebase.hashLength codebase
         join <$> traverse (renderReadme (ppe hqLen)) (Set.lookupMin readmes)
 
+isDoc :: Monad m => Var v => Codebase m v Ann -> Referent -> m Bool
+isDoc codebase ref = do
+  ot <- loadReferentType codebase ref
+  pure $ isDoc' ot
+
+isDoc' :: (Var v, Monoid loc) => Maybe (Type v loc) -> Bool
+isDoc' typeOfTerm = do
```
hojberg, line 307: Extracted so we can use this in our new function to filter terms that are docs.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -810,9 +828,53 @@ renderDoc ppe width rt codebase r = do
...
+  let width = defaultWidth
+  let ppe = PPE.fromNamesDecl hqLength printNames
+  docs <- concat <$> for docRefs (renderDoc ppe width runtime codebase . Referent.toReference)
+  liftIO $ traverse_ (renderDocToHtmlFile directory) docs
+
+  where
+    cleanPath :: FilePath -> FilePath
+    cleanPath filePath =
+      filePath <&> \case
+        '#' -> '@'
```
hojberg, line 855: We prefer `@` in urls (since `#` is reserved for url fragments), so we replace `#` to `@` to get a clean filename.

parser-typechecker/src/Unison/Codebase/Editor/Input.hs
```diff
@@ -108,50 +108,51 @@ data Input
...
-  | ListDependentsI (HQ.HashQualified Name)
-  | DebugNumberedArgsI
-  | DebugTypecheckedUnisonFileI
-  | DebugDumpNamespacesI
-  | DebugDumpNamespaceSimpleI
-  | DebugClearWatchI
-  | QuitI
-  | UiI
-  deriving (Eq, Show)
+    | UndoI
```
hojberg, line 111: these were all on a weird indentation level.

parser-typechecker/src/Unison/Server/Backend.hs
```diff
@@ -810,9 +828,53 @@ renderDoc ppe width rt codebase r = do
...
+          cleanPath $ HQ'.toString docName <> ".html"
+
+      in directoryPath </> docFileName
+
+    renderDocToHtmlFile :: FilePath -> (HashQualifiedName, UnisonHash, Doc.Doc) -> IO ()
+    renderDocToHtmlFile directory (docName, _, doc) =
+      let
+        docFilePath' = docFilePath directory docName
+      in
+      Lucid.renderToFile docFilePath' (DocHtml.toHtml doc)
```
pchiusano, line 877: If Lucid doesn't let you write to files outside the current directory (kind of an odd limitation...), I'd probably just call `Lucid.renderBS`

https://hackage.haskell.org/package/lucid-2.10.0/docs/Lucid.html#v:renderBS

And then use this to write that to any file you want

https://hackage.haskell.org/package/bytestring-0.11.1.0/docs/src/Data.ByteString.Lazy.html#writeFile

pchiusano (approved):
Awesome! I left one comment which could be addressed by a later PR if you want, but I'm going to merge as is.

parser-typechecker/src/Unison/Server/Backend.hs
hojberg: @pchiusano I think its probably and overall good security wise, but we should handle the error better.

aryairani:
@hojberg Re. testing, you could create a transcript test that will generate some good representative docs to html in some output directory in the repo, and then we can check it in and CI will check for unexpectedly changed output on disk like it does for all the other transcripts. (They all have the side effect of output files.)

I thought of it when I was trying to link examples to the new features in the release notes. We could link the transcript markdown like I did for the other three, and link the HTML also.

hojberg:
@aryairani nice! thats a great idea 🎉

PR #2585, opened 2021-11-02
Docs.AsHtml: Support linking between docs

hojberg:
## Problem
HTML Docs were referencing each other by reference, but the files generated are based on names and namespaces.

## Solution
* Add a way for docs to lookup a referent, get a name and convert it to
  a html href.
* Add a reference type as a data attribute to syntax refs

## Notes
* Also ran the codeformatter

### DIFF
--- a/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
+++ b/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
@@ -4,18 +4,25 @@
 module Unison.Server.Doc.AsHtml where
 
 import Data.Foldable
+import Data.Map (Map)
+import qualified Data.Map as Map
 import Data.Maybe
 import Data.Text (Text)
 import qualified Data.Text as Text
 import Lucid
 import qualified Lucid as L
 import Unison.Codebase.Editor.DisplayObject (DisplayObject (..))
+import Unison.Name (Name)
+import qualified Unison.Name as Name
+import Unison.Referent (Referent)
+import qualified Unison.Referent as Referent
 import Unison.Server.Doc
 import Unison.Server.Syntax (SyntaxText)
 import qualified Unison.Server.Syntax as Syntax
 
 data NamedLinkHref
   = Href Text
+  | DocLinkHref Name
   | ReferenceHref Text
   | InvalidHref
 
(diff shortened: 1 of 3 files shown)


parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -119,32 +139,32 @@ mergeWords sep = foldr merge_ []
...
+    Section d ds ->
+      toText sep d <> sep <> listToText ds
+    UntitledSection ds ->
+      listToText ds
+    Column ds ->
+      listToText ds
+    Word w ->
+      w
+    _ ->
+      ""
```
hojberg, line 167: this is just the formatter running.

parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -250,11 +270,14 @@ toHtml document =
...
-                case normalizeHref InvalidHref href of
+                case normalizeHref docNamesByRef href of
                   Href h ->
                     a_ [class_ "named-link", href_ h, rel_ "noopener", target_ "_blank"] $ currentSectionLevelToHtml label
+                  DocLinkHref name ->
+                    let href = "/" <> Text.replace "." "/" (Name.toText name) <> ".html"
+                     in a_ [class_ "named-link doc-link", href_ href] $ currentSectionLevelToHtml label
                   ReferenceHref ref ->
-                    a_ [class_ "named-link", data_ "ref" ref] $ currentSectionLevelToHtml label
+                    span_ [class_ "named-link", data_ "ref" ref, data_ "ref-type" "term"] $ currentSectionLevelToHtml label
```
hojberg, line 280: When we only have a reference, we shouldn't be using `<a>` since it wont have an `href` (the [https://github.com/unisonweb/website](website) build step makes these into Unison Share links).

hojberg: Technically we don't know if this is a term (see `data_ "ref-type" "term"`), but didn't want to take on that effort here as well.

parser-typechecker/src/Unison/Server/Syntax.hs
```diff
@@ -231,8 +231,8 @@ segmentToHtml (Segment segmentText element) =
         | isFQN = nameToHtml (Name.unsafeFromText sText)
         | otherwise = L.toHtml sText
    in case ref of
-        Just r ->
-          span_ [class_ className, data_ "ref" r] content
+        Just (r, refType) ->
+          span_ [class_ className, data_ "ref" r, data_ "ref-type" refType] content
```
hojberg, line 235: Unison Share has urls with `types` or `terms` in them, we want to make sure enough information is provided to construct such urls.

pchiusano (approved):

PR #2590, opened 2021-11-03
Measure integration test times

satotake:
## Overview

We added `integration-tests` in #2543 .
Some tests can be slow (#2559) .
We should monitor durations to judge if we need to improve or not.
Simple time measuring was added in this PR.

It may be better to use criterion. But I think it will be too much.

## Loose ends

- #2543 
- #2559


pchiusano (approved):

pchiusano:
Thanks!

PR #2598, opened 2021-11-05
Move many ucm-related modules into unison-cli library component

mitchellwrosen:
## Overview

This PR moves many modules related to the implementation of `ucm` out of `parser-typechecker` and into the `unison-cli` library component

Here's what the package looks like now:

Screenshot from 2021-11-05 11-15-14

## Loose ends

There are more modules that should be teased apart from `parser-typechecker`, but I only moved the ones that didn't put up a fight. (This way, we don't have to rely on `git`'s heuristic rename detection, as the before/after files match 100%).


.github/workflows/ci.yaml
```diff
@@ -91,11 +91,15 @@ jobs:
       - name: build dependencies
         run: stack --no-terminal build --fast --only-dependencies
       - name: build
-        run: stack --no-terminal build --fast
+        run: stack --no-terminal build --fast --no-run-tests --test
```
aryairani, line 94: What does this do, build tests but not run them?

aryairani (approved):

.github/workflows/ci.yaml
mitchellwrosen: Yeah; without it, on the `stack test` line(s) below, stack will stupidly rebuild _library_ components of packages with test suites (and their dependencies, and so on).

It's a stack caching bug caused by stack treating every component of a package as one atomic unit.

mitchellwrosen: We didn't have it before because we only had one test suite, which was an executable component. Now, there are a couple packages with actual test suites that are run with `stack test`.

ChrisPenner (approved):
Looking over this makes me curious what the purpose/semantics of the `HandleCommand` vs `HandleInput` layer really is, and how those differ from whatever our shared code for the cli and api will be.

I figured there'd be at least some work in `HandleCommand` that was useful/used for the server, but apparently not?

Which I think means this is a good move to do, then we should probably just clarify what the "Backend" is, and pull the shared pieces of it out of the `Server` module 👍🏼

PR #2610, opened 2021-11-08
Make PrettyPrintEnv only contain named things

mitchellwrosen:
## Overview

This PR changes the type of `PrettyPrintEnv` from

```haskell
data PrettyPrintEnv = 
  PrettyPrintEnv {
    terms :: Referent -> Maybe (HashQualified Name),
    types :: Reference -> Maybe (HashQualified Name)
  }
```

to

```haskell
data PrettyPrintEnv = 
  PrettyPrintEnv {
    terms :: Referent -> Maybe (HQ'.HashQualified Name),
    types :: Reference -> Maybe (HQ'.HashQualified Name)
  }
```

The functions contained within a `PrettyPrintEnv` were already morally returning `HQ'.HashQualified`, i.e. things that are not `HashOnly`, but at some point along the way (specifically, in `suffixedTypeName`/`suffixedTermName`), we were throwing that fact away.

Now, we retain a named `HQ'.HashQualified` all the way through.

### DIFF
--- a/unison-core/src/Unison/HashQualified.hs
+++ b/unison-core/src/Unison/HashQualified.hs
@@ -49,10 +49,9 @@ toName = \case
 --   [.foo.bar, foo.bar] -> [foo.bar, .foo.bar]
 sortByLength :: [HashQualified Name] -> [HashQualified Name]
 sortByLength hs = sortOn f hs where
-  f (NameOnly n) = (countDots n, 0, Left n)
-  f (HashQualified n _h) = (countDots n, 1, Left n)
+  f (NameOnly n) = (length (Name.reverseSegments n), 0, Left n)
+  f (HashQualified n _h) = (length (Name.reverseSegments n), 1, Left n)
   f (HashOnly h) = (maxBound, 0, Right h)
-  countDots n = Text.count "." (Text.dropEnd 1 (Name.toText n))
 
 hasName, hasHash :: HashQualified Name -> Bool
 hasName = isJust . toName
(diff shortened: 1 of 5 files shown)


unison-core/src/Unison/HashQualified.hs
```diff
@@ -49,10 +49,9 @@ toName = \case
 --   [.foo.bar, foo.bar] -> [foo.bar, .foo.bar]
 sortByLength :: [HashQualified Name] -> [HashQualified Name]
 sortByLength hs = sortOn f hs where
-  f (NameOnly n) = (countDots n, 0, Left n)
```
mitchellwrosen, line 52: No behavior change intended here, just noticed we were round-tripping through `Text` to count the segments in a name, which is no longer necessary.

pchiusano (approved):
😎

aryairani (approved):
Alright cool.

Note that what we lose here is the ability for the PPE to dictate how to render an unnamed Reference.  We call `HQ.take todoHashLength (HQ.fromReferent r)` externally, but you could previously do that inside the PPE if you wanted to.  Not sure we'd ever want to though.

mitchellwrosen:
@aryairani I'm not sure I understand the difference between printing hashes from within the PPE or not here. Is the idea that you could close over the hash length?

I can see a PPE returning either `Maybe HQ'` (as in this PR) or `HQ`, but `Maybe HQ` (as it was prior to this PR) seemed wrong, because you can always render a Reference/Referent without a name, so when would you ever return a Nothing?

aryairani:
> @aryairani I'm not sure I understand the difference between printing hashes from within the PPE or not here. Is the idea that you could close over the hash length?

Yeah; I guess that's not too interesting since the PPE wraps pure functions.

> I can see a PPE returning either `Maybe HQ'` (as in this PR) or `HQ`, but `Maybe HQ` (as it was prior to this PR) seemed wrong, because you can always render a Reference/Referent without a name, so when would you ever return a Nothing?

Good point!

Issue #939, opened 2019-11-06
view and edit should emit use statements at the top level


atacratic:
At the moment the pretty-printer only emits use statements at the start of blocks within a term.  It would be better if it could also emit use statements at the top level.

So instead of giving
```
foo : base.Int -> base.Int -> base.Int
foo x y = 
  use base.Int +
  x + y
```

It should give
```
use base Int

foo : Int -> Int -> Int
foo x y = 
  use base.Int +
  x + y
```

In this example the benefit is in eliding FQNs in type signatures.  

There is also a benefit if we are viewing or editing multiple definitions at once - we can do a use for names that appear only once in each definition (and therefore are not used in those definitions individually).

PR #2612, opened 2021-11-08
Use suffixed names inside ability declarations in codebase UI

### LINKED ISSUE #2567: Long names are sometimes picked for source on share
See:

CleanShot 2021-10-29 at 14 33 19@2x

Current theory is that the `pending` namespace on Share is causing, but needs more investigation.

pchiusano:
Fixes #2567 

Due to an implementation quirk, this weirdness was only present in the codebase UI (and on Unison Share). I modified the code so that both UCM and the codebase UI go through an identical code path, and added an explict export list to `DeclPrinter`.

To get the output of `ability` declarations further cleaned up, I think the pretty-printer could really use the ability to emit top-level `use` clauses https://github.com/unisonweb/unison/issues/939 which @atacratic has mentioned.

### DIFF
--- a/parser-typechecker/src/Unison/DeclPrinter.hs
+++ b/parser-typechecker/src/Unison/DeclPrinter.hs
@@ -1,7 +1,7 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 {-# LANGUAGE OverloadedStrings #-}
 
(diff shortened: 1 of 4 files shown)


hojberg (approved):
Sweet!!! 🎉

PR #2613, opened 2021-11-09
Doc.AsHtml: Fix folded and inlineCode

hojberg:
## Overview
Add a few css classes to the folded elements and fix inlineCode such
that it supports more than 1 css class


parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -96,14 +96,16 @@ foldedToHtml attrs isFolded =
             if isFolded
               then open_ "open" : attrs
               else attrs
-       in details_ attrsWithOpen $ summary_ [] $ sequence_ $ summary ++ details
+       in details_ attrsWithOpen $ do
+         summary_ [ class_ "folded-content"] $ sequence_ summary
+         div_ [ class_ "folded-content"] $ sequence_ details
```
hojberg, line 101: This fixes a bug with folded which ended up with the wrong html structure, where details were a child of summary instead of a sibling.

pchiusano (approved):

Issue #2163, opened 2021-07-01
Is new runtime using pointer equality or local compiler numbering when comparing functions?


pchiusano:
```
        (x -> x) === (x -> x)
        ⧨
        false
```

I think all pure functions including `Universal.===` should not depend on local machine stuff like pointer equality or local compiler numbering. Otherwise it can result in broken distributed programs, like you send a value to another node and suddenly it can’t be found in a Map there.

I think we previously had a few places where we fixed this up. But maybe we’re misunderstanding something about this example?

hojberg:
I think these are dupes? https://github.com/unisonweb/unison/issues/2162

PR #2615, opened 2021-11-09
Try to only float a single copy of a definition

dolio:
This small patch changes the lambda floating pass to try not to float multiple copies of a lambda that are identical. This will improve #2163. However, it might result in some differences in how things are eventually printed. In the transcripts, it seems one changed from `'(...)` to `(_ -> ...)`.

It's probably a good idea to look at some docs or something that might be affected by this before merging. But I don't know which docs would be likely to be affected.


aryairani:
@dolio Who do you want to review this?

dolio:
I guess @pchiusano . Or if he knows someone else who could check some stuff to make sure that the changes in decompiled output aren't a problem.

ceedubs (approved):
I've confirmed that this fixes hashing of thunks for me as well.

PR #2616, opened 2021-11-09
Make `dependents` only report named things (fixes #1525)

### LINKED ISSUE #1525: `dependents` includes deleted names
Transcript (input):
```
    ```unison
    a = 1
    b = a
    ```

    ```ucm
    .> add
    .> names b
    .> delete.term b
    .> ls
    .> dependents a
    ```

    Problem: `dependents a` still lists b by both name and hash, 
    even though the name is deleted, and without flagging that b is gone.
```

The output of the final command is
```

  Dependents of #jk19sm5bf8:
  
    Reference     Name
    #51ahuj5sm6   b
```
[transcript-and-output.zip](https://github.com/unisonweb/unison/files/4571780/transcript-and-output.zip)
master@6a36f1c

Arya's comment via slack: "we’ll have to think a little bit about what it means for something to be deleted, since it could still be referenced historically, but, I don’t think there’s any reason not to note that it’s deleted"

mitchellwrosen:
## Overview

This PR makes `dependents` only display dependents that have a name in the current namespace. Fixes #1525.

## Implementation notes

Previously, dependents of a type/term were given (possibly) historical names in the output of the dependents command, which was confusing.

Now, the behavior has changed in two ways, per #1525:

- We only pair dependent references with their name in the most recent namespace, if any.
- We do not show unnamed dependents at all, deferring that to an unimplemented `dependents.all` command.

Screenshot from 2021-11-10 13-51-05
Note: wording changed to "doesn't have any named dependents"
Screenshot from 2021-11-10 13-51-32


unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1311,24 +1311,28 @@ notifyUser dir o = case o of
               "",
               "Paste that output into http://bit-booster.com/graph.html"
             ]
-  ListDependents hqLength ld names missing ->
+  ListDependents hqLength ld results ->
     pure $
-      if names == mempty && missing == mempty
-        then c (prettyLabeledDependency hqLength ld) <> " doesn't have any dependents."
+      if null results
+        then prettyLd <> " doesn't have any dependents."
```
```suggestion
        then prettyLd <> " doesn't have any named dependents."
```

aryairani (approved):
Looks good.

PR #2617, opened 2021-11-09
Doc.AsHtml avoid pre in inlineCode

hojberg:
## Overview
pre elements are block elements and they are not allowed as children of
p (paragraph) elements. In fact the browser will automatically close an
open p if it encounters a block element in an assumption that the
developer had forgotten to close and this causes a very strange behavior
for inline code.

Also note that this "auto close" doesn't occur in client side rendering.

thanks to @dolio for helping me debug this :D

### DIFF
--- a/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
+++ b/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
@@ -43,7 +43,7 @@ embeddedSource ref =
 
 inlineCode :: [Text] -> Html () -> Html ()
 inlineCode classNames =
-  pre_ [classes_ ("inline-code" : classNames)] . code_ []
+  code_ [classes_ ("inline-code" : classNames)]
 
 codeBlock :: [Attribute] -> Html () -> Html ()
 codeBlock attrs =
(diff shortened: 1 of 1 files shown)


parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -97,8 +97,8 @@ foldedToHtml attrs isFolded =
               then open_ "open" : attrs
               else attrs
        in details_ attrsWithOpen $ do
-         summary_ [ class_ "folded-content"] $ sequence_ summary
-         div_ [ class_ "folded-content"] $ sequence_ details
+            summary_ [class_ "folded-content"] $ sequence_ summary
+            div_ [class_ "folded-content"] $ sequence_ details
```
hojberg, line 101: Formatting.

parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -43,7 +43,7 @@ embeddedSource ref =
 
 inlineCode :: [Text] -> Html () -> Html ()
 inlineCode classNames =
-  pre_ [classes_ ("inline-code" : classNames)] . code_ []
+  code_ [classes_ ("inline-code" : classNames)]
```
hojberg, line 46: https://github.com/unisonweb/unison/pull/2617

pchiusano (approved):

PR #2626, opened 2021-11-11
Docs.AsHtml: Fix doc folding symmetry

hojberg:
## Overview
Folded source includes the summary in the details, because we currently
don't have a good way to separate a signature from a definition or even
a shortened type (as seen in the Fuzzy Finder) from its definition.

This commit changes the non-source version of Folded to also include its
summary in the details part, such that the 2 Folded implementations at
least are symmetric (even though they are both semantically incorrect in
the eyes of html and the `<details>` element).

### DIFF
--- a/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
+++ b/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
@@ -97,8 +97,8 @@ foldedToHtml attrs isFolded =
               then open_ "open" : attrs
               else attrs
        in details_ attrsWithOpen $ do
-            summary_ [class_ "folded-content"] $ sequence_ summary
-            div_ [class_ "folded-content"] $ sequence_ details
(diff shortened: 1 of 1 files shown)


parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -196,7 +196,7 @@ toHtml docNamesByRef document =
               Code code ->
                 span_ [class_ "rich source inline-code"] $ inlineCode [] (currentSectionLevelToHtml code)
               CodeBlock lang code ->
-                div_ [class_ "rich source code", class_ $ textToClass lang] $ codeBlock [] (currentSectionLevelToHtml code)
+                div_ [class_ $ "rich source code " <> textToClass lang] $ codeBlock [] (currentSectionLevelToHtml code)
```
hojberg, line 199: Fixes a small bug where the `lang` would be added to the classes without a space.

pchiusano (approved):

PR #2633, opened 2021-11-13
Make lookupWatchCache take a Reference.Id rather than a Reference

mitchellwrosen:
## Overview

This PR amends the type of `lookupWatchCache` from

```haskell
lookupWatchCache :: (Monad m) => Codebase m v a -> Reference -> m (Maybe (Term v a))
```

to

```haskell
lookupWatchCache :: (Monad m) => Codebase m v a -> Reference.Id -> m (Maybe (Term v a))
```

because attempting to look up a builtin in the watch expression cache will always result in `Nothing`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,8 +60,8 @@ jobs:
           # recent branch cache.
           # Then it will save a new cache at this commit sha, which should be used by
           # the next build on this branch.
-          key: stack-work-2_${{matrix.os}}-${{github.sha}}
-          restore-keys: stack-work-2_${{matrix.os}}
+          key: stack-work-3_${{matrix.os}}-${{github.sha}}
+          restore-keys: stack-work-3_${{matrix.os}}
 
       # Install stack by downloading the binary from GitHub. The installation process is different for Linux and macOS,
       # so this is split into two steps, only one of which will run on any particular build.
--- a/unison-cli/src/Unison/Codebase/Editor/HandleCommand.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleCommand.hs
@@ -199,11 +199,10 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
     ClearWatchCache -> lift $ Codebase.clearWatches codebase
     FuzzySelect opts display choices -> liftIO $ Fuzzy.fuzzySelect opts display choices
 
-  watchCache (Reference.DerivedId h) = do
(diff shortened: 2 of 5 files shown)


unison-cli/src/Unison/Codebase/Editor/HandleCommand.hs
```diff
@@ -199,11 +199,10 @@ commandLine config awaitInput setBranchRef rt notifyUser notifyNumbered loadSour
     ClearWatchCache -> lift $ Codebase.clearWatches codebase
     FuzzySelect opts display choices -> liftIO $ Fuzzy.fuzzySelect opts display choices
 
-  watchCache (Reference.DerivedId h) = do
-    m1 <- Codebase.getWatch codebase WK.RegularWatch h
-    m2 <- maybe (Codebase.getWatch codebase WK.TestWatch h) (pure . Just) m1
-    pure $ Term.amap (const ()) <$> m2
-  watchCache Reference.Builtin{} = pure Nothing
+  watchCache :: Reference.Id -> IO (Maybe (Term v ()))
```
mitchellwrosen, line 202: `Codebase.lookupWatchCache`'s implementation was duplicated inside the definition of `watchCache` over here. Now it's not.

.github/workflows/ci.yaml
```diff
@@ -60,8 +60,8 @@ jobs:
           # recent branch cache.
           # Then it will save a new cache at this commit sha, which should be used by
           # the next build on this branch.
-          key: stack-work-2_${{matrix.os}}-${{github.sha}}
-          restore-keys: stack-work-2_${{matrix.os}}
+          key: stack-work-3_${{matrix.os}}-${{github.sha}}
+          restore-keys: stack-work-3_${{matrix.os}}
```
aryairani, line 64: Aside: Any ideas on this?

aryairani (approved):

.github/workflows/ci.yaml
mitchellwrosen: Do you mean any ideas about why the cache gets busted sometimes?

aryairani: Yeah

aryairani: Or more like: is there anything we can do to prevent the cache from getting busted sometimes, without significantly increasing average build time?

mitchellwrosen: I'm not sure, maybe cabal wouldn't have these problems 🤷‍♂️

Issue #2599, opened 2021-11-05
Add benchmarks for `Rope`, `Bytes`, and `Text`


pchiusano:
To support further performance tuning of these data structures

Swarnlataaa:
I would like to contribute to this, please guide me.

Issue #2631, opened 2021-11-12
runtime benchmark in parser-typechecker package currently doesn't build


aryairani:
@dolio Could you take a look and see if we want to keep this and, if so, to see what it would take to get it building again?  Then we can add it to CI to keep it happy.

@mitchellwrosen Do you have thoughts about where building benchmarks should go in the CI script slash could you add it once it's building :)

The output of `stack build --fast --bench` is [here](https://gist.github.com/aryairani/80f6205ecf28ce7479c29be0e6df708a).

dolio:
I think maybe we should just get rid of it. I'm not sure it's really that useful. I could probably fix it, but it was written before the more complicated runtime environment, and I'm not sure the few micro benchmarks in there are worth it.

aryairani:
Totally your call IMO.

aryairani:
It builds now

PR #2636, opened 2021-11-14
Add Text benchmarks

satotake:
## Overview

I am trying #2599.
I am not sure what is the focuses here.
So, as a starting point, I added basic benchmarks for `Unison.Util.Text` vs. `Data.Text`.
If this issue does not target `Data.Text`, I will remove them.
And I have not done `Bytes` or `Rope` yet. 

- It is not so helpful to increase `n` here. It might be good to limit the patterns.
- Some operations of `Text` underlying `Rope` (`size`, `at`) are faster than `Data.Text`

I ignored `parser-typechecker/benchmarks/runtime/Main.hs` tentatively.
It was not updated and I tried to fix the regression.
But I gave it up when I found that we need to use `CombIx` in `Let`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/2636)

### DIFF
--- a/parser-typechecker/benchmarks/Benchmarks/Util/Bytes.hs
+++ b/parser-typechecker/benchmarks/Benchmarks/Util/Bytes.hs
@@ -1,0 +1,6 @@
+module Benchmarks.Util.Bytes where
+
+import Criterion.Main
+
+run :: Benchmark
+run = bgroup "Bytes" [] -- TODO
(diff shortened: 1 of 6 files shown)


aryairani:
@satotake We have #2631 to fix `benchmarks/runtime/Main.hs`, sorry about that.

satotake:
@aryairani no worries, sorry for missing that.

PR #2643, opened 2021-11-16
Doc.AsHtml: Don't render block elements in `<p>`

hojberg:
## Overview
We missed quite a few block element handing when rendering `<p>`s,
resulting in weird empty tags and padding.

This ensures we handle them correctly. Especially `UntitledSection` needed special treatment.


pchiusano (approved):

hojberg:
@pchiusano we might want to consider if this should have been a parser fix instead?

PR #2654, opened 2021-11-19
Fix line breaks in guards with long patterns

runarorama:
This changes the way pattern matching is pretty-printed. Instead of rendering as a table, it tries to format things as a human might.

Before:

    foo : 'Nat
    foo =
      go x =
        'match (a -> a) x with
          SomethingUnusuallyLong lijaefliejalfijelfj aefilaeifhlei liaehjffeafijij 
          | lijaefliejalfijelfj == aefilaeifhlei                                   ->
          0
          | lijaefliejalfijelfj == liaehjffeafijij                                 ->
          1
      go (SomethingUnusuallyLong "one" "two" "three")

After:

    foo : 'Nat
    foo =
      go x =
        'match (a -> a) x with
          SomethingUnusuallyLong lijaefliejalfijelfj aefilaeifhlei liaehjffeafijij 
            | lijaefliejalfijelfj == aefilaeifhlei   -> 0
            | lijaefliejalfijelfj == liaehjffeafijij -> 1
      go (SomethingUnusuallyLong "one" "two" "three")

## Controversial decisions

I abandoned trying to line up arrows in pattern matches that have a mix of cases with guards and without. Since cases with guards can be _much_ longer than cases without, the lining up of the arrows just looks terrible most of the time, and it's really fiddly to make things line up.

We could at some point add some logic to e.g. line up groups of patterns with similar lengths. I.e. we could check if the guard plus the pattern is significantly longer (by 10 chars or something arbitrary) than the surrounding patterns, and just not line up the arrows in that case.


aryairani (approved):
Awesome!

PR #2667, opened 2021-11-23
Include self-references in dependents index

mitchellwrosen:
## Overview

This PR amends the logic of `saveDeclComponent`/`saveTermComponent` to include self-references in the dependents index, for both literal self-references (e.g. `foo = foo`), and self-component-references (e.g. `foo = bar; bar = foo`).

This is an initial step towards resolving #2649.  

To preserve the behavior of existing code that is querying the index for dependencies/dependents, this PR adds a post-processing step in Haskell that filters the self-references out.

The next step we have planned is to generalize the API of getting dependencies/dependents to allow the caller to choose whether they want self-references back or not, and continue from there.

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
@@ -101,7 +101,6 @@ import Control.Monad.Trans (MonadTrans (lift))
 import Control.Monad.Trans.Maybe (MaybeT (MaybeT), runMaybeT)
 import Control.Monad.Writer (MonadWriter, WriterT, runWriterT)
 import qualified Control.Monad.Writer as Writer
-import Data.Bifoldable (bifoldMap)
 import Data.Bifunctor (Bifunctor (bimap))
 import Data.Bitraversable (Bitraversable (bitraverse))
 import Data.ByteString (ByteString)
@@ -114,8 +113,7 @@ import Data.Functor.Identity (Identity)
 import Data.Map (Map)
 import qualified Data.Map as Map
 import qualified Data.Map.Merge.Lazy as Map
-import Data.Maybe (catMaybes, isJust)
-import Data.Monoid (First (First, getFirst))
+import Data.Maybe (isJust)
 import Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import Data.Set (Set)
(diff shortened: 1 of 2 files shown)


pchiusano (approved):

codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
```diff
@@ -523,21 +521,25 @@ saveTermComponent h terms = do
...
-                  getTypeSRef (C.ReferenceBuiltin t) = C.ReferenceBuiltin (tIds Vector.! fromIntegral t)
-                  getTypeSRef (C.Reference.Derived h i) = C.Reference.Derived (oIds Vector.! fromIntegral h) i
+                  getTypeSRef = \case
+                    C.ReferenceBuiltin t -> C.ReferenceBuiltin (tIds Vector.! fromIntegral t)
+                    C.Reference.Derived h i -> C.Reference.Derived (oIds Vector.! fromIntegral h) i
                   getSTypeLink = getTypeSRef
-                  getSTermLink :: S.Term.TermLink -> Maybe S.Reference
-                  getSTermLink = getFirst . bifoldMap (First . getTermSRef) (First . Just . getTypeSRef)
+                  getSTermLink :: S.Term.TermLink -> S.Reference
+                  getSTermLink = \case
```
ChrisPenner, line 535: Good simplification haha. Pointfree isn't always better 😄

mitchellwrosen: It's also no longer a foldMap-type thing 😛

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -645,31 +646,60 @@ addToDependentsIndex dependency dependent = execute sql (dependency :. dependent
...
-getDependentsForDependency dependency = query sql dependency where sql = [here|
-  SELECT dependent_object_id, dependent_component_index
-  FROM dependents_index
-  WHERE dependency_builtin IS ?
-    AND dependency_object_id IS ?
-    AND dependency_component_index IS ?
-|]
-
+getDependentsForDependency dependency =
+  filter isNotSelfReference <$> query sql dependency
```
ChrisPenner, line 652: Wouldn't it be better to filter out self-references in the query itself?

mitchellwrosen: It would be better, but I don't think it matters much. I was planning to make more efficient SQL in the next step of generalizing the query, so we don't have to rewrite it in place twice.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
```diff
@@ -523,21 +521,25 @@ saveTermComponent h terms = do
...
                   tpRefs' = Foldable.toList $ C.Type.dependencies tp
-                  getTermSRef :: S.Term.TermRef -> Maybe S.Reference
-                  getTermSRef (C.ReferenceBuiltin t) = Just (C.ReferenceBuiltin (tIds Vector.! fromIntegral t))
-                  getTermSRef (C.Reference.Derived (Just h) i) = Just (C.Reference.Derived (oIds Vector.! fromIntegral h) i)
-                  getTermSRef _selfCycleRef@(C.Reference.Derived Nothing _) = Nothing
+                  getTermSRef :: S.Term.TermRef -> S.Reference
+                  getTermSRef = \case
+                    C.ReferenceBuiltin t -> C.ReferenceBuiltin (tIds Vector.! fromIntegral t)
+                    C.Reference.Derived mh i ->
+                      C.Reference.Derived (maybe oId (\h -> oIds Vector.! fromIntegral h) mh) i
```
ChrisPenner: Might actually still be worth it to split this into the `Nothing` and `Just` cases so you can add a comment on the Nothing case specifying that we include self-references in the dependency index, similar to how the old version documented that this was the self-reference case with the `_selfCycleRef@`

ChrisPenner (approved):
I'd prefer the filtering happens in the SQL query than in a post-processing state, might as well use the DB for what it's good at.

This will also mean that the index is out of date, what's the plan for fixing that? I don't see anything mentioned and would prefer that it be documented somewhere that the index might be out of date.

Even if it just gets dumped into a ticket for migrating base or something that'd be a good thing to have.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -645,31 +646,60 @@ addToDependentsIndex dependency dependent = execute sql (dependency :. dependent
     ON CONFLICT DO NOTHING
   |]
 
+-- | Get non-self, user-defined dependents of a dependency.
 getDependentsForDependency :: DB m => Reference.Reference -> m [Reference.Id]
-getDependentsForDependency dependency = query sql dependency where sql = [here|
```
aryairani, line 649: Nit: I hate how ormolu breaks this one line into four, that's why there's ormolu_disabled in this file.  https://github.com/tweag/ormolu/issues/696

aryairani (approved):

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
mitchellwrosen: I had a hard time retaining the formatting and also including a second thing in the where-clause. :/

aryairani:
@mitchellwrosen Would it make more sense to merge this into `topic/rehash-codebase` rather than `trunk`? My reasoning being that, with the filters, I think it doesn't change observable behavior yet anyway other than partially filling in the index.

Or did we have a plan to do more for #2649 that would be observable before `topic/rehash-codebase`? (e.g. rework the APIs and lift the filters?) in which case should it just go into that PR?

I don't see any harm in merging it into `trunk` though, especially with the migration forthcoming, to complete everyone's indexes (to @ChrisPenner's point).

mitchellwrosen:
I think it's probably better to keep topic/rehash-migration smaller and merge this ahead of time, even though the index won't be repaired until topic-rehash-codebase lands. But I could go either way.

aryairani:
I meant to say "I don't see any harm", and edited my previous comment to reflect.

PR #2671, opened 2021-11-24
Optimize relation Eq/Ord instances a bit

mitchellwrosen:
## Overview

While looking over a profile of `update` I noticed `Branch.==` was taking some time, which I thought was odd. Apparently we call `Causal.consDistinct` many times, for reasons I don't yet totally understand, when doing an `update`.

Anyway, one easy win here is to just perform Eq/Ord operations on one dimension, since the others contain the same data. This PR implements that, and arbitrarily chooses the first dimension.

Memory usage drops quite a bit on this transcript, which just calls `update` 32 times in a row. Note that the numbers here include all of them cpu/mem attributed to just starting the app up and getting into the transcript file, but I'm not sure of an easy way to omit that phase when generating a report.

Screenshot from 2021-11-24 10-33-09

### DIFF
--- a/lib/unison-util-relation/src/Unison/Util/Relation.hs
+++ b/lib/unison-util-relation/src/Unison/Util/Relation.hs
@@ -107,10 +107,12 @@ module Unison.Util.Relation
 where
 
 import qualified Control.Monad as Monad
+import Data.Function (on)
 import qualified Data.List as List
 import qualified Data.Map as M
 import qualified Data.Map as Map
 import qualified Data.Map.Internal as Map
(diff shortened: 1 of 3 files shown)


ChrisPenner (approved):
Looks good!

Mostly unrelated to this change, but I think we should probably not export the `domain` and `range` setters since they have invariants on them, maybe we should replace with `_domain` and `_range` then define `domain` and `range` to be only getters?

mitchellwrosen:
Agree!

pchiusano (approved):
Nice catch.

PR #2674, opened 2021-11-24
Merge a couple util modules

### LINKED ISSUE #2673: Merge containers util modules
Right now we have multiple copies of some util modules (e.g. `Unison.Util.Set`, `U.Util.Set`), which exist in different packages and are drifting apart.

I think we should move these into the `unison-prelude` package, as they are just generally useful `containers` utils, which `unison-prelude` already depends on (to re-export the types).

mitchellwrosen:
## Overview

This PR merges

- `codebase2/util/src/U/Util/{Map,Set}.hs`
- `unison-core/src/Unison/Util/{Map,Set}.hs`

into

- `lib/unison-prelude/src/Unison/Util/{Map,Set}.hs`

Fixes #2673


ChrisPenner (approved):
LGTM 👍🏼

aryairani (approved):

PR #2681, opened 2021-11-24
Fix delete.namespace.force msg spacing

ceedubs:


ceedubs:
Hmm I thought that this was a straightforward enough change that I could get buy without running the tests. I'll have to look into why this failed next week. It's 🦃 time.

mitchellwrosen:
Ah, it's a transcript failure due to the typo fix. You can run this locally:

```
stack build
stack exec transcripts
git add .
```

I'd do it for you, but I don't have access to your clone of `unisonweb` :)

aryairani:
Because `git add .` can so easily add scratch files or ide temp files, I wouldn't recommend recommending that command. 😅

ceedubs:
@mitchellwrosen maybe I'm misunderstanding, but I think that the build failure is correctly pointing out that I assumed that my change would add the space to the transcript output but it _doesn't_. Taking a brief look at some of the pretty-printing code it seems like adding a space to the end of the string literal like this is probably not the right approach. It seems like using `<>` should take care of this for me, but maybe it has to do with `IP.patternName` not adding in a `P.wrap` or something?

unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -322,7 +322,7 @@ notifyNumbered o = case o of
           "",
           endangeredDependentsTable ppeDecl endangerments,
           "",
-          P.wrap "If you want to proceed anyways and leave those definitions without names, use"
+          P.wrap "If you want to proceed anyways and leave those definitions without names, use "
             <> IP.patternName IP.deleteNamespaceForce
```
```suggestion
          P.wrap ("If you want to proceed anyways and leave those definitions without names, use"
            <> IP.patternName IP.deleteNamespaceForce)
```
This might work, but I'm not sure.

aryairani:
@ceedubs Did you manually edit `delete-namespace.output.md` and check that in?

In any case, the build failure is just saying "the transcript output CI produced doesn't match the transcript output in VCS". 

Duplicate / leading / trailing spaces are discarded by `P.wrap`, so the code that's checked in isn't producing the output you want either.  I think `(<>)` will do that for you if it's inside of `P.wrap`, though I wouldn't bet money on it.

ceedubs:
> Did you manually edit delete-namespace.output.md and check that in?

@aryairani yes that's exactly what I did.

I'll take a closer look at this sometime probably 😅 .

aryairani:
I'm resisting OCD to fix it for you before this somehow rots in PR purgatory.

aryairani:
Ping @ceedubs

ceedubs:
Ah right. Evasive missing space, we meet again.

I'll take a look.

ceedubs:
@aryairani should be good now

aryairani (approved):
Super!

PR #2682, opened 2021-11-25
Fix Star3 deleteD2, deleteD3

mitchellwrosen:
## Overview

I believe this PR fixes a small bug in `Star3`.

A `Star3 fact d1 d2 d3` represents a set of `(fact, d1)`, a set of `(fact, d2)`, and a set of `(fact, d3)`. Additionally, it caches the set of `fact` itself, which is otherwise not recoverable in `O(1)` time.

So when deleting a single tuple like `(x, y)`, we need to additionally possibly delete `x` from the cached set of facts, if it is no longer related to any `d1`, `d2`, or `d3`.

## Test coverage

I have not added any tests, as `Star3` is not currently under test, so it was extra effort. I think I probably should add a test suite, but I was hoping the code would just look correct at a glance (=

### DIFF
--- a/parser-typechecker/src/Unison/Util/Star3.hs
+++ b/parser-typechecker/src/Unison/Util/Star3.hs
@@ -166,15 +166,26 @@ deleteD3 :: (Ord fact, Ord d1, Ord d2, Ord d3)
           => (fact, d3)
           -> Star3 fact d1 d2 d3
           -> Star3 fact d1 d2 d3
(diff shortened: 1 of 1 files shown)


aryairani:
@pchiusano Is it a bug or a feature if a Star3 contains a `fact` not represented in any of its relational dimensions?

aryairani (approved):
¯\\\_(ツ)\_/¯

PR #2686, opened 2021-11-29
Match UI release zip file of unisonLocal.zip

hojberg:
## Overview
The Codebase UI changed its main artifact of ucm.zip to unisonLocal.zip,
make sure we build using the right artifact.


ceedubs (approved):
🍔

PR #2687, opened 2021-11-29
add toText functions for Socket, Handle, ThreadId

stew:
## Overview

This adds toText builtins for Handle ThreadId Socket, Being able to log text representations of these types is very useful for debugging


dolio (approved):
Looks fine.

I think the build problems are just hash changes in transcripts that need to be recommitted.

ceedubs (approved):

Issue #2693, opened 2021-12-01
Segfault in ucm for watch expression with larger lists


hagl:
The following code causes a segfault in ucm on MacOS
```
list = List.range 0 2000
> List.map id list
```
Here is the output of ucm:
```
.>

  I found and typechecked these definitions in ~/projects/unison/bug.u. If you
  do an `add` or `update`, here's how your codebase would change:

    ⍟ These new definitions are ok to `add`:

      list : [Nat]

  Now evaluating any watch expressions (lines starting with `>`)...
  Ctrl+C cancels.

Segmentation fault: 11
```

If I add the `list` to the workspace first and then use the watch expression it works fine. Also if I inline the list term or use a smaller-sized list.

aryairani:
@dolio Any ideas on this? (Can you even repro on linux? I haven’t tried on mac yet.)

dolio:
It doesn't seem to be happening here, no. Strange.

Actually, scratch that. It just doesn't happen every time, I think.

dolio:
Okay, I think I figured out what's happening.

It actually works the first time you run it. But then unison caches a gigantic list literal and tries to use that instead the next time it happens in a watch. That's the thing that causes a problem.

dolio:
Should be fixed now.

PR #2704, opened 2021-12-02
Fix potential stack overrun

dolio:
Extremely large literals could cause functions to run off the end of the stack, because the function that ensures enough stack is available before a call was using a fixed extension size (not anticipating such large literals).


dolio:
Fixes #2693

ceedubs (approved):
Nice thanks @dolio!

pchiusano (approved):

pchiusano:
Nice!

PR #2725, opened 2021-12-05
Fix journal mode parsing

mitchellwrosen:
Fix journal mode parsing.


aryairani (approved):
Weird. Has it always returned lower-case?  I guess I just copied the upper-case names from [the spec](https://sqlite.org/pragma.html#pragma_journal_mode).

mitchellwrosen:
I think those uppercase names are my mistake :)

aryairani:
Oh phew yes lol. https://github.com/unisonweb/unison/blob/377d7cc96a4a120c2603c02747e1bcdccb985f0b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs#L230-L239

PR #2740, opened 2021-12-08
Add Debug.trace builtin

dolio:
This adds a new debug tracing function with type `Text -> a -> ()`. The text value is an extra label that shows up in the output. The function attempts to decompile the arbitrary value and pretty print it, but if that isn't possible, it prints out the raw representation.


dolio:
CI wasn't running properly on the commits. Not sure why (something about the release section in some file). I guess we'll see if the merge works.

pchiusano (approved):
Awesome! Really looking forward to this.

PR #2750, opened 2021-12-09
Monomorphize `v` to `Symbol` for builtins

mitchellwrosen:
## Overview

This PR monomorphizes `v` to `Symbol` for top-level builtin almost-CAFs that are expensive to compute. Without the type class constraint, they become true CAFs, and work is shared.


aryairani (approved):

pchiusano (approved):
Just discussed this with @mitchellwrosen and I think it's good to merge.

Basically, the only place we really need polymorphism over the variable type is in the ASTs and in typechecking, where we replace variables with more interesting variable types (tracking whether a variable is existential or universal). 

So this PR leaves those alone and is really just removing some gratuitous polymorphism that wasn't really needed for anything.

PR #2752, opened 2021-12-10
add a script to perform checks tests similar to CI

aryairani:
## Overview

Run `./scripts/check.sh` to run our 6 test suites, for fewer surprises in PRs.

The script short circuits after the first failing test suite, which is a bummer, because I'd like to know about all failing tests, but it is tough to find the exit status of earlier test suites if we run them all too.

### DIFF
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -2,7 +2,7 @@
 
 # eventually: ormolu -c `find . -name '*.hs'`
 true \
-  && stack build --fast \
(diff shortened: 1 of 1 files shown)


pchiusano (approved):

ChrisPenner (approved):
Probably worth adding the few suggestions Mitchell had over the call a while back (with regard to `stack build --test --no-run-tests`

PR #2753, opened 2021-12-10
Bugfix: make hashing effects order-independent

mitchellwrosen:
## Overview

- [x] Run this implementation by Paul
- [x] Close this and make the PR to `topic/rehash-codebase` instead

Previously, terms that are identical other than the order a set of effects is listed hashed differently. The following two terms should have the same hash, and now they do.

```
foo : Nat ->{Exception, IO} Nat
bar : Nat ->{IO, Exception} Nat
```


pchiusano (approved):

pchiusano:
Just went over this, one of the undefineds is going to be replaced with doing sorting, the other will be left as is.

mitchellwrosen:
#2777

PR #2765, opened 2021-12-17
Tweak Result for performance

mitchellwrosen:
## Overview

This PR makes `Result` fields strict, and makes sure `(<*>)` and `(>>=)` always get inlined.


dolio (approved):
Seems straight forward.

PR #2766, opened 2021-12-17
Refactor MT monad for performance

mitchellwrosen:
## Overview

This PR refactors the `MT` monad transformer to be more efficient.

Previously, each `>>=` would incur unnecessary allocation to reconstruct the `MEnv` type, which represented a 3-tuple of two read-only things and one read-write thing.

Now, the type is represented as a 3-argument function, so it can only return the argument that changes. 

In my testing this appears to make `update`, which seems to spend the majority of its time in the type checker, go roughly 7% faster and use 25% less memory.


dolio (approved):
Looks fine to me.

I wouldn't have necessarily expected it to make much of a difference. But maybe our type checking functions aren't concrete enough for GHC to just optimize out the tupling.

PR #2768, opened 2021-12-17
Add getRootBranchExists to the Codebase API

mitchellwrosen:
## Overview

(Context: this came up while working on a PR that moves `unison-codebase-sqlite` over to use `unison-sqlite`).

This PR adds a `getRootBranchExists` function to the codebase API. I noticed we had an inefficient `isBlank` function in its API, which pulled the whole root branch into memory, only to check whether it's empty.

I don't think it's possible with the current onboarding flow to end up asking whether a non-blank codebase is blank, but it looks like from a comment that this flow is intended eventually:

```
-- Previously Created is different from Previously Onboarded because a user can 
-- 1.) create a new codebase 
-- 2.) decide not to go through the onboarding flow until later and exit
-- 3.) then reopen their blank codebase 
```

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -37,8 +37,8 @@ module Unison.Codebase
 
     -- * Root branch
     getRootBranch,
+    getRootBranchExists,
     GetRootBranchError (..),
-    isBlank,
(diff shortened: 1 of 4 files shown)


ChrisPenner (approved):
Huh, interesting! Seems like a good optimization 👍🏼

PR #2783, opened 2022-01-06
Install a top level exception handler during compiled execution

dolio:
Some stuff @stew and @ceedubs were working on was getting weird errors during execution because compiled evaluation didn't have a top level exception handler. This should make it have a somewhat more normal message.

### DIFF
--- a/parser-typechecker/tests/Unison/Test/MCode.hs
+++ b/parser-typechecker/tests/Unison/Test/MCode.hs
@@ -72,7 +72,7 @@ env m = mapInsert (bit 24) m
 asrt :: Section
 asrt = Ins (Unpack Nothing 0)
      $ Match 0
-     $ Test1 1 (Yield ZArgs)
+     $ Test1 1 (Yield (BArg1 0))
                (Die "assertion failed")
 
 multRec :: String
--- a/parser-typechecker/src/Unison/Runtime/Interface.hs
+++ b/parser-typechecker/src/Unison/Runtime/Interface.hs
@@ -362,9 +362,16 @@ executeMainComb
   -> CCache
   -> IO ()
 executeMainComb init cc
-  = eval0 cc Nothing
+  = flip catch (putStrLn . toANSI 50 <=< handler)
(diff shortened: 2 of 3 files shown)


parser-typechecker/tests/Unison/Test/MCode.hs
```diff
@@ -72,7 +72,7 @@ env m = mapInsert (bit 24) m
 asrt :: Section
 asrt = Ins (Unpack Nothing 0)
      $ Match 0
-     $ Test1 1 (Yield ZArgs)
+     $ Test1 1 (Yield (BArg1 0))
```
ceedubs, line 75: Can you help me understand this change from `ZArgs` to `BArg1 0`?

ceedubs (approved):
I won't claim to totally understand this change, but it seems pretty straightforward, so if it has the desired effect 👍 from me. Thanks, @dolio!

parser-typechecker/tests/Unison/Test/MCode.hs
dolio: The tests were all implemented with this trivial function that tests its argument, and either does a zero-argument return or throws an (Haskell) exception. Then the tests just wrap it in something that will be successful if it terminates, and a failure if the exception is thrown. You can think of the zero-argument return as an optimized version of `a -> ()`, where we've optimized out building a boxed `()`.

The top level exception handler expects a value to be percolating up, though, for the default case. So the zero-argument return was making it segfault by reading outside the stack (because nothing was on the stack). So, instead I made the test function just return its boxed input in the successful case (that's the `BArg1 0`). Then the exception handler gets this value. It does nothing with it, but it's necessary to keep it from segfaulting.

aryairani:
Thanks!

parser-typechecker/tests/Unison/Test/MCode.hs
ceedubs: @dolio thanks that was a super helpful explanation

PR #2803, opened 2022-01-13
Fix universal eq/compare for links

### LINKED ISSUE #2431: Segmentation fault when displaying certain docs
```
You are running version: latest-59-gbcdb0d551' (built on 2021-09-21)
```

```
.> view .base.Exception.catchOnly.doc

  base.Exception.catchOnly.doc : base.Doc
  base.Exception.catchOnly.doc =
    use Any Any
    use Exception catchOnly
    use base.Exception raise
    use base.syntax docEval
    {{
    Tries to run code which potentially raises an {type base.Exception}, transforming it to
    an {type Either}. If an exception is raised, {catchOnly} catches only the given failure
    type {Type} as a {Left}, otherwise it re-raises the exception.

    # Examples

      {catchOnly} should wrap the type {type TlsFailure} in a {Left} when an
      {type base.Exception} of that type is raised.

      ```
      unsafeRun!
        '(catchOnly
          (typeLink TlsFailure) '(raise (Failure (typeLink TlsFailure) "Oh no" !Any)))
      ```

      Here's an example with no type match below. We're re-catching the raised exception
      using {catch} for rendering.

      ```
      catch
        '(catchOnly
          (typeLink Generic)
          '(raise (Failure (typeLink TlsFailure) "Exception type did not match" !Any)))
      ```
    }}

.> display .base.Exception.catchOnly.doc

  That term is ambiguous.

.> cd .base.Exception.catch
.base.Exception.catch      .base.Exception.catchMany  .base.Exception.catchOnly
.> cd .base.Exception.catchOnly
.base.Exception.catchOnly> display doc
Segmentation fault: 11
```

dolio:
It seems that optimizations cause problems together with the combination
of coercing and unsafe IO involved in reference equality. I'm not
exactly sure what goes wrong, but it causes segfaults. Defining a
noinline function for the equalities (instead of just `(==)@Reference`)
seems to fix the problem.

I got the same problem to happen for term links, so both of those use a
noinline function now. Other foreign values don't seem to be affected,
which I assume is due to them having pure primitive functions for
equality.

Fixes #2431

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Foreign.hs
+++ b/parser-typechecker/src/Unison/Runtime/Foreign.hs
@@ -40,11 +40,27 @@ data Foreign where
 promote :: (a -> a -> r) -> b -> c -> r
 promote (~~) x y = unsafeCoerce x ~~ unsafeCoerce y
 
+tylEq :: Reference -> Reference -> Bool
+tylEq r l = r == l
+{-# noinline tylEq #-}
+
+tmlEq :: Referent -> Referent -> Bool
+tmlEq r l = r == l
+{-# noinline tmlEq #-}
+
+tylCmp :: Reference -> Reference -> Ordering
+tylCmp r l = compare r l
+{-# noinline tylCmp #-}
+
+tmlCmp :: Referent -> Referent -> Ordering
+tmlCmp r l = compare r l
+{-# noinline tmlCmp #-}
+
 ref2eq :: Reference -> Maybe (a -> b -> Bool)
 ref2eq r
   | r == Ty.textRef = Just $ promote ((==) @Text)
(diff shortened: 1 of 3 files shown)


ceedubs (approved):
Nice @dolio. LGTM.

I'm interpreting this as something that might be a minor performance hit for boxed types but that shouldn't affect equality on primitive types like `Nat` and `Int`. Does that sound right?

dolio:
I didn't really change anything except avoiding some inlining in the implementation of equality for term and type links. So it shouldn't affect performance of anything else, and the performance of the affected cases don't really matter, because they were just crashing before.

ceedubs:
Ah that makes sense. And going from crashing to working is indeed the best performance improvement 😄. Ship it!

PR #2806, opened 2022-01-14
make computeFrontier more efficient, rename to computeDirty

mitchellwrosen:
## Overview

This PR just makes `computeFrontier` (now `computeDirty`) a little more efficient and compact, by only computing the dirty set, rather than the dirty set and frontier, since only the dirty set is needed.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/Propagate.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/Propagate.hs
@@ -39,6 +39,7 @@ import           Unison.Term                    ( Term )
 import           Unison.Util.Free               ( Free
                                                 , eval
                                                 )
+import           Unison.Util.Monoid             ( foldMapM )
 import qualified Unison.Util.Relation          as R
 import           Unison.Util.TransitiveClosure  ( transitiveClosure )
 import           Unison.Var                     ( Var )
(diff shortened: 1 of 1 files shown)


unison-cli/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -623,37 +624,25 @@ applyPropagate patch Edits {..} = do
...
   -> Names
-  -> m (R.Relation Reference Reference)
-computeFrontier getDependents patch names = do
-      -- (r,r2) ∈ dependsOn if r depends on r2
-  dependsOn <- foldM addDependents R.empty edited
-  -- Dirty is everything that `dependsOn` Frontier, minus already edited defns
-  pure $ R.filterDom (not . flip Set.member edited) dependsOn
- where
+  -> m (Set Reference)
+computeDirty getDependents patch names =
```
mitchellwrosen, line 639: This function is supposed to compute dirty references in the same way as the old version, it just doesn't bother maintaining the `Relation Reference Reference` ("dirty" and "frontier"), since just the `Set Reference` domain ("dirty") is needed by the caller.

ChrisPenner (approved):
Seems good to me 👍🏼

aryairani (approved):
Seems ok to me too.

PR #2809, opened 2022-01-14
Special-case overapplied binary operators

### LINKED ISSUE #2796: Pretty Printer change: overapplied operators should render with infix syntax
For functions like "compose" (.) or (<<) the renderer will not use infix syntax leading to somewhat confusing code. 

The following expression 
```
(f . g) x
```

renders as the following: 
```
(.) f g x 
```

This isn't a huge issue, it came up while documenting function application, but if it's easy to do that would be awesome.

runarorama:
Fixes #2796 by adding a special case to the term printer for overapplied operators.

Before:

```
(|>) f g a
```

After: 

```
(f |> g) a
```


pchiusano (approved):
Nice! Sorry I missed this.

Issue #2492, opened 2021-10-13
pretty printer roundtrip issue with nested constructors involving lists


ceedubs:
Given this (simplified) JSON type:

```haskell
unique type Json
  = JString Text
  | JObject [(Text, Json)]
  | JArray [Json]
```

This parses:

```haskell
myJson = JArray [ JObject
    [ ("ID", JString "40e4a748-2192-161a-0510-9bf59fe950b5"),
      ("Node", JString "t2.320")
    ]
  ]
```

Now we watch it to pretty-print it:

`> myJson`:

```haskell
Json.JArray
  [ Json.JObject
    [ ("ID", Json.JString "40e4a748-2192-161a-0510-9bf59fe950b5"),
      ("Node", Json.JString "t2.320") ] ]
```

But if we copy that output:

```haskell
> let
  Json.JArray
    [ Json.JObject
      [ ("ID", Json.JString "40e4a748-2192-161a-0510-9bf59fe950b5"),
        ("Node", Json.JString "t2.320") ] ]
```

We get:

```
  unexpected [
  expecting ,
    110 |       [ ("ID", Json.JString "40e4a748-2192-161a-0510-9bf59fe950b5"),
```

kanwren:
After doing some digging, this looks like a specific case of a more general and annoying round-trip problem: hanging the arguments to a function application might cause them to end up at the same indentation level as the function itself (due to parentheses, lists, ticks, etc.), which causes the parse error.

For example, if there is an application like this, in which the arguments are long and the line breaks:
```
f
  veryLongExpression1
  veryLongExpression2
```
This is fine, but if this expression needs to be parenthesized _exactly twice_, the following will happen:
```
((f
  veryLongExpression1
  veryLongExpression2))
```
This causes a parse error, since `f` is at the same column as the arguments. This isn't that the arguments need to be indented more, though that would work; it would also work to _dedent_ them by one space (or add a third pair of parens).

The issue above looks like the same thing, but with parens replaced by a list. As a contrived example:
```hs
ex : [Nat]
ex =
  [ Nat.and
      (123456789 + 123456789 + 123456789 + 123456789 + 123456789 + 123456789)
      (123456789 + 123456789 + 123456789 + 123456789 + 123456789 + 123456789) ]
```
This code fails to roundtrip, as it reads and pretty-prints to this:
```hs
ex : [Nat]
ex =
  use Nat +
  [ Nat.and
[...]

PR #2819, opened 2022-01-20
Fix incorrect hanging of boolean operators, and elide extra parentheses

### LINKED ISSUE #2818: Roundtrip error from naive hanging of boolean operators
Given the following example code:

```hs
example =
  ("a long piece of text to force unison to hang the line" == "")
    && ("a long piece of text to force unison to hang the line" == "")
```

This parses, but when viewed pretty-prints to this:
```hs
example : Boolean
example =
  use Text ==
  ("a long piece of text to force unison to hang the line" == "")
  &&
  ("a long piece of text to force unison to hang the line" == "")
```
which fails to parse. Instead, it should pretty-print to something like this, similar to other infix operators:
```hs
example : Boolean
example =
  use Text ==
  ("a long piece of text to force unison to hang the line" == "")
    && ("a long piece of text to force unison to hang the line" == "")
```

The pretty-printer here does not handle `&&` and `||` the same as other infix operators, and naively incorrectly hangs them when the line breaks, causing the roundtrip error.

kanwren:
## Overview

Per #2818, the pretty-printer does not handle `&&` and `||` the same as other infix operators, and naively incorrectly hangs them when the line breaks, causing a roundtrip error. Previously, breaking a long line with an `&&` or `||` would incorrectly print without indentation, like this:

```
("a long piece of text to hang the line" == "")
&&
("a long piece of text to hang the line" == "")
```

This fails to parse, and primarily happens becaues `&&` and `||` are special-cased and not handled the same as other infix operators. This modifies the pretty-printer to now correctly hang the line as:

```
("a long piece of text to hang the line" == "")
  && ("a long piece of text to hang the line" == "")
```

An additional side effect of the change removes redundant parentheses in adjacent boolean operators; see below.

Resolves #2818

## Implementation notes

This is implemented very similarly to how infix operators are handled, having a special case for a chain of like boolean operators and pretty-printing them all at once.

Another consequence of the old approach was that subsequent `&&`s and `||`s do not elide parentheses like infix operators do. For example, `a && b && c && d` would be rendered as `(((a && b) && c) && d)`. A side effect of the above change is that the pretty-printer will now search for a chain of `&&`s or `||`s and render it without the extra parentheses.

## Controversial decisions

Since infix operators are always left-assocative and at the same fixity, chains of infix operators usually drop all parentheses. For the sake of clarity, this handles chains of `&&` and `||` separately, so that parentheses are kept when `&&` and `||` are mixed. However, this is easy to change if it would be better to elide all parentheses, if this were more in line with unison's current decisions.

For example, take `a && b && c || d && e && f`:

- Previously, this would be printed as `((((a && b) && c) || d) && e) && f`
- Now, this would be printed as `((a && b && c) || d) && e && f`
- If the two cases were optionally merged, it would be printed as `a && b && c || d && e && f`

## Test coverage

A regression test can be added after the PR is created to ensure that splitting a line with a boolean operator will roundtrip. Let me know if more regression tests are necessary, or if changes need to be made elsewhere!

## Loose ends

While testing, I ran into a special case of #2492, which I do not address here. Consider this code:

```hs
hangExample1 =
  ((or
      ("a long piece of text to hang the line" == "")
      ("a long piece of text to hang the line" == ""))
    && true)
    || true
```

Before, this would be pretty-printed as:
```hs
hangExample1 : Boolean
hangExample1 =
  use Text ==
  ((Boolean.or
    ("a long piece of text to hang the line" == "")
    ("a long piece of text to hang the line" == ""))
  &&
  true)
  ||
  true
```
which is clearly a roundtrip error.

Now, it will be printed as the following:
```hs
hangExample1 : Boolean
hangExample1 =
  use Text ==
  ((or
    ("a long piece of text to hang the line" == "")
    ("a long piece of text to hang the line" == ""))
    && true)
    || true
```
This is more correct, but because this happens to put two parentheses before the `or`, it runs into the same error observed in #2492.

This would be my first contribution to unison; I'm open to any feedback/suggestions/alternatives!

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -64,3 +64,4 @@ The format for this list: name, GitHub handle, and then optional blurb about wha
 * Shawn Bachlet (@shawn-bachlet)
 * Solomon Bothwell (@solomon-b)
 * Sameer Kolhar (@kolharsam)
+* Nicole Prindle (@nprindle)
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -277,18 +277,6 @@ pretty0
        pf     = branch f
        branch tm = let (im', uses) = calcImports im tm
                    in uses $ [pretty0 n (ac 0 Block im' doc) tm]
-    And' x y ->
-      paren (p >= 10) $ PP.spaced [
-        pretty0 n (ac 10 Normal im doc) x,
-        fmt S.ControlKeyword "&&",
-        pretty0 n (ac 10 Normal im doc) y
-      ]
-    Or' x y ->
-      paren (p >= 10) $ PP.spaced [
-        pretty0 n (ac 10 Normal im doc) x,
-        fmt S.ControlKeyword "||",
-        pretty0 n (ac 10 Normal im doc) y
-      ]
     LetBlock bs e ->
       let (im', uses) = calcImports im term
       in printLet elideUnit bc bs e im' uses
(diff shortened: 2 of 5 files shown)


pchiusano (approved):
Great, thanks for submitting this, LGTM, could you just add yourself to CONTRIBUTORS.markdown and we'll get this merged?

kanwren:
Thanks! I've added myself to the list.

pchiusano:
Great, thanks! It should merge on its own after CI completes.

PR #2824, opened 2022-01-20
add custom error message for PatternArityMismatch

### LINKED ISSUE #2823: Reporting of pattern arity mismatch errors is weirdly slow on a large file
@stew and @ceedubs have noticed that when editing a scratch file with ~3500 LOC, all other typechecking errors are reported in a second or so, but if there's a pattern arity mismatch (`cases Foo x y -> ...` when that constructor has fields, say), it takes **literally 7 minutes** to report the error on a 16 core machine with 32 GB of RAM.

Things that we've tried:

* Disabling TDNR completely in the typechecker, we did that by changing `Term.prepareTDNR` to always return `Nothing` and not insert blanks and also by removing the call to `typeDirectedNameResolution` inside of `synthesizeAndResolve`. I think this this should do it?
* If you just move the definition with the arity mismatch above the fold so it's the only thing that's typechecked, it goes very fast.
* Adding 7000 other random definitions to the same file which has the arity mismatch still reports the arity mismatch in less than a second.

So it's not the sheer number of terms in the file, and it doesn't seem TDNR related, it's something about certain terms.

We don't have a good self-contained test case but contact @stew about reproducing.

stew:
When you have PatternArityMismatch in a large file, you used get a generic error message this prints a better error message.

This also eliminates the printing of the "path" in the generic error message, as it was taking several minutes on a large code base 

fixes: #2823

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -526,6 +526,15 @@ renderTypeError e env src = case e of
            [Type.var () (Var.named "e"), Type.var () (Var.named "a") ])
         (Type.var () (Var.named "o"))
 
+  Other (C.cause -> C.PatternArityMismatch loc typ num) ->
+    Pr.lines [
+      Pr.wrap "This pattern has the wrong number of arguments", "",
+      annotatedAsErrorSite src loc,
+      "The constructor has type ", "",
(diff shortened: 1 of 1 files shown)


pchiusano (approved):
Sweet. 

@stew and I worked on this together.

dolio (approved):
Looks fine to me.

I guess the path output is unlikely to be useful to people who aren't working on the typechecker anyway?

PR #2827, opened 2022-01-21
Fix pattern matching in docs

### LINKED ISSUE #2778: Docs pattern match issue in fenced codeblocks
It appears that expressions which use pattern matching in fenced codeblock elements in the Docs have their case statements reordered upon evaluation. For example the following should be `true`: 

````
myDoc = {{

  I am a doc

  ```
  match 1 with
    1 -> true
    _   -> false
  ```
}}
````

But this `display`s as 

```
.> display myDoc

  I am a doc
  
      _pattern = 1
      false
      ⧨
      false
```

A transcript with doc file to load is attached to replicate. A workaround I'm using where evaluation is not key for docs writing is to simply load the source into the Doc with `@source`. 

[docLoad.txt](https://github.com/unisonweb/unison/files/7803587/docLoad.txt)
[reorderingMatch.md](https://github.com/unisonweb/unison/files/7803588/reorderingMatch.md)

dolio:
It seems that docs can make multiple trips through the runtime with the decompiler in the middle. The decompiler was producing bad output for matching blocks, reordering cases in an invalid way and such, resulting in subsequent evaluation giving the wrong answer. This tweaks the way that decompilation information is collected to produce something valid in this case.

Fixes #2778

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Interface.hs
+++ b/parser-typechecker/src/Unison/Runtime/Interface.hs
@@ -280,15 +280,17 @@ intermediateTerm
   -> Term Symbol
   -> (SuperGroup Symbol, Map.Map Word64 (Term Symbol))
 intermediateTerm ppe ref ctx tm
-  = final
+  = first ( superNormalize
+          . splitPatterns (dspec ctx)
+          . addDefaultCases tmName
+          )
+  . memorize
   . lamLift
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
Nice!

Just to clarify, you're still working on a fix for retaining type signatures in decompiled code?

PR #2830, opened 2022-01-22
Fix pretty-printing of `if`s with multiline conditions

### LINKED ISSUE #518: Missing line-break before `then` after multi-line `if` condition
The pretty-printer emits the following code:
```
if
  use A x
  f x x then
  x
else y
```
Note the missing line-break before `then`.  

The problem will be something to do with the height checks in the `If'` branch of `pretty` in TermPrinter.hs.

kanwren:
## Overview

Per #518, ifs with multiline conditions would previously pretty-print incorrectly. For example:

```hs
if
  a = b
  a
then
  foo
else
  bar
```

would print as:

```hs
if
  a = b
  a then foo else bar
```

This fixes the pretty-printing of ifs, so that the above is now printed as

```hs
if
  a = b
  a
then foo
else bar
```

This does not change the printing of any `if`s that were already pretty-printed correctly (those with a single-line condition).

Fixes #518.

## Implementation notes

A new condition is added to the current `pretty0` handling of `If'` for when the condition part of an `if` is multiline.

The changes adhere to the old conventions as closely as possible, though that keeps things on the same line more often than I would personally. If any of the formatting is not ideal, it would be easy to change that as well while I'm at it.

## Test coverage

~~A regression test transcript has been added to make sure that `if` statements are pretty-printed as expected, for all 8 combinations of single-line or multi-line components.~~
The associated pretty-printer tests will be updated.


kanwren:
Ah, forgot about the termprinter test that referenced the issue; I'll fix that in a moment

pchiusano (approved):
Looks good, thanks! I'd just update that unit test and then get CI passing and we'll merge.

aryairani:
This is an exemplary pull request! Thank you.
P.S. Go Jackets!
P.P.S. I'm not sure why the `tls.md` transcript is failing?

kanwren:
Neither am I; that doesn't look like anything that I did? It fails for me when running locally on `trunk` too, which is strange.

aryairani:
@nprindle Disregard, it looks like an unrelated issue, we are tracking it at https://github.com/unisonweb/unison/issues/2834.

PR #2835, opened 2022-01-24
Fix some of the TLS error handling

stew:
this is a partial fix for #2834 more will have to come. Now there is a
cert there valid for 10 years and we load it and the private key with
now problem, the call to decode the cert now is checking for
exceptions and is properly handling the case when we successfully
decode 0 certs.

Now for some reason the last section of the transcript where we try to
do handshaking, we are Segfaulting on the one of the sides of the
handshake. I haven't figured out why yet.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Builtin.hs
+++ b/parser-typechecker/src/Unison/Runtime/Builtin.hs
@@ -1841,7 +1841,7 @@ declareForeigns = do
     \(config :: TLS.ServerParams,
       socket :: SYS.Socket) -> TLS.contextNew socket config
 
-  declareForeign "Tls.handshake.impl.v3" boxToEFBox . mkForeignTls $
+  declareForeign "Tls.handshake.impl.v3" boxToEF0 . mkForeignTls $
     \(tls :: TLS.Context) -> TLS.handshake tls
 
   declareForeign "Tls.send.impl.v3" boxBoxToEFBox . mkForeignTls $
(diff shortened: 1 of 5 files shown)


stew:
I'm going to ask that we merge this as-is (pending CI). It does NOT fix #2834, but does fix some of the bugs that are contributing to this test failing.

The test still currently fails due to a not yet understood bug in TLS.handshake that only seems to affect servers. We don't know of anyone currently using this code, so I think that it makes sense to merge this in order to unclog the CI pipeline until a more complete fix comes along, re-introducing the tests which are disabled by this PR

pchiusano (approved):

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1841,20 +1826,25 @@ declareForeigns = do
...
     \(tls :: TLS.Context,
       bytes :: Bytes.Bytes) -> TLS.sendData tls (Bytes.toLazyByteString bytes)
 
   let wrapFailure t = Failure Ty.tlsFailureRef (Util.Text.pack t) unitValue
       decoded :: Bytes.Bytes -> Either String PEM
-      decoded bytes = fmap head $ pemParseLBS  $ Bytes.toLazyByteString bytes
+      decoded bytes = case pemParseLBS $ Bytes.toLazyByteString bytes of
+        Right (pem : _) -> Right pem
+        Right _ -> Left "no PEM found"
+        Left l -> Left l
```
stew, line 1843: we obviously need to not just blindly call head here...

pchiusano:
I think this is fine as a temp fix, approved and it should merge automatically after CI passes.

Issue #2237, opened 2021-07-16
IO sandboxing to prevent access to arbitrary I/O during evaluation


pchiusano:
This needs design, but we want to be able to evaluate expressions without giving them access to arbitrary IO.  This will get used for making sure docs can’t do I/O even if the user subverts the typechecker, and it’ll also be used by the compute node server.

Here’s a sketch:

```
handlePure blah
with someHandler
```
I’m imagining that to typecheck `blah` in a `handlePure` block, we remove `{IO}` from the ambient abilities.

And then also at runtime, `handlePure` disables calling any (certain?) `IO` functions within `blah` - it can still make other kinds of requests, and `someHandler` may do `IO` on behalf of `blah`, but that’s what we want - this lets us create more limited abilities that have access to safer subsets of IO.

Alternately, maybe just a new builtin function:

```
IO.sandbox : (Request {g} a ->{g2} r) -> ‘{g} a ->{g2} r
```

Runtime support would be the same.

Phase 1 of this is just a design discussion with @dolio

dolio:
Here's an alternate idea we discussed.

If the concern is just remotely loaded code, we could keep track of which definitions transitively reference primitive `IO` functions. Then, just refuse to deserialize (or serialize) anything that includes one of those operations (throw an exception). Now remote code can only have `IO` effects via an intermediate API that the server executing the code knows how to fulfill.

A possible downside is that this relies more on correctness of builtins and compilation. The actual code representation that is sent remotely uses variables, so you can't just arbitrarily craft bad stack references to try to access forbidden things. But, improperly wrapped builtins or compiler bugs could introduce bad stack references, so that is a conceivable attack vector. I'm unsure if the sandboxing wrapper would be completely immune to something like that or not.

pchiusano:
Thanks for the writeup. It seems like you could also (or instead?) have a builtin, `validatePure : a -> Boolean` which just checks if any transitive dependencies of the input calls a forbidden function. This would be more useful for the doc evaluation case where you don't really want to have to serialize the example embedded in the doc just to check that it's okay to evaluate.

[...]

PR #2836, opened 2022-01-25
IO sandboxing check

dolio:
Add a prim op for checking if a value references sensitive IO operations.

I wasn't exactly sure of the status of some of the builtins, so I may have been overzealous about marking some stuff off limits. Some folks should audit the choices of what is allowed and what is not anyway.

Addresses #2237, although maybe we still want to do the more elaborate solution as well.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Builtin.hs
+++ b/parser-typechecker/src/Unison/Runtime/Builtin.hs
@@ -1794,7 +1794,7 @@ declareForeigns = do
   declareForeign Tracked "Tls.ServerConfig.default" boxBoxDirect $ mkForeign
     $ \(certs :: [X.SignedCertificate], key :: X.PrivKey) ->
         pure $ (def :: TLS.ServerParams) { TLS.serverSupported = def { TLS.supportedCiphers = Cipher.ciphersuite_strong }
-                                         , TLS.serverShared = def { TLS.sharedCredentials = Credentials [((X.CertificateChain certs), key)] }
+                                         , TLS.serverShared = def { TLS.sharedCredentials = Credentials [(X.CertificateChain certs, key)] }
                                          }
 
   let updateClient :: X.CertificateStore -> TLS.ClientParams -> TLS.ClientParams
@@ -1843,21 +1843,6 @@ declareForeigns = do
   declareForeign Untracked "Ref.write" boxBoxTo0 . mkForeign $
     \(r :: IORef Closure, c :: Closure) -> writeIORef r c
 
-  let
-    defaultSupported :: TLS.Supported
-    defaultSupported = def { TLS.supportedCiphers = Cipher.ciphersuite_strong }
-
-  declareForeign Tracked "Tls.Config.defaultClient" boxBoxDirect
-    .  mkForeign $ \(hostName :: Util.Text.Text, serverId:: Bytes.Bytes) -> do
-       store <- X.getSystemCertificateStore
-       let shared :: TLS.Shared
-           shared = def { TLS.sharedCAStore = store }
-           defaultParams = (defaultParamsClient (Util.Text.unpack hostName) (Bytes.toArray serverId)) { TLS.clientSupported = defaultSupported, TLS.clientShared = shared }
-       pure defaultParams
-
-  declareForeign Tracked "Tls.Config.defaultServer" unitDirect . mkForeign $ \() -> do
-    pure $ (def :: ServerParams) { TLS.serverSupported = defaultSupported }
-
   declareForeign Tracked "Tls.newClient.impl.v3" boxBoxToEFBox . mkForeignTls $
     \(config :: TLS.ClientParams,
       socket :: SYS.Socket) -> TLS.contextNew socket config
@@ -1866,8 +1851,10 @@ declareForeigns = do
     \(config :: TLS.ServerParams,
       socket :: SYS.Socket) -> TLS.contextNew socket config
 
-  declareForeign Tracked "Tls.handshake.impl.v3" boxToEFBox . mkForeignTls $
-    \(tls :: TLS.Context) -> TLS.handshake tls
+  declareForeign Tracked "Tls.handshake.impl.v3" boxToEF0 . mkForeignTls $
+    \(tls :: TLS.Context) -> do
+        i <- contextGetInformation tls
+        traceShow i $ TLS.handshake tls
 
   declareForeign Tracked "Tls.send.impl.v3" boxBoxToEFBox . mkForeignTls $
     \(tls :: TLS.Context,
@@ -1875,11 +1862,14 @@ declareForeigns = do
 
   let wrapFailure t = Failure Ty.tlsFailureRef (Util.Text.pack t) unitValue
       decoded :: Bytes.Bytes -> Either String PEM
-      decoded bytes = fmap head $ pemParseLBS  $ Bytes.toLazyByteString bytes
+      decoded bytes = case pemParseLBS $ Bytes.toLazyByteString bytes of
+        Right (pem : _) -> Right pem
+        Right _ -> Left "no PEM found"
+        Left l -> Left l
       asCert :: PEM -> Either String X.SignedCertificate
       asCert pem = X.decodeSignedCertificate  $ pemContent pem
     in
-      declareForeign Tracked "Tls.decodeCert.impl.v3" boxToEFBox . mkForeign $
+      declareForeign Tracked "Tls.decodeCert.impl.v3" boxToEFBox . mkForeignTls $
         \(bytes :: Bytes.Bytes) -> pure $ mapLeft wrapFailure $ (decoded >=> asCert) bytes
 
   declareForeign Tracked "Tls.encodeCert" boxDirect . mkForeign $
--- a/parser-typechecker/src/Unison/Runtime/Builtin.hs
+++ b/parser-typechecker/src/Unison/Runtime/Builtin.hs
@@ -16,6 +16,8 @@ module Unison.Runtime.Builtin
   , builtinTypeBackref
   , builtinForeigns
   , numberedTermLookup
+  , Sandbox(..)
+  , baseSandboxInfo
   ) where
 
 import Control.Monad.State.Strict (State, modify, execState)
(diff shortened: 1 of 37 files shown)


dolio:
So, one thing I've noticed about this is that you need to do stuff like:

    validateSandboxed [termLink openFile.impl] ...

If you just try `openFile`, it won't work, because that's not the actual prim-op. I can make it so that if you do `validateSandboxed [termLink f]` it expands into all the sensitive prim-ops that `f` uses. That's less annoying, but it might mean that you can accidentally give permissions you didn't think about as carefully. Any thoughts on which trade-off would be preferred?

dolio:
Here's another thought.

I marked reference read/write as untracked, but it occurs to me that you could somehow have a mutable reference that contained a sensitive function. There's no way to see this by checking the value that reads the reference. So it seems that mutable references need to be disallowed by default, and allowing them essentially defeats the guarantees.

pchiusano:
> If you just try openFile, it won't work, because that's not the actual prim-op. I can make it so that if you do validateSandboxed [termLink f] it expands into all the sensitive prim-ops that f uses. That's less annoying, but it might mean that you can accidentally give permissions you didn't think about as carefully. Any thoughts on which trade-off would be preferred?

I'd say don't expand. You have to be explicit.

>  marked reference read/write as untracked, but it occurs to me that you could somehow have a mutable reference that contained a sensitive function. There's no way to see this by checking the value that reads the reference. So it seems that mutable references need to be disallowed by default, and allowing them essentially defeats the guarantees.

Yeah, I'd disable access to mutable references by default, unless they are added to the sandbox.

For the node server, it may be fine to allow them, since there's no way the remote code has any way to reference local mutable references, right?

dolio:
Yeah, somehow the sandboxed code would need to get a mutable reference that someone put a sensitive operation into. Presumably it would need to be something that wasn't sandboxed putting it in there. But, in principle, it's a way you can get a sensitive operation in a way that isn't trackable just by looking at the value.

pchiusano:
> Yeah, somehow the sandboxed code would need to get a mutable reference that someone put a sensitive operation into. Presumably it would need to be something that wasn't sandboxed putting it in there. But, in principle, it's a way you can get a sensitive operation in a way that isn't trackable just by looking at the value.

Yeah, it's a good catch, so good to track it, and people can make their own determination about whether to allow it in the sandbox they set up.

Is this all ready for review now or are you still working on it?

dolio:
I think it's fine to review. `Unison.Runtime.Builtin` is a major thing to look at, since it has all the decisions about which things are sandboxed.

pchiusano:
@dolio Cool. I'd suggest just adding a section to builtins.md rather than introducing a new transcript. The `builtins.md` already has machinery so it will cause a test failure on unexpected results, rather than just yielding a transcript change. You can follow the examples in that file, for instance -

````
## `Any` functions

```unison
> [Any "hi", Any (41 + 1)]

test> Any.test1 = checks [(Any "hi" == Any "hi")]
test> Any.test2 = checks [(not (Any "hi" == Any 42))]
```
````

**Comment by dolio** [comment]
I'm not sure what you're talking about. I didn't introduce a new transcript. I just added to existing stuff.

**Comment by pchiusano** [comment]
Oh, I see, I misread the diff. 

I still like my other point about structuring it so the test actually fails if it gives unexpected results. You can crib from `builtins.md` for inspiration about how to structure `codeops.md` so that the `test` command at the end will run the tests you've added and will fail the transcript if they don't succeed. Alternately, you can just move these tests into `builtins.md` if you prefer.

File: parser-typechecker/src/Unison/Runtime/Builtin.hs [hunk 1]
```diff
@@ -1582,191 +1607,191 @@ declareForeigns = do
...
       in openFile fname mode
 
-  declareForeign "IO.closeFile.impl.v3" boxToEF0 $ mkForeignIOF hClose
-  declareForeign "IO.isFileEOF.impl.v3" boxToEFBool $ mkForeignIOF hIsEOF
-  declareForeign "IO.isFileOpen.impl.v3" boxToEFBool $ mkForeignIOF hIsOpen
-  declareForeign "IO.isSeekable.impl.v3" boxToEFBool $ mkForeignIOF hIsSeekable
+  declareForeign Untracked "IO.closeFile.impl.v3" boxToEF0 $ mkForeignIOF hClose
+  declareForeign Untracked "IO.isFileEOF.impl.v3" boxToEFBool $ mkForeignIOF hIsEOF
+  declareForeign Untracked "IO.isFileOpen.impl.v3" boxToEFBool $ mkForeignIOF hIsOpen
+  declareForeign Untracked "IO.isSeekable.impl.v3" boxToEFBool $ mkForeignIOF hIsSeekable
```
pchiusano: @dolio I'd make all these `Tracked`. Basically any builtin that starts with `IO` should be `Tracked`. The user can always add more things to the sandbox, but they have no way to remove things from the sandbox that are `Untracked`. 

The only thing the user should be able to do with an empty sandbox is pure computation. They could write an infinite loop or consume all the RAM on the machine, but they shouldn't be able to learn anything about the file system, network sockets, etc.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1582,191 +1607,191 @@ declareForeigns = do
...
-  declareForeign "MVar.tryPut.impl.v3" boxBoxToEFBool
+  declareForeign Tracked "MVar.tryPut.impl.v3" boxBoxToEFBool
     . mkForeignIOF $ \(mv :: MVar Closure, x) -> tryPutMVar mv x
 
-  declareForeign "MVar.swap.impl.v3" boxBoxToEFBox
+  declareForeign Untracked "MVar.swap.impl.v3" boxBoxToEFBox
     . mkForeignIOF $ \(mv :: MVar Closure, x) -> swapMVar mv x
 
-  declareForeign "MVar.isEmpty" boxToBool
+  declareForeign Untracked "MVar.isEmpty" boxToBool
```
pchiusano: Anything in `MVar` or `TVar` I'd make `Tracked` as well.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1775,63 +1800,63 @@ declareForeigns = do
...
+  declareForeign Untracked "STM.retry" unitDirect . mkForeign
     $ \() -> unsafeSTMToIO STM.retry :: IO Closure
 
   -- Scope and Ref stuff
-  declareForeign "Scope.ref" boxDirect
+  declareForeign Untracked "Scope.ref" boxDirect
     . mkForeign $ \(c :: Closure) -> newIORef c
 
-  declareForeign "IO.ref" boxDirect
+  declareForeign Untracked "IO.ref" boxDirect
```
pchiusano: I guess it's okay (if somewhat pointless) to have these `Untracked`.

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1844,48 +1869,48 @@ declareForeigns = do
...
-  declareForeign "Code.serialize" boxDirect
+  declareForeign Untracked "Code.serialize" boxDirect
     . mkForeign $ \(sg :: SuperGroup Symbol)
         -> pure . Bytes.fromArray $ serializeGroup sg
-  declareForeign "Code.deserialize" boxToEBoxBox
+  declareForeign Untracked "Code.deserialize" boxToEBoxBox
     . mkForeign $ pure . deserializeGroup @Symbol . Bytes.toArray
-  declareForeign "Code.display" boxBoxDirect . mkForeign
+  declareForeign Untracked "Code.display" boxBoxDirect . mkForeign
     $ \(nm,sg) -> pure $ prettyGroup @Symbol (Util.Text.unpack nm) sg ""
```
pchiusano, line 1901: Does this function expose any details about names for hashes that may exist in the local codebase? Or is it just showing hashes and local variable names? If the latter then I think it's fine to be `Untracked`, otherwise `Tracked`.

pchiusano (changes requested):
Okay looks great, but left a few comments on things I think should be made `Tracked`.

parser-typechecker/src/Unison/Runtime/Builtin.hs
dolio: It appears to just show hashes and local variables right now. I don't know if there are plans to change that in the future, though.

pchiusano: K, then I think this is fine for now.

parser-typechecker/src/Unison/Runtime/Builtin.hs
dolio: I decided to make these tracked anyway, just for uniformity with making pretty much all other mutable stuff tracked.

parser-typechecker/src/Unison/Runtime/Builtin.hs
dolio: Do you care about `Handle/Socket.toText`? That just prints out some kind of information string about the value (not sure what exactly it prints out).

parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1487,11 +1487,11 @@ builtinLookup
   , ("Boolean.or", (Untracked, orb))
   , ("Boolean.and", (Untracked, andb))
 
-  , ("bug", (Untracked, bug "builtin.bug"))
-  , ("todo", (Untracked, bug "builtin.todo"))
-  , ("Debug.watch", (Untracked, watch))
-  , ("Debug.trace", (Untracked, gen'trace))
-  , ("unsafe.coerceAbilities", (Untracked, poly'coerce))
+  , ("bug", (Tracked, bug "builtin.bug"))
+  , ("todo", (Tracked, bug "builtin.todo"))
```
pchiusano: Let's leave `bug` and `todo` as `Untracked`.

Everything else looks good.

pchiusano (approved):
Okay, we did a final pass through to decide on `Untracked`/`Tracked` for a few more.

A rule of thumb we were using is that if the user can already (in principle) implement the function by reading bytes of the Unison process they have in memory, then no need to make it `Tracked`. Also, the user can in principle submit code that hasn't been typechecked, so you aren't gaining anything by sandboxing various functions that cast values to different types.

PR #2842, opened 2022-01-27
vim completion: handle leading {

ceedubs:
Before this change if you had `{Abor<cursor>` and triggered completion,
nothing would be found because the `{` was included in the search
string. This change special-cases a few characters (braces, parentheses)
so that they don't accidentally get included in the search term.

I think that a fix that is 100% compatible with the Unison spec would be
more involved. But this seems to handle the vast majority of uses. Also
I don't think that the Unison parser even is 100% compatible with the
Unison spec 🙂.


ChrisPenner (approved):
I'm no vimscript expert, but I'll trust that this works 😄

PR #2850, opened 2022-02-01
Remember more type annotations in decompiled output

### LINKED ISSUE #2795: Prioritized Doc Rendering Improvements
* Preserving the type signatures for Unison terms in fenced codeblocks. See for example the difference between the following term renderings: 

`````
test = {{
  ```
  t : Text
  t = "hi"

  t
  ```
  @source{t1}

}}

t1 = "hi"
`````

* Pattern match issue in fenced codeblocks - https://github.com/unisonweb/unison/issues/2778

dolio:
This keeps some type annotations around in the remembered code for decompiled output, so that it can show up in docs, for instance. Should fix #2795 I believe.

I didn't include a test case for #2795 because you can't actually use a triple backtick doc in a transcript.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/ANF.hs
+++ b/parser-typechecker/src/Unison/Runtime/ANF.hs
@@ -178,7 +178,7 @@ enclose keep rec (LetRecNamedTop' top vbs bd)
   lvbs = (map.fmap) (rec keep' . abstract keep' . ABT.substs xpnd) vbs
   lbd = rec keep' . ABT.substs xpnd $ bd
 -- will be lifted, so keep this variable
-enclose keep rec (Let1NamedTop' top v b@(LamsNamed' vs bd) e)
+enclose keep rec (Let1NamedTop' top v b@(unAnn -> LamsNamed' vs bd) e)
   = Just . let1' top [(v, lamb)] . rec (Set.insert v keep)
   $ ABT.subst v av e
   where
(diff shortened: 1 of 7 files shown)


rlmark:
@dolio, I noticed this works regular code execution, but when I added the `@typecheck` annotation to the doc, subsequent doc evaluated codeblocks all fail. This snippet should repro it in the absence of a transcript (due to backticks). 

````
docTest = {{

  ```
  myFunction: Nat -> Nat
  myFunction n = n + 1

  myFunction 6
  ```

  @typecheck```

  myFunction2: Nat -> Nat
  myFunction2 n = n + 2

  ```

  ```
  test :  a -> (a -> Boolean) -> {Abort} a
  test a pred =
    if pred a then a else abort

  toOptional! '(test 5 (_ -> false ))
  ```

}}

````

````
.> display docTest

  cannot decompile an application to a local recursive binding
````

The UI makes this a bit easier to see where the error is happening: 
Screen Shot 2022-01-31 at 6 32 49 PM

pchiusano:
@dolio for regression test you can put the test doc in a separate `fix2795/doc.u` file, then do `.> load unison-src/transcripts/fix2795/doc.u` in the transcript.

dolio:
@rlmark It seems like that's just a complaint about the `test` function in the third block. It appears to fail on trunk to me. I'm not sure why it thinks it's recursive, but I don't think it's because of the type annotations.

Edit: also, I can just delete the `@typecheck` block and it still fails. Unless I'm doing something wrong.

rlmark:
@dolio yeah, it appears to fail on trunk as well. I can file a different ticket for the "cannot decompile an application to a local recursive binding" Doc rendering issue. The type signatures are amazing, thank you!

rlmark (approved):
🎉 The issue I ran into appears on trunk as well. Ran this against the doc codebase and this is super helpful!

pchiusano:
Great! Merging this.

PR #2852, opened 2022-02-02
Fix incorrect keyword lexing in links in docs

### LINKED ISSUE #2642: keyword fragments appearing in doc links are treated as their Unison entities
If you're writing a `Doc` link, as in `{{I am a doc which links to a term called {abilityPatterns} in the codebase}}` the UCM will error with 

```
I couldn't find a type for Patterns.
  
      9 | {abilityPatterns}
```

Indicating that it's looking for an ability named Patterns. 

Likewise with terms that are prefixed with `type` as in `{{ I am a doc which links to a term called [Type signatures](typeSignatures)}}`

The UCM will fail to typecheck with a message about how a type named Signatures could not be found in the codebase. 

As long as your keyword-fragment is not at the top level of the codebase, you can work around this by qualifying the term, as in `languageReference.abilityPatterns`, but this may not be ideal in the long run.

kanwren:
## Overview

Resolves #2642

A doc containing a term link in like this:
```
{{A doc linking to {abilityFoo}}}
```
would previously eagerly lex the `ability` or `type` at the start of an identifier as a keyword, tricking the parser into thinking we're in a type link instead. The above would complain about not finding the ability `Foo`, rather than about not finding the term `abilityFoo`.

This fixes the lexer to handle such cases properly.

## Implementation notes

This was just a tiny lexer fix in exactly the same vein as #2727/#2847; the lexer was looking for a literal keyword string, and didn't check to see if it was followed by something that could delimit a keyword.

I've looked over all of the other places where the lexer reads a literal string (using `lit` and its variants), and they all now look correct; I'm hoping this is the last such bug.

## Test coverage

Adds a regression test transcript to test that `type` and `ability` are lexed properly in type and term links in docs. The current lexer tests don't include any tests on docs, so I didn't add any, but I could if it would be preferred.

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -397,7 +397,7 @@ lexemes' eof = P.optional space >> do
           pure s
 
     typeLink = wrap "syntax.docEmbedTypeLink" $ do
-      _ <- typeOrAbilityAlt lit <* CP.space
+      _ <- typeOrAbilityAlt wordyKw <* CP.space
       tok (symbolyId <|> wordyId) <* CP.space
 
     termLink = wrap "syntax.docEmbedTermLink" $
--- a/unison-src/transcripts/doc-type-link-keywords.md
+++ b/unison-src/transcripts/doc-type-link-keywords.md
@@ -1,0 +1,41 @@
+Regression test to ensure that `type` and `ability` in embedded doc links are
+lexed properly when they occur at the start of identifiers.
+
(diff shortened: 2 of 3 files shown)


aryairani:
@nprindle Sorry this slipped through the cracks — @rlmark is gonna review this today 😁 
We are very eager to get it merged.

kanwren:
No problem! I'm working on a few other fixes in this area as well at the moment, so there's no rush :)

rlmark (approved):
Awesome! @aryairani and I tested this against a local docs codebase and it looks great! Thank you!!!

PR #2853, opened 2022-02-03
Make term printer output more backticks if needed.

### LINKED ISSUE #2810: Roundtripping fenced codeblocks with more than 3 backticks fails to parse
This doesn't happen frequently, but I noticed this when updating `Docs` about `Docs`. If you have a `Doc` element that uses the triple backtick syntax and 3+n backticks to escape the codeblock, for example: 

``````
testDoc : Doc
testDoc = {{

 This is an executable code snippet 

  ````` raw
  ```
  1 + 1
  ```
  ````` }}
``````

the UCM will normalize your 3+n backticks to be 3 backticks, breaking your document. 

````
testDoc : Doc
testDoc =
      {{ This is an executable code snippet
      
      ``` raw
      ``` unison
      1 + 1
      ```
      ``` }}
````

For hand-wavey reasons involving the transcript parser's  treatment of 3+n backticks, I can't submit this with a transcript attached, but adding and editing the code snippet above should surface the issue.

runarorama:
Fixes #2810.

Solved the same way as we do with bold, italics, underscore, etc. We look at the rendered text inside the fenced block and find the longest fence it contains. Then we output one more backtick than that.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -1408,16 +1408,16 @@ prettyDoc2 ppe ac tm = case tm of
           S.DocDelimiter
           "}}"
     bail tm = brace (pretty0 ppe ac tm)
-    -- Finds the longest run of a character and return a run one longer than that
-    oneMore c inner = replicate num c
-     where
-      num =
-        case
-            filter (\s -> take 2 s == "__")
-              $ group (PP.toPlainUnbroken $ PP.syntaxToColor inner)
-          of
-            [] -> 2
-            x  -> 1 + (maximum $ map length x)
+    -- Finds the longest run of a character and return one bigger than that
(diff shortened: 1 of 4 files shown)


pchiusano:
LGTM, can you add a regression test though? I think you can just add it to the existing test harness you've got for all the other pretty-printer fixes.

runarorama:
I tried to add a regression test to the transcript, but the transcript parser got tripped up by the nested backticks.

pchiusano:
@runarorama ah. In that case you can just use the `load` command to load an external `.u` file, if that `.u` file can't be directly embedded in the transcript inside some backticks.

pchiusano (approved):

pchiusano:
GitHub doesn't tell me when there's a new commit on a PR I've commented on, so just saw that you added the regression test. Everything LGTM.

PR #2854, opened 2022-02-03
docs.as-html: use block elements for Style

hojberg:
## Overview

The `Style` variant is meant to wrap things to give them an appropriate
css class, but needs to be a block element so that it can contain other
block elements.

This fixes an issue with the new docs site where the css class weren't being applied.


pchiusano (approved):

pchiusano:
Is this still relevant or was it folded into the other PR?

hojberg:
still relevant - let me rebase

PR #2857, opened 2022-02-03
Fix: '"" should be parsed as delayed text

hagl:
The code `'""` currently causes a parse error:

```
unexpected '"'
expecting end of input
```

## Overview

I came across this problem in code like 
```
display = Optional.fold '("") Nat.toText
```
which is parsed correctly because the delayed text is in parentheses. But when editing this method the pretty-printer removes the parentheses and causes a compile error.

See this discussion on Slack
https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1643913764607739

## Implementation notes

The lexer will will now treat double-quotes as separator for delay and force tokens. (So `!""` is also be accepted by the lexer, but is then rejected by the type checker)

## Test coverage

I've added a test case

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -65,3 +65,4 @@ The format for this list: name, GitHub handle, and then optional blurb about wha
 * Solomon Bothwell (@solomon-b)
 * Sameer Kolhar (@kolharsam)
 * Nicole Prindle (@nprindle)
+* Harald Gliebe (@hagl)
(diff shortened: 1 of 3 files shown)


aryairani:
Hi @hagl, thanks for this! 
Would you add yourself to `CONTRIBUTORS.markdown` as well, assuming you agree to the terms there?

pchiusano:
Great!

pchiusano (approved):

PR #2878, opened 2022-02-09
docs.as-html: move tooltips to hidden spot in dom

hojberg:
## Problem
Rich tooltips use block elements, which, in the html spec, and in how browsers interpret it, are not allowed inside <p> elements. This results in the browser closing the <p> as soon as it encounters a block element, effectively moving that block element outside of the narrative flow.

## Solution
To fix this, when rendering docs as html, collect the contents of tooltips in a separate collection (using `State` and `Writer`) from the rest of the document, and insert them at the end of the DOM in a hidden div. Each tooltip trigger has a `data-tooltip-content-id` attribute that can be used to render the tooltip in-place when hovering with additional JavaScript.

## Notes
Paired with @pchiusano

### DIFF
--- a/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
+++ b/parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
@@ -1,13 +1,20 @@
-{- ORMOLU_DISABLE -} -- Remove this when the file is ready to be auto-formatted
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 -- | Render Unison.Server.Doc and embedded source to Html
 module Unison.Server.Doc.AsHtml where
 
+import Control.Monad.State.Class (MonadState)
+import qualified Control.Monad.State.Class as State
+import Control.Monad.Trans.State (evalStateT)
+import Control.Monad.Writer.Class (MonadWriter)
+import qualified Control.Monad.Writer.Class as Writer
+import Control.Monad.Writer.Lazy (runWriterT)
 import Data.Foldable
 import Data.Map (Map)
 import qualified Data.Map as Map
 import Data.Maybe
+import Data.Sequence (Seq)
 import Data.Text (Text)
 import qualified Data.Text as Text
 import Lucid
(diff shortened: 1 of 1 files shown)


parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -1,13 +1,20 @@
-{- ORMOLU_DISABLE -} -- Remove this when the file is ready to be auto-formatted
```
hojberg, line 1: There's a little bit of formatting noise due to this change.

parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -179,214 +188,263 @@ toText sep doc =
...
 toHtml docNamesByRef document =
-  let toHtml_ sectionLevel doc =
-        let -- Make it simple to retain the sectionLevel when recurring.
-            -- the Section variant increments it locally
-            currentSectionLevelToHtml =
-              toHtml_ sectionLevel
+  article_ [class_ "unison-doc"] $ do
+    content
+    div_ [class_ "tooltips", style_ "display: none;"] $ sequence_ tooltips
+  where
```
hojberg, line 194: Changed this from a `let..in` to.a `where`.

parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -179,214 +188,263 @@ toText sep doc =
...
 toHtml :: Map Referent Name -> Doc -> Html ()
 toHtml docNamesByRef document =
-  let toHtml_ sectionLevel doc =
-        let -- Make it simple to retain the sectionLevel when recurring.
-            -- the Section variant increments it locally
-            currentSectionLevelToHtml =
-              toHtml_ sectionLevel
+  article_ [class_ "unison-doc"] $ do
+    content
+    div_ [class_ "tooltips", style_ "display: none;"] $ sequence_ tooltips
```
hojberg, line 193: This is where the tooltips are added at the end of the DOM in a hidden div.

parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -179,214 +188,263 @@ toText sep doc =
...
-                      ico
-                      div_ [class_ "callout-content"] $ currentSectionLevelToHtml content
-              Table rows ->
-                let cellToHtml =
-                      td_ [] . currentSectionLevelToHtml
+              span_
+                [ class_ "tooltip-trigger",
+                  data_ "tooltip-content-id" tooltipId
+                ]
+                <$> currentSectionLevelToHtml triggerContent
```
hojberg, line 263: Each tooltip is given an id (a number incremented through `State`), which is added as an html id to the content and as a data attribute to tooltip triggers to tie the pair together.

parser-typechecker/src/Unison/Server/Doc/AsHtml.hs
```diff
@@ -179,214 +188,263 @@ toText sep doc =
...
+            Word word ->
```
hojberg, line 264: Changes below here are to satisfy the return type with `State` and `Writer`.

pchiusano (approved):
This came out pretty nice!

I wondered later if you could directly work with `HtmlT (WriterT . StateT)`; it might come out even nicer. But I'd leave it alone.

Issue #2712, opened 2021-12-03
round trip bug + type inference bug


stew:
Below is a transcript which show's a round trip problem but I kinda think that the reason that this is a round trip bug is that it's actually exposing an problem with ability inference. When the pretty printer prints `susan: Nat ->{} Nat -> ()` it prints it without the null ability marker: `susan: Nat -> Nat -> ()` which causes the error:
```
  The expression in red
  
                needs the abilities: {𝕖6}
    but was assumed to only require: {𝕖26, 𝕖5}
  
  This is likely a result of using an un-annotated function as an argument with concrete abilities. Try adding an annotation to the function definition whose body is red.
  
      7 |   mapWithKey susan pam
```

==TRANSCRIPT==

    ```unison
    unique type Map k v = Tip | Bin Nat k v (Map k v) (Map k v)
    
    mapWithKey : (k ->{e} a ->{e} b) -> Map k a ->{e} Map k b
    mapWithKey f m = Tip
    ```
    
    ```ucm
    .> add
    ```
    
    ```unison
    
    naiomi = 
      susan: Nat ->{} Nat -> ()
      susan a b = ()
      
      pam: Map Nat Nat
      pam = Tip
      
      mapWithKey susan pam
      
    ```
    
    ```ucm
    .> add
    .> edit naiomi
    .> undo
    ```
    
    ```ucm
    .> load scratch.u
    .> add
    ```

atacratic:
Related: #738 #717 #703 - and see also https://gist.github.com/atacratic/7a91901d5535391910a2d34a2636a93c#ability-inference-and-generalization.  

If the limitation in inference shown in this issue is just how it has to be, then I think that's a good case for remembering and faithfully pretty-printing the null ability list.

pchiusano:
Just to clarify, when `naiomi` is pretty-printed, the `{}` on the annotation of `susan` is elided by the pretty-printer. This causes round trip bugs.

pchiusano:
Since there's a decent workaround (just rename the namespace to not use `_`) this is lower priority.

aryairani:
> Since there's a decent workaround (just rename the namespace to not use `_`) this is lower priority.

@pchiusano Was this meant for this ticket? or for #2822

aryairani:
Hopefully will fix these definitions from [website](https://github.com/unisonweb/website:#8agaqj52kh):
* [ ] fundamentals.controlFlow.exceptionHandling.either

pchiusano:
[...]

Issue #2423, opened 2021-09-21
Spurious ability check failure


pchiusano:
I wouldn’t stop what you’re doing to work on this, but would love to fix at some point.
```
structural ability Split where
  skip! : x
  both : a -> a -> a

Split.append : '{Split, g} a -> '{Split, g} a -> '{Split, g} a 
Split.append s1 s2 _ = force (both s1 s2)

force a = !a

Split.zipSame : '{Split, g} a -> '{Split, g} b -> '{Split, g} (a, b)
Split.zipSame sa sb _ = 
  go : '{Split,g2} y -> Request {Split} x ->{Split,g2} (x,y)
  go sb = cases
    { a } -> (a, !sb)
    { skip! -> _ } -> skip!
    { both la ra -> k } -> 
      handle !sb with cases
        { _ } -> skip!
        { skip! -> k } -> skip!
        { both lb rb -> k2 } -> 
          force (
            (zipSame '(k la) '(k2 lb) `Split.append` 
             zipSame '(k ra) '(k2 rb)))

  handle !sa with go sb
```

Yields:

```  
The expression in red
  
                needs the abilities: {g186}
    but was assumed to only require: {Split, 𝕖181, g182}
  
  This is likely a result of using an un-annotated function as an argument with concrete abilities. Try adding an annotation to the function definition whose body is red.
  
     23 |              zipSame '(k ra) '(k2 rb)))
  
```

In this specific example, I can workaround by inlining the definition of `append`, like this typechecks - 

```
          force (
            both (zipSame '(k la) '(k2 lb))
                 (zipSame '(k ra) '(k2 rb)))
```

pchiusano:
Fixed by #2895

PR #2895, opened 2022-02-11
'Linearize' quantified ability variables for better inference

dolio:
This inserts some processing when ungeneralizing types that should lead to better inference with the newer ability algorithm. It rewrites types like:

    (a ->{e} b ->{e} c) ->{e} d

to:

    (a ->{e1} b ->{e2} c) ->{e1,e2} d

which behave a bit better. These types are inter-convertible (set all variables to `e` for one direction, and instantiate `e` to the whole row and use subtyping for the other), but the latter avoids some issues when inferring types involving higher-order functions.


pchiusano (approved):
Sweet. Can you come up with a regression test?

dolio:
Added test cases. This actually fixes #2423.

I was thinking about factoring out some of the variance-tracking traversal stuff, but then I realized that I might not actually be able to use it usefully in this scenario. So I guess I'll just submit this for now.

dolio:
Also fixes #2712.

pchiusano:
Amazing how much this little change improves and fixes! Merging.

PR #2898, opened 2022-02-13
Don't use `!` syntax for ctor application (`Any ()` and `Some ()`, not `!Any` or `!Some`)

### LINKED ISSUE #2402: pretty printer: Unit applied to type constructor is printed with !
`Some 3` is printed as `Some 3`, but `Some ()` is printed as `!Some`.

I'm guessing that there's some logic to pretty-print unit being applied to a thunk (a _function_) that is also being applied to type constructors.

[Here](https://gist.github.com/ceedubs/3bd3a09edc6b3a2bf0819f6d07398d6c) is a transcript with a couple of failing examples.

pchiusano:
Fixes #2402 

This fixes some peculiar use of `!` syntax in pretty-printing application of data constructors, in favor of just passing `()` directly. Like you won't see `!Some` or `!Left`, instead you'll now see `Some ()` and `Left ()`. Have a look at the diff in the transcripts to see more examples.

As I fixed this, I noticed indentation is wrong for multiline lists containing multiline expressions, and fixed that as well. Sorry didn't bother with a separate PR. Specifically, if we have:

```haskell
x = [ foo -- this doesn't parse, yet this is how it is current printed 
      x -- this line should be indented by 2 more
      y -- same here
    , blah ]
```

And the `foo x y` can't fit on one line, it's subsequent lines past the first need to be indented by 2 more than they are, because the list elements start at relative column 2 (after `[ `).

Added a regression test for this to the round trip harness.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -327,12 +327,17 @@ pretty0
 
     t -> l "error: " <> l (show t)
  where
+  goNormal prec tm = pretty0 n (ac prec Normal im doc) tm     
   specialCases term _go | Just p <- prettyDoc2 n a term = p
   specialCases term go = case (term, binaryOpsPred) of
     (DD.Doc, _) | doc == MaybeDoc ->
       if isDocLiteral term
       then prettyDoc n im term
       else pretty0 n (a {docContext = NoDoc}) term
+    (App' f@(Builtin' "Any.Any") arg, _) ->   
+      paren (p >= 10) $ goNormal 9 f `PP.hang` goNormal 10 arg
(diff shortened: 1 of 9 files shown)


ceedubs:
Sounds like this would fix #2402

aryairani:
Ooh, good catch, thanks for the fix.

I kinda think we should have a dedicated type for the arg instead of `()`; a unique unit. Would that be any harder to support? Or a bad idea? Otherwise there will always be this ambiguity on how to render polymorphic function calls, right?

pchiusano:
@aryairani yeah, that is definitely the choice I'd make today, but changing that now has bigger impact and is out of scope for this PR. :)

rlmark (approved):

PR #2908, opened 2022-02-16
Fix an issue computing sandbox info

dolio:
The sandboxed dependencies for combinators were being computed just from
the previously known combinators. So, if multiple new, interrelated
combinators were introduced simultaneously, their full transitive
dependencies wouldn't take into account anything in the new set.

To fix this, just iterate the dependency inference until a fixed point
is reached.


pchiusano (approved):
Cool.

PR #2914, opened 2022-02-17
patch Shellmet (instead of replacing it)

aryairani:
I remembered that we want to give back to open source, so this PR reverts #2900 (which had copied Shellmet.hs into Git.hs, with plans to stop depending on it altogether), and replaces it with a fork at https://github.com/unisonweb/shellmet/tree/topic/avoid-callCommand and an upstream PR at https://github.com/kowainik/shellmet/pull/27.

This approach also better handles the licensing requirements of Shellmet.


ChrisPenner (approved):
Good call 👍🏼

PR #2915, opened 2022-02-17
delete broken runtime benchmark

aryairani:
per discussion with @dolio


pchiusano (approved):

aryairani:
Shoot, I forgot to generate and check in the new `.cabal` file.

PR #2916, opened 2022-02-17
add integration-tests to hie.yaml

aryairani:
This has been missing for a while. I'm not sure how to make a test for it?

### DIFF
--- a/hie.yaml
+++ b/hie.yaml
@@ -51,6 +51,9 @@ cradle:
     - path: "unison-cli/src"
       component: "unison-cli:lib"
 
+    - path: "unison-cli/integration-tests"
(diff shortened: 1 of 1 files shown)


mitchellwrosen (approved):

mitchellwrosen:
LGTM; pretty sure hls is smart enough to get on without this file. So I'd also approve of deleting it :)

ChrisPenner (approved):

aryairani:
> LGTM; pretty sure hls is smart enough to get on without this file. So I'd also approve of deleting it :)

I'd like that to be true, but it keeps not being true.  I'm not really sure how to tell. We could try it again?

PR #2917, opened 2022-02-17
switch to Windows code page 65001 in exe:unison and exe:tests

aryairani:
I think it's only needed in these two spots, as the other executables just shell out to `stack exec unison`.
    
I'm not actually clear on how this interacts with (or could be replaced by `hSetEncoding`, which we do have sprinkled around already, but not sufficiently to let Windows builds work. If there's a cleaner solution, I'd like to use it. _i.e._ Am I papering over a more correct multi-platform fix by using the Windows-only `System.IO.Codepage`?

See https://hackage.haskell.org/package/code-page-0.2.1/docs/src/System.IO.CodePage.html#withCodePage

### DIFF
--- a/unison-cli/package.yaml
+++ b/unison-cli/package.yaml
@@ -51,6 +51,7 @@ library:
 tests:
   tests:
     dependencies:
+      - code-page
       - easytest
       - here
       - shellmet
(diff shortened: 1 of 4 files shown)


pchiusano:
Should this be behind a cpp flag, only taking effect if the OS is windows? I was under the impression that codepages for console output was just a janky windows only thing and consoles on other platforms can just print arbitrary unicode. Do you know what effect it has when it's set on other platforms?

aryairani:
@pchiusano The CPP flag is inside the library calls.

aryairani:
> [System.IO.CodePage](https://hackage.haskell.org/package/code-page-0.2.1/docs/System-IO-CodePage.html): a cross-platform module that exports functions which adjust code pages on Windows, and do nothing on other operating systems.

pchiusano (approved):

PR #2920, opened 2022-02-17
Traverse type references in ANF functions

dolio:
Previously the references in FCon and FReq didn't matter when loading into the cache, but due to tag packing, they now do. Strictly speaking, they probably still don't matter for FReq, but might as well be safe.

Not sure how to actually manufacture a test case for this, but it's been confirmed to fix a problem.


ceedubs (approved):
Thanks @dolio!

PR #2929, opened 2022-02-18
Fix universal equality

runarorama:
This duplicates some code from another place where `EQLU` is implemented. I'm guessing it was an oversight.


dolio (approved):
Strange, not sure what sequence of events led to that mistake.

PR #2932, opened 2022-02-18
vim-completion: don't include bang

ceedubs:
Similar to #2842.

When triggering auto-completion on something like `!currentTime`, we
don't want to include the `!` in the term that we search for.


aryairani (approved):
rubber stamp

PR #2946, opened 2022-02-23
Doc Embed: Add Video and FrontMatter

hojberg:
## Overview
Expand the Doc Embed variant to support embedding videos and
front-matter. The latter will be rendered at the top of the output of the
html file rendered via the `docs.to-html` command.

This is needed as part of the effort to move the blog to Unison Doc.

Paired with @pchiusano 

## Implementation notes
Add a few new types that `Doc2.Embed` wraps and add patterns for them and add support for them
when rendering docs to html.

## Caveat

Haven't tested this yet. Not actually sure the doc syntax to embed. @pchiusano any pointers?

### DIFF
--- a/parser-typechecker/package.yaml
+++ b/parser-typechecker/package.yaml
@@ -100,6 +100,7 @@ library:
     - mwc-random
     - NanoID
     - lucid
+    - yaml
     - semialign
     - servant
     - servant-docs
--- a/parser-typechecker/src/Unison/Runtime/IOSource.hs
+++ b/parser-typechecker/src/Unison/Runtime/IOSource.hs
@@ -145,6 +145,18 @@ doc2UntitledSectionId = constructorNamed doc2Ref "Doc2.UntitledSection"
 doc2ColumnId = constructorNamed doc2Ref "Doc2.Column"
 doc2GroupId = constructorNamed doc2Ref "Doc2.Group"
 
+doc2MediaSourceRef :: R.Reference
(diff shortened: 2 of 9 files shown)


parser-typechecker/package.yaml
```diff
@@ -100,6 +100,7 @@ library:
     - mwc-random
     - NanoID
     - lucid
+    - yaml
```
hojberg, line 103: Used for rendering front matter to the top of the file when converting docs to html files.

parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -358,6 +373,54 @@ metadata.isPropagated = IsPropagated.IsPropagated
...
+
+-- Useful for embedded data into a Doc, like title, date, tags etc:
+--   {{ 
+--     embed Any (
+--       FrontMatter [ ("title", "The Hobbit")
+--                   , ("subTitle", "There and Back Again")
+--                   , ("tags", "book")
+--                   ]
+--       )
+--   }}
```
hojberg: These examples are **not** accurate.

pchiusano:
@hojberg here's the syntax -

```
x = {{
  Some doc with {{ Special (Embed (Any (Video blah blah)))) }}
}}
```

pchiusano (approved):
You might need a transcript refresh for things that use `mergeio`

PR #2959, opened 2022-03-02
Enable completion for compile.output

ceedubs:
This allows auto-completion for the first argument to `compile.output` (the main method).

Full disclosure: I don't know what I'm doing with `InputPattern` and copy/pasted from other examples. But it seems to work.

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -1798,7 +1798,7 @@ makeStandalone =
   InputPattern
     "compile.output"
     []
-    []
+    [(Required, exactDefinitionTermQueryArg), (Required, noCompletions)]
     ( P.wrapColumn2
         [ ( "`compile.output main file`",
             "Outputs a stand alone file that can be directly loaded and"


aryairani (commented):
LGTM

aryairani (approved):

aryairani:
I forget whether it's hard to autocomplete with a specific type only (e.g. `'{IO} ()`) but maybe that would be good?

ceedubs:
@aryairani that could be nice. I don't see any existing samples that are doing that.

If we were to implement that at some point, we might want to make sure that `'{IO} ()` and `'{Exception, IO} ()` both work.

pchiusano:
We should rename `compile.output` to just `compile` as well (or make `compile` the primary name and `compile.output` an alias).

To do the filtering you could just take the list of terms you get back, pull out all their types, and do `Typechecker.isSubtype ty mainType` where `mainType` is `MainTerm.builtinMain`. That will also correctly handle the subtyping if you don't use all `Exception` and/or `IO`, though perhaps that isn't what you want since it will catch pure functions that return `()`. Or maybe that's fine since pure functions don't generally return `()`.

PR #2960, opened 2022-03-02
Use softhang if a delayed block is the last argument to a function

pchiusano:
This has been bugging me for a while. 

When a delayed computation block is passed to a function as the last argument in a context where the ambient precedence is low enough, we can elide parentheses around it and use a "soft hang" to put the `'let` on the same line as the function call. This looks nice and is typically how you'd write the code in your scratch file. **So now the pretty-printer will produce code like:**

```Haskell
forkAt usEast 'let
  x = thing1
  y = thing2
  ...
```

instead of the ugly but effective current rendering. 👎 (😢 🎺 sad trombone)

```Haskell
forkAt 
  usEast 
  ('let
    x = thing1
    y = thing2
    ...)
```

I added this as a round trip test along with a few corner cases. It looks super!

Note: @runarorama you added softhangs when a delayed block is the RHS of a binding (like `x = 'let`) and this is applying the same sort of strategy to a different spot in the pretty-printer.

This will probably need a refresh after #2958 if that gets merged first.

### DIFF
--- a/parser-typechecker/src/Unison/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/TermPrinter.hs
@@ -7,7 +7,7 @@ module Unison.TermPrinter where
 
 import Unison.Prelude
 
-import Control.Lens ((^.))
+import Control.Lens ((^.), unsnoc)
 import Control.Monad.State (evalState)
 import qualified Control.Monad.State as State
 import Data.List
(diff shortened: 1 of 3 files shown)


runarorama (approved):

pchiusano:
FYI, CI passed, but mergify doesn't wait for the newly added windows build to finish. Tracking in https://github.com/unisonweb/unison/issues/2961

PR #2965, opened 2022-03-03
Rename compile.output to compile

ceedubs:
As [suggested](https://github.com/unisonweb/unison/pull/2959#issuecomment-1057059547) by @pchiusano.

There was another suggestion to filter to only terms that match subtypes
of `{Exception, IO} ()`. So far I haven't done that, because I'm not
very familiar with these types and am having trouble figuring out how to
get type information out of relations and referents.

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -1797,7 +1797,7 @@ makeStandalone :: InputPattern
 makeStandalone =
   InputPattern
     "compile"
-    []
+    ["compile.output"]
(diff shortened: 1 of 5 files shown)


unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -1796,11 +1796,11 @@ ioTest =
 makeStandalone :: InputPattern
 makeStandalone =
   InputPattern
-    "compile.output"
+    "compile"
     []
```
pchiusano:
```suggestion
    "compile"
    ["compile.output"]
```

pchiusano (approved):
Cool. I suggest leaving `compile.output` as an alias - added a suggestion for that. Next release we can remove the alias.

PR #2968, opened 2022-03-03
Remove the access control header to limit CORS

hojberg:
# Overview

Remove the star header of Access-Control-Allow-Origin; disallowing cross domain requests.


pchiusano (approved):

ceedubs (approved):
Seems like the right thing to me. But wanted to check with @runarorama since I think that you originally added this.

Issue #2794, opened 2022-01-10
Allow updates to individual elements of a cycle


pchiusano:
This has come up a bunch of times for @rlmark who is working on docs (which are often cyclic with backreferences and so forth) and also @ceedubs has hit it too when working on Unison Cloud.

Right now, when you do an `edit foo`, if `foo` is part of a cycle with `bar` and `baz`, it will dump both `foo` and `bar` into your text buffer, as of https://github.com/unisonweb/unison/pull/2563 by @mitchellwrosen. However, if you then delete or put a fold in the middle of a cycle, and the `update`, you get weird results.

For example, suppose we have `foo` and `bar` in a cycle. We then `edit foo`, get `foo` and `bar` in our buffer (in that order), then put a fold after `foo` and then `update`. The result will be that:

* `foo` will contain a reference to the old `bar`, which is part of the old `foo`/`bar` cycle.
* So we've create a new `foo` that reference `bar` from the old cycle.

```
foo -> (old_foo <-> old_bar)
```

Question is: can we just have it always do the right thing? Idea would be that when you `update` a cycle, it acts as if you had all the cycle elements in your buffer, even if you deleted some or moved them below the fold.

@mitchellwrosen and @aryairani this might be high value / easy to fix.

mitchellwrosen:
We were just talking about this on Fri. It's nice to have an explicit ticket for the work now too :)

Some more discussion about it here
https://github.com/unisonweb/unison/issues/2583

mitchellwrosen:
> Question is: can we just have it always do the right thing? Idea would be that when you update a cycle, it acts as if you had all the cycle elements in your buffer, even if you deleted some or moved them below the fold.

I just want to point out if "do the right thing" might not always be as if you had all the cycle elements in your buffer.

Consider:

```
foo _ = !bar + 1
bar _ = !foo + 1
```

Add that, then put this back in your buffer and save:

```
foo = "hey"
```

I think in this case, we would want to update `foo`, and show `bar` in the output of `todo`, like a normal dependent of a term whose type had changed, so propagation couldn't proceed.

If instead we acted as if the whole cycle was in your buffer, we'd have

```
foo = "hey"
bar _ = !foo + 1
```

which doesn't type check.

aryairani:
@mitchellwrosen Should this ticket be closed as Done?

mitchellwrosen:
@aryairani I don't think so, no - probably all the bullets in #3445 should be addressed.

[...]

PR #2974, opened 2022-03-08
add transcript demonstrating an update bug

mitchellwrosen:
## Overview

This PR adds a transcript that demonstrates when we update a term from a test to a non-test, we retain its `isTest` metadata.

This will be fixed by the work in #2649 / #2794


ChrisPenner (approved):
Nice, would you mind indicating in the transcript somewhere that although this is the current behaviour, that it's undesirable and should be fixed? Otherwise it can be tough to tell why a transcript like this exists in the future, Thanks 😄

aryairani (approved):

PR #3003, opened 2022-03-27
Switch parsing and `find` to by default use only names in the current namespace

### LINKED ISSUE #2988: Self-contained namespaces (replaces archived namespaces idea)
So I want to close #1340 as being a bad approach (in addition to adding a bunch of ugly underscores to your namespace tree) and propose something better instead.

**Problem:** My namespace tree is a mess of multiple versions of libraries and definitions, even though for a given project / context, I don't really want everything that appears anywhere in my tree to be visible. In addition to just making search results a lot noisier, suffix-based resolution will find me things in places I don't really intend, and I'll get suffix-based resolution ambiguities just because somewhere in my namespace tree I have an old version of a library. This is no fun.

It's a FEATURE to be able to manage multiple projects in the same namespace tree and switch between them instantly. Let's keep that. But experience has shown us that we DO still want the ability to restrict what names are visible. So here is a dead simple proposal for how to do that, which has some nice properties:

__Nothing outside the current namespace is visible__

So you cannot `view` or `edit` or `find` anything outside your current namespace (we might introduce `find.at`, `view.at` etc to do the equivalent of `cd`, `find`, `back`). Each namespace is a completely isolated universe of code and you can never accidentally reference anything outside of it.

The only commands that can talk about things outside the current namespace are `fork`, `merge`, and `cd`.

Let's see how this plays out.

Rather than having various versions of libraries strewn about your namespace, each "project" (I'm using this term informally, but will connect this to the formal projects proposal at the end) will have a `lib` namespace for its external dependencies:

```
distributed/
  trunk/
    lib/
      base/
    Readme : Doc
    ability Remote
    type Seq
    ...

  releases/
    a2/
      lib/
        base/
      Readme : Doc
      ability Remote
      type Seq
```

Boom. Done. The `distributed` namespace is totally self-contained, as are the various subnamespaces that correspond to branches (like `trunk` and releases like `a1` and `a2`). If I want to add a new dependency to `distributed`, for now I can just:

```
.> cd distributed.trunk.lib
.distributed.trunk.lib> pull <somelibrary> json
```

Bammo. If I really want to install multiple versions of a library, I can always do `pull <somelibrary> json.v6` in my `lib` namespace. (Eventually some project-specific commands for installing and upgrading libraries would be great rather than this manual `cd` and `pull` business. Suggestion `install share.stew.json` will just do the correct `pull` command from share and stick it in `lib` under `stew.json`)

That's really it.

__Benefits:__

* The various forks and feature branches of my library no longer interfere with each other. I can't like, accidentally reference something from another fork or get bad suffix-based name resolution just because my tree has multiple projects and forks.
* Optimization: UCM can avoid even loading things into memory that are not under the current path. Hello instant startup times!
* I can easily create "template" projects with whatever curated collection of dependencies, and fork copies of those projects anytime I wanna start something new. I'd suggest that codebase initialization create a `scratch` project with a `lib` directory that includes just `base`.
* The feature naturally coaxes people to produce self-contained namespaces, since there's no way to reference definitions outside the current namespace!

__Drawbacks:__

* You can't really just `cd` into some random subnamespace of your project and mess around there. This feature was always a footgun anyway--for instance you pretty much never want to produce a patch in a random subnamespace of your project. Resurrecting the `namespace` blocks (implemented better though) will be a good substitute for the more legit concern of just not wanting to manually prefix all the symbols in your scratch file.
* You can no longer reference definitions willy nilly without understanding what libraries they are coming from. On the one hand I kind of liked this, but it doesn't scale and I don't think we're doing people any favors by making this the easy path. With this proposal, you now have to think about your dependencies more carefully and there's a bit more ceremony to bringing in a new dependency. As I mentioned, a `find.at` or `find.global` command would be nice for discovering libraries, though `cd .`, `find`, and then `back` works too.
* Not exactly a drawback, but `update` constructor mapping propagation relies on there being names that exist for the old and new versions of the constructors. If we do the thing proposed in https://github.com/unisonweb/unison/issues/2248 (where for each update, we create an `old` subnamespace containing old versions of things, prior to the change, and then `todo` is just reporting on things that depend on something in `old`), then this just works out and constructor mapping should work fine.

## Some other ideas I thought about

One idea I had was to have a `lib` namespace and a `src` namespace inside every project:

```
distributed/
  trunk/
    lib/
    src/
      Readme : Doc
      ability Remote
      ...
```

This allows people to have projects with a `lib` namespace that doesn't correspond to external dependencies. Which doesn't seem all that useful. It also is awkward to deal with the `src` namespace that prefixes literally all symbols in the project... even if you have `namespace` blocks, it is weak having to put `namespace src where` as boilerplate at the top of every file.

Another idea I considered was having `src` and `lib`, and then have `edit`, `add`, and `update` magically always target the `src` namespace without you having to manually prefix all the symbols in your scratch file. I didn't like this idea. It feels too magical and what if I really do want to just edit a thing at the root which isn't nested under `src`.

## Implementation notes

On startup, it probably pays to remember the last path you were in (perhaps reading it from a text file we store next to the reflog), and just load that namespace into RAM, rather than loading the entire namespace into RAM. So the loop state for `HandleInput` is a path and a `Branch`, but the branch corresponds to the `Branch` at that path (rather than what we do now where we store the root `Branch` and then a path). The codebase should be able to efficiently save a subnamespace of the root.

## How does this interact with the projects proposal?

Easy peasy. You'd just have a project definition at the root of your project, at the same level as `trunk` and `releases` and other feature branches:

```
distributed/
  project : Project
  trunk/
  releases/
```

Now, rather than manually mucking with the `lib` subnamespace of each fork, you use some more specific project-oriented commands for this purpose. But the idea is the same.

pchiusano:
This fixes #2988 

It does so by tweaking a couple places in `HandleInput` and `Backend` to use only names under the current namespace rather than searching globally. I added `find.global` as an extra option, and if `find` returns no results, there's now a hint to suggest using `find.global`.

image

## Interesting decisions

This doesn't try to do anything fancy like only load the current namespace into RAM. That would be cool, but could be done later and isn't a behavior change, just a memory optimization.

I also left the "pretty-printing names" alone - the pretty-printer is still given all names globally, and will pick globally unique suffixes. This might be overkill, but I was thinking this "degrades gracefully" if people have some namespaces that aren't self-contained. If I didn't do this, then the user would be shown hashes when viewing definitions that aren't within a self-contained namespace. 

## What now? 

I haven't played with this much, but it seems nice so far. I'm going to dogfood this branch for my daily Unison development and try to sand any rough edges. I'd encourage other folks to try the same if they want.

## Loose ends

We should probably change where `base` gets fetched - I'd put it in `.scratch.base` rather than `.base`.

And then we should update docs if/when this gets merged.


unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -399,9 +399,15 @@ viewByPrefix =
     )
 
 find :: InputPattern
-find =
+find = find' "find" False
+
+findGlobal :: InputPattern
+findGlobal = find' "find.global" True
+
+find' :: String -> Bool -> InputPattern
```
aryairani, line 426: Maybe a new type here instead of Bool?

aryairani: IsVerbose, and Global vs Local?
What was isVerbose again anyway? `ls` vs `find`?

aryairani (approved):

unison-cli/src/Unison/CommandLine/InputPatterns.hs
pchiusano: `isVerbose` is whether to include hashes in the results.

There was an unused `Bool` flag that I just repurposed for `find` vs `find.global`. I'm not actually sure what the intent of it was... since it was totally ignored in `HandleInput`. :)

I can do some `newtype`s though.

sullyj3:
What about having find take an optional namespace argument? Then `find.global foo` becomes `find . foo`, and you simultaneously eliminate a lot of unnecessary cding around

pchiusano:
https://github.com/unisonweb/unison/issues/2988#issuecomment-1080650808

Might be a good tweak to make this work nicer. I like the idea of not needing to do anything fancy for the library install process and retaining the full DAG of library dependencies (which can then be used for name resolution) rather than flattening it.

pchiusano:
I’ve been dogfooding this all week and it’s been great. No complaints. I don’t even notice it really, which is more than I could say for the old behavior.

I’m going to fix up the failing transcripts (reliant on old behavior) and merge. 

I might add the ranking of names idea based on ‘lib’ segments also, I think that will help make this even more scalable.

ceedubs:
@pchiusano I've been using this branch for about an hour or two of Unison development, and so far it has felt right and I think that it will solve a lot of the common issues that I run into.

One observation: I think that we either want `names` to not be limited to the current namespace or add a `names.global` command to mirror `find.global`.

pchiusano:
@ceedubs good call on `names` I've added `names.global` and a helpful hint suggesting its use.

pchiusano:
Okay, this is ready to go, will merge after CI. A couple final notes:

* I did add `names` and `names.global` commands, with `names` just giving aliases in the current namespace and `names.global` giving them globally.
* I implemented ranked suffix-based resolution based on number of `lib` segments https://github.com/unisonweb/unison/issues/2988#issuecomment-1080650808 via the function `Name.searchByRankedSuffix`.
* Added a bit of verbiage in `suffixes.md` transcript describing and exercising the new behavior.
* A number of transcripts had to be changed as they were reliant on being able to resolve names in some great aunt / uncle or cousin namespace. Nothing exciting here.

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -500,7 +500,7 @@ renderTypeError e env src = case e of
             "\n\n",
             annotatedAsErrorSite src termSite,
             case expectedType of
-              Type.Var' (TypeVar.Existential {}) -> "\nThere are no constraints on its type."
+              Type.Var' (TypeVar.Existential {}) -> "\nThere are no constraints on its type. "
```
pchiusano, line 503: Missing a space in this message. It was super bugging me.

unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -949,8 +949,8 @@ loop = do
 
                 fixupOutput :: Path.HQSplit -> HQ.HashQualified Name
                 fixupOutput = fmap Path.toName . HQ'.toHQ . Path.unsplitHQ
-            NamesI thing -> do
-              ns0 <- basicParseNames
+            NamesI global thing -> do
+              ns0 <- if global then pure basicPrettyPrintNames else basicParseNames
```
pchiusano, line 953: `NamesI` picks up a `IsGlobal` flag, which is used to control which scope names to use (the local parsing names, or the global pretty print names).

Maybe `basicPrettyPrintNames` should be renamed to `globalPrettyPrintNames` and `basicParseNames` to `localParseNames`?

unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -963,7 +963,7 @@ loop = do
                   types' = Set.map go types
                     where
                       go r = (r, NamesWithHistory.typeName hqLength r printNames)
-              respond $ ListNames hqLength (toList types') (toList terms')
+              respond $ ListNames global hqLength (toList types') (toList terms')
```
pchiusano, line 966: `ListNames` picks up an `IsGlobal` flag as well, so it can print a helpful tip about using `names.global` or not, if the user has already just used `names.global`.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
pchiusano: Tried out a new type, did not love it, went with type alias.

pchiusano:
> We should probably change where base gets fetched - I'd put it in .scratch.base rather than .base.

Actually I no longer agree with this. Where it is is fine. You can always just `fork .base .myproject.lib.base` if you want to make `base` available to your project.

PR #3012, opened 2022-03-30
Fix a bug caused by renaming during ANF

dolio:
Sometimes the ANF translation would flatten `let` blocks, leading to a binding like:

    let v = u
    E

And this is handled by renaming to avoid some complications. However, the renaming code was not properly renaming everything. `E` is first translated to ANF, yielding a context and a body, and only the body had renaming done on it. This is wrong both because the context could contain `v`, and because the context could (conceivably) shadow `v` (although I'm not actually sure if that can happen). This PR fixes these issues.


ceedubs (approved):
Thanks @dolio! I've confirmed that this fixed the issue that I was seeing.

PR #3024, opened 2022-04-01
Fix a bug with code serialization.

dolio:
Get group was associating variables with local functions incorrectly. I had anticipated the right approach, but mistakenly didn't use it.


pchiusano (approved):
Nice regression tests.

PR #3031, opened 2022-04-06
Make POp serialization more reliable

dolio:
There was a huge list of all the ops associated with numbers used to serialize them, but there was no way to check that every `POp` was in the list (and some were actually missing). Instead, use a function from `POp` to number so that the coverage checker can complain if a new op is added. Then the list is generated from the `Enum`/`Bounded` instances.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/ANF.hs
+++ b/parser-typechecker/src/Unison/Runtime/ANF.hs
@@ -822,6 +822,10 @@ litRef (C _) = Ty.charRef
 litRef (LM _) = Ty.termLinkRef
 litRef (LY _) = Ty.typeLinkRef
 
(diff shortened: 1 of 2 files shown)


pchiusano:
Is this something @ceedubs should do a smoke test on before merging?

pchiusano (approved):

dolio:
In theory they'll never bump into this code, because only built-ins have a `POp` in them, and they never ask to serialize those.

But it wouldn't hurt, of course.

ceedubs:
It's easy enough for me to run my shell script against this branch, so I'll give it a try. Just gotta wait for a stack build.

ceedubs (approved):
All good!

PR #3033, opened 2022-04-07
Fix move.namespace documentation

Sgeo:


ceedubs (approved):
Good catch. Thank you @Sgeo!

PR #3063, opened 2022-04-29
restore `-Wtype-defaults`

aryairani:
Un-suppressing this warning, which can lead to runtime surprises.

### DIFF
--- a/parser-typechecker/tests/Unison/Test/ANF.hs
+++ b/parser-typechecker/tests/Unison/Test/ANF.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE PatternGuards #-}
-{-# LANGUAGE TypeApplications #-}
 
 module Unison.Test.ANF where
 
--- a/parser-typechecker/tests/Unison/Test/ANF.hs
+++ b/parser-typechecker/tests/Unison/Test/ANF.hs
@@ -6,7 +6,6 @@ module Unison.Test.ANF where
 
 import Control.Monad.Reader (ReaderT (..))
 import Control.Monad.State (evalState)
-import Data.Int (Int64)
 import qualified Data.Map as Map
 import qualified Data.Set as Set
 import Data.Word (Word64)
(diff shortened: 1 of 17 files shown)


aryairani:
@dolio Could you take a look at this—specifically `ANF.hs` and `Test/ANF.hs`—to see if you agree with the annotations, or would annotate it differently? For these that I didn't fully understand, I just tried to put in what it seemed like was currently being selected as the default, even though I wasn't sure if that's what was actually desired. Feel free to add commits.

dolio:
I think those spots look fine.

I noticed that there seem to be some indentation changes in there. Is the indentation check not in CI yet? Or was it already wrong somehow?

Edit: I guess what I'm getting at is: is this adding something that will break in the future, or fixing something that would have broken in the future? It seems like the indenting must have been done on different versions of the tool, and I don't know which is the one that's going to be canonical eventually.

parser-typechecker/tests/Unison/Test/ANF.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE TypeApplications #-}
```
ChrisPenner: Is this not yet in our package defaults? We should add it :)

ChrisPenner (approved):

aryairani:
> I think those spots look fine.
> 
> I noticed that there seem to be some indentation changes in there. Is the indentation check not in CI yet? Or was it already wrong somehow?
> 
> Edit: I guess what I'm getting at is: is this adding something that will break in the future, or fixing something that would have broken in the future? It seems like the indenting must have been done on different versions of the tool, and I don't know which is the one that's going to be canonical eventually.

Yeah, good catch, thanks. I'm having a hard time telling what version haskell-language-server (which I use) embeds as a Haskell library, and I guess some people are using standalone binaries of various versions. So I'm not quite sure what to do about this or what version we should be using. 😬 

We don't have any enforcement in CI yet, because there was a concern that we'd stymie new contributors. Though I think the closer that we get to homogenous formatting, the more annoying it will be to see these in diffs.

parser-typechecker/tests/Unison/Test/ANF.hs
aryairani:
```suggestion
```

aryairani: Good point — it looks like it is already in our package defaults, and I probably just added it here in a desperate attempt to make HLS work better in some moment. This will probably keep happening though until I have the list of package defaults memorized, and stop trying such things.

Also it looks like there are lots of existing spots where we explicitly enable extensions that are already in our package defaults, so we should make a PR for that, assuming it doesn't break some standalone ormolu that we've decided we depend on and which isn't aware of package defaults. 😬

PR #3068, opened 2022-05-10
Add built-in nanosecond-precision clocks

runarorama:
This adds a few builtins for nanosecond-precision clocks.

1. `realtime` returns the civic time as nanoseconds since epoch.
2. `monotonic` returns a number of nanoseconds since an arbitrary point in the past, guaranteed to monotonically increase on subsequent calls. This is unaffected by time changes and other processes changing the system clock. 
3. `processCPUTime` returns the amount of time spent in the Unison process running the program.
4. `threadCPUTime` returns the amount of time spent in the OS thread running the program.


parser-typechecker/src/Unison/Runtime/Builtin.hs
```diff
@@ -1470,6 +1483,12 @@ boxNatToEFBox =
   where
     (arg1, arg2, nat, stack1, stack2, fail, result) = fresh7
 
+-- Nat -> Either Failure b
+-- natToEFBox :: ForeignOp
+-- natToEFBox = inNat arg nat result $ outIoFail stack1 stack2 fail result
+--   where
+--     (arg, nat, stack1, stack2, fail, result) = fresh6
+
```
dolio, line 1485: Maybe this commented code could go?

dolio (approved):
Looks fine to me, aside from the commented code that might have snuck in.

pchiusano:
@dolio are all these implementable in Chez land? 

(This is something we should always ask ourselves for new builtins I think)

dolio:
Yeah, actually.

https://cisco.github.io/ChezScheme/csug9.5/system.html#./system:h10

PR #3075, opened 2022-05-18
Add examples to `patch`'s help

### LINKED ISSUE #3030: patch help should describe optional argument
I believe that the `patch` command accepts an optional patch argument, but it isn't described in `help patch`:

```
  patch
  `patch` rewrites any definitions that depend on definitions with type-preserving edits to use the
  updated versions of these dependencies.
```

plumenator:
Close #3030

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -293,11 +293,24 @@ patch =
     []
     I.Visible
     [(Required, patchArg), (Optional, namespaceArg)]
-    ( P.wrap $
+    ( P.lines
+      [ P.wrap $
         makeExample' patch
-          <> "rewrites any definitions that depend on "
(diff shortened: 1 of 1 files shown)


ceedubs:
Thanks a lot for the PR, @plumenator!

I think that there should be an example for the no-arg version that uses the default `patch` in the current namespace.

It's news to me that you can also pass in a namespace. That's a thing @aryairani? If so, should the example not have the square braces in `[path]` since this is the example where `path` is being provided and there's another example where it's not?

plumenator:
The `patch` arg is required. As for the `path` example, I followed the `todo` example:

https://github.com/unisonweb/unison/blob/0a0e089e1d61fe0e3bd80d0604787e2c54e8d572/unison-cli/src/Unison/CommandLine/InputPatterns.hs#L133

ceedubs (approved):
@plumenator sorry I think that I was confused both in my comment on this PR and in the original issue that I created. Thanks for setting me straight and for fixing the docs!

PR #3094, opened 2022-06-03
Add package.yaml workaround to stack rebuild issue

mitchellwrosen:
Upstream issue: https://github.com/commercialhaskell/stack/issues/5503

In short, `stack` later than 2.3.1 rebuilds too often:

```
stack build
stack build
```

usually unregisters a few packages and rebuilds, due to detected changes to auto-generated `Paths_*` modules. So, @aryairani found this workaround, which I cherry-picked from another branch onto trunk.

### DIFF
--- a/codebase2/util/package.yaml
+++ b/codebase2/util/package.yaml
@@ -3,13 +3,20 @@ github: unisonweb/unison
 
 library:
   source-dirs: src
+  when:
+    - condition: false
+      other-modules: Paths_unison_util
 
 benchmarks:
   bench:
+    when:
+      - condition: false
+        other-modules: Paths_unison_util
     dependencies:
       - criterion
       - sandi
       - unison-util
+      - unison-util-base32hex
     main: Main.hs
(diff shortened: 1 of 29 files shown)


ChrisPenner (approved):
Huh; TIL.

Glad we've got a fix for that; Here's hoping that means I can `stack build --fast --test` to and then later `stack build --fast` without rebuilding everything now? 🤞🏼

codebase2/util/package.yaml
```diff
@@ -3,13 +3,20 @@ github: unisonweb/unison
...
 benchmarks:
   bench:
+    when:
+      - condition: false
+        other-modules: Paths_unison_util
     dependencies:
       - criterion
       - sandi
       - unison-util
+      - unison-util-base32hex
```
mitchellwrosen: Oops, this should be deleted, this package doesn't exist on trunk

pchiusano (approved):

PR #3100, opened 2022-06-06
Change `delete.namespace` to not print a diff

pchiusano:
It's not really useful information and takes a while to compute (like tens of seconds sometimes). You can always `ls` or `find` within the namespace before deleting if you're not sure what you're deleting.

Here's the new behavior.

image

A super common use of `delete.namespace` is deleting a feature branch after it's been merged to trunk. Seeing a massive "diff" of all the definitions in your library (and also your library's dependencies) is not helpful.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
@@ -760,25 +760,20 @@ loop = do
                 Just (Branch.head -> b0) -> do
                   endangerments <- computeEndangerments b0
                   if null endangerments
-                    then doDelete b0
+                    then doDelete
                     else case insistence of
                       Force -> do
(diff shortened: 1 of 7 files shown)


pchiusano:
Maybe if diffs always can be computed instantly and then excerpted if they're too big, then we can consider resurrecting the old behavior. But I think how it is now isn't very usable and we should merge this.

mitchellwrosen (approved):

ceedubs:
@pchiusano was `delete.namespace` on a subnamespace previously checking that you weren't removing the names for any terms that are in use in the current namespace? For example would `foo> delete.namespace bar` potentially say something like "bar.baz is referenced by foo.x"?

If so, has that behavior been affected by this change?

pchiusano:
@ceedubs nope that behavior is still the same. This is literally just about quieting the output.

PR #3102, opened 2022-06-06
Limit batch size of pushes to share

mitchellwrosen:
## Overview

This PR limits the number of entities included in an upload request to Share to 500.

Depends on #3097 

Still TODO (in another PR): limit size of pulls from Share


aryairani (approved):
LGTM

PR #3117, opened 2022-06-15
update to ghc 9.2.3

aryairani:
pins monad-validate at https://github.com/hasura/monad-validate/pull/5 and sets ghc 9.2.3 via `resolver: nightly-2022-06-15`

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3117)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -29,19 +29,21 @@ packages:
 - lib/unison-pretty-printer
 
 #compiler-check: match-exact
-resolver: lts-18.28
+resolver: nightly-2022-06-15
 
 extra-deps:
 - github: unisonweb/configurator
   commit: e47e9e9fe1f576f8c835183b9def52d73c01327a
 - github: unisonweb/shellmet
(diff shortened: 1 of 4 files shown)


aryairani:
I guess enlil will need the same treatment at some point.

aryairani:
hls 1.7.0 doesn't support 9.2.3 anyway

mitchellwrosen:
> hls 1.7.0 doesn't support 9.2.3 anyway

Oh, yeah... but it _does_ support 9.2.2 according to: https://github.com/haskell/haskell-language-server/releases

Maybe we should pin to 9.2.2 instead, then?

aryairani:
Yeah, I will try that. 

There are some deprecation warnings to deal with regardless, including calls to `pTrace` etc. that I'm not sure how to handle? Apparently it doesn't want to be in production code at all.  I can disable deprecation warnings in that module, but...

Also hls has 9.2.3 forthcoming https://github.com/haskell/haskell-language-server/pull/2936

PR #3125, opened 2022-06-16
Update type and term indices after push and pull 

tstat:
## Overview

Modifies `saveSyncEntity` to update the term and type indices, allowing push and pull to complete successfully. 

## Implementation notes

Moves the relevant index updates into `saveTermComponent` and `saveDeclComponent` so that we don't have to remember to update the indices at all call sites. Instead we have one blessed way to save terms/decls. 

In order to keep `unison-codebase-sqlite` hash agnostic it now takes the hashing functions as a parameter (`HashHandle`) and a new library is introduced (`unison-codebase-sqlite-hashing-v2`) that provides the `v2->h2` hash handle and wraps calls from `unison-codebase-sqlite`.

There really isn't much of an api exposed by `unison-codebase-sqlite-hashing-v2` at this time, but modules are organized based on the type that it is inserting into the database, rather than the query/operation distinction that exists in `unison-codebase-sqlite`.

## Test coverage

No tests were written as part of this PR. Push and pull were manually verified to complete successfully and the indices were manually examined and queried against.

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/HashHandle.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/HashHandle.hs
@@ -15,7 +15,7 @@ data HashHandle = HashHandle
     toReference :: C.Term.Type Symbol -> C.Reference,
     -- | Hash type's mentions
     toReferenceMentions :: C.Term.Type Symbol -> Set C.Reference,
-    -- | Hash decl
+    -- | Hash the type of a single constructor in a decl component. The provided hash argument is the hash of the decl component.
     toReferenceDecl :: Hash -> C.Type.TypeD Symbol -> C.Reference,
     -- | Hash decl's mentions
     toReferenceDeclMentions :: Hash -> C.Type.TypeD Symbol -> Set C.Reference
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/HashHandle.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/HashHandle.hs
@@ -1,0 +1,22 @@
+module U.Codebase.Sqlite.HashHandle
+  ( HashHandle (..),
+  )
+where
+
+import qualified U.Codebase.Reference as C
+import U.Codebase.Sqlite.Symbol (Symbol)
+import qualified U.Codebase.Term as C.Term
+import qualified U.Codebase.Type as C.Type
+import U.Util.Hash (Hash)
+import Unison.Prelude
+
+data HashHandle = HashHandle
+  { -- | Hash type
+    toReference :: C.Term.Type Symbol -> C.Reference,
+    -- | Hash type's mentions
+    toReferenceMentions :: C.Term.Type Symbol -> Set C.Reference,
+    -- | Hash decl
+    toReferenceDecl :: Hash -> C.Type.TypeD Symbol -> C.Reference,
+    -- | Hash decl's mentions
+    toReferenceDeclMentions :: Hash -> C.Type.TypeD Symbol -> Set C.Reference
+  }
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
@@ -11,13 +11,13 @@ module U.Codebase.Sqlite.Operations
     expectCausalBranchByCausalHash,
 
     -- * terms
-    saveTermComponent,
+    Q.saveTermComponent,
     loadTermComponent,
     loadTermByReference,
     loadTypeOfTermByTermReference,
 
     -- * decls
-    saveDeclComponent,
+    Q.saveDeclComponent,
     loadDeclComponent,
     loadDeclByReference,
     expectDeclTypeById,
@@ -55,11 +55,11 @@ module U.Codebase.Sqlite.Operations
     dependentsOfComponent,
 
     -- ** type index
-    addTypeToIndexForTerm,
+    Q.addTypeToIndexForTerm,
     termsHavingType,
 
     -- ** type mentions index
-    addTypeMentionsToIndexForTerm,
+    Q.addTypeMentionsToIndexForTerm,
     termsMentioningType,
 
     -- ** name lookup index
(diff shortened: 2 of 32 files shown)


codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
```diff
@@ -877,10 +565,13 @@ s2cBranch (S.Branch.Full.Branch tms tps patches children) =
           boId <- Q.expectBranchObjectIdByCausalHashId chId
           expectBranch boId
 
-saveRootBranch :: C.Branch.CausalBranch Transaction -> Transaction (Db.BranchObjectId, Db.CausalHashId)
-saveRootBranch c = do
+saveRootBranch ::
+  HashHandle ->
```
aryairani, line 569: Kind of a bummer how HashHandle has to be passed through these functions that won't trigger it, just because the low level functions are generic and might need it in some other contexts.

codebase2/codebase-sqlite/unison-codebase-sqlite.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.34.4.
+-- This file has been generated from package.yaml by hpack version 0.34.7.
```
aryairani, line 3: Can `nix` give you a stack with hpack 0.34.4? (: nix question

aryairani (approved):
Great!

aryairani:
I forget, does this PR already cover enlil, provided it uses the right version of the dependency?

codebase2/codebase-sqlite/unison-codebase-sqlite.cabal
tstat: Oh yes, I will downgrade hpack.

tstat:
There is a small follow-up PR for enlil.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
tstat: Ah, so some paths do not need to check if they should flush the temp entity table? If so, maybe we should split or redefine some of the lower level functions so that they don't always attempt to flush.

mitchellwrosen: > Ah, so some paths do not need to check if they should flush the temp entity table?

I think all paths do need to check if they should flush the temp entity table

codebase2/codebase-sqlite/U/Codebase/Sqlite/HashHandle.hs
```diff
@@ -0,0 +1,22 @@
...
+import qualified U.Codebase.Term as C.Term
+import qualified U.Codebase.Type as C.Type
+import U.Util.Hash (Hash)
+import Unison.Prelude
+
+data HashHandle = HashHandle
+  { -- | Hash type
+    toReference :: C.Term.Type Symbol -> C.Reference,
+    -- | Hash type's mentions
+    toReferenceMentions :: C.Term.Type Symbol -> Set C.Reference,
```
mitchellwrosen, line 17: A thought for another day: do we need `toReferenceMentions` / `toReferenceDeclMentions` in this handle, or are the implementable via `toReference` / `toReferenceDecl`?

codebase2/codebase-sqlite/U/Codebase/Sqlite/HashHandle.hs
```diff
@@ -0,0 +1,22 @@
...
+    -- | Hash decl
```
mitchellwrosen:
```suggestion
    -- | Hash the type of a single constructor in a decl component. The provided hash argument is the hash of the decl component.
```

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -142,35 +144,60 @@ module U.Codebase.Sqlite.Queries
     syncToTempEntity,
     insertTempEntity,
     saveTempEntityInMain,
+    expectTempEntity,
```
mitchellwrosen, line 145: Travis's note to self: these might not need to be exported

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -647,19 +704,26 @@ saveCausal self value parents = do
...
 -- | `tryMoveTempEntityDependents #foo` does this:
 --    0. Precondition: We just inserted object #foo.
 --    1. Look up the dependents of #foo
 --    2. Delete #foo as dependency from temp_entity_missing_dependency. e.g. (#bar, #foo), (#baz, #foo)
 --    3. For each like #bar and #baz with no more rows in temp_entity_missing_dependency,
 --        insert_entity them.
-tryMoveTempEntityDependents :: Hash32 -> Transaction ()
-tryMoveTempEntityDependents dependency = do
+tryMoveTempEntityDependents ::
+  -- | Move TempEntity to main
```
mitchellwrosen:
```suggestion
```

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1624,3 +1651,407 @@ elaborateHashes hashes = do
...
+               in Set.fromList $
+                    map getTermSRef tmRefs
+                      ++ map getSTermLink tmLinks
+                      ++ map getTypeSRef (tpRefs ++ tpRefs')
+                      ++ map getSTypeLink tpLinks
+         in Set.map (,self) dependencies
+  traverse_ (uncurry addToDependentsIndex) dependencies
+  for_ ((snd <$> terms) `zip` [0 ..]) \(tp, i) -> do
+    let self = C.Referent.RefId (C.Reference.Id oId i)
+        typeForIndexing = toReference tp
```
mitchellwrosen, line 1772: Note: the crowd prefers something like `HashHandle.toReference hh`

parser-typechecker/src/Unison/Codebase/SqliteCodebase/Migrations/MigrateSchema1To2.hs
```diff
@@ -50,6 +50,7 @@ import qualified U.Codebase.Sqlite.Patch.Full as S
 import qualified U.Codebase.Sqlite.Patch.TermEdit as TermEdit
 import qualified U.Codebase.Sqlite.Patch.TypeEdit as TypeEdit
 import qualified U.Codebase.Sqlite.Queries as Q
+import U.Codebase.Sqlite.V2.HashHandle (v2HashHandle)
```
mitchellwrosen, line 53: Note: we realize this hash handle isn't actually going to be used in the migration. (none of `saveCausal`/`saveDbBranchUnderHashId`/`saveDbPatch` use it)

mitchellwrosen (approved):
:tada:

PR #3126, opened 2022-06-16
Add `Sha1` builtin

dfreeman:
## Overview

We discussed this briefly in Slack last weekend. While SHA-1 is a broken algorithm from a cryptographic perspective, it's still useful to have available for existing protocols that require its use (e.g. the [WebSocket handshake](https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.2)).

## Implementation notes

This introduces a new builtin `crypto.HashAlgorithm.Sha1` using the existing `declareHashAlgorithm` machinery in `parser-typechecker`.

In principle there's no reason this couldn't be implemented in pure Unison (and in fact I did just that as an exercise), but that's likely significantly slower, and there's already a fairly established pattern of exposing well-known cryptographic primitives as builtins.

## Test coverage

I've mirrored the transcript coverage that the other hash builtins have using test vectors from [the same source](https://www.di-mgt.com.au/sha_testvectors.html) as the existing ones.

## Loose ends

I'm not 100% clear about the relationship between `.base` and `.builtin`, and whether there are any additional steps necessary to integrate this. If so, I'd welcome guidance 🙂

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -68,3 +68,4 @@ The format for this list: name, GitHub handle
 * Harald Gliebe (@hagl)
 * Phil de Joux (@philderbeast)
 * Travis Staton (@tstat)
+* Dan Freeman (@dfreeman)
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -593,7 +593,7 @@ hashBuiltins =
     B "crypto.hmac" $ forall1 "a" (\a -> hashAlgo --> bytes --> a --> bytes),
     B "crypto.hmacBytes" $ hashAlgo --> bytes --> bytes --> bytes
   ]
-    ++ map h ["Sha3_512", "Sha3_256", "Sha2_512", "Sha2_256", "Blake2b_512", "Blake2b_256", "Blake2s_256"]
+    ++ map h ["Sha3_512", "Sha3_256", "Sha2_512", "Sha2_256", "Sha1", "Blake2b_512", "Blake2b_256", "Blake2s_256"]
   where
     hashAlgo = Type.ref () Type.hashAlgorithmRef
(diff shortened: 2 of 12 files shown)


dolio (approved):
Looks pretty straight forward to me.

dolio:
For reference, `.builtin` is where these sorts of things get auto-added when running the `builtins.merge` command in UCM.

To get them in base, you should submit a PR to it. The simplest way I know how to prepare that is to run `builtins.merge` locally, then `move.term` the added thing into the appropriate place in a forked base. I think there might be a way to directly create a term alias for a new builtin without `builtins.merge`, but I don't actually know how.

pchiusano:
Awesome, thanks @dfreeman !

Could you add yourself to CONTRIBUTORS.markdown and we'll get this merged!

pchiusano:
@dfreeman once this is merged you can add an alias to base or at least file a ticket at https://github.com/unisonweb/base so we don't lose track of adding it for the next release.

But basically, for builtins like this, you can do what @dolio said above, or do -

```
.base> fork trunk sha1
.base.sha1> alias.term ##Sha1 crypto.HashAlgorithm.Sha1
```

Also you can create a doc called `crypto.HashAlgorithm.Sha1.doc` as well.

And then push that somewhere and open a ticket on base that says where to pull from.

pchiusano (approved):

dfreeman:
> Could you add yourself to CONTRIBUTORS.markdown and we'll get this merged!

Done! Sorry for the delay.

Once this is merged and I'm back home from traveling later this week I'll plan to open a ticket for base as well. Thanks for the guidance!

pchiusano:
@dfreeman sounds good, I created https://github.com/unisonweb/base/issues/95 to track.

pchiusano:
Applied ready-to-merge, it should merge after CI completes. Thanks again @dfreeman!

PR #3135, opened 2022-06-19
`do` parses as `'let` now, and this is used by pretty-printer

pchiusano:
You can still use `'let`, but `do` can be used instead:

```haskell
main : '{IO, Exception} ()
main = do 
  printLine "Greetings, earthling! 👽 🪐"
  printLine "You can still use 'let if you really want."
```

And the pretty-printer will opportunistically use this syntax.

More examples: 

image

I played around with this manually and also updated the round trip tests to exercise it. I'd say the testing is pretty good.

## Controversial decisions

It seemed like `do` was the winner when we last bikeshedded this, so I went with that. But maybe since then people have had other ideas. In any case, it's easy to change whenever we want.

### DIFF
--- a/parser-typechecker/src/Unison/Lexer.hs
+++ b/parser-typechecker/src/Unison/Lexer.hs
@@ -929,6 +929,7 @@ lexemes' eof =
             <|> openKw "cases"
             <|> openKw "where"
             <|> openKw "let"
+            <|> openKw "do"
           where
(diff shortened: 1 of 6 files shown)


ChrisPenner (approved):
I'm a fan 👍🏼

runarorama (approved):

Issue #2848, opened 2022-01-31
Program using ability-kinded type parameter incorrectly typechecks


pchiusano:
```haskell
unique type Loc blah = Loc
unique ability Oog where oog : Nat

l : Loc {}
l = Loc 

foo : Loc {Oog} -> a -> Nat
foo l a = 99

> foo l 9
```

so `l : Loc {}` yet somehow that typechecks as a `Loc {Oog}`?

If I had to guess, it has something to do with the fact that the constructor `Loc` has type `forall a . Loc a`.

pchiusano:
Here's another clue. This program does not typecheck as expected, which tells me it's something with ability checking specifically:

```Haskell
unique type Loc blah = Loc
unique ability Oog where oog : Nat

l : Loc Int 
l = Loc 

foo : Loc Nat -> a -> Nat
foo l a = 99

> foo l 9
```

This gives the expected error:

```
  The 1st argument to `foo`
  
            has type:  Loc Int
      but I expected:  Loc Nat
```

pchiusano:
Here's another one, showing that it doesn't matter whether it's a phantom type or not:

```Haskell
unique type Loc blah = Loc ('{blah} Nat)
unique ability Oog where oog : Nat

l : Loc {} 
l = Loc '0 

foo : Loc {Oog} -> a -> Nat
foo l a = 99

> foo l 9
```

This incorrectly typechecks as the original example does.

PR #3149, opened 2022-06-24
Some ability checking adjustments

### LINKED ISSUE #3037: type-checking false positive with effect type
```
structural type Runner g = Runner (forall a. '{g} a -> {} a)

pureRunner : Runner {}
pureRunner = Runner base.force

-- this compiles, but shouldn't the effect type parameter on Runner be invariant?
runner : Runner {IO}
runner = pureRunner
```

```
    
      structural type Runner g
      pureRunner : Runner {}
      runner     : Runner {IO}
```

dolio:
Fixed an erroneous case when equating two sets of abilities. Normally we try to unify pieces of them with pieces of the opposite side to render them equal. However, in certain cases where that wasn't possible, we would just not do anything, and also not fail. This code tries some additional unification cases, and properly throws an error if nothing will work.

This caused some knock-on problems with certain test examples. For instance, if you annotate a handler as `Request {...} r -> ...`, then variables made up in the function body will fail to unify with the rigidly defined row. So ability checking for cases has been modified to avoid making up superfluous variables.

Fixes #3037 and #2848

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -1215,17 +1215,33 @@ checkCases scrutType outType cases@(Term.MatchCase _ _ t : _) =
   scope (InMatch (ABT.annotation t)) $ do
     mes <- requestType (cases <&> \(Term.MatchCase p _ _) -> p)
     for_ mes $ \es ->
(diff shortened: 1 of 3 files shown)


pchiusano (approved):

pchiusano:
Nice!

PR #3157, opened 2022-06-26
Move syntax-related modules into Unison.Syntax (close #3156)

### LINKED ISSUE #3156: Collect syntax-related modules under `Unison.Syntax`
This would be a preparatory step towards https://github.com/unisonweb/unison/issues/499, if we choose to do it.

The possible advantage of doing this as a first step is twofold:

1. It would make https://github.com/unisonweb/unison/issues/499 easier, "make the change easy, then make the easy change", yadda yadda.
2. On the off chance that we don't add a second syntax for awhile, we avoid the clutter of passing a `Syntax` record around in the meantime. Though due to (1) this still helps us even if we want to do that immediately.

Here's one way we could do the reorg:
```
renamed:    parser-typechecker/src/Unison/Lexer.hs -> parser-typechecker/src/Unison/Syntax/Lexer.hs
renamed:    parser-typechecker/src/Unison/Parser.hs -> parser-typechecker/src/Unison/Syntax/Parser.hs
renamed:    parser-typechecker/src/Unison/TermParser.hs -> parser-typechecker/src/Unison/Syntax/Parser/Term.hs
renamed:    parser-typechecker/src/Unison/TypeParser.hs -> parser-typechecker/src/Unison/Syntax/Parser/Type.hs
renamed:    parser-typechecker/src/Unison/Parsers.hs -> parser-typechecker/src/Unison/Syntax/Parsers.hs
renamed:    parser-typechecker/src/Unison/DeclPrinter.hs -> parser-typechecker/src/Unison/Syntax/Printer/Decl.hs
renamed:    parser-typechecker/src/Unison/NamePrinter.hs -> parser-typechecker/src/Unison/Syntax/Printer/Name.hs
renamed:    parser-typechecker/src/Unison/TermPrinter.hs -> parser-typechecker/src/Unison/Syntax/Printer/Term.hs
renamed:    parser-typechecker/src/Unison/TypePrinter.hs -> parser-typechecker/src/Unison/Syntax/Printer/Type.hs
```

Resulting in the module structure:
```
$ tree parser-typechecker/src/Unison/Syntax/
parser-typechecker/src/Unison/Syntax/
├── Lexer.hs
├── Parser
│   ├── Term.hs
│   └── Type.hs
├── Parser.hs
├── Parsers.hs
└── Printer
    ├── Decl.hs
    ├── Name.hs
    ├── Term.hs
    └── Type.hs
```

Though someone more familiar with the code could probably improve on that. I'll put some notes below with questions.[1]

Also, this would prepare us for even another incremental improvement before we get to https://github.com/unisonweb/unison/issues/499: we could use `Unison.Syntax` to define the interface we want in our eventual `Syntax` type (though in a module not in a type yet). Then we could switch the rest of the code over to only interact with the parser/prettyprinter through that interface, forbidding it to import anything from `Unison.Syntax.*` except for `Unison.Syntax` itself. That way when we do make a `Syntax` type the only thing to do will be to thread it around.

I'd suggest we save that for a later issue and PR though, not here.

What do you folks think?

[1] If we do go with this idea, here are my questions:
- `Unison.FileParser` probably go in `Unison.Syntax` as well, right?
- Should we rename `Unison.FileParsers` since it does a lot more than parsing like typechecking and evaluation?
- Should we go ahead and call this stuff `Unson.Syntax.Default` instead of `Unison.Syntax`? I don't have an opinion one way or another.
- I'm not sure if `Unison.Parser.Ann` is default-syntax specific, in which case it should be moved into the same module hierarchy as all the above, or will be used by all the syntaxes, in which case it should probably stay a level up.
- Also not sure what to do about `Unison.PrettyPrintEnv*` and `Unison.PrintError`. They're not moved in the above suggested changes, down to do whatever with them.

seagreen:
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3157)

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Path/Parse.hs
+++ b/parser-typechecker/src/Unison/Codebase/Path/Parse.hs
@@ -23,11 +23,11 @@ import Data.List.Extra (stripPrefix)
 import qualified Data.Text as Text
 import Unison.Codebase.Path
 import qualified Unison.HashQualified' as HQ'
-import qualified Unison.Lexer as Lexer
 import Unison.NameSegment (NameSegment (NameSegment))
 import qualified Unison.NameSegment as NameSegment
 import Unison.Prelude hiding (empty, toList)
 import qualified Unison.ShortHash as SH
+import qualified Unison.Syntax.Lexer as Lexer
 
 -- .libs.blah.poo is Absolute
 -- libs.blah.poo is Relative
--- a/parser-typechecker/src/Unison/FileParser.hs
+++ b/parser-typechecker/src/Unison/FileParser.hs
@@ -13,20 +13,20 @@ import qualified Text.Megaparsec as P
 import qualified Unison.ABT as ABT
 import Unison.DataDeclaration (DataDeclaration, EffectDeclaration)
 import qualified Unison.DataDeclaration as DD
-import qualified Unison.Lexer as L
 import qualified Unison.Name as Name
 import qualified Unison.Names as Names
 import qualified Unison.Names.ResolutionResult as Names
 import qualified Unison.NamesWithHistory as NamesWithHistory
-import Unison.Parser
 import Unison.Parser.Ann (Ann)
 import Unison.Prelude
+import qualified Unison.Syntax.Lexer as L
+import Unison.Syntax.Parser
+import qualified Unison.Syntax.Parser.Term as TermParser
+import qualified Unison.Syntax.Parser.Type as TypeParser
 import Unison.Term (Term)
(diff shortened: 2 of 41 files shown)


seagreen:
This really is just a draft, if it seems promising I'll make the corresponding renames within the tests as well, and also write up a real PR message.

aryairani (commented):
Seems fine / good to me; thanks @seagreen

seagreen:
Closed in favor of https://github.com/unisonweb/unison/pull/3319

PR #3161, opened 2022-06-28
Adjust effect variable 'tweaking' strategy

### LINKED ISSUE #3159: Spurious ability check failures
```haskell
structural type Hashed a = Hashed Bytes

structural ability Fold g a where
  fork : '{Fold g a, g} r -> Hashed r 
  foldMap : (a ->{g} r) -> r -> (r ->{} r ->{g} r) -> Hashed r 
  foldLeft : r -> (r -> a ->{g} r) -> Hashed r
  await : Hashed r -> r

sum : (a ->{} Nat) ->{Fold g a} Hashed Nat
sum by = foldMap by 0 (+)
```

This yields:

```
  The expression in red needs the {Fold {g13, g14} a6} ability, but this location only has access to the {Fold g9 a6} ability,
  
     11 | sum by = foldMap by 0 (+)
                   ^^^^^^^^^^^^^^^^
```

Removing the `{g}` from `(a ->{g} r)` in `foldMap` fixes it, but I really want to allow `g` in the mapping function there.

dolio:
This is a part of the type checker that mediates between some old inferred types and ones that work better in the new type checker. The idea was to turn, e.g. the type of `foldRight`:
    
    (a ->{g} b ->{g} b) -> b -> [a] ->{g} b
    
into:
    
    (a ->{g1} b ->{g2} b) -> b ->[a] ->{g1,g2} b
    
Which makes up distinct variables for each negative occurrence, which in turn means that solving those negative occurrences doesn't interfere in an order-dependent way. This is always sound, because you can instantiate `g` to the whole row, and use subtyping on the negative occurrences.
    
However, having multi-variable rows in invariant positions doesn't work out very well, and this tweak prevents you from declaring such situations to only involve a single variable. So, this patch modifies the tweak to only occur if `g` only occurs in positions that are either covariant or contravariant. It may be "invariant" in that it occurs both covariantly _and_ contravariantly, but it may not occur in a single position that is considered invariant.

Fixes #3159

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -1632,7 +1632,7 @@ tweakEffects ::
   Type v loc ->
   M v loc ([v], Type v loc)
 tweakEffects v0 t0
-  | isEffectVar v0 t0 =
+  | isEffectVar v0 t0 && isVariant v0 t0 =
       rewrite (Just False) t0 >>= \case
         ([], ty) ->
           freshenTypeVar v0 >>= \out -> finish [out] ty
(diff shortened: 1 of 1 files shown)


pchiusano (approved):

pchiusano:
Ima merge this, but would like to revisit our ability inference at some point. I think we're in agreement that we should just be less ambitious with inference. User has to write more type signatures, but the places where they need to do that are clear, and they get predictable results.

PR #3172, opened 2022-07-01
Don't consider names outside the current namespace when pretty-printing, unless necessary

pchiusano:
This changes how the pretty-print environment gets built - rather than munging the names, it just produces two separate pretty-print environments, one for the local names, and one for the global names. If a symbol has no name locally, it looks in the global names, otherwise it uses the local name. 

Thus, if a namespace doesn't happen to be self-contained, we still get a name for everything, but we don't make the rendering of local names worse out of fear of conflicting with something outside the current namespace.

Previously, we would do weird stuff to try to shadow the global names with local ones (which is also costly). So I think this will give faster results and also not show needless name segments in some cases. (I don't have a good test case handy though)


pchiusano:
Transcript refresh shows some improvements which makes me feel better about this change.

ChrisPenner (approved):
Yeah this looks better;

In general I don't think it makes much sense to have a "Names" object with anything other than full parse names, since that gives us the most info and we can just intelligently suffixify at the moment of pretty-printing like you do here.

~Note that as-is this won't affect share-next, since the names are generated from an index there, but we can perform the same improvement there too.~

Added the relevant changes to share-next here: https://github.com/unisonweb/unison/pull/3194

aryairani (approved):
Nice. I'm gonna merge it.

PR #3173, opened 2022-07-01
Add builtins for efficient Text matching

dolio:
This PR adds some builtin functionality for matching strings in the form of a `Pattern` type an functions on it. At the moment it is limited to `Text` matching, but this is only because the leaf level patterns all match `Text`. The type is general enough that leaves for matching on `Bytes` could also be added.

The operations allowed are essentially regular expressions. The functions for using a pattern include a simple boolean `isMatch`, and a more complicated `run` function that gives back a list of captures and any remainder of the string.

Some tests of the underlying functionality are included.


parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -468,8 +468,38 @@ builtinsSrc =
     B "Text.unsnoc" $ text --> optionalt (tuple [text, char]),
     B "Text.toCharList" $ text --> list char,
     B "Text.fromCharList" $ list char --> text,
+    -- Todo: implement me @dolio
```
pchiusano:
```suggestion
```

dolio: Oh! Right.

parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -468,8 +468,38 @@ builtinsSrc =
...
+    B "Text.reverse" $ text --> text,
+    B "Text.toUppercase" $ text --> text,
+    B "Text.toLowercase" $ text --> text,
     B "Text.toUtf8" $ text --> bytes,
     B "Text.fromUtf8.impl.v3" $ bytes --> eithert failure text,
+    -- TODO dolio these need implementation
```
pchiusano:
```suggestion
```

parser-typechecker/src/Unison/Builtin.hs
```diff
@@ -468,8 +468,38 @@ builtinsSrc =
...
+    -- Pattern.many : Pattern a -> Pattern a
+    B "Pattern.many" $ forall1 "a" (\a -> pat a --> pat a),
+    B "Pattern.replicate" $ forall1 "a" (\a -> nat --> nat --> pat a --> pat a),
+    B "Pattern.capture" $ forall1 "a" (\a -> pat a --> pat a),
+    B "Pattern.join" $ forall1 "a" (\a -> list (pat a) --> pat a),
+    B "Pattern.or" $ forall1 "a" (\a -> pat a --> pat a --> pat a),
+    -- Pattern.run : Pattern a -> a -> Optional ([a], a)
+    B "Pattern.run" $ forall1 "a" (\a -> pat a --> a --> optionalt (tuple [list a, a])),
+    B "Pattern.isMatch" $ forall1 "a" (\a -> pat a --> a --> boolean),
+    -- end todo
```
pchiusano:
```suggestion
```

pchiusano (approved):

parser-typechecker/src/Unison/Builtin.hs
pchiusano: I fixed just right here on GH.

dolio: Now I have to merge.

PR #3178, opened 2022-07-01
Topic/text patterns

pchiusano:
Minor oversight - `builtins.merge` was not giving a name to `Pattern` builtin type. I fixed that and updated transcripts, which now have different hashes for empty namespace.


dolio (approved):
Looks fine to me.

PR #3179, opened 2022-07-03
2-3x speed improvement for text matching

pchiusano:
I have some bigger ideas for how we could further improve performance of the regex engine, but this is a simple win that took parsing of a million line csv file from 50s to under 20s.

### DIFF
--- a/parser-typechecker/src/Unison/Util/Text.hs
+++ b/parser-typechecker/src/Unison/Util/Text.hs
@@ -29,6 +29,9 @@ one, singleton :: Char -> Text
 one ch = Text (R.one (chunk (T.singleton ch)))
 singleton = one
 
+appendUnbalanced :: Text -> Text -> Text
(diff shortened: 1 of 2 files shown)


dolio:
Would this be a problem if it ends up in the result of the matching function? Or is it not a big deal?

pchiusano:
I think it's not a big deal. Everything still works fine. It'll just have a single chunk at the top of the tree, which is actually not a bad configuration - it makes a subsequent take or drop of fewer than that chunk size fast (this is also why it improves speed here). The thing you don't want is like a linear chain of 1 character chunks at the top of the tree, but I can't see a way to make such a thing.

Also the balancing is only relevant for random access into the text, which is sort of a special case anyway.

dolio:
Okay. That was my suspicion. Just wanted to make sure.

dolio (approved):
Seems fine.

PR #3180, opened 2022-07-03
Upgrade terminal-size to fix apple m1 terminal text-wrapping bugs

### LINKED ISSUE #2737: Line wrapping after every word
When running unison, lines wrap after every word (sometimes):

Screen Shot 2021-12-07 at 7 02 14 AM

If I type something like `help`, I get the same thing:

Screen Shot 2021-12-07 at 7 03 25 AM

It looks like, for whatever reason, the line wrapper thinks I have a terminal width of zero (or something).

Any help to fix this issue is appreciated, let me know what additional information is required to resolve this issue.

## Build Steps

I just grabbed unison's source:

```
git clone <unison url> --recursive --depth=1
cd unison
```

and built it with:

```
stack build --compiler ghc-8.10.7 --system-ghc
```

Then ran unison with:

```
stack exec unison --compiler ghc-8.10.7 --system-ghc
```

(The flags are there to tell `stack` where to look for `ghc`, I don't think they have anything to do with the issue.)

Oh, and if you're curious:

```
$ stack exec tests --compiler ghc-8.10.7 --system-ghc
...
✅  857 tests passed, no failures! 👍 🎉
```

Looks like all tests pass, which implies this is most likely something wrong on my end (or something on my end that `ucm` isn't picking up).

## System

Hardware: MacBook Air (M1, 2020)
OS: macOS Monterey 12.0.1
Terminal: iTerm Build 3.4.14
GHC: 8.10.7
Stack: Version 2.7.3 aarch64
Unison: Commit `138799f92` (latest as of posting)

### LINKED ISSUE #3136: `ucm transcript`: segmentation fault
- 1.0.0-M3
- macOS 12.4, aarch64

Running `ucm` works fine, but running `ucm transcript something.md` will always seg fault (provided that something.md exists, doesn't seem to matter what's inside though)

Screen Shot 2022-06-19 at 2 49 18 PM

alvaroc1:
## Overview

Fixes text wrapping bug on apple m1 (previous version reported 0 width terminal) ~~and history-navigation using arrow keys.~~ (edit: potential fix for this was not included on this PR)

Fixes #2737

Maybe fixes #3136 ?

## Implementation notes

terminal-size: 0.3.2.1 -> 0.3.3

Which fixes a bug on apple m1 that causes terminal with to be reported as 0 and i think it also causes some segfaults

Explicit version on `extra-deps`. But i don't know much about stack, not sure if that's the best way.

## Test coverage

Works for me :P

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -44,9 +44,6 @@ extra-deps:
   commit: 563e96238dfe392dccf68d93953c8f30fd53bec8
   subdirs:
     - ki
-# lts 18.28 provides 0.8.2 but we need un-released version that fixes apple m1 bug
-- github: judah/haskeline
-  commit: 7a83e7659211049f3bd167e7a13bc1cb4c42c91b
 - guid-0.1.0@sha256:a7c975be473f6f142d5cc1b39bc807a99043d20b1bb0873fdfe7a3ce84d2faf1,1078
 - prelude-extras-0.4.0.3@sha256:1c10b0123ea13a6423d74a8fcbaeb2d5249b472588abde418a36b47b7c4f48c8,1163
 - sandi-0.5@sha256:b278d072ca717706ea38f9bd646e023f7f2576a778fb43565b434f93638849aa,3010
--- a/stack.yaml
+++ b/stack.yaml
@@ -57,6 +57,7 @@ extra-deps:
 - recover-rtti-0.4.0.0@sha256:2ce1e031ec0e34d736fa45f0149bbd55026f614939dc90ffd14a9c5d24093ff4,4423
 - lock-file-0.7.0.0@sha256:3ad84b5e454145e1d928063b56abb96db24a99a21b493989520e58fa0ab37b00,4484
 - http-client-0.7.11
+- terminal-size-0.3.3
 
 ghc-options:
  # All packages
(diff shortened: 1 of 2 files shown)


stack.yaml
```diff
@@ -57,6 +57,7 @@ extra-deps:
 - recover-rtti-0.4.0.0@sha256:2ce1e031ec0e34d736fa45f0149bbd55026f614939dc90ffd14a9c5d24093ff4,4423
 - lock-file-0.7.0.0@sha256:3ad84b5e454145e1d928063b56abb96db24a99a21b493989520e58fa0ab37b00,4484
 - http-client-0.7.11
+- terminal-size-0.3.3
```
aryairani, line 61:
```suggestion
# lts 18.28 provides 0.3.2.1 but we need at least 0.3.3
- terminal-size-0.3.3
```

aryairani: we wanna add explanations for all extra-deps so we can know when they're redundant

aryairani (approved):
Thanks @alvaroc1 !

stack.yaml
```diff
@@ -44,6 +44,9 @@ extra-deps:
   commit: 563e96238dfe392dccf68d93953c8f30fd53bec8
   subdirs:
     - ki
+# lts 18.28 provides 0.8.2 but we need un-released version that fixes apple m1 bug
+- github: judah/haskeline
+  commit: 7a83e7659211049f3bd167e7a13bc1cb4c42c91b
```
aryairani: According to https://github.com/judah/haskeline/pull/163 (from aug 2021) that PR was included in haskeline 0.8.1.3. So if we are using 0.8.2, shouldn't that be newer / good enough?

What was happening when you didn't include this (older?) unreleased version?

pchiusano:
Sweet, thanks @alvaroc1 !

PR #3198, opened 2022-07-06
Fix a problem with equating ability rows where one is concrete

### LINKED ISSUE #3196: Unable to instantiate an ability variable to a concrete ability when inside an ability list
```
unique type W abilities = W

w1 : W {Abort}
w1 = W 
w2 : W {g} -> W {g}
w2 w = todo "????"  

> w2 w1
```

This complains:

```
  The expression in red needs the {g24} ability, but this location only has access to the {Abort} ability,
  
      9 | > w2 w1
               ^^
```

It should be able to instantiate `g` to `Abort` here.

Oddly, if I instead write the signature as:

```
w2 : W g -> W g
w2 w = todo "????"  
```

then it works

dolio:
There was logic to equate rows like `{g,Foo}` with `{Bar,Foo}`. However, it was only actually working if the rows were passed into the equating function in the right order. Fixed the logic so that it behaves symmetrically.

Fixes #3196


pchiusano (approved):
Thank you thank you!!! 🙏

Issue #3183, opened 2022-07-04
UCM up-arrow navigation broken (trunk - apple m1)


alvaroc1:
In the video, after using `find`, I am just hitting the up-arrow several times to browse previous commands and the output is messed up.

https://user-images.githubusercontent.com/107395/177197948-50456981-19db-4a80-8c5c-8c1283f99a97.mov

Other terminal navigation like down-arrow is also messed up.

Workaround: Starting ucm with `TERM=linux`:
```
$ TERM=linux unison --codebase .
```

Probably something to do with `haskeline`? but I think we're currently on the latest version.
Source of the workaround: https://github.com/judah/haskeline/issues/162

PR #3216, opened 2022-07-10
Downgrade haskeline (somehow it fixes ucm navigation issues on arm64)

alvaroc1:
## Overview

I don't understand how, but it fixes ucm navigation issues on Apple M1 (arm64): #3183 

## Implementation notes

Downgrades version of haskeline:
- 0.8.2 -> 0.8.1.3

Version 0.8.2 is supposed to also have the fix, but for some reason the bug manifests itself for me on that version.


pchiusano (approved):
Thanks for getting to the bottom of this!

PR #3225, opened 2022-07-12
Don't sandbox watch expressions

### LINKED ISSUE #3223: Don't run watch expressions sandboxed
Or if that's too hard, as a temporary fix, I think it's okay to just not mark `trace` as tracked.

dolio:
This makes watch expressions do an unsandboxed eval.

The doc, display and test commands still run sandboxed, although test watches will be initially evaluated unsandboxed.

Fixes #3223.


pchiusano (approved):

pchiusano:
Nice. That was easy.

Issue #3192, opened 2022-07-05
Pushing to a non-"public" gives a "no permissions" error instead of suggesting public


hojberg:
<img width="748" alt="CleanShot 2022-07-05 at 15 52 25@2x" src="https://user-images.githubusercontent.com/2371/177405494-f5296558-0968-4365-b081-fc3faa8e4de5.png">

When pushing to a non-"public" namespace on Share Next we should suggest the user to use the "public" namespace instead, explain why, and give an example of a full command based on their previous attempt.

aryairani:
First step: Come up with a good replacement message.

mitchellwrosen:
<noting that within reason we still want old clients' error messages to make sense after some "projects" work - will `public` continue to be a thing?>

hojberg:
After the project work, `public` will no longer be a thing. I think its worth a dedicated error message in this first release because its so specific.

Here's a suggestion:

> Unison Share currently only supports sharing public code. This is done by hosting code in a public namespace under your handle.
> It looks like you were trying to push `textExtra` to the `hojberg` handle. Try nesting under `public` like so:
> push.create hojberg.public.textExtra .textExtra
> 
> If you like, I can just go ahead and do that for you? y/n

The last bit with the prompt would be really awesome, but I understand that we can't do prompts right now in UCM.

PR #3226, opened 2022-07-13
Suggest public on "no permissions" error

tstat:
#3192


pchiusano (approved):
Paired with @tstat on this, looks great!

ChrisPenner (approved):
It would be nice to use the user's actual handle in the message, but that's an extra thing to wire up and it's not easily accessible in UCM right now, so this is good 👍🏼

PR #3236, opened 2022-07-14
Ignore things with lib in their name in docs.to-html

pchiusano:
`docs.to-html` now ignores stuff in `lib`. With this change I was able to run it successfully on the whole website codebase.

image

image

We might want to do something fancier with this command for later, but since it's mostly for our internal use I think this is fine as a quick thing.


hojberg (approved):

rlmark (approved):
tested locally works great

PR #3237, opened 2022-07-14
add PRAGMA busy_timeout = 1000 to sqlite connections

aryairani:
We were getting HTTP 500 errors from enlil during concurrent pull, possibly due to SQLITE_BUSY during WAL cleanup conflicting with an assumption of `runReadOnlyTransaction` that it will not encounter SQLITE_BUSY except in cases of programmer error.

Sqlite lets you install a simple timeout for SQLITE_BUSY with https://www.sqlite.org/pragma.html#pragma_busy_timeout, which we install here on every connection.

Travis and I verified (with a small test) that the busy handler is not invoked if SQLITE_BUSY is returned due to concurrent writers. So, it seems that we can set it on connection acquisition without worrying that it introduces an unnecessary delay for concurrent writers.

### DIFF
--- a/lib/unison-sqlite/src/Unison/Sqlite/Connection.hs
+++ b/lib/unison-sqlite/src/Unison/Sqlite/Connection.hs
@@ -119,6 +119,7 @@ openConnection name file = do
   conn0 <- Sqlite.open file `catch` rethrowAsSqliteConnectException name file
   let conn = Connection {conn = conn0, file, name}
   execute_ conn "PRAGMA foreign_keys = ON"
+  execute_ conn "PRAGMA busy_timeout = 1000"
   pure conn
 
 -- Close a connection opened with 'openConnection'.


tstat (approved):

pchiusano (approved):

mitchellwrosen:
Just to emphasize a point above I found interesting - installing a "busy handler" does not mean any SQLITE_BUSY your application may have seen will instead be handled internally by a sleep in SQLite.

It's smarter, and will deliver SQLITE_BUSY to the application anyway, if it either detects a potential deadlock, or (much more commonly) a write transaction can't be made serializable due to an intervening write. (This last one is actually a SQLITE_BUSY_SNAPSHOT error, if you have extended error codes enabled).

So, it seems pretty good... I'm surprised the documentation doesn't emphasize its usage more.

PR #3238, opened 2022-07-14
Fix constructor name for records nested in a subnamespace

### LINKED ISSUE #2628: record constructor duplicates namespace
## current behavior

If you declare a record type like `unique type foo.bar.baz.MyRecord = { value : Nat }`, then you'll end up with a type `foo.bar.baz.MyRecord`, and a constructor `foo.bar.baz.MyRecord.foo.bar.baz.MyRecord`.

## expectation

I'd expect the constructor name to be `foo.bar.baz.MyRecord.MyRecord`.

## transcript

Below is the transcript output of [this raw transcript](https://gist.github.com/ceedubs/1b92a69d718023568c98b8a48ebf546c).

```unison
unique type foo.bar.baz.MyRecord = {
  value : Nat
}
```

```ucm
.> add

    unique type foo.bar.baz.MyRecord
    foo.bar.baz.MyRecord.value        : MyRecord -> Nat
    foo.bar.baz.MyRecord.value.modify : (Nat ->{g} Nat)
                                        -> MyRecord
                                        ->{g} MyRecord
    foo.bar.baz.MyRecord.value.set    : Nat
                                        -> MyRecord
                                        -> MyRecord

.> find : Nat -> MyRecord

  1. foo.bar.baz.MyRecord.foo.bar.baz.MyRecord : Nat -> MyRecord
  

```

## other notes

This doesn't happen with non-record types.

The `add` output doesn't include the constructor name, so it's easy to miss that this happened. This is probably reasonable behavior, because the output could be pretty verbose for a union type with many constructors.

pchiusano:
See #2628

### DIFF
--- a/parser-typechecker/src/Unison/FileParser.hs
+++ b/parser-typechecker/src/Unison/FileParser.hs
@@ -346,7 +346,8 @@ dataDeclaration mod = do
           sepBy1 (reserved "," <* optional semi) $
             liftA2 (,) (prefixVar <* reserved ":") TypeParser.valueType
         _ <- closeBlock
-        pure ([go name (snd <$> fields)], [(name, fields)])
+        let lastSegment = name <&> (\v -> Var.named (Name.toText $ Name.unqualified (Name.unsafeFromVar v)))
+        pure ([go lastSegment (snd <$> fields)], [(name, fields)])
   (constructors, accessors) <-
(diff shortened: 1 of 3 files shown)


aryairani:
Oh wow. Thanks @ceedubs and @pchiusano.

aryairani (approved):
👍

Issue #3253, opened 2022-07-18
Simplify `Action` monad


tstat:
The `Action` monad contains a free monad that we are not benefiting from. We only have a single interpretation, and we have a `MonadUnliftIO` constraint on our free monad, which limits the possible interpretations to readers over IO.

In order to reduce the complexity, we should replace the free monad with a concrete monad that provides the effects we want.

PR #3255, opened 2022-07-19
monomorphize Action/Command to IO

mitchellwrosen:
## Overview

This PR removes the `m` type variable from `Action`, `Command`, and related types, as part of the effort to eliminate `Command` and simplify the "handle input" loop. See #3253.


pchiusano (approved):
Nice!

ChrisPenner (approved):

PR #3262, opened 2022-07-21
delete unused root branch watching code

mitchellwrosen:
## Overview

This PR deletes some unused root branch watching code. We may want to implement this feature some day, if we still have in-memory state when we decide that running multiple simultaneous UCMs should work, but it's not clear that any of the existing stubs are really worth keeping around for when/if that happens. Signed off by @aryairani


pchiusano (approved):

ChrisPenner (approved):
Agreed that it's better to just remove than have a deceptively broken combinator around 👍🏼

aryairani (approved):

PR #3264, opened 2022-07-21
Simplify Command and HandleCommand

### LINKED ISSUE #3253: Simplify `Action` monad

tstat:
## Overview
Closes #3253 

As the issue notes, the `Action` monad is a free monad, but we don't benefit from this by rewriting syntax or have multiple interpreters. So, we are paying a complexity cost without reaping any benefits. Our intention is to (eventually) remove `Action` entirely, replacing it with a concrete monad that provides the effects we want. This PR does not remove `Action`, but greatly simplifies it. We see this PR as a stepping stone that can help break up the work. 

We also add a couple of effects that we think will be useful: `with` and `reset` to provide nicer syntax for scoping resource acquisition, and an effect to short-circuit the processing of a single input.

## Implementation notes

We remove most of the effect handlers and interpret to `Cli` instead of `IO` (`Cli` is our concrete monad candidate).

## Test coverage

We rely upon the existing tests.

## Loose ends

A follow-up PR will remove `Action` entirely, replacing it with `Cli`.


pchiusano (approved):
yay!

mitchellwrosen:
The transcript change is (now understood to be) expected. We accidentally fixed a bug in the transcript runner. Before, this transcript would print "I had nothing to do".

```
unique type A = A

.> add

unique type A = A

.> update
```

Now, it generates a new UUID for the unique type, just like `ucm` does.

aryairani (approved):

Issue #3166, opened 2022-06-30
"marshalToForeign: unhandled closure" on mutually recursive ability handler


runarorama:
The runtime crashes on the following (using `Stream` from Base):

```
Stream.map : (a -> b) -> '{Stream a} r -> '{Stream b} r
Stream.map f stream = handle !stream with Stream.map.handler f

Stream.map.handler : (a -> b) -> Request {Stream a} r -> '{Stream b} r
Stream.map.handler f = cases 
  {Stream.emit a -> resume} -> 'let
    Stream.emit (f a)
    !(Stream.map f resume)
  {r} -> 'r

> !(Stream.toList (Stream.map increment (Stream.fromList [1,2,3])))
```

The crash looks like:

```
Encountered exception:
marshalToForeign: unhandled closure: DataU1 ##Nat 1245184 0 CallStack ( from HasCallStack ) :
   error
```

pchiusano:
Interestingly, no crash here:

```
Stream.map : (a -> b) -> '{Stream a} r -> '{Stream b} r
Stream.map f stream = handle !stream with Stream.map.handler f

Stream.map.handler : (a -> b) -> Request {Stream a} r -> '{Stream b} r
Stream.map.handler f = cases 
  {r} -> 'r
  {Stream.emit a -> resume} -> do 
    Stream.emit (f a)
    Stream.map f resume ()

> let
    s1 = do emit 10 
    s2 = Stream.map (a -> a * 10) s1
    Stream.toList! s2 
```

dolio:
I think I'm finally about done fixing this. Should have a PR soon.

PR #3266, opened 2022-07-21
Fix issues with continuations and pending arguments

dolio:
There were multiple problems with how the continuation operations worked related to over-applied functions. In either case, it would result in said arguments getting effectively removed from the stack, and the functions that were supposed to consume them would be under applied.

- When capturing a continuation, the pending arguments are captured.  However, how many there were was not recorded in the continuation.  When restoring the continuation, the stack should have as many pending  arguments as were originally captured, but we were just setting the  pending arguments to 0.
  * The number of arguments can be calculated from the captured  continuation, but storing the numbers is likely faster. The numbers are only discared and re-computed when moving to the serialized representation, which allows the storage format to remain the same.
- Similarly, it is possible to _restore_ a continuation while arguments are pending, in which case the restored continuation/stack as to account for that. The previous code wasn't.
- Also, since pending arguments are part of the continuation, Reset should block the capture of those arguments. Instead, the previous code was just leaving them in the pending position. The new code for Reset modifies the stack pointers, and records how many arguments  are pending in the continuation mark.
  * This information is not redundant, so the serialization format had to be modified. I've incremented the version and added support for parsing the old version, although those values could actually be erroneous.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/ANF.hs
+++ b/parser-typechecker/src/Unison/Runtime/ANF.hs
@@ -1016,7 +1016,7 @@ data Value
 
 data Cont
   = KE
-  | Mark [Reference] (Map Reference Value) Cont
+  | Mark Word64 Word64 [Reference] (Map Reference Value) Cont
   | Push Word64 Word64 Word64 Word64 GroupRef Cont
   deriving (Show)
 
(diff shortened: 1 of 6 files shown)


dolio:
Fixes #3166

pchiusano (approved):
Nice!

PR #3269, opened 2022-07-22
Fix a variable handling bug in the pattern compiler

### LINKED ISSUE #3244: "guard variables do not match body" does not include line number
I got started on a practice problem and ran into a compiler error.  It's not clear why the error is occurring or what line number caused the error.
 
```
use Either Right Left

unique type Error = Error Text

unique type FrameCell = SingleRoll Nat | Spare Nat | Strike | Empty | Skip
unique type FrameStatus = NotStarted | Scored Nat | PendingScore | PendingRolls

unique type Frame = {
  frameNumber: Nat,
  cell1: FrameCell,
  cell2: FrameCell,
  bonus1: FrameCell,
  status: FrameStatus
}

-- WIP
parseFrames : [Nat] -> Either Error [Frame]
parseFrames rolls =
  go t =
    (frame, _) = t
    create c1 c2 c3 = Frame frame c1 c2 c3 PendingScore

    match t with
      (frame, 10 +: xs) -> (create Strike Skip Skip) List.+: (go (frame + 1, xs))
      (frame, [a, b] ++ xs)
        | a + b == 10 -> (create (SingleRoll a) (Spare b) Skip) List.+: (go (frame + 1, xs))
        | a + b < 10 -> (create (SingleRoll a) (SingleRoll b) Skip) List.+: (go (frame + 1, xs))
        | otherwise -> throw (Error "Bad roll")
      _ -> throw (Error "TODO")

  toEither '(go (1, rolls))

> parseFrames [6,4,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```

dolio:
The pattern compiler was assuming that the variables bound in a guard always match those bound in the rest of the branch, but apparently that isn't the case. This fixes the case where they at least have the same length. I'm not sure what it would mean for them to _not_ have the same length, so an error is still reported in those cases.

Fixes #3244

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Pattern.hs
+++ b/parser-typechecker/src/Unison/Runtime/Pattern.hs
@@ -698,6 +698,8 @@ mkRow sv (MatchCase (normalizeSeqP -> p0) g0 (AbsN' vs b)) =
     g = case g0 of
       Just (AbsN' us g)
         | us == vs -> Just g
+        | length us == length vs ->
+            Just $ renames (Map.fromList (zip us vs)) g
         | otherwise ->
             internalBug "mkRow: guard variables do not match body"
       Nothing -> Nothing
--- a/unison-src/transcripts/fix3244.md
+++ b/unison-src/transcripts/fix3244.md
@@ -1,0 +1,21 @@
+```ucm:hide
+.> builtins.merge
+```
+
+This tests an previously erroneous case in the pattern compiler. It was assuming
+that the variables bound in a guard matched the variables bound in the rest of
+the branch exactly, but apparently this needn't be the case.
+
+```unison
+
+foo t =
+  (x, _) = t
+  f w = w + x
+
+  match t with
+    (x, y)
+      | y < 5 -> f x
+      | otherwise -> x + y
+
+> foo (10,20)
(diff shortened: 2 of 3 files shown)


pchiusano:
My understanding was that the parser will produce an AST that bind the same variables in the guard as in the body, and it binds them in the same order. It looks from the code in `TermParser.hs` like this is the case (just search for `MatchCase`). So given that, is it possible there's some other bug that this is covering up in the pattern compiler?

image

image

pchiusano (approved):

dolio:
It's actually not a bug. Not exactly, at least. It's caused by a previous compilation step. In the example:

```
foo t =
  (x, _) = t
  f w = w + x

  match t with
    (x, y)
      | y < 5 -> f x
      | otherwise -> x + y
```

`f` get floated, so it becomes:

```
f x w = w + x
```

This means that the body of the first case becomes:

```
   f x x'
```

But the outer `x` doesn't occur in the guard, so it doesn't get freshened.

The freshening causes some weird output when decompiled, but if I successfully get the decompile info earlier, it won't make a difference, I think.

pchiusano:
cool

PR #3279, opened 2022-07-29
add some windows build tips to development.md

aryairani:
If we could capture the actual symlink error here, that would be good, but I couldn't bring myself to deconfigure my VM to get it.

### DIFF
--- a/development.markdown
+++ b/development.markdown
@@ -76,6 +76,9 @@ its location on the command line.
 This codebase uses symlinks as a workaround for some inconveniences in the `here` package. Support for symlinks in Windows is relatively new, and isn't enabled by default. As a result, your cloned copy of the code probably won't build.
 
 First you'll need to enable "Developer Mode" in your Windows settings.
+
+	See https://consumer.huawei.com/en/support/content/en-us15594140/
+
 Then you'll need to enable symlink support in your `git` configuration, e.g.
 
     `git config core.symlinks true`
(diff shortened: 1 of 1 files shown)


development.markdown
```diff
@@ -68,3 +68,23 @@ its location on the command line.
     `cabal v2-install --project-file=contrib/cabal.project unison`
 
 * The install directory can be modified with the option `--installdir: ...`
+
+## Building on Windows
+
+### I get an error about unison/sql/something
+
+This codebase uses symlinks as a workaround for some inconveniences in the `here` package. Support for symlinks in Windows is relatively new, and isn't enabled by default. As a result, your cloned copy of the code probably won't work.
```
aryairani:
```suggestion
This codebase uses symlinks as a workaround for some inconveniences in the `here` package. Support for symlinks in Windows is relatively new, and isn't enabled by default. As a result, your cloned copy of the code probably won't build.
```

development.markdown
```diff
@@ -68,3 +68,23 @@ its location on the command line.
...
 
 * The install directory can be modified with the option `--installdir: ...`
+
+## Building on Windows
+
+### I get an error about unison/sql/something
+
+This codebase uses symlinks as a workaround for some inconveniences in the `here` package. Support for symlinks in Windows is relatively new, and isn't enabled by default. As a result, your cloned copy of the code probably won't build.
+
+First you'll need to enable "Developer Mode" in your Windows settings.
```
aryairani, line 78:
```suggestion
First you'll need to enable "Developer Mode" in your Windows settings.

	See https://consumer.huawei.com/en/support/content/en-us15594140/

```

pchiusano (approved):

PR #3284, opened 2022-08-01
Fix erroneous 'compiled' runtime state

### LINKED ISSUE #3280: Universal equality fails across ucm sessions
If you serialize a value and then deserialize it in another UCM session, univseral equality can produce false negatives.

Talking to @dolio, this seems to be that the constructor tags are currently serialized, but are dependent on the order in which types are loaded into the code cache.

Below are two transcripts and a shell command to run to reproduce the error.

## save this as serialize.md

````
```ucm
.> alias.type ##Nat Nat
.> alias.type ##Sequence List
```

```unison
unique type Foo = Foo Nat
```

```ucm:hide
.> add
```

```unison
structural type Bar = Bar (List Nat)
```

```ucm:hide
.> add
.> pull unison.public.base.releases.M4 lib.base
```

```unison
serializeToFile : '{Exception, IO} ()
serializeToFile =
  valueToSerialize = Bar [1, 2, 3]
  Debug.trace "valueToSerialize" valueToSerialize
  writeToFile file = valueToSerialize |> Value.value |> Value.serialize |> putBytes file
  do bracket '(openFile (FilePath "serialized-value") Write) closeFile writeToFile
```

```ucm:hide
.> add
.> compile serializeToFile serializeToFile
```
````

## save this as deserialize.md

````
```ucm:hide
.> alias.type ##Nat Nat
.> alias.type ##Sequence List
```

```unison
structural type Foo = Foo (List Nat)
```

```ucm:hide
.> add
.> pull unison.public.base.releases.M4 lib.base
```

```unison
deserializeFromFile : '{Exception, IO} ()
deserializeFromFile =
  deserialize file =
    bytes = getBytes file 10000000
    match Value.deserialize bytes with
      Right v -> match Value.load v with
        Right v -> v
  do
    result = bracket '(openFile (FilePath "serialized-value") Read) closeFile deserialize
    Debug.trace "deserialized value" result
    if result === Foo [1, 2, 3] then () else bug "they are not equal!"
```

```ucm:hide
.> add
.> compile deserializeFromFile deserializeFromFile
```
````

## then run this

```sh
ucm transcript serialize.md && ucm run.compiled serializeToFile.uc && ucm transcript deserialize.md && ucm run.compiled deserializeFromFile.uc
```

Among the output you should see something like:

```
trace: valueToSerialize
DataB1 #seb0f 3866624 (Foreign (Wrap ##Sequence _))

trace: deserialized value
DataB1 #seb0f 3932160 (Foreign (Wrap ##Sequence _))

💔💥

I've encountered a call to builtin.bug with the
following value:

  "they are not equal!"
```

dolio:
When generating a saved runtime state, we would try to serialize all the
type information needed. However, we were actually filtering by the
packed tags instead of the type number. The packed tags have 48 bits of
type information followed by 16 bits of constructor tag, so the type
numbering would never actually match.

Fixes #3280

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/MCode.hs
+++ b/parser-typechecker/src/Unison/Runtime/MCode.hs
@@ -37,7 +37,7 @@ where
 
 import Control.Applicative (liftA2)
 import Data.Bifunctor (bimap, first)
-import Data.Bits (shiftL, (.|.))
+import Data.Bits (shiftL, shiftR, (.|.))
 import Data.Coerce
 import Data.List (partition)
 import qualified Data.Map.Strict as M
@@ -1407,7 +1407,7 @@ sectionTypes (Match _ br) = branchTypes br
 sectionTypes _ = []
 
 instrTypes :: Instr -> [Word64]
-instrTypes (Pack _ w _) = [w]
+instrTypes (Pack _ w _) = [w `shiftR` 16]
 instrTypes _ = []
 
 branchDeps :: Branch -> [Word64]


pchiusano:
Could we add the transcript test that @ceedubs put together?

dolio:
I don't think so. The transcripts produce two compiled outputs that have to each be run separately in sequence.

dolio:
@ceedubs To see this fix to 3280, you need to re-generate the .uc file for the program that's going to load the remote values. It will then contain types it _should_ have known about (because they're used in the saved code), but weren't actually being saved. The mismatch was caused by the remote code re-introducing those types with new numbers, but the saved code using the old, forgotten numbers.

pchiusano:
Okay, word.

pchiusano (approved):

pchiusano:
@ceedubs can you confirm this fixes things for you, then merge if so?

ceedubs (approved):
👍 works for me.

I'd love if we could get CI to run these tests, because:

1. This fails in a really subtle way and made me go mad for hours.
2. It took me quite a while to automate a way to reproduce this, so I hate for it to go to waste :)

Is it possible to add some lines to https://github.com/unisonweb/unison/blob/trunk/.github/workflows/ci.yaml#L157 to catch this? We'd probably want to pull base in a different way and might want to change the transcripts to put the output file in a `tmp` directory.

pchiusano:
Going to merge and created https://github.com/unisonweb/unison/issues/3285 to track

PR #3296, opened 2022-08-04
Float local function definitions to their own reference during compilation

dolio:
This PR implements a simple method of floating local function definitions to proper top-level combinators during compilation. Essentially it runs the old floater, then calls a function for hashing the letrec. This should make serialization and sandboxing more fine grained. Since a local pure function gets turned into its own reference, it shouldn't be flagged as sandboxed just because the enclosing function uses some I/O.

@aryairani Can you (or someone else who would know) check that I used the right hashing function? I used `hashTermComponentsWithoutTypes`, which seems like it's the only analogue of something I was using in the past for this sort of thing. However, it seems like an odd module for that functionality to be (something about converting between versions).

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/ANF.hs
+++ b/parser-typechecker/src/Unison/Runtime/ANF.hs
@@ -353,16 +353,18 @@ float ::
   Term v a ->
   (Term v a, [(Reference, Term v a)], [(Reference, Term v a)])
 float tm = case runState go0 (Set.empty, [], []) of
-  (bd, (_, ctx, dcmp)) -> let
-      m = hashTermComponentsWithoutTypes . Map.fromList $ fmap deannotate <$> ctx
-      trips = Map.toList m
-      f (v, (id, tm)) = ((v, id), (v, idtm), (id, tm))
-        where idtm = ref (ABT.annotation tm) (DerivedId id)
-      (subvs, subs, tops) = unzip3 $ map f trips
-      subm = Map.fromList subvs
-    in ( letRec' True [] . ABT.substs subs . deannotate $ bd
-       , fmap (first DerivedId) tops
-       , dcmp <&> \(v, tm) -> (DerivedId $ subm Map.! v, tm))
+  (bd, (_, ctx, dcmp)) ->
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
Nice.

Will wait for @aryairani on whether that's the right hashing function.

ceedubs:
Will wait for Arya's input, but I've verified that this prevents the sandbox validation false positives when inlining `'{Remote} a` programs!

aryairani:
@dolio TL;DR: This seems reasonable, although I'd be curious what the lifecycle of the hashes you're getting out is.

Basically the hashing functions operate on a dedicated copy of the data type definition for terms, that only contain the data that gets incorporated into the hash. It was a way to decouple from the typechecker representation and hopefully have fewer hashing snafus down the line.

The `Unison.Hashing.V2.Convert` module converts between the typechecker representation and the hashing representation, and in the case of `hashTermComponentsWithoutTypes`, records some info that's otherwise lost in conversion to the hashing representation, filling it back into the typechecker-representation terms that are returned.

If the the runtime or post-typechecking representation of terms also change in the future, we could have a separate conversion function from that to the hashing representation.  I'll also note that depending on the use case, you may not care to use the same hashing function as the codebase uses.

Lastly here's the Haddock from `hashTermComponentsWithoutTypes`, which I'm guessing is fine for your use case?
> This shouldn't be used when storing terms in the codebase, as it doesn't incorporate the type into the hash.
> this should only be used in cases where you just need a way to identify some terms that you have, but won't be
> saving them.

dolio:
Yeah, I read the docs. What this code does is split out all local lambdas into separate definitions and refer to them by hash. But it's only a runtime thing. The hashes are used because the runtime references are based on the same hashing as for the codebase, but none of these are going to be added to the codebase, so it seemed fine.

Mainly I wasn't sure if I was using something that was involved in translating between codebase versions or something. But if the 'Convert' part is about converting between two currently used internal formats, that makes sense.

aryairani:
> But if the 'Convert' part is about converting between two currently used internal formats, that makes sense.

Yeah, exactly.

aryairani (approved):

PR #3301, opened 2022-08-08
Missed Name instructions as a source of combinator deps

dolio:
When making a standalone file, we trace through the combinators to only serialize only the ones that are necessary to run the program. However, the code was missing `Name` as a source of such references.

Previously this never mattered (I think). `Name` is used in compiling abilities. Essentially `handle e with h` turns into something similar to `name x = e in h x`, but `e` was very likely to just be referencing a separate section of the same definition as the `handle`. However, the new lambda lifting can now float this to a completely separate combinator that gets missed when making a standalone file.

Probably fixes a bug that @ceedubs was encountering.


ceedubs (approved):
Thanks for the quick turnaround, @dolio. This seems to resolve the issue!

PR #3303, opened 2022-08-09
Fix serialization issue with code that uses abilities

ceedubs:
We were using session-specific constructor tags to identify abilities,
but after 4918b2ef272884f96c76c9f408f3e77595de0ad8 that information was
no longer being populated. This should change it so we properly pack the
tags.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/MCode.hs
+++ b/parser-typechecker/src/Unison/Runtime/MCode.hs
@@ -918,12 +918,16 @@ emitFunction rns _ _ _ (FCon r t) as =
     rt = toEnum . fromIntegral $ dnum rns r
 emitFunction rns _ _ _ (FReq r e) as =
   -- Currently implementing packed calling convention for abilities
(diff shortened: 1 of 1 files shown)


dolio:
To elaborate a little more...

Normal data gets constructed with tags like `<rtag><ctag>`, where `rtag` is a numbering of type references, and `ctag` is the constructor tag. Ability requests use the same data constructor machinery, but were using just `<rtag>`.

When I did #3284, I didn't notice this. The old code actually works correctly for abilities, where the whole 64-bit word corresponds to the reference numbering. So, by fixing the data case, it broke the ability case. The solution is to make abilities use `<rtag><rtag>` tags, so that the type information is in the expected spot, and the 'constructor' tag is also the reference number, because the ability handler is doing cases on the constructor tag to determine which ability is being requested.

This only affects local runtime state (and therefore standalone binaries), so remote serialization should already be fine.

dolio (approved):

PR #3317, opened 2022-08-12
Improve error messages for ability equality errors

### LINKED ISSUE #3300: Exception with no useful information when using types with curly braces
```
structural type Row c = Row

check : Row {Boolean} -> ()
check row = ()

value : Row {Text, Boolean}
value = Row

> check value
```
```
.alvaro.public> 
Encountered exception:
unpossible
CallStack ( from HasCallStack ) :
   error
```

i have no idea if the code should type-check or not, but the error should provide more details

dolio:
This changes the error generated when failing an ability equality check, and detects it for some better error messages.

The messages try to figure out when an ability equality error arose from applying functions, so it can indicate the function and argument involved, together with some more type context available. In other situations, an error about abilities/types not being equal is generated, rather than the old message which was just reporting as if there was an ambient-vs-needed error, which doesn't make much sense in these scenarios.

This also fixes #3300, which was caused by the old error message not expecting the "needed" abilities in a failure to ever be empty, while an equality failure can be due to an empty row not being equal to a non-empty row, and the choice of what goes in the 'needed' spot being arbitrary for an equality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3317)

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -396,49 +396,49 @@ renderTypeError e env src curPath = case e of
   AbilityCheckFailure {..}
     | [tv@(Type.Var' ev)] <- ambient,
       ev `Set.member` foldMap Type.freeVars requested ->
-      mconcat
-        [ "I tried to infer a cyclic ability.",
-          "\n\n",
-          "The expression ",
-          describeStyle ErrorSite,
-          " was inferred to require the ",
-          case length requested of
-            1 -> "ability: "
-            _ -> "abilities: ",
-          "\n\n    {",
-          commas (renderType' env) requested,
-          "}",
-          "\n\n",
-          "where `",
-          renderType' env tv,
-          "` is its overall abilities.",
-          "\n\n",
-          "I need a type signature to help figure this out.",
-          "\n\n",
-          annotatedAsErrorSite src abilityCheckFailureSite,
-          debugSummary note
-        ]
+        mconcat
+          [ "I tried to infer a cyclic ability.",
+            "\n\n",
+            "The expression ",
+            describeStyle ErrorSite,
+            " was inferred to require the ",
+            case length requested of
+              1 -> "ability: "
+              _ -> "abilities: ",
+            "\n\n    {",
+            commas (renderType' env) requested,
+            "}",
+            "\n\n",
+            "where `",
+            renderType' env tv,
+            "` is its overall abilities.",
+            "\n\n",
+            "I need a type signature to help figure this out.",
+            "\n\n",
+            annotatedAsErrorSite src abilityCheckFailureSite,
+            debugSummary note
+          ]
   AbilityCheckFailure {..}
     | C.InSubtype {} :<| _ <- C.path note ->
-      mconcat
-        [ "The expression ",
-          describeStyle ErrorSite,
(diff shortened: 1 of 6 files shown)


aryairani:
@dolio Who would you like to review this?

dolio:
I'm not sure. Is anyone familiar with the error context parsing stuff? I was kind of cobbling together stuff while only partially understanding it on that part.

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -467,6 +467,79 @@ renderTypeError e env src curPath = case e of
...
         debugSummary note
       ]
+  AbilityEqFailure {..} ->
+    mconcat
+      [ "I could not solve an ability equation when checking the expression ",
+        describeStyle ErrorSite,
+        "\n\n",
+        Pr.wrap $
+          mconcat
+            [ "When trying to match ",
```
aryairani, line 484:
```suggestion
            [ "while I was trying to match ",
```

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -467,6 +467,79 @@ renderTypeError e env src curPath = case e of
         annotatedAsErrorSite src abilityCheckFailureSite,
         debugSummary note
       ]
+  AbilityEqFailure {..} ->
+    mconcat
+      [ "I could not solve an ability equation when checking the expression ",
```
aryairani: Is there a user-friendlier description we could use here? Or do we like this terminology?

aryairani:
@dolio By "error context parsing stuff", do you mean like in `parser-typechecker/src/Unison/Typechecker/Extractor.hs` etc.?  I wrote a bunch of that a long time ago, but it sounds completely foreign now. (@pchiusano may have some ideas?)

I'd say we can merge this much if you're happy with it, and then still come back and change terminology / formatting as needed to make it newbie-friendly.

aryairani (approved):

pchiusano:
@dolio can you change to this wording:

```
I found an abilities mismatch here:

  <code excerpt>

<detailed explanation you already have>
```

Otherwise looks good.

parser-typechecker/src/Unison/PrintError.hs
pchiusano: Yeah, I'd just do: "I found an abilities mismatch here:"

dolio:
How do the messages look now? You can't see the colors in the transcripts, but does the text look okay?

parser-typechecker/src/Unison/PrintError.hs
pchiusano: am neutral on this wording change, going to leave it as is

pchiusano:
Yup, thanks @dolio! Applied ready-to-merge.

parser-typechecker/src/Unison/PrintError.hs
pchiusano: Elaborating a bit, the start of the sentence already establishes the compiler as talking in first person, which is the style we like for our messages. But no need to insert first person in every clause of the sentence and in this particular case I'm not sure it reads any better.

"I found a mismatch here... while I was trying to..." 
vs 
"I found a mismatch here... when trying to..."

Issue #2049, opened 2021-06-01
Make sure it's possible to catch all exceptions


pchiusano:
We want to be able to implement `bracket` and `finally`. There's exceptions in pure code, due to calling `bug` or `todo` or an incomplete pattern match, and then also async exceptions due to a thread being killed.

Needs some more thought about how to handle. I think it could be good if all these things could be caught by an `Exception` handler, where:

```
ability Exception where raise : Failure -> x
```

@dolio let's discuss this

pchiusano:
@dolio so my proposal is this - `bug` and `todo` should behave like this:

```
unique type Bug =
unique type Todo =

bug : x -> y
bug x = 
  f x = Exception.raise (Failure (typeLink Bug)) "bug" (Any x)
  coerceAbilities f x

todo : x -> y
todo x = 
  f x = Exception.raise (Failure (typeLink Todo)) "todo" (Any x)
  coerceAbilities f x
```

Now these can be caught in `Exception`.

dolio:
This would work with the current ability implementation. But if we ever switch to a system where ability usage corresponds to under-the-hood information passing, then these `coerceAbilities` don't actually make any sense. They'll just produce invalid code.

So, it seems like this walks us more toward not fixing the ability soundness issues.

pchiusano:
What we could do in that situation is just have `Exception.catch` be defined to also catch the exception thrown by `bug` and `todo`. 

But `Exception.catch` would be compiled to a builtin that gets selected when compiling a definition with the same hash as the current [`Exception.catch`](https://share.unison-lang.org/latest/namespaces/unison/base/;/terms/Exception/catch). This wouldn't box us in re: ability soundness changes.

We've talked about being able to do this sort of thing more generally. Like we might compile some pure Unison code to a more efficient builtin, but then the pure Unison source of that definition is still available to look at so people can understand what the code does.

ceedubs:
You've mentioned `bug` and `todo`. Does the scope of this issue also include things like interrupts? Something [that has come up](https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1638803980221100) is a `main` method that should perform some cleanup (connections, file handles, etc) when it is stopped via ctrl-c.

I'll paste some of the Slack discussion in case it goes away.

For example if you run the following program and hit ctrl-c, you'll never see "close".

```
main : '{Exception, IO} ()
main _ =
  open : '{Exception, IO} ()
  open _ = printLine "open"
[...]

PR #3365, opened 2022-08-25
Implement a primop for catching runtime exceptions

dolio:
This PR implements a new primitive operation:

    tryEval : '{IO} a ->{IO, Exception} a

It runs the specified `IO` operation, but catches all internal runtime exceptions and rethrows them using the `Exception` ability. They can then be caught by using an ordinary ability handler.

Addresses #2049

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3365)


pchiusano (approved):
LGTM!

PR #3372, opened 2022-08-29
Fix a bug in the `tryEval` wrapper code

dolio:
It was wrapping the successful result in a `Some` by mistake

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3372)


ceedubs (approved):
@dolio thanks! This seems to resolve the issue.

Would it be easy to add a test-case for the non-failure case to the fix2049.md transcript?

ceedubs (approved):
🎉

ceedubs:
I think that this is ready to merge once the CI build passes

PR #3374, opened 2022-08-30
Fix a bug in the argument handling of the runtime

### LINKED ISSUE #2409: runtime crash: applying non-function: Enum ##Boolean 1
I tried for a while, but couldn't figure out how to create a smaller reproduction of this issue.

[Here](https://gist.github.com/ceedubs/ce02f577dc60c7857cac03d3c24527f1) is the raw transcript. Below is the result of running it.

## background: data structures and functions

```unison
{{
A data structure that holds the arguments that you would pass to foldLeft.

The one difference is that there is an extra "extract" function at the end that is to be run after
the foldLeft completes. This can be useful when the fold needs to maintain some state during the
fold but you don't want to include it in the final result.
}}
unique type Fold' g a b x = Fold' (x -> {g} a -> {g} x) x (x -> {g} b)

{{
This is essentially a {Fold'} where the `x` type parameter is an existential type.

Because Unison doesn't directly support existential types, there is a bit of noise required to
craft them from universal type parameters.
}}
unique type Fold g a b = Fold (∀ g2 r. (∀ x. Fold' g a b x -> {g2} r) -> {g2} r)

Fold.fromFold' : Fold' g a b x -> Fold g a b
Fold.fromFold' fold = Fold.Fold (f -> f fold)

{{ Helper function for creating a {type Fold} without manually constructing a {type Fold'}}}
Fold.mkFold : (t -> {g} a -> {g} t) -> t -> (t -> {g} b) -> Fold g a b
Fold.mkFold step init extract =
  Fold.fromFold' (Fold'.Fold' step init extract)

{{ Check whether the predicate holds for all elements }}
folds.all : (a -> {g} Boolean) -> Fold g a Boolean
folds.all predicate =
  Fold.mkFold (b -> a -> b && (predicate a)) true id

{{ Run a fold on a {type .base.Stream}}}
Fold.Stream.fold : Fold g a b -> '{g, Stream a} r -> '{g} b
Fold.Stream.fold =
  run: Fold' g a b x -> '{g, Stream a} r -> '{g} b
  run =
    cases Fold'.Fold' step init extract -> stream -> _ -> extract !(.base.Stream.fold step init stream)
  cases
    Fold f -> stream -> f (f' -> run f' stream)
```

```ucm

  change:
  
      unique type Fold g a b
      unique type Fold' g a b x
      Fold'.doc            : base.Doc
      Fold.Stream.fold     : Fold g a b
                             -> '{g, Stream a} r
                             -> '{g} b
      Fold.Stream.fold.doc : base.Doc
      Fold.doc             : base.Doc
      Fold.fromFold'       : Fold' g a b x -> Fold g a b
      Fold.mkFold          : (t ->{g} a ->{g} t)
                             -> t
                             -> (t ->{g} b)
                             -> Fold g a b
      Fold.mkFold.doc      : base.Doc
      folds.all            : (a ->{g} Boolean)
                             -> Fold g a Boolean
      folds.all.doc        : base.Doc

```
## the error

Now the following code compiles, but when run via `test>` results in the runtime error `applying non-function: Enum ##Boolean 1`:

```unison
folds.all.tests.stream =
  pred = n -> (n Nat.> 2)
  res : 'Boolean
  res = Fold.Stream.fold (folds.all pred) (Stream.range 1 5)
  check (!res Universal.== false)
```

```ucm

  change:
  
      folds.all.tests.stream : [Test.Result]

```
## similar code that runs fine

However, similar code using `List` works just fine. So it would seem that the issue arises somewhere with abilities and/or delayed thunks.

```unison
Fold.List.fold : Fold g a b -> [a] -> {g} b
Fold.List.fold =
  run: Fold' g a b x -> [a] -> {g} b
  run =
    cases Fold'.Fold' step init extract -> as -> extract (.base.List.foldLeft step init as)
  cases
    Fold.Fold f -> as -> f (f' -> run f' as)

test> folds.all.tests.list =
  pred = n -> (n Nat.> 2)
  res : Boolean
  res = Fold.List.fold (folds.all pred) [1, 2, 3, 4, 5]
  check (res Universal.== false)
```

```ucm

  change:
  
      Fold.List.fold       : Fold g a b -> [a] ->{g} b
      folds.all.tests.list : [Test.Result]
  
  `>`)... Ctrl+C cancels.

    10 |   pred = n -> (n Nat.> 2)
    
    ✅ Passed : Proved.

```

dolio:
The argument handling was sometimes failing to correctly offset some of the arguments. This would lead to partial applications with arguments missing and shifted, causing subsequent problems.

Fixes #2409

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3374)

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Stack.hs
+++ b/parser-typechecker/src/Unison/Runtime/Stack.hs
@@ -214,7 +214,7 @@ uargOnto stk sp cop cp0 (ArgN v) = do
         | i < 0 = return ()
         | otherwise = do
             (x :: Int) <- readByteArray stk (sp - indexPrimArray v i)
-            writeByteArray buf (sz - 1 - i) x
+            writeByteArray buf (boff - i) x
             loop $ i - 1
   loop $ sz - 1
   when overwrite $
@@ -224,6 +224,7 @@ uargOnto stk sp cop cp0 (ArgN v) = do
     cp = cp0 + sz
     sz = sizeofPrimArray v
     overwrite = sameMutableByteArray stk cop
+    boff | overwrite = sz - 1 | otherwise = cp0 + sz
 uargOnto stk sp cop cp0 (ArgR i l) = do
   moveByteArray cop cbp stk sbp (bytes l)
   pure $ cp0 + l
@@ -255,9 +256,10 @@ bargOnto stk sp cop cp0 (ArgN v) = do
         | i < 0 = return ()
         | otherwise = do
             x <- readArray stk $ sp - indexPrimArray v i
-            writeArray buf (sz - 1 - i) x
+            writeArray buf (boff - i) x
             loop $ i - 1
   loop $ sz - 1
+
   when overwrite $
     copyMutableArray cop (cp0 + 1) buf 0 sz
   pure cp
@@ -265,6 +267,7 @@ bargOnto stk sp cop cp0 (ArgN v) = do
     cp = cp0 + sz
     sz = sizeofPrimArray v
     overwrite = stk == cop
+    boff | overwrite = sz - 1 | otherwise = cp0 + sz
 bargOnto stk sp cop cp0 (ArgR i l) = do
   copyMutableArray cop (cp0 + 1) stk (sp - i - l + 1) l
   pure $ cp0 + l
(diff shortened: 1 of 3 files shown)


aryairani (approved):

ceedubs (approved):
Thanks @dolio!

PR #3377, opened 2022-08-31
Uniformize translation between internal/external closure representations

### LINKED ISSUE #3171: `view` and UCM console output render reversed arguments for nested delayed computations
The following expression 

```
stream10 = Stream.range 0 10
```

In a watch expression and when `view`ed will render with the arguments reversed 

```
 8 | > stream10
          ⧩
          _ -> Stream.range! 10 0
```

See [slack thread from a Unison community member who found this](https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1656618988988889)

The problem is not unique to `Stream` or abilities. If you wrap a regular `List.range` function in the same way, you'll see the issue surface. 

```
List.myRange n m _ = List.rangeClosed n m

> List.myRange 0 10
```

```
    6 | > List.myRange 0 10
          ⧩
          _ -> List.rangeClosed 10 0
```

dolio:
Some previous work fixed cases where internal, backwards data segments were not reversed when producing external representations of closures. However, not all data segments got this treatment; mostly it was data types and unboxed data segments.

This PR fixes the pattern synonyms used for the translation to always reverse when going between internal an external representations, so that the external representations have segments with the values ordered forwards. These synonyms were also used when decompiling, so it fixes problems where arguments to decompiled partial applications were reversed. I think no other issues would actually arise from the old implementation, because it was self-consistent. However, it could be confusing to look at (and possibly stuff like the pure unison port was expecting a different convention).

This technically modifies the serialization format, but the old format could also be seen as a bug, so I haven't incremented the version. I can if that's deemed important, though.

Fixes #3171

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3377)


pchiusano:
This seems good but we are going bump the version number. Reason is that an old client could totally serialize some closures and we don't the server to silently flip arguments around and result in some weird runtime errors.

Leave it up @dolio about whether to try to still read the old version, or just fail fast with a "unsupported old version" error of some sort.

pchiusano (approved):
Approved but hold off on merging until version gets bumped.

aryairani:
Does this mean it's ready to merge now?

PR #3412, opened 2022-09-13
Refactor: pull `update` handler out of `HandleInput.hs`

mitchellwrosen:
## Overview

This PR continues the effort to break up `HandleInput.hs`. First, I extracted the `update` handler to a new module, and then I pulled out all of its dependencies into their own modules and organized them a bit.

One small (quasi-)functional change to call out, since this work started out as the cycle update bugfix: previously, when unhashing a term/decl component, we would create variables that would look like this:

```haskell
refNamed :: Var v => Reference -> v
refNamed ref = Var.named ("ℍ" <> Reference.toText ref)
```

(it's a `User` variable type that begins with a freaky H).

I replaced that with a new constructor of `Var.Type`:

```haskell
data Type
  = User Text
  | UnnamedReference Reference.Id
  | ...
```

It's really the same thing, just more structured and easier to pattern match on. Other than that, this PR only moves code around.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3412)


ChrisPenner (approved):
Cool, I'll trust that it's all just moving things around;

thanks for making it a separate PR, it'll make it much easier to review the new work when it comes around 🙌🏼

PR #3413, opened 2022-09-13
Fix the way builtins were building Failure values

dolio:
The way builtin operations were constructing `Failure` values was wrong in, I believe, every case. Sometimes only 2 fields were being provided, and in the rest the 3rd field was not being wrapped in `Any`. This fixes these problems, along with some refactoring.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3413)


stew (approved):

ceedubs:
@dolio Thanks! This makes my tests pass, but the Windows build failed. I think that it might be a matter of a file op getting a different error message on Windows? https://github.com/unisonweb/unison/actions/runs/3048600314/jobs/4913830616#step:15:4280

ceedubs (approved):
🎉

PR #3420, opened 2022-09-15
io.test alias, completion, and doc fix

### LINKED ISSUE #2677: io.test lies about looking in the most recently typechecked file
```unison
aTest : '{IO, Exception} [Test.Result]
aTest _ =
  printLine "hi"
  check (1 == 2)
```

```
.testzone> 

  I found and typechecked these definitions in ~/code/unison/http.u. If you do an `add` or
  
      aTest : '{IO, Exception} [Test.Result]

.testzone> io.test aTest

  😶
  
  I looked for a function `aTest` in the most recently typechecked file and codebase but couldn't
  find one. It has to have the type:
  
    aTest : '{IO, Exception} [Test.Result]

.testzone> add

    aTest : '{IO, Exception} [Test.Result]


  I added some default metadata.

.testzone> io.test aTest
hi

    New test results:
  
  ✗ aTest   
  
  🚫 1 test(s) failing
  
  Tip: Use view aTest to view the source of a test.

```

### LINKED ISSUE #2678: rename io.test to test.io?
I just noticed that there is an `io.test` feature in ucm. Should it be renamed `test.io` so it's more discoverable when looking doing tab completion with `test`?

ceedubs:
## Overview

This does 3 things for `io.test`:

- Adds a `test.io` alias (resolves #2678)
- Changes the doc to not say that it searches the most recently saved
  scratch file, since it doesn't (resolves #2677)
- Adds term tab completion similar to that of the `run` command.

## Test coverage

Some transcript tests for the tab completion

## Loose ends

At some point we should probably make `io.test` allow you to run it on a term recently saved to the scratch file but not yet added to the codebase. But until then its help shouldn't say that it can do this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3420)

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -2103,13 +2103,12 @@ ioTest :: InputPattern
 ioTest =
   InputPattern
     "io.test"
-    []
+    ["test.io"]
     I.Visible
-    []
+    [(Required, exactDefinitionTermQueryArg)]
     ( P.wrapColumn2
         [ ( "`io.test mytest`",
-            "Runs `!mytest`, where `mytest` is searched for in the most recent"
-              <> "typechecked file, or in the codebase."
+            "Runs `!mytest`, where `mytest` is a test that requires the IO ability and has been added to the codebase."
           )
         ]
     )
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -2108,7 +2108,7 @@ ioTest =
     [(Required, exactDefinitionTermQueryArg)]
     ( P.wrapColumn2
         [ ( "`io.test mytest`",
-            "Runs `!mytest`, where `mytest` is a test that requires the IO ability and has been added to the codebase."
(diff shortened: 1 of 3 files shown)


unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2103,13 +2103,12 @@ ioTest :: InputPattern
...
-    []
+    ["test.io"]
     I.Visible
-    []
+    [(Required, exactDefinitionTermQueryArg)]
     ( P.wrapColumn2
         [ ( "`io.test mytest`",
-            "Runs `!mytest`, where `mytest` is searched for in the most recent"
-              <> "typechecked file, or in the codebase."
+            "Runs `!mytest`, where `mytest` is a test that requires the IO ability and has been added to the codebase."
```
aryairani: Does `mytest` have to require the `IO` ability for `test.io`/`io.test mytest` to work?

aryairani (approved):

unison-cli/src/Unison/CommandLine/InputPatterns.hs
ceedubs: I thought that it did at first, but I guess the problem there was that it wasn't a delayed program. It looks like `io.test` works for things that don't declare `IO` or `Exception`, so this wording isn't technically correct. I'll update it.

pchiusano (approved):

PR #3443, opened 2022-09-22
Update cycle element first draft

mitchellwrosen:
## Overview

This PR adds a working first-stab at the cycle update issue, with lots of missing functionality, which is being tracked in #3445.

See the transcripts for details on what's implemented. In brief, a basic update of some subset of a cycle "just works" ("just works" (i.e. the resulting namespace will have updates to all of the members of the cycle, not just the ones that were explicitly updated by the user):

In the codebase:
```
ping _ = !pong + 1
pong _ = !ping + 2
```

In a scratch file:
```
ping _ = !pong + 3
```

After updating:
```
ping _ = !pong + 3
pong _ = !ping + 2 -- note: not !#horriblehash + 2
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3443)


unison-cli/src/Unison/Cli/TypeCheck.hs
```diff
@@ -0,0 +1,88 @@
+module Unison.Cli.TypeCheck
```
mitchellwrosen, line 1: This stuff was just moved out of `HandleInput.hs`.

unison-core/src/Unison/Names.hs
```diff
@@ -66,7 +66,7 @@ import qualified Unison.LabeledDependency as LD
 import Unison.Name (Name)
```
mitchellwrosen, line 66: Only cosmetic/irrelevant changes in this file. I went through and replaced `Reference` with either `TermReference` or `TypeReference`, where appropriate.

unison-cli/src/Unison/Cli/TypeCheck.hs
ChrisPenner, line 1: Good call extracting these out.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
         & Path.resolve @_ @_ @Path.Absolute currentPath'
         & tShow
 
+getSlurpResultForUpdate :: Set Name -> Cli r (SlurpResult Symbol)
+getSlurpResultForUpdate requestedNames = do
+  (slurp, termRefToNames, nameToTermRefs) <- do
+    slurpCheckNames <- Branch.toNames <$> Cli.getCurrentBranch0
```
ChrisPenner, line 207: I notice this same set of names is being constructed twice, both in the call location and again inside the function; maybe pass them in instead? Building names can sometimes get expensive on large branches or at the root.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+        nameToTermRefs = Names.refTermsNamed slurpCheckNames . Name.unsafeFromVar
+
+    pure (slurp, termRefToNames, nameToTermRefs)
+
+  slurp1 <- do
+    Cli.Env {codebase} <- ask
+    unisonFile0 <- Cli.expectLatestTypecheckedFile
+
+    -- Here, we identify whether there are any "implicit terms", which are terms that should be brought into the latest
+    -- typechecked Unsion file and re-typechecked, which also re-discovers components.
```
ChrisPenner, line 227: What does it mean to `re-discover` components? Is this sentence just saying that we find the components of all things being updated?

EDIT: Ahh, I see, the example clarifies things.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+    -- The user re-defines "ping" in their scratch file thus:
+    --
+    --   ping = wham + 4   (reference = #newping)
+    --
+    -- Note that, pre-update, we had two components [ping,pong] and [wham]. But after the update, since the new `ping`
+    -- refers to the old `wham`, which refers to the old `pong`, which refers to the old `ping`, we really want to end
+    -- up with a single component in the end, [ping,pong,wham].
+
+    -- First, compute an initial slurp, which will identify the initial set of definitions we are updating ({"ping"}).
+    -- Also, this will be the slurp that we fall back on, in case re-typechecking another Unison file with implicit
```
ChrisPenner, line 246: Out of curiosity, when can this fail? Could you add a transcript for the failure state please?

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+    --
+    --   - Both:
+    --     - Either:
+    --         - (A) A component-mate of a term that's being updated.
+    --         - Both:
+    --           - (B) A dependent of a term that's being updated.
+    --           - (C) A dependency of the new term.
+    --     - (D) Not being updated.
+    --
+    -- FIXME Additionally, we have two more temporary requirements.
```
ChrisPenner, line 278: `FIXME`

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+    --
+    --   - (E) The term has exactly one name in the namespace.
+    --   - (F) No other terms share its name.
```
ChrisPenner, line 281: Re: (E); is there a problem if the term has _more_ than one name in the namespace? 🤔 
In that case I'd assume one name would be picked, then upon update the 'aliases' would also be updated.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+    -- This works around two fixable issues:
+    --
+    --   1. If the term has no names in the namespace, then we can't successfully put it into the Unison file anyway and
+    --      forward it through the typecheck + slurp process, because slurping currently assumes that it can freely
+    --      convert back-and-forth between vars and names.
+    --
+    --   2. If more than one term has its name, then putting this name in the Unison file and proceeding to do an update
+    --      would have the undesirable side-effect of resolving the conflict.
+    --
+    -- FIXME don't bother for type-changing updates
```
ChrisPenner, line 292: FIXME

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+    -- In our running example, the full list of component-mates (A) of the terms being updated is:
+    --
+    --   [ #pingpong.ping, #pingpong.pong ]
+    --
+    -- And because #pingpong.ping is being updated, due to (D), only #pingpong.pong remains.
+    --
+    -- Furthermore, #wham is both a dependent of #pingpong (B), and a dependency of #newping, so it too is an implicit
+    -- term.
+    --
+    -- FIXME: this currently only looks at old components; doesn't search for new cycles.
```
ChrisPenner, line 303: FIXME

What are the implications of not looking for new cycles?

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+                        --   Set.disjoint { "ping" } { "ping" } is false, so don't add to the map.
+                        --
+                        -- Running example, second time through (processing #pingpong.pong):
+                        --
+                        --   Set.disjoint { "ping" } { "pong" } is true, so add
+                        --   #pingpong.pong => (<#pingpong.ping + 2>, { "pong" })) to the map.
+                        notBeingUpdated = Set.disjoint namesBeingUpdated
+                        names = termRefToNames ref
+                     in if notBeingUpdated names
+                          then case Set.asSingleton names {- (E) -} of
```
ChrisPenner, line 351: Here's where it seems like it might be okay for there to be multiple names 🤔

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -279,3 +603,24 @@ propagatePatchNoSync ::
...
   Cli.time "propagatePatchNoSync" do
     Cli.stepAtNoSync' (Path.unabsolute scopePath, Propagate.propagateAndApply patch)
+
+------------------------------------------------------------------------------------------------------------------------
+-- Tiny helper bijection type
+--
+-- This is semantically a set of `(a, b)` tuples, where no `a` nor `b` appears more than once. An `a` can be looked up
+-- given its associated `b`, and vice-versa.
+
+data Bij a b
```
ChrisPenner, line 613: Don't have to do this now, but this seems like a useful type; I'd be in favour of moving it to a util package at some point, or just lumping it into `util-relation` as a new module.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+                  updatedNameToOldRefs
+                    & Map.toList
+                    & foldMap
+                      ( \(name, oldRefs) ->
+                          oldRefs
+                            & Set.toList
+                            & map
+                              ( \oldRef ->
+                                  case bijLookupL name nameToInterimRef of
+                                    Nothing -> undefined
```
ChrisPenner, line 440: `undefined`; Can this be replaced by some CallStack'd error call or something similar please?

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+              UnisonFileId
+                { dataDeclarationsId = UF.dataDeclarationsId' (Slurp.originalFile slurp0),
+                  effectDeclarationsId = UF.effectDeclarationsId' (Slurp.originalFile slurp0),
+                  -- Running example:
+                  --
+                  --   fresh1 = fresh3 + 4
+                  --   fresh2 = fresh1 + 2
+                  --   fresh3 = fresh2 + 3
+                  terms = Map.elems refToGeneratedNameAndTerm,
+                  -- In the context of this update, whatever watches were in the latest typechecked Unison file are
```
ChrisPenner, line 458: What happens to updates of test watches which are part of a component? E.g. If ping & pong were tests and the edit to `ping` happened inside of a test watch?

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+            --   "fresh2" -> "pong"
+            --   "fresh3" -> "wham"
+            let generatedNameToName :: Map Symbol Symbol
+                generatedNameToName =
+                  refToGeneratedNameAndTerm & Map.remap \(ref, (generatedName, _term)) ->
+                    ( generatedName,
+                      case bijLookupR ref nameToInterimRef of
+                        Just name -> name
+                        Nothing ->
+                          let (_term, name) = implicitTerms Map.! ref
```
ChrisPenner, line 480:
```suggestion
                          let (_term, name) = implicitTerms ^?! ix ref
```

The lensy version has a `HasCallStack` on it 😄

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -189,12 +201,324 @@ handleUpdate input optionalPatch requestedNames = do
...
+                           in name
+                    )
+
+            let renameTerm ::
+                  (Symbol, Term Symbol Ann, Type Symbol Ann) ->
+                  (Symbol, Term Symbol Ann, Type Symbol Ann)
+                renameTerm (generatedName, term, typ) =
+                  ( generatedNameToName Map.! generatedName,
```
ChrisPenner, line 488:
```suggestion
                  ( generatedNameToName ^?! ix generatedName,
```

unison-src/transcripts/cycle-update-1.md
```diff
@@ -0,0 +1,27 @@
+Update a member of a cycle, but retain the cycle.
```
ChrisPenner, line 1: What do you think of either:

1. putting all of these transcripts in the same file (you can nest the cases under different namespaces)
2. including a description of the behaviour in the name

I think it might be hard to get a sense of how cycle-updates work at a glance if I have to open and scan over 6 different transcript files.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: Yeah, it's being tracked in the minimegaticket

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: Oh, good catch. (E) isn't necessary.

unison-src/transcripts/cycle-update-2.output.md
```diff
@@ -0,0 +1,67 @@
...
+      ping : 'Nat
+
+```
+```ucm
+.> update
+
+  ⍟ I've updated these names to your new definition:
+  
+    ping : 'Nat
+    pong : 'Nat
```
ChrisPenner, line 55: Is it possible to add a note or indicator on `pong` here to say why/how it was updated? I think this will cause confusion for folks otherwise; "Why is pong updated? I didn't edit pong, did something break?"

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: Also being tracked in the minimegaticket

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: Being tracked in the minimegaticket. The implication is demonstrated in `cycle-update-4.md`: https://github.com/unisonweb/unison/pull/3443/files#diff-4de0eaae368ac2293fd5604a14f9e6a1d2978cbef88addf75880959fe528810c

unison-src/transcripts/cycle-update-3.output.md
```diff
@@ -0,0 +1,66 @@
...
+
+.> view ping pong
+
+  ping : Nat
+  ping = 3
+  
+  pong : 'Nat
+  pong _ =
+    use Nat +
+    !#4t465jk908.1 + 2
```
ChrisPenner, line 64: This is probably a whole different issue, but do we do anything to warn the user that this is happening? Or is `pong` at least added to the update frontier or something?

Even in that case I don't see any warning being output by UCM; we really need to strive to avoid any possible workflow that orphans references like this :'( 

Any ideas of something we could do to improve UX here?

unison-src/transcripts/cycle-update-4.md
```diff
@@ -0,0 +1,27 @@
+Not yet working: using update to establish a cycle.
```
ChrisPenner, line 1: 👍🏼

unison-src/transcripts/cycle-update-4.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+```unison
+ping : 'Nat
+ping _ = !pong + 1
+```
+
+```ucm
+.> update
+.> view ping pong
```
ChrisPenner, line 26: Maybe add a comment here about what you expect to see when it _is_ eventually working 😄

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: I'm not in favor of moving it out yet; it's really just to locally reduce the number of random mappy/setty data structures in this algorithm by one.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: Yeah- whoops! Meant to add a better error message.

unison-src/transcripts/cycle-update-6.md
```diff
@@ -0,0 +1,28 @@
+Not yet working: properly updating implicit terms with conflicted names.
```
ChrisPenner, line 1: AFAICT this is an alias, not a conflicted name; is that right?

unison-src/transcripts/cycle-update-1.md
mitchellwrosen: Hmm, I'm not inclined to do either. I think smaller test cases are better than one giant one, which requires somewhat carefully considering what's happened above (so it's hard to add to). And naming each case seems kinda tedious; it's much easier to describe the test inside the file with a full sentence.

unison-src/transcripts/cycle-update-2.output.md
mitchellwrosen: Yeah, I think we'd prefer `pong` not to appear at all, just like other dependents that we propagate the update to. That's being tracked in the minimegaticket.

ChrisPenner (approved):
Good work on this, this is a really hard problem.

I appreciate you splitting it up into pieces and delivering value one chunk at a time 👍🏼 

I left a good number of notes and questions to address before merging, but otherwise looks good 😄

unison-src/transcripts/cycle-update-2.output.md
ChrisPenner: I think my preferred behaviour would be that `pong` would appear more like an update propagation, since that's _kind of_ what it is; but if it's in the megaticket that's good for now 👍🏼

unison-src/transcripts/cycle-update-3.output.md
mitchellwrosen: Yeah, this case actually isn't different from a normal type-changing update, without complications due to cycles. In both cases, ucm allows the update, then you can run `todo` and see what references the old thing. Agree that it's probably useful to know if one would cause a `todo` entry before applying an update.

As of https://github.com/unisonweb/unison/pull/3400 at least you'll see the busted thing.

unison-src/transcripts/cycle-update-6.md
mitchellwrosen: Whoops, ya. And it's being rejected for a bunk reason as you pointed out `(E)`. I'll make it a real conflicted term with a merge.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: This fails when the user makes a type-incompatible update; the existing transcripts do cover that case

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: > If ping & pong were tests and the edit to ping happened inside of a test watch?

You can't perform an `update` on a watched term

unison-src/transcripts/cycle-update-2.output.md
mitchellwrosen: Yeah, that's what I said, or tried to say :)

mitchellwrosen:
I think this is good to go; I'm tracking all of the nits and stuff in the minimegaticket.

tstat (commented):
LGTM

PR #3466, opened 2022-09-28
update: discover new cycles

mitchellwrosen:
## Overview

This PR implements the "new component discovery" feature of `update`.

In brief, if you have a codebase that contains

```
foo _ = 500
bar _ = !foo + 100
```

and then update `foo` to

```
foo _ = !bar + 200
```

then you'll end up with a single component `foo <-> bar`:

```
foo _ = !bar + 200
bar _ = !foo + 100
```

Prior to this PR, you'd end up with this unfortunate situation:

```
foo _ = !bar + 200
bar _ = !#oldfoo + 100
```


codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1394,6 +1395,101 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+getDependenciesBetweenTerms oid1 oid2 =
+  queryListCol sql (oid1, oid2) <&> Set.fromList
+  where
+    -- Given the example above, we'd have tables that look like this.
+    --
+    -- First, the `paths` table finds all paths from source `A` to sink `B`, exploring depth-first. As a minor
+    -- optimization, we seed the search not with `A`, but rather the direct dependencies of `A` (namely `X` and `Y`).
+    --
+    -- +-paths-------------------------------+
+    -- +-level-+-object_id-+-object_id_array-+
```
aryairani: Maybe easier to understand if `object_id_array` was called `path`? or `path_so_far` Or maybe we can talk through it on Weds :)

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -242,6 +243,15 @@ getSlurpResultForUpdate requestedNames slurpCheckNames = do
     -- terms in it fails.
     let slurp0 = slurp unisonFile0
 
+    -- Grab some interim info out of the original slurp.
```
aryairani, line 246: Maybe we can expand a little bit about why this interim info, on Weds

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1394,6 +1395,101 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+--     Y---Z
+--
+-- ...then `getDependenciesBetweenTerms A B` would return the set {X Y Z}
+getDependenciesBetweenTerms :: ObjectId -> ObjectId -> Transaction (Set ObjectId)
+getDependenciesBetweenTerms oid1 oid2 =
+  queryListCol sql (oid1, oid2) <&> Set.fromList
+  where
+    -- Given the example above, we'd have tables that look like this.
+    --
+    -- First, the `paths` table finds all paths from source `A` to sink `B`, exploring depth-first. As a minor
```
aryairani: finds all paths from source `A` (not just ones to `B`)
note: we can't know that a path doesn't end in `B` until we get there.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1394,6 +1395,101 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+    sql = [here|
+      with recursive paths(level, object_id, object_id_array) as (
+        select
+          0,
+          dependents_index.dependency_object_id,
+          ''
+        from dependents_index
+          join object on dependents_index.dependency_object_id = object.id
+        where dependents_index.dependent_object_id = ?
+          and object.type_id = 0
```
aryairani: add a comment to remind what `type_id = 0` is there for

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1394,6 +1395,101 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+          dependents_index.dependency_object_id,
+          dependents_index.dependent_object_id
+            || ','
+            || paths.object_id_array
+        from paths
+          join dependents_index
+            on paths.object_id = dependents_index.dependent_object_id
+          join object on dependents_index.dependency_object_id = object.id
+        where object.type_id = 0
+          and not dependents_index.dependent_object_id = dependents_index.dependency_object_id
```
aryairani: maybe add a filter to not recurse whenever the dependency is the target

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1394,6 +1395,101 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+        select
+          0,
+          dependents_index.dependency_object_id,
+          ''
+        from dependents_index
+          join object on dependents_index.dependency_object_id = object.id
+        where dependents_index.dependent_object_id = ?
+          and object.type_id = 0
+          and not dependents_index.dependent_object_id = dependents_index.dependency_object_id
+        union
```
aryairani: question about `level` vs `union`

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1394,6 +1395,101 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+        from paths
+        where paths.object_id = ?
+        union all
+        select
+          substr(object_id_array, 0, instr(object_id_array, ',')),
+          substr(object_id_array, instr(object_id_array, ',') + 1)
+        from elems
+        where object_id_array != ''
+      )
+      select cast(object_id as integer) as object_id
```
aryairani: do we need `select distinct` here?

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
aryairani: even forgetting `level`, if we have two paths that meet up, we want to record both ways you got there, because you want all the elements from both paths in the result

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1408,6 +1409,108 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+    --
+    -- +-object_id-+
+    -- |         X |
+    -- |         Z |
+    -- |         Y |
+    -- +-----------+
+    --
+    -- Notes
+    --
+    -- (1) We only care about term dependencies, not type dependencies
```
ChrisPenner: Why is that? Is it not possible to end up creating a new cycle with types for some reason? 
Or is `getDependenciesBetweenTypes` intended to be added in a later PR?

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
```diff
@@ -1408,6 +1409,108 @@ getDependencyIdsForDependent dependent@(C.Reference.Id oid0 _) =
...
+    -- (2) No need to search beyond the sink itself, since component dependencies form a DAG
+    -- (3) An explicit cast from e.g. string '1' to int 1 isn't strictly necessary
+    sql :: Sql
+    sql = [here|
```
ChrisPenner, line 1480: just a nit, but for consistency would you mind using ALL_CAPS for the SQL keywords? I know it doesn't matter, but it does generally help me parse sql better that way.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
ChrisPenner: Agreed, I'd love to know what the "iterim" info is used for, and if possible incorporate that description in the name or docs.

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -294,64 +311,126 @@ getSlurpResultForUpdate requestedNames slurpCheckNames = do
...
-                        nameIsUnconflicted name = Set.size (nameToTermRefs name) == 1
-                        names = termRefToNames ref
-                     in if notBeingUpdated names
-                          then case Foldable.find nameIsUnconflicted names {- (E) -} of
-                            Nothing -> acc
-                            Just name -> Map.insert ref (term, name) acc
-                          else acc
-                )
-                Map.empty
+        Codebase.withConnection codebase \conn ->
```
ChrisPenner, line 320: I think you can use `Codebase.runTransaction` here.

ChrisPenner (approved):
LGTM, thanks for the transcript 👍🏼

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
aryairani: Yeah, in a later PR when we remember what propagating changes through types should mean — maybe requiring kind checking, too. In the end, I imagine it will look similar to this PR though.

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
mitchellwrosen: I think I made this better, see [58e2bf6](https://github.com/unisonweb/unison/pull/3466/commits/58e2bf6541f2d3b226ec11eca1488bb227f71a19)

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: The "interim" info is used all over the place, it's simply the name I picked for the the just-typechecked file that seeds this whole algorithm of discovering more terms from the codebase that form components with the updated ones (including component-mates and all terms along all paths from new version to old version). It doesn't have to be let-bound; we could have just as easily reached into the `hashTermsId` field of the latest typechecked unison file of the slurp result. So, I guess what I mean is - I'm not sure that this particular let-binding would be the place to clear up any confusion resulting from its choice of name. Is there something specific about the algorithm I could document better? (I assume "yes" - just call out as appropriate)

codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
mitchellwrosen: Yeah. This is a `UNION ALL` now as we discussed in the call. I think the `level` bit is probably still worth keeping, though?

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: Yes, thanks, I forgot about that

mitchellwrosen:
macOS-11.0 CI failed with a lot of errors like:

```
utf8.md EXCEPTION!!!: Unable to acquire lock file: "/Users/runner/.local/share/unisonlanguage/credentials.json.lock"
```

I guess I should restart and pray?

aryairani:
@mitchellwrosen I guess it was restart and pray.

aryairani:
@mitchellwrosen Go ahead and merge if it's still ready

PR #3473, opened 2022-09-30
refactor: leave getTermComponentWithTypes in Transaction

mitchellwrosen:
## Overview

This PR changes the type (in the codebase record-of-functions) of `getTermComponentWithTypes` from

```haskell
getTermComponentWithTypes :: Hash -> m (Maybe [(Term v a, Type v a)]),
```

to

```haskell
getTermComponentWithTypes :: Hash -> Transaction (Maybe [(Term v a, Type v a)]),
```

thus allowing the caller to put together a transaction that uses this function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/unisonweb/unison/3473)

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
@@ -54,6 +54,7 @@ import Unison.Referent (Referent)
 import qualified Unison.Referent as Referent
 import qualified Unison.Result as Result
 import Unison.Runtime.IOSource (isTest)
+import qualified Unison.Sqlite as Sqlite
 import Unison.Symbol (Symbol)
 import Unison.Term (Term)
 import qualified Unison.Term as Term
@@ -322,7 +323,9 @@ getSlurpResultForUpdate requestedNames = do
           --   [ (<#pingpong.pong + 1>, <Nat>),
           --     (<#pingpong.ping + 2>, <Nat>)
           --   ]
-          terms <- Codebase.unsafeGetTermComponent codebase oldHash
+          terms <-
+            Codebase.withConnection codebase \conn ->
+              Sqlite.runTransaction conn (Codebase.unsafeGetTermComponent codebase oldHash)
           pure $
             terms
               -- Running example:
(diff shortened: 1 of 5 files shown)


unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
```diff
@@ -322,7 +323,9 @@ getSlurpResultForUpdate requestedNames = do
           --   [ (<#pingpong.pong + 1>, <Nat>),
           --     (<#pingpong.ping + 2>, <Nat>)
           --   ]
-          terms <- Codebase.unsafeGetTermComponent codebase oldHash
+          terms <-
+            Codebase.withConnection codebase \conn ->
+              Sqlite.runTransaction conn (Codebase.unsafeGetTermComponent codebase oldHash)
```
aryairani, line 328: I wonder if we could get much use out of a combinator that takes a `Codebase` and a `Transaction` and calls `Codebase.withConnection` and `Sqlite.runTransaction`.

unison-cli/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -498,8 +499,12 @@ propagate patch b = case validatePatch patch of
     unhashTermComponent' :: Hash -> Cli r (Map Symbol (Reference.Id, Term Symbol Ann, Type Symbol Ann))
     unhashTermComponent' h = do
       Cli.Env {codebase} <- ask
-      liftIO (Codebase.getTermComponentWithTypes codebase h) <&> foldMap \termsWithTypes ->
-        unhash $ Map.fromList (Reference.componentFor h termsWithTypes)
+      maybeTermsWithTypes <-
+        liftIO do
+          Codebase.withConnection codebase \conn ->
+            Sqlite.runTransaction conn (Codebase.getTermComponentWithTypes codebase h)
```
aryairani, line 505: 👀

aryairani (approved):

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update.hs
mitchellwrosen: I'm sure we could; there's also `runReadOnlyTransaction` / `runWriteTransaction` that we might want to wrap. Fairbairn threshold 🤷

Issue #3045, opened 2022-04-13
aarch64 support


kubukoz:
Hi, I'm wondering if there are any plans to provide ARM binaries - e.g. for M1 macs. Are there? 😅

ceedubs:
This has come up a few times as a useful thing that we'll want to do. As far as I know GitHub Actions don't support building as an M1 Mac, but please let me know if you know of a good solution to this. We may be able to run builds on another CI that can build as an M1 Mac, but it might be a while before we get around to that.

aryairani:
There is no official binary distribution of `stack` for darwin arm64 at the moment either. There is an unofficial binary provided by ghcup folks, but it segfaults for me. This is the first thing that would have to change before we'd have a timeline for M1 binaries.

@kubukoz Are you having some kind of trouble with the x86 build? Or just excited about arm64 like me?

kubukoz:
It's just the efficiency I'm worried about 😄

aryairani:
Gotcha — I'm no expert, but I'm operating under the assumption that Haskell code compiled with the arm64 backend could actually be slower than when compiled with the more mature x86 backend and then translated through Rosetta, so I think this could be the more efficient option for a while too. Plus, we're working hard to make ucm efficient in other ways :)

tfausak:
For what it's worth, you should be able to build darwin arm64 binaries by adding macos-14 to your build matrix. That's what I do for Gild: https://github.com/tfausak/cabal-gild/blob/a2249ed/.github/workflows/ci.yml#L43

However I use Cabal rather than Stack. I think Stack support arm64 since version 2.13.1.

aryairani:
We'll probably try to provide arm64 binaries through https://github.com/Homebrew/homebrew-core in not too long.

ceedubs:
It would be nice to provide arm64 binaries independent of homebrew. For example it would be cool to be able to run Unison programs on Raspberry Pi and similar ARM machines running linux. Though I guess I don't know what builtins might not be supported on those platforms.

aryairani:
Supposedly arm linux available by the end of the year https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/

PR #3483, opened 2022-10-05
network: 3.1.1.1 -> 3.1.2.7

ceedubs:
## Overview

I finally got around to trying to build Unison on my M1 Mac with an ARM
version of Stack instead of through x86 emulation, and I ran into [this
issue](https://github.com/haskell/network/pull/510). Bumping the
`network` dependency a bit seems to resolve the issue.

A step toward #3045.

Stack made a bunch of unrelated changes to the lock file, mostly just
moving fields around but in some cases updating their values. I don't
know why it moved stuff around, and I can't find any information about
how to make it only update a single package. Is this just a thing that
Stack does?

## Test coverage

All of the tests and transcripts run fine with this change, and after
kicking the tires a bit I didn't see any issues.

## Loose ends

We don't have aarch64 builds in CI, so we won't have any automated tests for
whether this continues to work on M1 Macs.

Also I don't understand the relationship between Stack and Cabal. Does any Cabal stuff need to be updated for the dependency change?

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -69,6 +69,8 @@ extra-deps:
 - co-log-core-0.3.1.0
 # lts 18.28 provides 0.3.2.1 but we need at least 0.3.3
 - terminal-size-0.3.3
+# lts 18.28 provides 3.1.1.1 but we need at least 3.1.2.7
+- network-3.1.2.7
 
 ghc-options:
  # All packages
--- a/stack.yaml.lock
+++ b/stack.yaml.lock
@@ -5,166 +5,173 @@
 
 packages:
 - completed:
+    sha256: d4fd87fb7bfc5d8e9fbc3e4ee7302c6b1500cdc00fdb9b659d0f4849b6ebe2d5
+    name: configurator
     size: 15989
     url: https://github.com/unisonweb/configurator/archive/e47e9e9fe1f576f8c835183b9def52d73c01327a.tar.gz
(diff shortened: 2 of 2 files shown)


aryairani:
> Stack made a bunch of unrelated changes to the lock file, mostly just moving fields around but in some cases updating their values. I don't know why it moved stuff around, and I can't find any information about how to make it only update a single package. Is this just a thing that Stack does?

I'm not sure; I'm guessing it's parsing, updating, and re-printing the whole lock file, possibly reordered due to it being a different version of stack as during the last update. I've seen this happen and didn't give it any thought.

> We don't have aarch64 builds in CI, so we won't have any automated tests for whether this continues to work on M1 Macs.

We should probably add these, OTOH we hate waiting for CI and we are already maxing out the available runners.

> Also I don't understand the relationship between Stack and Cabal. Does any Cabal stuff need to be updated for the dependency change?

I don't fully either, but I think Cabal just cares about "network" and Stack is changing the version that's available.

ceedubs:
> We should probably add these, OTOH we hate waiting for CI and we are already maxing out the available runners.

I also think that you can't get ARM runners on GitHub Actions for free. You have to set up your own infra for runners and hook it up. We might want to do something like that eventually, but I'm not sure when we'll get to it.

aryairani (approved):
I meant to approve this yesterday — thanks!

PR #3491, opened 2022-10-11
Add putTermComponent to codebase interface

mitchellwrosen:
## Overview

This PR adds a `putTermComponent` helper to the codebase interface.

Previously, we only had `putTerm`, which:

- Buffers the term in-memory, because it might be part of a component, and we need the whole component buffered before we can write a row to SQLite.
- Tries flushing the buffer (which flushes complete components, and then other components that depend on the flushed ones)

`putTermComponent` is similar, it just takes the whole component as input, and only tries flushing the buffer once, after buffering each term individually.

So, not a huge change, just a step towards eliminating `putTerm` and replacing it with some ideal `putTermComponent` that doesn't do any buffering.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/SqliteCodebase/Conversions.hs
+++ b/parser-typechecker/src/Unison/Codebase/SqliteCodebase/Conversions.hs
@@ -204,6 +204,13 @@ term2to1 h lookupCT =
           V2.Term.PConcat -> V1.Pattern.Concat
         a = Ann.External
 
+termComponent1to2 ::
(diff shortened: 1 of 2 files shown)


aryairani (approved):

ChrisPenner (approved):
👍🏼 

We should probably discuss whether we want to continue storing components as single giant blobs in the future, we need the component to hash and save things of  course, but it'd generally be more efficient to unpack things if the component definitions were stored independently.

That's a question for later though, good stuff 👍🏼

PR #3507, opened 2022-10-14
Fix pretty-printing of empty docs

### LINKED ISSUE #3506: Empty docs display incorrectly
```unison
x = {{}}
```

```ucm
base> display x
```

Outputs:

```
Annotated.Empty
```

Should output nothing at all.

runarorama:
Fixes #3506.

The case for the `Annotated.Empty` constructor was simply missing from `displayTerm'`, as it's not an `Apps'` but just a `Constructor'`. Consequently it was falling through to the regular term pretty-printer.

Further, the `displayPretty` function was using an `App` pattern as well, which is incorrect, as the `Empty` constructor takes no arguments. So if it had been making it into `displayPretty`, this would cause an infinite loop.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/IOSource.hs
+++ b/parser-typechecker/src/Unison/Runtime/IOSource.hs
@@ -331,7 +331,7 @@ prettyAppendId = constructorNamed prettyAnnotatedRef "Pretty.Annotated.Append"
 
 prettyTableId = constructorNamed prettyAnnotatedRef "Pretty.Annotated.Table"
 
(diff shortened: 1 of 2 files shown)


aryairani (approved):
lgtm

pchiusano:
I know this was a quick fix but I'd have liked to have seen a transcript regression test / demonstration of the behavior, which would have taken about 1 minute to put together. :)

aryairani:
We could still use the transcript from the original issue as a test

PR #3512, opened 2022-10-16
Fewer parens inside list types

### LINKED ISSUE #1453: Surplus parentheses around function types in lists
Add the following to the codebase:

```
x = [a -> a + 1]
```

Unison will say it added `x : [(Nat -> Nat)]`. That type has altogether too many parens.

runarorama:
Fixes https://github.com/unisonweb/unison/issues/1453.

The precedence for printing parentheses around function types is `0`, so it would always print no matter what the outer type is. I set the precedence inside the `[]` of a list type to `-1`.

I added a roundtrip test for types `[a]`, `[a -> b]`, and `[a ->{g} b]`.


aryairani:
Should we add a test for this, or are there some already that don't show up in the diff?

aryairani (approved):
Sweet

ChrisPenner (approved):

PR #3519, opened 2022-10-18
Pull out unison-util-bytes into its own package

mitchellwrosen:
## Overview

This PR extracts `Unison.Util.Bytes` into its own package, `unison-util-bytes`. The purpose is to continue extracting all of the surface-syntax related parsing/printing modules out of `unison-parser-typechecker`. This type is a dependency of the lexer (`Bytes` is a lexeme), so this package will be depended on by both `unison-parser-typechecker` and `unison-syntax` (see the next PR: #3521)


aryairani (approved):
Great

tstat (approved):

PR #3525, opened 2022-10-19
Backticky symbol parsing

### LINKED ISSUE #1519: impossible to create docs under symboly namespace
```ucm
    .> alias.term ##Nat.+ Nat.+
    .> alias.term ##Nat.+ Nat.plus
    ```
    ```unison
    Nat.+.doc = [: @Nat.plus very useful for adding. :]
    ```

```
scratch.u:1:1:
unexpected Nat.+.
expecting ability, type, or use
    1 | Nat.+.doc = [: @Nat.plus very useful for adding. :]
```

I know this is not new, but not sure where the dupe is.

### LINKED ISSUE #1589: Add or co-opt syntax for escaped identifier names
Inspired by:
https://unisonlanguage.slack.com/archives/CLG4P0GSD/p1590663931259900
#1523 - `.` is accepted as namespace name, but leads to issues
#1586 - `add` crashes on the name `.`
#1587 - consider disallowing namesegments with `.` in them
#1362 - syntax error when trying to define | as an operator (and some other weirdness)

### Motivation
Already, with `.` (and possibly `||` and `&&`?), we are running into parsing complications with somewhat-or-more- reserved characters in Unison identifiers.  With pluggable syntax (#499), we will have the additional problem that different syntaxes have different reserved characters, and different reserved words.  It is and will continue to be important to have a way to render and reference names created in one syntax but illegal in yours.

A `Name` should be represented by a `Seq NameSegment`, where `NameSegment` is a `newtype` of `Text` or `String`.  Any `NameSegment` consisting of printable characters should be allowed, and every pluggable syntax (including Unison's current one) should have a syntax to unambiguously parse and display it.

Some escaping options we've considered so far:

#### Backslashes
  - `foo.\<.\>.bar` means `Relative ["foo", "<", ">", "bar"]`
  - `foo.\<\.\>.bar` means `Relative ["foo", "<.>", "bar"]`
  - `.` means `Absolute []`
  - @aryairani doesn't love this one.

#### Backticks
  - ``foo.`<`.`>`.bar`` means `Relative ["foo", "<", ">", "bar"]`
  - ``foo.`<.>`.bar`` means `Relative ["foo", "<.>", "bar"]`
  - `` foo.`type` ``  means `Relative ["foo", "type"]`
  - `.` means `Absolute []`
  - `` `.` `` means `Relative ["."]`
  - Pro: Scala uses similar syntax
  - Pro/Con: backticks would no longer mean "infix wordy operator"
    - these aren't really being used well currently anyway; e.g. ``x `contains` set`` 🤔, also the 
pretty-printer doesn't generate them.
  - what about backticks in identifier names?
    - moar backticks!
      - markdown style:
        - ``` `` `.` `` ``` means ``Relative ["`.`"]``; leading and trailing space is ignored
        - ```` ``` ``. ``` ```` means ```Relative ["``."]```
      - scala style? something to do with triple-backticks though, and not super expandable
    - backslashes? 🤦🏻‍♂️
      - `` `\`.\`` ``   means ``Relative ["`.`"]``
      - `` `\`\`.` ``   means ```Relative ["``."]```
#### Parentheses
  - `foo.(<).(>).bar.` means `Relative ["foo", "<", ">", "bar"]`
  - `foo.(<.>).bar` means `Relative ["foo", "<.>", "bar"]`
  - `foo.(type)`  means `Relative ["foo", "type"]`
  - `.` means `Absolute []`
  - `(.)` means `Relative ["."]`
  * We already use this syntax to make operators first-class, although we say `(Int.+)`, not `Int.(+)`.
  * Q: how hard would it be to parse distinctly from parenthesized expressions?
  - Q: What about parens in identifier names?  I guess we disallow that.

### Task

Choose a syntax from above or from the forthcoming discussion below, and update:
- [ ] `InputPatterns` parser(s)
- [ ] auto-completion parser(s)
- [ ] `TermParser`
- [ ] `TermPrinter` (and/or open a separate ticket and ask @atacratic nicely for help) 😁

### Known loose ends:
- Per #1586, configurator doesn't like `.` namesegments either.  or maybe a lot of namesegments.  How should we escape these / replace configurator (#1472, #1129)?

### LINKED ISSUE #2970: support symboly namesegments in scratch file parser

### PR DESCRIPTION
**Pull request by mitchellwrosen**
## Overview

This PR adds two features:

1. Names now consist of 1+ wordy or symboly name segments. Previously, they consisted of 0+ wordy name segments followed by 1 wordy or symboly name segment. That means `foo.++.doc` is now valid, the primary purpose of the PR.

2. Historical naming accidents `.` and `()` are now handled better - they are valid name segments, so long as they belong to a name segment escaped in backticks, e.g. `` `.` `` or `` `()()()` ``. Reserved symbols and names (such as `=` and `match`) are also valid name segments, if they are escaped (`` `=` ``, `` `match` ``).

- - -

Fixes #1519
Fixes #1589
Fixes #2970

### CODE REVIEW

File: unison-cli/src/Unison/CommandLine/Completion.hs [hunk 1]
```diff
@@ -241,22 +244,24 @@ completeWithinNamespace compTypes query currentPath = do
...
+--   -- ".base."
+--   -- ".base.List"
+--   Just ("" : pathPrefix, querySegment) -> (Path.AbsolutePath' . Path.Absolute . Path.fromList . fmap NameSegment $ pathPrefix, NameSegment querySegment)
+--   -- ""
+--   -- "base"
+--   -- "base.List"
+--   Just (pathPrefix, querySegment) ->
+--     ( Path.RelativePath' . Path.Relative . Path.fromList . fmap NameSegment $ pathPrefix,
+--       NameSegment querySegment
+--     )
```
aryairani: reminder to delete when ready

unison-src/tests/methodical/dots.u
```diff
@@ -1,11 +1,11 @@
 
 -- You can define an operator called dot
-(.) f g x = f (g x)
+(`.`) f g x = f (g x)
```
aryairani, line 3: This is interesting; it makes me think that there are some unambiguous syntaxes we could accept without ``` `` ```, e.g. if it's in parens, all symbols, with no spaces, like the `(.)` above, or a group of symboly chars that are separated with spaces or other tokens, from any other identifiers, thus not a path-qualified name, like the `id . id` below, but this is still fine/good for now.

unison-src/tests/methodical/dots.u
```diff
@@ -14,15 +14,15 @@ id2 = id . id
 -- foo = id .zonk
 
 -- You can define qualified functions
-(base.function..) f g x = f (g x)
+(base.function.`.`) f g x = f (g x)
```
aryairani, line 17: Even with the relaxed rules proposed in the previous comment, this would still need the ``` `` ```s, because it's got other `.` in it, which would mean it's a path-qualified name.

unison-syntax/src/Unison/Syntax/Lexer.hs
```diff
@@ -889,7 +889,10 @@ lexemes' eof =
           <|> layoutKeywords
       where
         keywords =
-          symbolyKw "."
+          -- yes "wordy" - just like a wordy keyword like "true", the literal "." (as in the dot in
+          -- "forall a. a -> a") is considered the keyword "." so long as it is either followed by EOF, a space, or some
+          -- non-wordy character (because ".foo" is a single identifier lexeme)
+          wordyKw "."
```
aryairani, line 895: Why not `symboly`, because the "eof/space/non-wordy-character" definition of wordy just _is_ what we need here?

mitchellwrosen: Yeah, the definition of `wordyKw` is what we need here. I can't remember exactly what bug `symbolyKw` was causing at the moment, unfortunately, but it's not correct

aryairani:
@mitchellwrosen Could you add a regression test for #2970 and #1519 and then we'll get it merged.

aryairani (approved):

PR #3531, opened 2022-10-20
Generate stack traces for calls to `bug` and similar.

dolio:
This modifies the runtime to track a reference to the currently executing combinator, and uses it and the continuation to produce a (unison) stack trace when calls to the error instruction are made. This includes builtins like `bug` and pattern match failures.

I don't have a good place to put this information when these exceptions are caught by `tryEval`, so I'm just discarding them right now. Maybe that can be improved in the future.

One additional limitation is that I think you often won't see the function that directly calls `bug`, because it will be a tail call. Not sure what to do about that at the moment. You could work around it by putting something useless _after_ the call to bug.

Example:

```
bomb : Nat -> Text
bomb = cases
  0 -> bug "whoops"
  n -> "1+" ++ bomb (n-1)

> bomb 5
```
image

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Exception.hs
+++ b/parser-typechecker/src/Unison/Runtime/Exception.hs
@@ -6,10 +6,11 @@ import Data.Text
 import GHC.Stack
 import Unison.Runtime.Stack
 import Unison.Util.Pretty as P
+import Unison.Reference (Reference)
 
 data RuntimeExn
   = PE CallStack (P.Pretty P.ColorText)
-  | BU Text Closure
+  | BU [(Reference,Int)] Text Closure
   deriving (Show)
 
 instance Exception RuntimeExn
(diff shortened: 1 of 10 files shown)


pchiusano:
> One additional limitation is that I think you often won't see the function that directly calls bug, because it will be a tail call. Not sure what to do about that at the moment. 

You could just not consider a call to `bug` in tail position to be a tail call. Basically, compile it like `bug blah; ()`. Since a call to `bug` is going to unwind the stack anyway, treating it as a tail call doesn't really matter.

But this is still great as is!

pchiusano (approved):
Awesome!

hojberg:
@dolio is the stack trace truncated? I think it'd be cool if it had maybe max 10 or 15 items in the list (counting recursion as 1 perhaps) and give users a command to see the full trace.

PR #3540, opened 2022-10-25
Fix to `Pattern.or` and `Pattern.replicate` to properly do backtracking when given composite patterns

### LINKED ISSUE #3530: Unexpected capture with Pattern.optional and Pattern.join
This seems like a bug to me, but it could be my expectations being off. Consider the following code:

```haskell
  p = Pattern.capture (
    Pattern.optional (
      Pattern.join [
      , literal "a"
      , literal "b"
      ]
    )
  )
  Pattern.run p "ac"
```

I would expect it to return `Some ([], "ac")`, because the inner `Pattern.join` expects an `a` followed by a `b`, so I would expect the entire `optional` segment to return an empty string.

However, it actually returns `Some (["a"], "c")`. Am I off or is the `Pattern` implementation?

@dolio @pchiusano

pchiusano:
Fixes #3530 thanks to @dolio for pairing on this!

Previously, patterns like:

```
Pattern.or (Pattern.join [p1, p2]) p3
```

Would not correctly backtrack the state (including both captures and the position) if `p1` succeeded. `Pattern.replicate` had a similar issue.

This PR fixes that and adds a regression (unit) test for the specific case reported by @ceedubs and a number of other related test cases.

I'm shocked this wasn't noticed before, but I'm guessing it's a combination of:

* Many "in the wild" patterns only need "1 token of lookahead" and are written as such. (Including examples in the docs and the prior unit tests 😳)
* API hasn't gotten a ton of usage yet

Note that if you do `capture (optional p)`, that will always capture at least the empty string (if `p` fails, then `optional` succeeds, consuming 0 characters, so the empty string consisting of 0 characters is captured), so the original bug report was slightly off in its expectations.

### DIFF
--- a/parser-typechecker/src/Unison/Util/Text/Pattern.hs
+++ b/parser-typechecker/src/Unison/Util/Text/Pattern.hs
@@ -47,11 +47,46 @@ cpattern p = CP p (run p)
 
 run :: Pattern -> Text -> Maybe ([Text], Text)
 run p =
-  let cp = compile p (\_ _ -> Nothing) (\acc rem -> Just (reverse acc, rem))
-   in \t -> cp [] t
+  let cp = compile p (\_ _ -> Nothing) (\acc rem -> Just (s acc, rem))
+      s = reverse . capturesToList . stackCaptures
+   in \t -> cp (Empty emptyCaptures) t
 
+data Stack = Empty !Captures | Mark !Captures !Text !Stack
+
+type Captures = [Text] -> [Text]
+
+stackCaptures :: Stack -> Captures
+stackCaptures (Mark cs _ _) = cs
+stackCaptures (Empty cs) = cs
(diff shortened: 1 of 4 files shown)


ceedubs (approved):
I won't claim to understand all of the changes or their implications, but the tests look good to me.

It would be nice if there were a way to test this closer to the Unison code. Is it possible to test `Pattern` via a transcript?

pchiusano:
@ceedubs I went ahead and added a transcript regression test as well. :) We do also have transcript integration tests for the pattern stuff, so I added an example there:

https://github.com/unisonweb/unison/pull/3540/files#diff-9766e8bd8574eb3b1a896c857f18b737b01f11ab9c43082f3a28dcdd0e94ad00

ceedubs (approved):
Nice thanks @pchiusano!

PR #3546, opened 2022-10-28
run ucm main in an unbound thread

mitchellwrosen:
## Overview

This PR changes ucm's main to run in an unbound thread. When main is bound (the default), all of its foreign calls are run on the same OS thread, and its capability can't run any other haskell threads. `ucm` has a little bit of concurrency (background web server, LSP thread, maybe one other), so we often end up paying for an OS thread context switch when scheduling between main and any other thread. This PR fixes that.

### DIFF
--- a/parser-typechecker/unison-parser-typechecker.cabal
+++ b/parser-typechecker/unison-parser-typechecker.cabal
@@ -1,6 +1,6 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.35.0.
+-- This file has been generated from package.yaml by hpack version 0.34.4.
 --
 -- see: https://github.com/sol/hpack
 
@@ -300,9 +300,9 @@ library
     , x509-system
     , yaml
     , zlib
-  default-language: Haskell2010
   if flag(optimized)
     ghc-options: -funbox-strict-fields -O2
+  default-language: Haskell2010
 
 test-suite parser-typechecker-tests
   type: exitcode-stdio-1.0
@@ -488,6 +488,6 @@ test-suite parser-typechecker-tests
     , x509-system
     , yaml
     , zlib
-  default-language: Haskell2010
   if flag(optimized)
     ghc-options: -funbox-strict-fields -O2
+  default-language: Haskell2010
--- a/unison-cli/unison/Main.hs
+++ b/unison-cli/unison/Main.hs
@@ -23,7 +23,7 @@ import ArgParse
     parseCLIArgs,
   )
 import Compat (defaultInterruptHandler, onWindows, withInterruptHandler)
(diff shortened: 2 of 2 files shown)


parser-typechecker/unison-parser-typechecker.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.35.0.
+-- This file has been generated from package.yaml by hpack version 0.34.4.
```
mitchellwrosen, line 3: We should standardize all of our tooling so these diffs go away 😓. Use nix-provided `stack`, `hpack`, `ghc`,  etc., probably?

tstat (approved):

ChrisPenner (commented):
Interesting, TIL;

I notice the docs for `runInUnboundThread` mention:

> Use this function only in the rare case that you have actually observed a performance loss due to the use of bound threads. A program that doesn't need its main thread to be bound and makes heavy use of concurrency (e.g. a web server), might want to wrap its main action in runInUnboundThread.

Have you observed performance losses? Do we have any way of benchmarking this to see if it improves things?

I wouldn't necessarily classify UCM as a _heavy_ user of concurrency, but we do communicate between threads periodically.

mitchellwrosen:
I haven't done any benchmarking, no.

parser-typechecker/unison-parser-typechecker.cabal
aryairani: How do we do that?

aryairani: and then update development.markdown

aryairani: Also I'm surprised an older version overwrote a newer version, was that intentional?

aryairani:
If the docs say we should only switch once we've observed a performance loss, but we haven't observed a performance loss, then why do we think we want to switch?  Is there any downside?

parser-typechecker/unison-parser-typechecker.cabal
mitchellwrosen: Yes, because otherwise the cabal file doesn't get generated. I had to delete it manually and rebuild to regenerate

mitchellwrosen:
I'm not sure why the docs are worded like that. They don't provide justification for why you would not want to run main in an unbound thread. Perhaps at one time it was thought to be a safer default because if you write a program that makes FFI calls that rely on thread-local state, like OpenGL calls, then your program will misbehave and crash. I think it's the wrong default for the reasons provided in the PR description, but I'm not sure it's really worth proving with a benchmark at this time, so feel free to close.

PR #3548, opened 2022-10-29
Comment out builtins from the edit command

### LINKED ISSUE #3361: (unimportant) editing an alias for a built-in adds invalid code to scratch file
Let's say that you create an alias for a builtin term:

```ucm
.> alias.term ##Nat.+ plus
```

Now you want to load your whole namespace to a scratch file so you run `.> find` and then `edit 1-24` (where `24` is the number for the last declaration spat out by `find`).

Inside your scratch file you'll now have a line like:

```unison
builtin plus : ##Nat -> ##Nat -> ##Nat
```

This is invalid Unison code, so your scratch file won't compile.

Generally I wouldn't expect users to run into this. I've run into it a few times when I've needed to use a built-in that hasn't yet been added to the base library, so I've created an alias for it.

I really don't know which part of this I think should change.

It seems reasonable that `find` includes this alias.

Maybe `edit` should reject trying to edit a builtin, and that would be fine when editing a single term, but it would be kind of annoying if a single built-in alias caused `edit 1-24` to completely fail and forced the user to do something like `edit 1-12 14-24`.

The pretty-printer is giving a pretty reasonable printing of `plus`.

The only option that comes to mind that doesn't seem terrible is to add support for defining an alias in scratch files and special-casing builtin aliases to do something like: `alias plus = ##Nat.+`.

It's hard for me to imagine this ever being a priority for someone to work on, but I figured I might as well document the behavior.

runarorama:
Fixes https://github.com/unisonweb/unison/issues/3361

Simply comments out the output when the destination is a file.

### DIFF
--- a/unison-cli/src/Unison/CommandLine/OutputMessages.hs
+++ b/unison-cli/src/Unison/CommandLine/OutputMessages.hs
@@ -2077,9 +2077,10 @@ displayDefinitions outputLoc ppe types terms =
           case dt of
             MissingObject r -> missing n r
             BuiltinObject typ ->
-              P.hang
-                ("builtin " <> prettyHashQualified n <> " :")
-                (TypePrinter.prettySyntax (ppeBody n r) typ)
+              (if isJust outputLoc then P.indent "-- " else id) $
+                P.hang
+                  ("builtin " <> prettyHashQualified n <> " :")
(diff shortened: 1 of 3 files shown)


pchiusano:
Transcript test please. 😀 Otherwise looks good.

pchiusano (approved):
👍

PR #3550, opened 2022-11-01
do an in-sqlite before-check before reconstructing FF-path proof client-side

mitchellwrosen:


pchiusano (approved):
LGTM!

aryairani (approved):

aryairani:
Merging for now based on Slack conversation, but reminder to come write up whatever notes you would have wanted to capture if we'd gone through a full PR. @mitchellwrosen

Issue #2945, opened 2022-02-23
Extra lambda-lifting remants in decompiled form for docs


pchiusano:
An example with this
```
foreach (x -> Ref.modify r (v -> v Text.++ x)) [“cat”,“hat”,“saw”]
```
expands to:
```
foreach ((r x -> Ref.modify r ((x v -> v ++ x) x)) r) [“cat”, “hat”, “saw”]
```

With what looks like a remnant of the lambda lifting pass.

PR #3553, opened 2022-11-02
Improve decompilation via undoing lambda lifting transformations

### LINKED ISSUE #3265: Lambda lifting leftovers in decompiled output
image

From: https://share.unison-lang.org/@unison/code/latest/namespaces/public/distributed/latest/;/types/@3uarccgfv6hh5312b0389kus776uqr6vkm6c8u46bd95h0k3gglmnrjaq2408vfbbgg10l3irij7avve368b9iqa6u8b6defoicher0

Not super high priority (workaround is to use a `@source{blah}` in the doc instead of an inline example) but I wonder if there's an easy fix.

dolio:
This PR adds code to the compilation pipeline to improve decompiled output. Some passes over the code are necessary before we can save decompiled output, so instead the new code does _reverse_ operations before saving the code for decompilation. E.G. when compilation encloses a lambda and produces:

    (w x y z -> ...) w x

this code beta reduces all such occurrences before saving it in the decompilation info. It tries to undo exactly the transforms that compilation does up to that point, which are reasonably detectable. It could technically reverse someone manually writing code in the odd way, however (though it should always produce semantically equivalent code, excepting some ability-related corner cases that can't be detected at this level).

Fixes #3265 and #2945


pchiusano (approved):
Nice!

Issue #3273, opened 2022-07-25
`patch` command should ignore `lib` subnamespace


pchiusano:
A typical process is you update one of your dependencies, then want to apply the patch from its new release to your codebase. Like you'd type `patch lib.something.shiny_new_version.patch` at your project root.

That does not necessarily mean you want to apply that patch to all the stuff in `lib`. For that you could `cd` into `lib`.

ceedubs:
I just noticed something related. If you run `replace foo bar` you end up with todo items within your `lib` namespace. This probably boils down to the same issue, but I thought that it would be worth explicitly calling out `replace` behavior for whomever works on this and writes test transcripts.

PR #3564, opened 2022-11-03
Refactor: make isType in Transaction rather than IO

mitchellwrosen:
## Overview

This PR moves `isType` into `Transaction` rather than `IO`


ChrisPenner (approved):
I agree that moving most of these into Transaction is the  right thing to do.

It's starting to seem a bit arbitrary now which combinators exist in the Codebase record-of-functions vs just in the Codebase module; do we have any guiding principles there or should we be moving things out of the record-of-function as we can?

I think the record of functions is mostly superfluous now except for holding the connection pool and various caches; so combinators that don't need those can probably be moved out as we go 👍🏼

mitchellwrosen:
Agree @ChrisPenner! (these are just minimal redactors I'm making while implementing #3273 but a dedicated cleanup ticket would be nice)

Issue #3565, opened 2022-11-03
try out constructing deepTypes/deepTerms more directly


aryairani:
We think we can get a quick space and time win by constructing `deepTypes`/`deepTerms` more directly, without forcing the same fields in child branches.

As a follow-on effort, we could try filtering out transitive dependencies.  i.e.
- lib/text - yes
- lib/text/lib - no
I say "try" because we don't know what the full consequences of doing so will be. We should audit the use sites of `deepTypes`/`deepTerms` to understand this better before committing such a change to trunk.

aryairani:
Fwiw I tried this super quick version and it was worse: https://github.com/unisonweb/unison/compare/trunk...arya/fast-deep
It got to 40 gb of vm allocations before it killed itself 😬 
The trunk version completed in half the time that the experimental version took to crash.

aryairani:
This version was better than the one linked in the previous comment (finished before killing itself), but still 2x as slow as trunk: https://github.com/unisonweb/unison/compare/trunk...arya/fast-deep2
cc @pchiusano

hojberg:
YAY its done!

Issue #123, opened 2016-10-13
Research optimal Data.Trie usage for Index


sfultong:
Currently Index.lookupGT traverses more of the Tries than probably necessary.

Email Data.Trie maintainer, and possibly implement PR for Data.Trie lookupGT/lookupLT functions.

sfultong:
@pchiusano FYI, mail to maintainer of this package failed.

Delivery to the following recipient failed permanently:

     wren@community.haskell.org

chreekat:
That's probably because of community.haskell.org, and probably not because of the maintainer. :) Try https://github.com/wrengr/bytestring-trie, or posting to haskell-cafe where they are fairly active.

sfultong:
Thanks! I didn't realize there was a github clone of the darcs repo

kanishka-azimi:
How did you detect it traverses more Tries than necessary?

PR #3570, opened 2022-11-04
lazy version of more direct implementation of deepTerms

aryairani:
tl;dr: In the @stew "slow-update.md" test case, 

| |total time|max memory residency| comment|
|-|-|-|-|
|trunk|309s|6.5GB| |
| ea5a2 (just eliminate recursion)| 381s |1.4GB | slower|
|7bb0a |175s | 0.5|buggy |
|**ccdcc (lib<2 or unseen)**|**190s**|**0.7GB**|best so far|
|lib<3 |330s|1.1gb|slower|

___
Implements #3565, i.e. 

* Not force all the child `deep{Terms,Types,TermMetadata,TypeMetadata,Edits,Paths}` when computing for a given node, to thereby not holding them all in memory going forward. There are only a small subset of "current namespaces" we care about, so it's wasteful to construct these big non-shared values on the child branches.

The new implementations just do a breadth-first search on the simpler fields, and assemble the appropriate names as they descend.

* Not descend into nested `lib` namespaces if the namespace has been seen before
lib/text - yes
lib/text/lib/base - maybe no

This definitely affects the `names` command, and probably affects all parsing. This could use more testing, but it didn't affect any transcripts other than the one mentioned below.

## Implementation notes

* Adds `isEmpty0 :: Bool` as a field on `Branch0`; the previous implementation forced `deep{Terms,Types,Edits}` on all child namespaces. This was also transitively used by `nonEmptyChildren`.
* Avoid `Name.cons`, which contrary to the name, adds to the slow end of a linked list. Uses `Name.fromReverseSegments` instead.
* Removes some intermediate `Relation` / `Relation4` / `Star3` rebuilding (though we're not sure of the impact, due to laziness? It wasn't helping anything though.)

## Test coverage

Added a `deep-names.md` demonstrating `names` only showing transitive deps if they are unique.
Updated `suffixes.md` to specify that names for transitive dependencies (things under two `lib`s) are not available.

## Future work

> There are only a small subset of "current namespaces" we care about.

* It would be nice to hold none of them in memory, and stream them from relational namespaces in the database or whatever.

* Where is `deepEdits'` used? I didn't try to update this function. Answer: It's used when diffing a patch as part of diffing a branch.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -221,22 +221,35 @@ deriveDeepTerms branch =
     makeDeepTerms :: Branch0 m -> [(Referent, Name)]
     makeDeepTerms branch = go [(mempty, branch)] mempty
       where
-        go :: forall m. [([NameSegment], Branch0 m)] -> [(Referent, Name)] -> [(Referent, Name)]
+        -- `reversePrefix` might be ["Nat", "base", "lib"], and `b0` the `Nat` sub-namespace.
+        -- Then `R.toList` might produce the NameSegment "+", and we put the two together to
+        -- construct the name `Name Relative ("+" :| ["Nat","base","lib"])`.
+        go ::
+          forall m.
+          [([NameSegment], Branch0 m)] ->
+          [(Referent, Name)] ->
+          [(Referent, Name)]
         go [] acc = acc
         go ((reversePrefix, b0) : work) acc =
           let terms :: [(Referent, Name)]
-              terms = map (second (Name.fromReverseSegments . (NonEmpty.:| reversePrefix))) (R.toList (Star3.d1 (_terms b0)))
+              terms =
+                map
+                  (second (Name.fromReverseSegments . (NonEmpty.:| reversePrefix)))
+                  (R.toList (Star3.d1 (_terms b0)))
            in go (deepChildrenHelper reversePrefix b0 <> work) (terms <> acc)
 
 -- | Derive the 'deepTypes' field of a branch.
-deriveDeepTypes :: Branch0 m -> Branch0 m
+deriveDeepTypes :: forall m. Branch0 m -> Branch0 m
 deriveDeepTypes branch =
   branch {deepTypes = R.fromList (makeDeepTypes branch)}
   where
     makeDeepTypes :: Branch0 m -> [(TypeReference, Name)]
     makeDeepTypes branch = go [(mempty, branch)] mempty
       where
-        go :: forall m. [([NameSegment], Branch0 m)] -> [(TypeReference, Name)] -> [(TypeReference, Name)]
+        go ::
+          [([NameSegment], Branch0 m)] ->
+          [(TypeReference, Name)] ->
+          [(TypeReference, Name)]
         go [] acc = acc
         go ((reversePrefix, b0) : work) acc =
           let types :: [(TypeReference, Name)]
@@ -244,14 +257,17 @@ deriveDeepTypes branch =
            in go (deepChildrenHelper reversePrefix b0 <> work) (types <> acc)
 
 -- | Derive the 'deepTermMetadata' field of a branch.
-deriveDeepTermMetadata :: Branch0 m -> Branch0 m
+deriveDeepTermMetadata :: forall m. Branch0 m -> Branch0 m
 deriveDeepTermMetadata branch =
   branch {deepTermMetadata = R4.fromList (makeDeepTermMetadata branch)}
   where
     makeDeepTermMetadata :: Branch0 m -> [(Referent, Name, Metadata.Type, Metadata.Value)]
     makeDeepTermMetadata branch = go [(mempty, branch)] mempty
       where
-        go :: forall m. [([NameSegment], Branch0 m)] -> [(Referent, Name, Metadata.Type, Metadata.Value)] -> [(Referent, Name, Metadata.Type, Metadata.Value)]
+        go ::
+          [([NameSegment], Branch0 m)] ->
+          [(Referent, Name, Metadata.Type, Metadata.Value)] ->
+          [(Referent, Name, Metadata.Type, Metadata.Value)]
         go [] acc = acc
         go ((reversePrefix, b0) : work) acc =
           let termMetadata :: [(Referent, Name, Metadata.Type, Metadata.Value)]
@@ -262,14 +278,17 @@ deriveDeepTermMetadata branch =
            in go (deepChildrenHelper reversePrefix b0 <> work) (termMetadata <> acc)
 
 -- | Derive the 'deepTypeMetadata' field of a branch.
-deriveDeepTypeMetadata :: Branch0 m -> Branch0 m
+deriveDeepTypeMetadata :: forall m. Branch0 m -> Branch0 m
 deriveDeepTypeMetadata branch =
   branch {deepTypeMetadata = R4.fromList (makeDeepTypeMetadata branch)}
   where
     makeDeepTypeMetadata :: Branch0 m -> [(TypeReference, Name, Metadata.Type, Metadata.Value)]
     makeDeepTypeMetadata branch = go [(mempty, branch)] mempty
       where
-        go :: forall m. [([NameSegment], Branch0 m)] -> [(TypeReference, Name, Metadata.Type, Metadata.Value)] -> [(TypeReference, Name, Metadata.Type, Metadata.Value)]
+        go ::
+          [([NameSegment], Branch0 m)] ->
+          [(TypeReference, Name, Metadata.Type, Metadata.Value)] ->
+          [(TypeReference, Name, Metadata.Type, Metadata.Value)]
         go [] acc = acc
         go ((reversePrefix, b0) : work) acc =
           let typeMetadata :: [(TypeReference, Name, Metadata.Type, Metadata.Value)]
@@ -280,14 +299,14 @@ deriveDeepTypeMetadata branch =
            in go (deepChildrenHelper reversePrefix b0 <> work) (typeMetadata <> acc)
 
 -- | Derive the 'deepPaths' field of a branch.
-deriveDeepPaths :: Branch0 m -> Branch0 m
+deriveDeepPaths :: forall m. Branch0 m -> Branch0 m
 deriveDeepPaths branch =
   branch {deepPaths = makeDeepPaths branch}
   where
     makeDeepPaths :: Branch0 m -> Set Path
     makeDeepPaths branch = go [(mempty, branch)] mempty
       where
-        go :: forall m. [([NameSegment], Branch0 m)] -> Set Path -> Set Path
+        go :: [([NameSegment], Branch0 m)] -> Set Path -> Set Path
         go [] acc = acc
         go ((reversePrefix, b0) : work) acc =
           let paths :: Set Path
@@ -298,14 +317,14 @@ deriveDeepPaths branch =
            in go (deepChildrenHelper reversePrefix b0 <> work) (paths <> acc)
 
 -- | Derive the 'deepEdits' field of a branch.
-deriveDeepEdits :: Branch0 m -> Branch0 m
+deriveDeepEdits :: forall m. Branch0 m -> Branch0 m
 deriveDeepEdits branch =
   branch {deepEdits = makeDeepEdits branch}
   where
     makeDeepEdits :: Branch0 m -> Map Name EditHash
     makeDeepEdits branch = go [(mempty, branch)] mempty
       where
-        go :: forall m. [([NameSegment], Branch0 m)] -> Map Name EditHash -> Map Name EditHash
+        go :: [([NameSegment], Branch0 m)] -> Map Name EditHash -> Map Name EditHash
         go [] acc = acc
         go ((reversePrefix, b0) : work) acc =
           let edits :: Map Name EditHash
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -700,7 +700,7 @@ deleteTermName _ _ b = b
 deleteTypeName :: TypeReference -> NameSegment -> Branch0 m -> Branch0 m
 deleteTypeName r n b
   | Star3.memberD1 (r, n) (view types b) =
-    over types (Star3.deletePrimaryD1 (r, n)) b
+      over types (Star3.deletePrimaryD1 (r, n)) b
 deleteTypeName _ _ b = b
 
 lca :: Monad m => Branch m -> Branch m -> m (Maybe (Branch m))
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -28,7 +28,6 @@ module Unison.Codebase.Branch
 
     -- * Branch tests
     isEmpty,
-    isEmpty0,
     isOne,
     before,
     lca,
@@ -86,8 +85,10 @@ where
 
 import Control.Lens hiding (children, cons, transform, uncons)
 import Data.Bifunctor (second)
+import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map as Map
 import qualified Data.Semialign as Align
+import qualified Data.Sequence as Seq
 import qualified Data.Set as Set
 import Data.These (These (..))
 import U.Codebase.Branch (NamespaceStats (..))
@@ -194,6 +195,11 @@ branch0 terms types children edits =
       _children = children,
       _edits = edits,
       -- These are all overwritten immediately
+      isEmpty0 =
+        R.null (Star3.d1 terms)
+          && R.null (Star3.d1 types)
+          && Map.null edits
+          && all (isEmpty0 . head) children,
       deepTerms = R.empty,
       deepTypes = R.empty,
       deepTermMetadata = R4.empty,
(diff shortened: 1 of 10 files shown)


pchiusano:
Other misc datapoint: I ran `find` with no args in `distributed.main` before (with `main`) and after (with this PR), and memory usage went from > 2GB max to 700MB max.

aryairani:
@pchiusano `deep{Terms,Types}` get called from `Branch.toNames` which impacts `find` and probably all `PPE` construction.

pchiusano:
One note is that if we merge this, people will need to explicitly include dependencies in `lib` that they are directly using. They can't just be like "oh, I'm using the frobnicate library, that will pull in base for me". 

This seems okay to me and is the direction we're headed anyway with new dependency management stuff, but it would need to be documented.

The other concern I have is how this will affect code browsing. I'd still like to be able to view code for my dependencies, both with the `view` command and in the local UI and on Share. If we merge this, will we just start seeing hashes everywhere?

aryairani:
We're going to switch from "don't descend into lib twice" to "don't descend into the same namespace hash twice" and see if that's reasonable performance without causing namelessness chaos in ucm and especially on share.

pchiusano:
@aryairani When you do the deduping, make sure you pick the shorter names or somehow do it breadth first. (At the very least, visit direct dependencies first) Otherwise base could get a name like httpclient.lib.http.lib.base.

aryairani:
@pchiusano It turns out that we only want to de-dupe stuff that's under `lib`? or maybe under `lib.?.lib`?

mitchellwrosen:
@aryairani We should sync up on the latest plan for this work. If we omit `lib` from `deep*` things, then there wouldn't be any additional work to address #3273, which I was otherwise going to handle with special treatment of `lib` in the propagate algorithm. 😅

aryairani:
@mitchellwrosen We still want that. As it stands we will include one level of `lib` in `deep*`, plus maybe another fraction (previously unseen transitive ilbs), but still don’t want to propagate a patch into any of that.

Eventually (maybe?) `lib` goes away altogether , and that means throwing away the work, but I think we still want it in the interim, and maybe it’s just a one-line change anyway?

pchiusano (approved):
Did you do the breadth first thing? It seems important that at least direct dependents get the short names.

aryairani:
> Did you do the breadth first thing? It seems important that at least direct dependents get the short names.

Yes

aryairani:
converted back to draft to fix a bug cody found

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +222,161 @@ branch0 terms types children edits =
...
+    "lib" : _ -> modifying _2 (+ 1)
+    _ -> pure ()
+  flip
+    Monoid.foldMapM
+    (Map.toList (nonEmptyChildren b0))
+    \(ns, b) -> do
+      let h = namespaceHash b
+      result <-
+        ifM
+          (uses _1 (Set.notMember h) ||^ uses _2 (< 2))
```
ChrisPenner: It'd be nice to have some names for readability here I think :smile:

```suggestion
          (uses (\(seenHashes, libDepth) -> Set.notMember h seenHashes || libDepth < 2))
```

Or to give names to the fields of DeepState?

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +222,161 @@ branch0 terms types children edits =
...
+                  (Name.fromReverseSegments . (NonEmpty.:| reversePrefix))
+                  (fst <$> _edits b0)
+           in do
+                children <- deepChildrenHelper reversePrefix b0
+                go (work <> children) (edits <> acc)
+
+type DeepState m = State (Set (NamespaceHash m), Sum Int)
+
+-- | Helper for knowing whether to descend into a child branch or not.
+-- Accepts child namespaces with previously unseen hashes, and any nested under fewer than 2 `lib` segments.
```
ChrisPenner: > Accepts child namespaces with previously unseen hashes

If I'm understanding this correctly, this is going to just ignore a bunch of valid names in certain situations isn't it?

E.g. if I have 

`foo.bar` where the namespace bar has hash #123
`foo.bear` where the namespace bear ALSO has hash #123, 
won't this strategy avoid adding all the names inside `foo.bear`? If so, this will have some pretty undesirable effects 😬

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +222,161 @@ branch0 terms types children edits =
...
+deepChildrenHelper :: [NameSegment] -> Branch0 m -> DeepState m [([NameSegment], Branch0 m)]
+deepChildrenHelper reversePrefix b0 = do
+  case reversePrefix of
+    "lib" : _ -> modifying _2 (+ 1)
```
ChrisPenner: Since we're in the state monad, we'll continue to accumulate this number as we go; I don't think I understand exactly how your recursive `go` works, but if it follows either of BFS or DFS I think this strategy is incorrect, since one branch of the tree affects the state for ALL branches traversed after it.

Maybe using a reader monad with `local` and explicitly recursing rather than tail-recursion would be a good fix to prevent the state from leaking outside of its subtree?

ChrisPenner: I managed to find a transcript that demonstrates the issue, notice that if we've encountered at least two libs, then AFTER that we find two identical namespaces, the second will be ignored even though it's not inside a `lib`;

So in the final names call I expect to see both `b.b.b.x` and `c.c.c.x`, but only `b.b.b.x` is shown.

    ```unison
    lib.x = 1
    a.lib.x = 2
    b.b.b.x = 5
    c.c.c.x = 5
    ```

    ```ucm

      change:
      
          a.lib.x : ##Nat
          b.b.b.x : ##Nat
          c.c.c.x : ##Nat
          lib.x   : ##Nat

    ```
    ```ucm
    .> add

        a.lib.x : ##Nat
        b.b.b.x : ##Nat
        c.c.c.x : ##Nat
        lib.x   : ##Nat

    .> names x

      Term
      Hash:   #k09e57l61r
      Names:  b.b.b.x
      
      Tip: Use `names.global` to see more results.

    ```

ChrisPenner:
Glad to see this is yielding results! I think the current implementation has a bug or two that I posted a transcript for in a comment;

> It would be nice to hold none of them in memory, and stream them from relational namespaces in the database or whatever.

Yeah, I think we should probably just do this, I don't think it'd be too terribly hard and would remove one more usage of deep*.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Yeah this is the bug I converted back to Draft for 😞

ChrisPenner: Let me know if you'd like a hand fixing it, I have some ideas 😄

aryairani:
Starting to feel pessimistic about this approach; the version that dedups based on hash but doesn't accidentally prune a ton of stuff seems to be way slower than `trunk`. Not sure why yet. @pchiusano 

If we just hard prune below lib depth 2, that might work, but we did experience some weirdness there that we'd have to work around with `view` etc.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Good idea

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: That should only be the case if foo.bar and foo.bear are nested under at least 2 levels of `lib`, which seems okay? It means we will have a name for PPE but might be missing a name for parsing that we don't need anyway. We just have to do like Paul commented and alert users that they need to add all their direct dependencies.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Ok, this is fixed in the latest:

````
```unison
lib.x = 1
a.lib.x = 2
b.b.b.x = 5
c.c.c.x = 5
```

```ucm

  change:
  
      a.lib.x : ##Nat
      b.b.b.x : ##Nat
      c.c.c.x : ##Nat
      lib.x   : ##Nat

```
```ucm
.> add

    a.lib.x : ##Nat
    b.b.b.x : ##Nat
    c.c.c.x : ##Nat
    lib.x   : ##Nat

.> names x

  Term
  Hash:   #k09e57l61r
  Names:  b.b.b.x c.c.c.x
  
  Tip: Use `names.global` to see more results.

```
````

File: parser-typechecker/src/Unison/Codebase/Branch.hs (hunk 2 above)
Comment (ChrisPenner) [reply parser-typechecker/src/Unison/Codebase/Branch.hs]: > That should only be the case if foo.bar and foo.bear are nested under at least 2 levels of lib, which seems okay?

The previous version accumulated "seen libs" rather than only applying them for nested subnamespaces which is why I was concerned, since this would apply for _all_ namespaces not just libs, but now that you've fixed that and only apply the skip when in nested libs this concern should go away 👍🏼

**Comment by tstat** [comment]
In the transcript provided by Stew we can see ~8GB or ~95% of the heap is allocated and retained by forcing deepTerms on trunk. This is brought down to ~200MB in this PR. This is a big improvement that will give us some breathing room as we work towards relational namespaces.

# Trunk d43fb46a98166695a4692b19d29457975312342a
trunk

# This PR ccdcc44b7b5fadd6d06e1b38727772f5106aea28
fastdeep2

**Review by tstat — APPROVED** [review]

**Comment by pchiusano** [comment]
@tstat amazing!

File: parser-typechecker/src/Unison/Codebase/Branch.hs [hunk 4]
```diff
@@ -618,13 +700,13 @@ addTypeName r new md =
 deleteTermName :: Referent -> NameSegment -> Branch0 m -> Branch0 m
 deleteTermName r n b
   | Star3.memberD1 (r, n) (view terms b) =
-      over terms (Star3.deletePrimaryD1 (r, n)) b
+    over terms (Star3.deletePrimaryD1 (r, n)) b
```
mitchellwrosen: Pre-0.4.0.0 `ormolu` is back from the grave

parser-typechecker/src/Unison/Codebase/Branch/Type.hs
```diff
@@ -65,6 +69,11 @@ data Branch0 m = Branch0
     -- Every level in the tree has a history.
     _children :: Map NameSegment (Branch m),
     _edits :: Map NameSegment (EditHash, m Patch),
+    -- | Checks whether a Branch0 is empty, which means that the branch contains no terms or
+    -- types, and that the heads of all children are empty by the same definition.
+    -- This is not as easy as checking whether the branch is equal to the `empty0` branch
+    -- because child branches may be empty, but still have history.
+    isEmpty0 :: Bool,
```
mitchellwrosen, line 75: I think I see - this is part of `Branch` so that it's `O(n)` rather than `O(m)`, where `n` is the number of direct children, and `m` is the number of transitive children? (And we expect `m` to be large-ish, say thousands, for average-sized Unison libraries, given self-contained transitive deps?)

If so it would be good to capture that in this comment, I think.

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +220,163 @@ branch0 terms types children edits =
...
+        go (e@(reversePrefix, _, b0) Seq.:<| work) acc = do
+          let edits :: Map Name EditHash
+              edits =
+                Map.mapKeysMonotonic
+                  (Name.fromReverseSegments . (NonEmpty.:| reversePrefix))
+                  (fst <$> _edits b0)
+          children <- deepChildrenHelper e
+          go (work <> children) (edits <> acc)
+
+type DeepState m = State (Set (NamespaceHash m))
```
mitchellwrosen, line 349: Opinion: these aliases (maybe all aliases) hurt more than they help. But feel free to ignore.

mitchellwrosen: What is this set, though?

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +220,163 @@ branch0 terms types children edits =
...
+
+type DeepChildAcc m = ([NameSegment], Int, Branch0 m)
```
mitchellwrosen, line 353: Could you document what these fields mean?

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +220,163 @@ branch0 terms types children edits =
 -- | Derive the 'deepTerms' field of a branch.
 deriveDeepTerms :: Branch0 m -> Branch0 m
 deriveDeepTerms branch =
-  branch {deepTerms = makeDeepTerms (_terms branch) (nonEmptyChildren branch)}
+  branch {deepTerms = R.fromList (makeDeepTerms branch)}
```
mitchellwrosen, line 223: Just curious: was memory usage (or something) worse with a `Relation Referent Name` accumulator, rather than this `[(Referent, Name)]` accumulator that's `R.fromList`'d at the end?

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +220,163 @@ branch0 terms types children edits =
...
+          isUnseenTransitiveDep <- State.gets (Set.notMember h)
+          pure
+            if isDirectDependency || isUnseenTransitiveDep
+              then Seq.singleton (ns : reversePrefix, libDepth', head b)
+              else Seq.empty
+        State.modify' (Set.insert h)
+        pure result
+  Monoid.foldMapM go (Map.toList (nonEmptyChildren b0))
+
+-- | Flattens a Metadata.Star into a 4-tuple.
```
mitchellwrosen: I see this is almost `Metadata.starToR4`, except that returns this list in `R4`'d form. Maybe this belongs in that module anyway though, as `Metadata.starToR4List` or something?

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +220,163 @@ branch0 terms types children edits =
...
+                  (Name.fromReverseSegments . (NonEmpty.:| reversePrefix))
+                  (fst <$> _edits b0)
+          children <- deepChildrenHelper e
+          go (work <> children) (edits <> acc)
+
+type DeepState m = State (Set (NamespaceHash m))
+
+type DeepChildAcc m = ([NameSegment], Int, Branch0 m)
+
+-- | Helper for knowing whether to descend into a child branch or not.
```
mitchellwrosen, line 355: Something I'm not totally understanding: `libDepth` tracking. I'm wondering whether the logic here allows one to use the `lib` name segment within their library (e.g. `lens.prisms.lib.whobaloo` or something, where `lens.lib.*` would be len's direct dependencies).

It seems to me that the (e.g.) terms we want to include in a branch's `deepTerms` are:

- shallow terms (i.e. this branch's terms that happen to have a 1 segment name)
- all non-`lib` childrens' deep terms (i.e. this branch's terms that happen to have a 2+ segment name)
- all `lib` childrens' shallow terms (i.e. direct dependencies' terms that happen to have a 1 segment name)
- all `lib` childrens' non-`lib` childrens' deep terms (i.e. direct dependencies' terms that happen to have a 2+ segment name)

And none of that requires any stateful tracking of how many `"lib"` we've seen.

unison-cli/src/Unison/Codebase/Editor/Output.hs
```diff
@@ -323,7 +323,8 @@ isFailure o = case o of
   BadMainFunction {} -> True
   CreatedNewBranch {} -> False
   BranchAlreadyExists {} -> True
-  FindNoLocalMatches {} -> True
+  -- we do a global search after finding no local matches, so let's not call this a failure yet
```
mitchellwrosen, line 326: Unrelated to this PR but this whole function is weird, and this change highlights that. I think we should get rid of it and stop classifying stanzas in transcripts as "I expect success" or "I expect an error" - couldn't the golden test output just be inspected, and committed if it looks right, without switching on this bool?

mitchellwrosen (approved):

unison-cli/src/Unison/Codebase/Editor/Output.hs
aryairani: I think the idea was to have a sense of overall success or failure for a transcript; e.g. we currently run `transcripts` and count the number of successes and failures, and list the failures, so you can go back and check.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Yeah good call, I do think it belongs in that module.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: > I'm wondering whether the logic here allows one to use the lib name segment within their library

True, it doesn't; but we haven't come up with a better alternative

> It seems to me that the (e.g.) terms we want to include in a branch's deepTerms are

I didn't understand the alternative 😅 and also did think it still required tracking

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -211,80 +220,163 @@ branch0 terms types children edits =
...
+deepChildrenHelper (reversePrefix, libDepth, b0) = do
+  let libDepth' = case reversePrefix of
+        "lib" : _ -> libDepth + 1
+        _other -> libDepth
+  let go :: (NameSegment, Branch m) -> DeepState m (Seq (DeepChildAcc m))
+      go (ns, b) = do
+        let h = namespaceHash b
+        result <- do
+          let isDirectDependency = libDepth' < 2
+          isUnseenTransitiveDep <- State.gets (Set.notMember h)
```
ChrisPenner: `isUnseenTransitiveDep` implies that we know we're in a lib at this point, but that's not necessarily true is it? Couldn't this just be a regular ol' user namespace with a libDepth of 0?

The behaviour seems right, but I'd maybe go with `isUnseenNamespace`

ChrisPenner (approved):
Part of me still wonders whether we should include transitive deps in names at all, since it allows you to use definitions from your transitive deps in your application code without explicitly depending on that library, but maybe that's a feature and not a bug 🤷🏼‍♂️ 

It might be kind of weird that if I want to refer to the version of the Time object from lib `foo` I can do `lib.foo.lib.time.Time`, but it might not work to do `lib.bar.lib.time.Time` if the versions of time are the same, since the second would be culled 🤔 ; but I think it's worth merging anyways because of the huge perf gainz 💪🏼 .

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: It is annoying that Haskell type aliases / declarations can't be local to functions.

parser-typechecker/src/Unison/Codebase/Branch/Type.hs
aryairani: Good catch, I copied the comment from elsewhere without thinking.

aryairani: It's part of `Branch0` here because its definition previously depended on `deepTerms`/`deepTypes` the whole way down the tree, and we want to force as few of those as possible because the result is quite big. However, I didn't test whether they were being forced to any problematic degree, and maybe they weren't. Nonetheless, it removes one usage of `deepTerms`/`deepTypes`, which I think is probably a good thing.

I'm struggling to confirm the complexity especially in the presence of laziness. It will check `isEmpty0` on each child, however a first evaluation could force up to the entire subtree, `O(m)` time and memory where `m` is the number of descendants. But that memory usage is probably smaller by a constant factor than the thunk before it was forced. So maybe up to `O(m)` time and negative memory?

What would be a good comment in this case?

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: IIRC the thinking was that `[]` was easier to understand and might be significantly better and couldn't be worse, especially in the `Relation4` case and possibly in the `Relation` case too.

@tstat Do you remember a better answer?

aryairani:
> Part of me still wonders whether we should include transitive deps in names at all, since it allows you to use definitions from your transitive deps in your application code without explicitly depending on that library, but maybe that's a feature and not a bug 🤷🏼‍♂️

My first instinct was also that we shouldn't. However, we want a seamless way to view the source of at least your direct dependencies and eventually of your indirect dependencies as well without needing to force a context switch.

> It might be kind of weird that if I want to refer to the version of the Time object from lib `foo` I can do `lib.foo.lib.time.Time`, but it might not work to do `lib.bar.lib.time.Time` if the versions of time are the same, since the second would be culled 🤔 ; but I think it's worth merging anyways because of the huge perf gainz 💪🏼 .

Yeah it is weird. We will have to teach people to make direct dependencies if they want to reference them in new code.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Ah, maybe I understand?

It's not captured in this thread, but we'd decided we want to include all the direct dependencies (<=1 lib) and at least one shallow-ish copy of transitive dependency names to allow `view` to continue to work at any depth without showing references for the time being.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: It's the set of visited namespace hashes.

parser-typechecker/src/Unison/Codebase/Branch.hs
aryairani: Sorry... I'm not actually sure how to fix this the ormolu issue.
```suggestion
      over terms (Star3.deletePrimaryD1 (r, n)) b
```

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -624,7 +700,7 @@ deleteTermName _ _ b = b
 deleteTypeName :: TypeReference -> NameSegment -> Branch0 m -> Branch0 m
 deleteTypeName r n b
   | Star3.memberD1 (r, n) (view types b) =
-      over types (Star3.deletePrimaryD1 (r, n)) b
+    over types (Star3.deletePrimaryD1 (r, n)) b
```
aryairani:
```suggestion
      over types (Star3.deletePrimaryD1 (r, n)) b
```

unison-cli/src/Unison/Codebase/Editor/Output.hs
aryairani: Maybe "transcript failed" or "transcript succeeded" also doesn't mean much locally, maybe we'd need to use `git` to decide if there was a "failure", and maybe that's a good idea.

parser-typechecker/src/Unison/Codebase/Branch/Type.hs
mitchellwrosen: Oops, yeah -- disregard my complexity "analysis" -- the real difference is (clearly) that this caches the `Bool` whereas a top-level definition would compute it anew on every call.

Maybe since we didn't measure, a good comment would be "this seems smart to cache in the data structure in case we ask for it multiple times". But not saying that is also good with me.

aryairani: Ok, I think I improved it.

PR #3573, opened 2022-11-05
Refactor: make dependentsOfComponent in Transaction, not IO

mitchellwrosen:
- [x] Depends on #3572 
  - [x] Depends on #3571 

---

## Overview

This PR makes `dependentsOfComponent` in `Transaction` rather than `IO`. And with that, update propagation now occurs within a single transaction.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -399,7 +399,7 @@ dependents c selector r =
     . Set.map Reference.DerivedId
     <$> dependentsImpl c selector r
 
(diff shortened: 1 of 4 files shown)


ChrisPenner (approved):
Did you make any significant changes or is it just simple refactoring?

I'll rubber stamp because it seems straight-forward, let me know if you want a more in-depth review 👍🏼

mitchellwrosen:
> Did you make any significant changes or is it just simple refactoring?

No significant changes; probably "Hide whitespace changes" would make that a lot more clear since a big chunk of the lines changed were due to indentation

PR #3574, opened 2022-11-05
Refactor: remove a couple functions from the Codebase record

mitchellwrosen:
- [x] Depends on #3573 
  - [x] Depends on #3572 
    - [x] Depends on #3571 
- [x] Depends on #3576 

---

## Overview

This PR removes a bunch of functions from the `Codebase` record, which simplifies their call sites. There's no need to go through the `Codebase` record indirection, when simply running a transaction with a connection in scope (usually/always? provided by a `Codebase` record) will do.


ChrisPenner (approved):
Reminder to please update Enlil if it uses any of these combinators too please.

PR #3576, opened 2022-11-07
cleanup: add a couple cached transaction helpers

mitchellwrosen:
## Overview

This PR just adds a couple helpers to unify/simplify how we create transactions that hit a cache before SQLite.


ChrisPenner (approved):
Looks good 👍🏼 

Maybe add a note to the helpers that they should be used exclusively on READ-ONLY transactions.

aryairani:
cc @tstat just to be aware of our various caches

PR #3578, opened 2022-11-07
Refactor: remove getShallowCausalByHash from Codebase record

mitchellwrosen:
- [x] Depends on #3577 
  - [x] Depends on #3574 
    - [x] Depends on #3573 
      - [x] Depends on #3572 
        - [x] Depends on #3571 
    - [x] Depends on #3576

---

## Overview

This PR removes `getShallowCausalByHash` from the `Codebase` record. It also does not "transform" the returned `Causal Transaction` (returning `Causal IO`), but just leaves it in `Causal Transaction`. If a caller needs to run the embedded transactions to go back in history (which no current callers do), then they can `runTransaction` themselves.


ChrisPenner (approved):
Just skimmed it, but LGTM

PR #3589, opened 2022-11-09
Implement commands for building/running via scheme

dolio:
This PR implements support for using chez scheme to run unison programs. Much of the machinery is written in unison itself, so the stuff here is for wrappers around/support of that. Here's an overview of the parts:

- The new `chez-libs` directory contains some written scheme files that we'll need to distribute with unison. These are not complete yet, but they're working enough for some basic purposes. They implement primops and other underlying machinery that makes it easier to generate scheme definitions from unison intermediate code.
- I've added multiple commands for interacting with the scheme compiler
  * One command downloads the unision implementation that generates scheme and puts it in a standard location in the codebase
  * One command generates some scheme library files _from_ unison. These are more along the lines of unison implementations of builtin functionality, like the above scheme files. They're the same for every program, so they can be generated once and cached for a while.
  * One command generates scheme for a 'main' function and just has chez immediately execute it, similar to `run`. You just see output from chez, though, so it's not quite as integrated as the actual `run` command. It should be possible to beef it up in the future, though; this was just easy to hook up, and can run certain things a lot faster (like adding up 1 a million times).
  * One command is like the above, but generates a compiled binary. I haven't quite gotten this one working, because the actual sequence of commands for producing a working binary is eluding me so far, but most of the rest of the work is shared with the above 'run' function.
- You don't need to run all the commands above in sequence. Just doing a `run.scheme` will download the compiler and generate the libraries. But the dedicated commands will force updates, while the run/compile commands will check if you already have something.

I tried to factor some of my stuff out of `HandleInput` into a separate file, but a fair amount seemed like that wasn't actually possible without doing more refactoring of other commands that I'm not very familiar with.

I also implemented serialization of byte arrays, because the unison stuff gets its intermediate representations via binary serialization, and some of the compiler code uses byte arrays. So to compile itself, it requires byte arrays to be serializable.

Some stuff definitely won't work. The implementation of abilities is overly naive, because I used the simplest possible implementation of delimited continuations just to get something barely working. So, stuff like nested handlers with multiple abilities needs this to be fleshed out. But I think it's good enough for simple exception stuff in the mean time. There are also lots of missing builtins and such, but this was an exercise in getting things wrapped so that some amount of trial could be done by people other than me. Fleshing things out comes next.

I'm not exactly sure how to go about getting the more static scheme files installed, or where they should be. Right now I've been manually putting them somewhere that is probably not where they should actually go, and that path should be changed. I think that's the purview of Arya's team. I did put some logic in so that the config file can affect where we look for that, in case people want to tweak, though.


dolio:
This is getting close to being able to be merged, I think. Here are some outstanding things I can think of that I could use some assistance with:

1. Scheme library locations mentioned above. Where should they go, and how do we get them there?
2. Some further explanation on how to use the commands is probably required, but I'm not sure where to put it. Like, chez scheme needs to be installed and on the path. Where should that be explained? The ucm help for the commands doesn't seem ideal.
3. @mitchellwrosen I had moved some stuff you changed, so I tried to move your changes to corresponding places. It had to do with putting some type lookup stuff in a transaction monad. Can you take a look and make sure I didn't screw it up? I think the places to look are `TypeCheck.hs` and `TermResolution.hs`.

Possibly I'll come up with some more later.

pchiusano:
@dolio AMAZING!!! I can't wait.

Can you show us at least an informal transcript demoing usage of it? Or is it possible to have a proper transcript that exercises this? 

Idea: maybe you could add a `run.native` command which does the equivalent of `compile.native` followed by just running the resulting executable / scheme.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
         _ -> Left $ showPatternHelp makeStandalone
     )
 
+runScheme :: InputPattern
+runScheme =
+  InputPattern
+    "run.scheme"
```
pchiusano: @dolio oh I see you've already added this! Nice. So we can have transcripts that use this!! Epic.

Can you rename this to `run.native` please? I get that `scheme` is more accurate but it's meant to be an implementation detail as far as the user is concerned.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+    ( \case
+        [main] ->
+          Input.ExecuteSchemeI <$> parseHashQualifiedName main
+        _ -> Left $ showPatternHelp runScheme
+    )
+
+compileScheme :: InputPattern
+compileScheme =
+  InputPattern
+    "compile.scheme"
```
pchiusano:
```suggestion
    "compile.native"
```

same deal here.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+    []
+    I.Visible
+    [(Required, exactDefinitionTermQueryArg), (Required, noCompletionsArg)]
+    ( P.wrapColumn2
+        [ ( "`compile.scheme main file`",
```
pchiusano:
```suggestion
        [ (IP.makeExample compileScheme ["main", "file"],
```

Rather than hardcoding the backticked command name.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+    ( \case
+        [main, file] ->
+          Input.CompileSchemeI file <$> parseHashQualifiedName main
+        _ -> Left $ showPatternHelp compileScheme
+    )
+
+schemeLibgen :: InputPattern
+schemeLibgen =
+  InputPattern
+    "compile.scheme.genlibs"
```
pchiusano:
```suggestion
    "compile.native.genlibs"
```

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+    []
+    I.Visible
+    []
+    ( P.wrapColumn2
+        [ ( "`compile.scheme.genlibs`",
```
pchiusano:
```suggestion
        [ (IP.makeExample schemeLibgen [],
```

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+            "Generates libraries necessary for scheme compilation.\n\n\
+            \There is no need to run this before `compile.scheme`, as\
```
pchiusano:
```suggestion
            \There is no need to run this before" <> P.group (IP.makeExample compileScheme []) <> " as\
```

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+    )
+    ( \case
+        [] -> pure Input.GenSchemeLibsI
+        _ -> Left $ showPatternHelp schemeLibgen
+    )
+
+fetchScheme :: InputPattern
+fetchScheme =
+  InputPattern
+    "compile.scheme.fetch"
```
pchiusano:
```suggestion
    "compile.native.fetchlib"
```

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+    []
+    I.Visible
+    []
+    ( P.wrapColumn2
+        [ ( "`compile.scheme.fetch`",
```
pchiusano:
```suggestion
        [ ( IP.makeExample fetchScheme [],
```

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2116,6 +2116,92 @@ makeStandalone =
...
+            "Fetches the unison library for compiling to scheme.\n\n\
+            \This is done automatically when `compile.scheme` is run\
```
pchiusano:
```suggestion
            \This is done automatically when " <> IP.makeExample compileScheme [] <> " is run\
```

pchiusano (approved):
Left some comments on renaming the commands. Otherwise will wait for feedback from Arya's team.

I'd like to see a transcript added to this PR to demo / test if that's possible, which it seems like it is?

Very exciting!!

dolio:
Do you just want a transcript _somewhere_ in the repo? If I put it in one of the usual directories, we'll have to figure out how to get chez installed for CI, right? Maybe that can be done, but I don't really know much about how to do it.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
dolio: Yeah, I actually figure out how to do this first, because when you call the scheme binary with a program file, it just runs it. To compile it, you need to actually start the repl and feed it commands to compile your stuff, so it's more complicated.

dolio:
Oh, one more thing I should mention.

Apparently chez doesn't produce single, self-contained binaries like I thought when beginning this.  It actually produces files that are kind of similar to our `.uc` files, but they are `.boot` instead. It's possible to get a sort of stand-alone program, but the way you do that is actually to produce a boot file, put it in a particular location, and make a copy of the scheme binary with a matching name. If the scheme binary is named `my-program`, then it will look for `my-program.boot` and execute it.

However, on unix, the `.boot` locations don't include the directory that the program is in. So that's a bit more complicated. Apparently on Windows it does, so you could just put both the copy of the scheme binary and the boot file in the same directory, and it would run like a normal program. But on unix-based systems, you need to install the boot file somewhere else, or you need to pass command line arguments specifying what to do, so a script might still be needed for the latter.

pchiusano:
@dolio sure, you can just put it somewhere in the repo, and create a separate ticket for figuring out how to get it to run in CI. I feel like CI could just download a chez binary from somewhere on the internet but there might be some annoying stuff dealing with multiple platforms.

dolio:
Hmm, we might not want this in CI anyway, because it has to fetch all of base. At least, not until most of base is pre-packaged for starting new codebases.

PR #3595, opened 2022-11-10
Refactor: make updateNameLookup in Transaction, not IO

mitchellwrosen:
## Overview

This PR makes `updateNameLookup` in `Transaction`, not `IO`


ChrisPenner (approved):
Thanks for going through all of these 👍🏼 

This one is going to need an update in Enlil as well btw.

aryairani (approved):

mitchellwrosen:
@ChrisPenner Hm, I went to update `enlil` but I don't see this function used. Maybe it's dead code.

mitchellwrosen:
Oops, disregard - case-sensitive grep mishap - it is used. But not the `Codebase` version defined here, just the underlying query which accepts a cached decltype lookup function. So maybe we just remove this from the `Codebase` record.

Issue #3603, opened 2022-11-14
Universal.compare on lists should probably be lexicographic


pchiusano:
Right now `Universal.compare [1,2,3] [3]` returns `+1`, since the `[3]` list is shorter.

In general, I'd like `Universal.compare` to be a simple structural comparison done lexicographically.

Question: do we think it's okay to just change this behavior now? Or should we introduce a new builtin with the new behavior?

runarorama:
Seems OK to change it as the current behavior is probably not very useful.

hojberg:
Yay we fixed it!

PR #3606, opened 2022-11-15
Make universal compare do lexicographic comparison on lists

dolio:
Equality still checks length first, because that should be able to short circuit a bit faster.

Addresses #3603

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Machine.hs
+++ b/parser-typechecker/src/Unison/Runtime/Machine.hs
@@ -2159,7 +2159,7 @@ universalCompare frn = cmpc False
     cmpc tyEq (Foreign fl) (Foreign fr)
       | Just sl <- maybeUnwrapForeign Rf.listRef fl,
         Just sr <- maybeUnwrapForeign Rf.listRef fr =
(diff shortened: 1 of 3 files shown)


parser-typechecker/src/Unison/Runtime/Machine.hs
```diff
@@ -2159,7 +2159,7 @@ universalCompare frn = cmpc False
     cmpc tyEq (Foreign fl) (Foreign fr)
       | Just sl <- maybeUnwrapForeign Rf.listRef fl,
         Just sr <- maybeUnwrapForeign Rf.listRef fr =
-          comparing Sq.length sl sr <> fold (Sq.zipWith (cmpc tyEq) sl sr)
+          fold (Sq.zipWith (cmpc tyEq) sl sr)
```
pchiusano, line 2162: I think this isn't correct - it will have `[1,2,3]` equal to `[1,2,3,4]`, right? Since the zipWith cuts off at the shorter list.

I think putting the length comparison after the `zipWith` comparison would do it, but could you add a few tests to `builtins.md` transcript?

dolio: Oh right. Good catch.

pchiusano (approved):
Cool. LGTM!

PR #3611, opened 2022-11-16
Clean up applyPropagate

tstat:
## Overview

Codebase updates are very slow for Stew's codebase. I observe ~76 seconds to execute Stew's update transcript on trunk (6bacb88e614a9883a107bcf50365f8a9025647b3). Profiling revealed that the bulk of the time was spent in `applyPropagate`'s `clearPropagated` call.  `clearPropagated` deletes the propagated metadata tags for all patch targets by looping over the patch targets and deleting them from the metadata Star3.

The call to `clearPropagated` appears to be unnecessary: `replaceFacts` traverses the intersection of the domain of (term/type)Edits and the corresponding metadata Star3 and properly assigns the propagated metadata values (see  `replaceTerms`/`replaceTypes`). So, `clearPropagated` would only be useful if it corrected metadata for references not in this intersection. If a reference is not in the Star3 then there is nothing to correct. If a reference is in the Star3 but not in the domain of `termEdits` then it would be erroneous to label it as not propagated.

I observe ~21 seconds to execute Stew's update transcript with this change.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/Propagate.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/Propagate.hs
@@ -650,9 +650,9 @@ applyPropagate patch Edits {..} = do
             typeOf r t = fromMaybe t $ Map.lookup r termTypes
 
         replaceTerm :: Referent -> Referent -> Metadata.Star Referent NameSegment -> Metadata.Star Referent NameSegment
-        replaceTerm r r' s =
+        replaceTerm _r r' s =
           ( if isPropagatedReferent r'
-              then Metadata.insert (propagatedMd r') . Metadata.delete (propagatedMd r)
+              then Metadata.insert (propagatedMd r')
               else Metadata.delete (propagatedMd r')
           )
             $ s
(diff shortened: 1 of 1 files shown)


unison-cli/src/Unison/Codebase/Editor/Propagate.hs
```diff
@@ -632,23 +632,32 @@ applyPropagate patch Edits {..} = do
...
-          updateMetadatas Referent.Ref $
+          updateMetadatas projRef $
             Star3.replaceFacts replaceTerm termEdits terms0
         types :: Branch.Star Reference NameSegment
         types =
-          updateMetadatas id $
+          updateMetadatas Just $
             Star3.replaceFacts replaceType typeEdits _types
 
+        projRef :: Referent -> Maybe Reference
```
mitchellwrosen: Suggestion: `Referent.toTermReference`

mitchellwrosen (approved):

ChrisPenner (approved):

PR #3628, opened 2022-11-21
Indent after newline in pattern lists

### LINKED ISSUE #3627: Virtual semicolon introduced in long pattern lists
In someting like

```
foo = cases 
  x, y -> bar x y
```

If `x` and `y` are long enough to reach the end of the line, the pretty-printer adds a soft break after the `,`. This will cause a virtual semicolon to be introduced by the lexer and the pattern fails to parse.

runarorama:
Fixes #3627 

Long pattern lists were failing to roundtrip due to a virtual semicolon after a comma separator.

The fix is to just indent after the first newline in the pattern list to avoid confusing the lexer.


pchiusano (approved):

pchiusano:
Cool. Applied ready-to-merge for after CI completes.

PR #3634, opened 2022-11-22
Eta-reduce Doc term links to undo compiler expansion

### LINKED ISSUE #3582: Regression in links to constructor terms in docs
Doc links like `{{ {Some} }}` used to render as (from current Share version):

image

Now they get rendered as (latest trunk):

image

So it looks like the term has been eta-expanded. I'm guessing this happens at compile-time, since this is only occurring with new docs, not e.g. docs pulled from Base.

runarorama:
Fixes #3582

For a doc like this:

````
medoc = {{ 

```
x -> Some x
```

{Some}

}}
````

Rendering before: 

```
      x -> Some x
      ⧨
      Some

  Right (Term.Term (Any '(x -> Some x)))
```

Rendering after: 

```
      x -> Some x
      ⧨
      Some

  Some
```

That last `Some` is properly rendered as a link to the `Optional` type.

### DIFF
--- a/unison-src/transcripts/fix3634.md
+++ b/unison-src/transcripts/fix3634.md
@@ -1,5 +1,4 @@
 ```ucm:hide
-.> builtins.merge
 .> builtins.mergeio
 ```
 
--- a/unison-src/transcripts/fix3634.md
+++ b/unison-src/transcripts/fix3634.md
@@ -1,0 +1,22 @@
+```ucm:hide
+.> builtins.merge
+.> builtins.mergeio
+```
+
+
+```unison
+structural type M a = N | J a
+
+d = {{
+
+{{ docExample 0 '(x -> J x) }}
+
+{J}
+
+}}
+```
+
+```ucm
+.> add
+.> display d
+```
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by pchiusano** [comment]
Nice! A test which fails on trunk but passes here would be good though.

**Comment by runarorama** [comment]
Any ideas how to write such a test?

**Comment by pchiusano** [comment]
@runarorama a transcript that uses `display`. I guess it doesn't need to fail, but we can visually inspect the output on trunk and in this branch and confirm it looks as expected.

**Comment by runarorama** [comment]
I am not sure how to get a doc into a transcript, since docs require Base.

**Comment by runarorama** [comment]
`builtins.mergeio` does it, apparently

File: unison-src/transcripts/fix3634.md [hunk 1]
```diff
@@ -0,0 +1,22 @@
+```ucm:hide
+.> builtins.merge
```
pchiusano: fyi you don't need both of these. you can just `builtins.mergeio`.

pchiusano (dismissed):
@runarorama left one minor comment if you'd like to clean up. You can merge whenever.

pchiusano:
@runarorama check out `Server/Doc.hs` the `renderDoc` function, you'll want to update that as well, and at least test it manually using the `ui`.

runarorama:
image

pchiusano (approved):
😎

PR #3642, opened 2022-11-29
Chore: bump to lts-20.0

mitchellwrosen:
## Overview

Redoing #3631, this time landing on GHC 9.2.4


aryairani:
Superficial HLS test seems okay. I do have to open the package containing the definitions I want to jump to before I can jump to them, which seems kind of pointless, but I'm not sure it didn't already work that way.

aryairani (dismissed):

aryairani:
Reminder to create corresponding PR for enlil.

stack.yaml
```diff
@@ -52,26 +51,23 @@ extra-deps:
...
-# 2.3.27 bundles sqlite >=3.35.0, needed for 'delete returning'
-- direct-sqlite-2.3.27
-# not in lts-18.13
-- recover-rtti-0.4.0.0@sha256:2ce1e031ec0e34d736fa45f0149bbd55026f614939dc90ffd14a9c5d24093ff4,4423
-- lock-file-0.7.0.0@sha256:3ad84b5e454145e1d928063b56abb96db24a99a21b493989520e58fa0ab37b00,4484
-- http-client-0.7.11
-- lsp-1.5.0.0
-- lsp-types-1.5.0.0
-- text-rope-0.2@sha256:53b9b4cef0b278b9c591cd4ca76543acacf64c9d1bfbc06d0d9a88960446d9a7,2087
+  commit: 3b26f1e6c279e9bb5cd052c34c852aa8f8cfa17e
```
aryairani: Sanity check: did you ensure color renders properly in a Windows terminal, or what commit is this?

unison-cli/src/Unison/CommandLine/Welcome.hs
```diff
@@ -166,7 +166,7 @@ authorSuggestion :: P.Pretty P.ColorText
 authorSuggestion =
   P.newline
     <> P.lines
-      [ P.wrap "📜 🪶 You might want to set up your author information next.",
+      [ P.wrap "📜 You might want to set up your author information next.",
```
aryairani, line 169: Can we not use the 🖊️ character anymore?

unison-hashing-v2/src/Unison/Hashing/V2/Term.hs
```diff
@@ -174,8 +174,6 @@ instance Var v => Hashable1 (F v a p) where
                     B.Recorded (B.Resolve _ s) ->
                       [tag 2, Hashable.Text (Text.pack s)]
                 Ref (Reference.Builtin name) -> [tag 2, accumulateToken name]
-                Ref Reference.Derived {} ->
-                  error "handled above, but GHC can't figure this out"
```
aryairani, line 178: 🎉

stack.yaml
aryairani: I'm testing it, but it will take a while.

mitchellwrosen: This commit is morally the same: the PR that we were pointing at was rebased by the author atop the latest main so it continues to build. The PR still isn't merged :/

aryairani: image
This is what I get; let me check again on trunk.

mitchellwrosen: Oh. Sorry for making you do all of that, but I understand now.

I too went looking for "what is commit `d6c2643b0d5c19be7e440615c6f84d603d4bc648`?" when trying to fix compiler errors with `haskeline` due to the bump, and I saw it in the history of the open PR we were depending on:

image

But I got the history wrong. Looks like the author was not just force-pushing over their PR to keep it atop the main branch, they changed the implementation, too. 

image

Meanwhile @ChrisPenner (maybe independently) discovered that the original approach was correct, so pointed us at `d6c2643b0d5c19be7e440615c6f84d603d4bc648`.

So I believe the correct move is to grab `haskeline` master (since it builds), and revert https://github.com/judah/haskeline/commit/f827f10145380ec9e685a72e4b8a7a27ef6717ea on top of it (which is how `d6c2643b0d5c19be7e440615c6f84d603d4bc648` came to be).

unison-cli/src/Unison/CommandLine/Welcome.hs
mitchellwrosen: I haven't tested on GHC 9.2.4. On GHC 9.0.2, no, which was weird. But I didn't investigate too closely because I thought putting two emojis in a row here was a mistake, anyway.

stack.yaml
mitchellwrosen: @aryairani I've updated this dependency. Now it's pointing at our `unisonweb/haskeline` fork, branch `lts-20.0`. The contents of the branch are just `judah/haskeline` HEAD, plus [this commit](https://github.com/unisonweb/haskeline/commit/6df6d5a0711dd1056b6182559281b31ecfc24773), which is equivalent to [the previous commit we were using](https://github.com/judah/haskeline/commit/d6c2643b0d5c19be7e440615c6f84d603d4bc648) (although now rebased on top of `judah/haskeline` HEAD, because the old commit didn't build against `lts-20.0`).

Some confusion: we have a [unison branch](https://github.com/unisonweb/haskeline/tree/unison) (the default branch) in our `unisonweb/haskeline` fork, 7 commits ahead and 13 commits behind `aryairiani/haskeline`. I first tried merging `judah/haskeline` HEAD into this, but got a ton of conflicts. Is this branch just dead code at this point?

unison-cli/src/Unison/CommandLine/Welcome.hs
mitchellwrosen: Should I try putting it back?

ChrisPenner:
Has anything changed on the points listed from the last time I attempted moving to 9.2?

https://github.com/unisonweb/unison/pull/3304

> Don't want to upgrade to (yet) 9.2.3 because :
Wingman isn't yet supported via HLS on 9.2*, which means no case-splitting functionality. However there's active work to split off that functionality to a new plugin which would be supported for 9.2.3: https://github.com/haskell/haskell-language-server/issues/2971

It's okay if we decide that's not a priority, but we should consider it I think 😄, I use the case-splitting feature very often.

unison-cli/src/Unison/CommandLine/Welcome.hs
aryairani: I dunno, I guess it's supposed to be writing on paper, not sure if the paper alone makes sense. @rlmark ? :)

stack.yaml
aryairani: > Is this branch just dead code at this point?

I would guess so.

mitchellwrosen:
> It's okay if we decide that's not a priority, but we should consider it I think smile, I use the case-splitting feature very often.

I think it's mostly up to you then whether we upgrade now or later. Personally I've never gotten that feature to work with any hls.

stack.yaml
aryairani: screenshot of 20dacc9a0dca7f14997973f1d54da3a4796bb2c6  courtesy of @tarikozkanli
image

aryairani:
Current status is: 
- windows prompt is still broken (cc @mitchellwrosen)
- waiting on either case-splitting Wingman support, or for us to declare we don't care enough to keep holding off (cc @ChrisPenner)

ChrisPenner:
The windows prompt seems important, 

as for wingman, I'll miss the pattern splitting, but it wouldn't kill be me to go without it, copilot can help fill that gap I suppose.

I tried out the GHC 9.2.4 build and I was hoping it would speed up GHC compile times on M1 more than it actually did, so that's a bummer, but still probably good to keep _relatively_ up to date with GHC 👍🏼

ceedubs:
For what it's worth, I'm hoping that [these memory changes](https://well-typed.com/blog/2021/03/memory-return/) in GHC 9.2 will help address some memory issues in Unison Cloud related to the GHC runtime using way more memory than the allotted heap.

mitchellwrosen:
This PR is old and dead, which is a shame. It's sometimes quite a lot of work to bump the LTS, and we are on quite an old GHC at this point, so it would be good to upgrade.

I do sometimes wonder if it's worth it for our small team to support Windows. If even basic libraries like `haskeline` just don't work out of the box, and it's on us to debug and figure out - well, maybe the Haskell ecosystem just isn't good enough for cross-platform stuff :sweat:

ChrisPenner:
I agree that we should find a way to upgrade soon, the longer we wait the further behind we get. It's likely worth a week of off-sprint time to figure out the haskeline thing at this point.

aryairani:
We do want to support Windows, but yeah agree that the situation is pretty ridiculous.

aryairani:
@mitchellwrosen Ok, when I build this PR with 
```yaml
- github: judah/haskeline
  commit: d6c2643b0d5c19be7e440615c6f84d603d4bc648
```
(same as on trunk), then UCM works with `allow-newer: true`; without it, I'm just missing the bumped bounds that the newer versions of Haskeline have. 
```
In the dependencies for haskeline-0.8.0.0(-terminfo):
  base-4.16.3.0 from Stack configuration does not match >=4.9 && <4.15 (latest matching version is 4.14.3.0)
  bytestring-0.11.3.1 from Stack configuration does not match >=0.9 && <0.11 (latest matching version is 0.10.12.1)
needed due to unison-cli-0.0.0 -> haskeline-0.8.0.0
```

I confirmed that rebasing our preferred Happy Windows commit onto the latest Haskeline, and building that with the new LTS is no longer happy on Windows, so I guess something else broke in Haskeline that breaks the proposed-and-then-retracted fix from @. this was what you had done too that also didn't work back in December (https://github.com/unisonweb/unison/pull/3642/commits/20dacc9a0dca7f14997973f1d54da3a4796bb2c6)

I guess we should just use the commit we know, put it on a branch to satisfy #3945, and bump the bounds ourself?

aryairani:
superseded by #4008

PR #3647, opened 2022-11-30
Add instructions for emacs with lsp-mode

alanz:
Documentation change only

### DIFF
--- a/docs/language-server.markdown
+++ b/docs/language-server.markdown
@@ -47,8 +47,9 @@ Simply install the [Unison Language VSCode extension](https://marketplace.visual
 ```elisp
 (use-package unisonlang-mode
   )
-
-(require 'lsp-mode)
+  
+(use-package lsp-mode
+)
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-tcp-connection
                                    ;; Not sure if this is necessary, but it works
--- a/docs/language-server.markdown
+++ b/docs/language-server.markdown
@@ -42,6 +42,33 @@ Note that you'll need to start UCM _before_ you try connecting to it in your edi
 
 Simply install the [Unison Language VSCode extension](https://marketplace.visualstudio.com/items?itemName=unison-lang.unison).
 
+### Emacs with lsp-mode
+
+```elisp
+(use-package unisonlang-mode
+  )
+
+(require 'lsp-mode)
+(lsp-register-client
+ (make-lsp-client :new-connection (lsp-tcp-connection
+                                   ;; Not sure if this is necessary, but it works
+                                   (lambda (port)
+                                     (let* ((other-port 5757)
+                                            (arg1 (format "tcp-l:%s,fork,reuseaddr" port))
+                                            (arg2 (format "tcp:127.0.0.1:%s" other-port)))
+                                       (list "socat" arg1 arg2)
+                                     )))
+                  :activation-fn (lsp-activate-on "u")
+                  :initialized-fn (lambda (workspace)
+                                    (with-lsp-workspace workspace
+                                      (lsp--set-configuration (lsp-configuration-section "unisonlang"))))
+                  :priority -1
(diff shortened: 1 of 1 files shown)


stew:
When I try to add this stuff to my emacs inititalization I get:

```
Symbol's function definition is void: lsp-register-client
```

alanz:
`lsp-register-client` is defined in `lsp-mode`. I guess I should add

```elisp
(use-package lsp-mode
)
```

stew:
Now when I open a .u file I get this message:

```
LSP :: The following servers support current file but do not have automatic installation: unisonlang
You may find the installation instructions at https://emacs-lsp.github.io/lsp-mode/page/languages.
(If you have already installed the server check *lsp-log*).
```

alanz:
As per the overall language server instructions, you do not start it as normal, just connect to it.

So what I do is

```
M-x term
~/unisonlang/ucm
```
This starts a `ucm` session, which also listens on port 5757.

Then open a `xxx.u` file in the same directory you started `ucm`, and it should work.  You can also do `lsp-workspace-restart` once you start up `ucm`

stew:
Yeah, I still get this error whenever it tries to start lsp-mode in a .u buffer. even if ucm is running.

lsp-mode DOES work for other language servers, and LSP is currently working for me with unison using vscode

acowley:
I ran into the same situation as stew when following the instructions in this PR.

The approach shown below works in that I see type check errors, but the type information for local variables is not good (e.g. for the Unison [hello world](https://www.unison-lang.org/learn/at-a-glance/#hello-world) program, the use of the local variable `name` shows popups for `Author.name` and `CopyrightHolder.name`). I also get `The connected server(s) does not support method textDocument/completion.` when trying to complete names. If those are all expected, I can open a PR with this approach (which I modeled on `lsp-gdscript`).

```emacs-lisp
(lsp-register-client
   (make-lsp-client
    :language-id "unisonlang"
    :new-connection 
    (list
     :connect (lambda (filter sentinel name _environment-fn _workspace)
                (let* ((host "localhost")
                       (port 5757)
                       (tcp-proc (lsp--open-network-stream host port (concat name "::tcp"))))

                  (set-process-query-on-exit-flag tcp-proc nil)
                  (set-process-filter tcp-proc filter)
                  (set-process-sentinel tcp-proc sentinel)
                  (cons tcp-proc tcp-proc)))
     :test? (lambda () t))
    :major-modes '(unisonlang-mode)
    :server-id 'unisonlang
    ))
```

zetashift:
I don't see the textDocument/completion errors, but the type information for local variables indeed is off:
image

(my cursor is on the "name" variable)

My config:

```elisp
(use-package unisonlang-mode
  :mode "\\.u\\'"
  :ensure t
  :config
  
  (add-hook 'unisonlang-mode-hook 'lsp-deferred)
  (with-eval-after-load 'lsp-mode
    (add-to-list 'lsp-language-id-configuration
         '(unisonlang-mode . "unisonlang"))
    (setq-local lsp-tcp-connection-timeout 10)
    (lsp-register-client
     (make-lsp-client
      :new-connection
      (list
       :connect (lambda (filter sentinel name _environment-fn _workspace)
                  (let* ((host "localhost")
             (port 5757)
             (tcp-proc (lsp--open-network-stream host port (concat name "::tcp"))))

                    (set-process-query-on-exit-flag tcp-proc nil)
                    (set-process-filter tcp-proc filter)
                    (set-process-sentinel tcp-proc sentinel)
                    (cons tcp-proc tcp-proc)))
       :test? (lambda () t))
      :activation-fn (lsp-activate-on "unisonlang")
      :server-id 'unisonlang))))
```

zetashift:
I don't know if this should be a new issue but VSCode reports the same actually:

image

sellout:
I just put up #4922, which does this for Eglot instead of lsp-mode. Mostly because it’s simpler.

PR #3655, opened 2022-12-02
Blocks enforce that actions return `()`

### LINKED ISSUE #614: Blocks should enforce that non-terminal expressions have type `()`
We should disallow return types other than `()` in block statements that are not the final expression in the block.

It’s pretty easy to shoot yourself in the foot by forgetting to force a computation that you intended to get the effects out of.

e.g.

``` haskell
printLine "Are you sure you want to launch the missiles?"
launch = !readLine
if launch == "YES" then
  launchTheMissiles
  printLine "Missiles launched!"
else abort
```

If you forget to `!launchTheMissiles`, Unison happily ignores `launchTheMissiles` of type `() -> {IO} ()`.

So we should enforce that you do one of the following:
• Make the expression have type `{A} ()` for some ability `A` (possibly empty).
• Assign the result of the expression to a variable that you then have to use somewhere (it should be an error to have an unused binding).
• Explicitly ignore the result by saying `_ =`.

pchiusano:
@stew and I recently wasted another 30 minutes being bit by Unison gleefully allowing non-`()` actions in blocks, so I decided to take a crack at fixing this and it was quite easy. 

There is a minor change to 3 lines of the typechecker, a very minor update to the prettyprinter, and I updated the error message printer to detect this case and give a nice message. 

A number of transcripts were changed as they were reliant on the old behavior.

Closes #614

This modifies the typechecker so that code like this no longer typechecks: 

```haskell
main = do
  Stream.emit -- oops, underapplied!
  ()
```

image

Instead, you'll have to explicitly ignore, using something like:

```haskell
main = do
  _ = Stream.emit
  ()

-- alternately
main2 = do
  void Stream.emit
  ()
```

See [the transcript](https://github.com/unisonweb/unison/pull/3655/files#diff-ce00beb0766862294065f9c16ae5f8b2efd9f33da5fea875502d088072af97b9) for more examples.

Also updated a number of existing transcripts which were taking advantage of this loophole and now have to be explicit about where they're discarding values.

# Interesting stuff and scope of this PR

In working on this, I discovered that the way actions are represented in a block is just as bindings with a name of the form `_123` (a single underscore followed by 1 or more digits). A bit janky. I didn't have the stamina to seriously consider a more comprehensive fix for that, but I think the way you'd do it is with a new type of variable which makes this explicit ("this is an ignored variable"). But then that would have to become a codebase schema change, since you want that to survive a round trip to the codebase? Alternately, you could change the AST to make this explicit, but that seems more complex to me (and it's also a schema change). 

Anyway, I'm not touching that for this PR but it would be good future cleanup work.

I did factor out the logic into a new `Var.isAction` function which is used in both the pretty-printer and the typechecker to detect this situation.

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -19,7 +19,7 @@ import qualified Data.Set.NonEmpty as NES
 import qualified Data.Text as Text
 import qualified Text.Megaparsec as P
 import qualified Unison.ABT as ABT
-import Unison.Builtin.Decls (pattern TupleType')
+import Unison.Builtin.Decls (pattern TupleType', unitRef)
 import qualified Unison.Codebase.Path as Path
 import Unison.ConstructorReference (ConstructorReference, GConstructorReference (..))
 import Unison.HashQualified (HashQualified)
@@ -364,7 +364,7 @@ renderTypeError e env src curPath = case e of
   Mismatch {..} ->
     mconcat
       [ Pr.lines
-          [ "I found a value of type:  " <> style Type1 (renderType' env foundLeaf),
+          [ "I found a value  of type:  " <> style Type1 (renderType' env foundLeaf),
             "where I expected to find:  " <> style Type2 (renderType' env expectedLeaf)
           ],
         "\n\n",
@@ -375,13 +375,17 @@ renderTypeError e env src curPath = case e of
             -- , (,Color.ForceShow) <$> rangeForType foundType
             -- , (,Color.ForceShow) <$> rangeForType expectedType
             -- ,
-            (,Type1) <$> rangeForAnnotated mismatchSite,
+            (,Type1) <$> rangeForAnnotated foundLeaf,
             (,Type2) <$> rangeForAnnotated expectedLeaf
           ],
         fromOverHere'
           src
           [styleAnnotated Type1 foundLeaf]
           [styleAnnotated Type1 mismatchSite],
+        if giveUnitHint then 
+          "\nNote: actions within a block must have type " <> 
+             style Type2 (renderType' env expectedLeaf)    <> ".\n" 
+        else "",
         intLiteralSyntaxTip mismatchSite expectedType,
         debugNoteLoc
           . mconcat
@@ -400,6 +404,12 @@ renderTypeError e env src curPath = case e of
             ],
         debugSummary note
       ]
+      where
+        giveUnitHint = case expectedType of  
+          Type.Ref' u | u == unitRef -> case mismatchSite of 
+            Term.Let1Named' v _ _ -> Var.isAction v
+            _ -> False
+          _ -> False
   AbilityCheckFailure {..}
     | [tv@(Type.Var' ev)] <- ambient,
       ev `Set.member` foldMap Type.freeVars requested ->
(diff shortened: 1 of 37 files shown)


ceedubs:
Thanks, @pchiusano I'm super excited about this change!

One thing that Stew and I have been doing quite a bit that is affected by this is making inline comments via `Doc`. For example this no longer compiles:

```
main = do
  {{ hello this is a code comment }}
  ()
```

We could work around this with `void` or something. Also I've wondered whether this means that our code is doing extra work to create that `Doc` every time the function is called. So perhaps we'd be better off doing something like `void do {{ my comment }}`. Or maybe there's an altogether better way to comment code.

The other code that I'm seeing this flag is places where we are ignoring a `ThreadId` when forking or ignoring an `Either Failure ()` when we run something for its side effect. All of those are easy enough to work around with `_ =` or `void` or whatever.

The color highlighting of the spans in the error messages is really helpful. I could definitely see the error message being confusing for someone who didn't know what was going on, so if it's not too hard to give a better error message for this case, I think that would be helpful. Especially since people may encounter this with code that compiled before but doesn't now that they are trying to edit it again. But if it's a pain, then I think that this is probably worth merging in its current form any way.

Thanks this is great stuff!

pchiusano:
> We could work around this with void or something. Also I've wondered whether this means that our code is doing extra work to create that Doc every time the function is called. So perhaps we'd be better off doing something like void do {{ my comment }}. Or maybe there's an altogether better way to comment code.

Yes, it will do a bunch of pointless work to construct the document. Putting it behind a thunk would be better, at least until we have some optimization passes that could eliminate this sort of thing. But I think comments deserve a more first-class solution anyway.

For now maybe something like:

```
note : 'a -> ()
note d = ()

myFunction x y z = 
  note '{{ here's some comments, yo. }}
  x + y + z
```

pchiusano:
Okay, added a special case to the error message!

image

image

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -375,13 +375,14 @@ renderTypeError e env src curPath = case e of
             -- , (,Color.ForceShow) <$> rangeForType foundType
             -- , (,Color.ForceShow) <$> rangeForType expectedType
             -- ,
-            (,Type1) <$> rangeForAnnotated mismatchSite,
+            (,Type1) . startingLine <$> (rangeForAnnotated mismatchSite),
```
pchiusano, line 378: The mismatchSite for a `let` block includes the entire remainder of the block, which is excessive. Clipping this to just the first line.

pchiusano (commented):
Okay, this is good to merge I'd say, and has a nicer error message now.

image

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -364,7 +364,7 @@ renderTypeError e env src curPath = case e of
   Mismatch {..} ->
     mconcat
       [ Pr.lines
-          [ "I found a value of type:  " <> style Type1 (renderType' env foundLeaf),
+          [ "I found a value  of type:  " <> style Type1 (renderType' env foundLeaf),
```
runarorama, line 367: Nice.

runarorama (approved):

runarorama:
I've added `remark : '{g} -> ()` to Base:

```
    sumOf n acc = 
      if n > 0 then 
        remark '"Add n to the sum so far, and continue"
        remark '"with the next smaller n"
        sumOf (n - 1) (acc + n)
      else
        remark '"We're done. Return the sum"
        acc
    sumOf 1000 0
```

PR #3656, opened 2022-12-02
Update URL refering to newly pushed share code

hojberg:
Share recently changed its URL scheme to prefix "code" with a "p" (for pages). Update the suggested url to reflect this change.

The new URLs have already been launched, and we have redirects in place to support old urls.


ceedubs (approved):
👍 looks good to me.

aryairani (approved):

PR #3667, opened 2022-12-07
Cleanup: use Cli monad in sync http functions

### LINKED ISSUE #3327: Refactor some of the sync helper functions to use Cli monad
Now that we have #3281, I think we can simplify some of the helper functions used in the Share push/pull code paths, which are currently just in IO.

Specifically, they could benefit from early return on some failure conditions, and an implicit environment that contains the more boilerplatey arguments like an HTTP manager.

mitchellwrosen:
## Overview

Fixes #3327

This PR refactors the sync client functions to use the `Cli` monad. They were written before `Cli` existed, but now we can leverage `Cli` for passing around some boilerplate args (like the http manager and codebase record), and we can use our typed/named short circuit mechanism, rather than rely on the rather nasty pattern of sprinkling `catchSyncErrors` where we don't want particular exception types to escape, but otherwise throwing exceptions internally.


pchiusano (approved):

ChrisPenner (commented):
Hrmm, this is one area where I could see us pulling out a lot of this logic into a helper function that might need to run outside of Cli; e.g. if we support a local headless API; or if I need to call sync from the LSP, so `IO` seems more adaptable in that scenario (I've already found myself needing to convert some helpers _out_ of Cli in the past so I can use them in LSP for example).

I guess I'm not convinced about the benefits of switching over, especially since it looks like the new version is longer.

mitchellwrosen:
I think the main benefit of the refactor is utilizing our type-safe short-circuiting method. For now, it cleans up a lot of code, and when/if we want to use these methods in some monad that isn't `Cli`, I think we still have a clear path forward.

aryairani (approved):
Seems good to me at a high level.

@mitchellwrosen is gonna do another pass of manual testing (listing anything worth listing here) and then merge.

If we need to generalize some pieces for LSP or other purposes later, we can still do that.

Issue #3668, opened 2022-12-07
Doc rendering crash


pchiusano:
Suspect this is a compilation / decompilation bug, but certain docs can trigger a crash in doc rendering:

```
Encountered exception:
can't hashComponents if bindings have free variables: [ "User "v"" ]
    [ "User "d""
    , "User "d2""
    , "User "d"-1"
    ] CallStack ( from HasCallStack ) :
   error
```

The message is a bit confusing. Doc rendering uses the same evaluation cache as watch expressions. So for the backticked blocks in the doc, it hashes them and checks if it's already evaluated that hash. So I suspect compilation or decompilation is producing an expression that's not closed.

[Slack discussion](https://unisoncomputing.slack.com/archives/CF8D3DCP3/p1670383673148639?thread_ts=1670357266.932529&cid=CF8D3DCP3)

ceedubs:
I think that this was fixed by #3671.

PR #3671, opened 2022-12-07
Fix a corner case in new 'beta-reduced' decompilation

dolio:
The check was looking for occurrences of `f x y z ...` to make sure it was safe to eliminate variables that had been added by enclosure. However, it wasn't finding occurrences of `f` by itself, because that is not considered an application by the predicate that was being used.

I also made the 'fix-up' code default to not eliminating any variables when there are *no* subsequent occurrences of `f`. Previously this would allow eliminating as many variables as possible. However, this can produce ill-scoped definitions. Avoiding this would require knowing which variables are in scope when reducing an unused definition, but that would require a more complicated rewriting process than is currently used.


dolio:
Fixes #3668

pchiusano (approved):

PR #3672, opened 2022-12-07
Fix CI to run some (previously skipped) unit tests in unison-parser-typechecker

pchiusano:
CI does the following to run unit tests:

```
stack --no-terminal build --fast --test unison-parser-typechecker
```

This appears to set the current directory to unison-parser-typechecker, which means that paths relative to the project root no longer work.

image

I tried figuring out how to convince stack not to set the current directory like that but gave up and instead settled for adding `..` to the paths in the test suite.

Annoyingly, for people who use cabal, it seems that cabal doesn't change the current directory when running subproject test suites, so you'll have to `cd` into the subproject when running those, maybe? Not sure.


pchiusano:
Also, this PR fixes some of the files in light of: https://github.com/unisonweb/unison/pull/3655

dolio (approved):
Looks fine to me.

PR #3677, opened 2022-12-08
Add a few more `TMap` variants to `tmap.u` benchmark

pchiusano:
A few things I tried:

* Different branching factors (16 instead of 256)
* Sparse children pointers (a `Map` instead of a `List`)

The original 256-way trie still seems the fastest (480ms vs 735ms), at around 20k inserts + lookups per second.

Also tried `TMap` with hashed keys, which didn't seem to make a difference on performance.


dolio (approved):
Good to have more variants.

Not too surprised these examples are slower, because they add a lot of arithmetic overhead by making the key chunking even smaller.

pchiusano:
I also added some more concurrency tests of the various map implementations.

PR #3687, opened 2022-12-09
Cleanup: break up `unison-util` sin bin

mitchellwrosen:
## Overview

This PR breaks up (and eliminates) the old `unison-util` sin-bin package. Most modules were simply moved over to `unison-prelude` where I think they belong (e.g. `Unison.Util.Alternative`). I moved `U.Util.Cache` (now `Unison.Util.Cache`) into its own package, `unison-util-cache`.


ChrisPenner (approved):
Seems fine, is there a particular motivation?
The current layout hasn't really ever caused me issues 🤷🏼‍♂️

mitchellwrosen:
> Seems fine, is there a particular motivation?

Just cleaning up.

PR #3691, opened 2022-12-12
Cleanup: eliminate second NameSegment type

mitchellwrosen:
## Overview

We had two `NameSegment` types floating around; this PR eliminates one of them.


ChrisPenner (approved):
Nice.

PR #3698, opened 2022-12-15
Cleanup: tidy up v2-hashing package and hide it behind an interface

mitchellwrosen:
## Overview

This PR finishes off some code organization work that we had only ever discussed, but never implemented, when extracting out the hashing modules into their own package, `unison-hashing-v2`.

The `unison-hashing-v2` package is meant to hold data structures that directly correspond to the content that we hash; a user of this library is meant to first convert their type (e.g. `Unison.Term.Term`) into the corresponding hashable structure (e.g. `Unison.Hashing.V2.Term.Term`), then hash that.

We left `unison-hashing-v2` in a state that exposes too many innards, though. Its users should only have access to the data structures that have hashable instances, not any of the guts that can be used to hash something. So, this PR essentially just hides some of the implementation of `unison-hashing-v2`, and exposes only one module, `Unison.Hashing.V2`.

A concrete list of changes:

- A new `unison-hash` package was created to hold `Unison.Hash`. This was extracted from `unison-core`, because it is a common dependency of the (otherwise isolated) hashing packages. `unison-hashing-*` should _not_ depend on `unison-core`, but they currently do (pre- and post-PR), for bad reasons.
- A new `unison-hashing` package was created to be a common dependency of `unison-hashing-*` packages. Currently, it just contains one module, `Unison.Hashing.ContentAddressable`, which is a rename of `Unison.Hashing.V2.Hashable` (to avoid confusion with `Unison.Hashing`). `unison-hashing-v2` re-exports this for convenience, as users of `unison-hashing-v2` will always need access to it.
- `U.Util.Hash` was folded into `Unison.Hash`, and their confusing API differences were fixed up.
- `U.Util.Hash32` was renamed to `Unison.Hash32` for consistency.
- `Unison.Hashing.V2.Token` is now monomorphic. The `h` type variable was just `Hash` everywhere.
- Similarly, the `Unison.Hashing.V2.Accumulate` type class was deleted, because its only instance was `Hash`.
- One bad (but benign) `unison-hashing-v2` dependency on `Unison.DataDeclaration.ConstructorId` (from `unison-core`) was corrected. Now there's a `Unison.Hashing.V2.ConstructorId` module.
- `NameSegment` was pulled out of `Unison.Hashing.V2.Branch` and put into its own `Unison.Hashing.V2.NameSegment` module (just for consistency).
- A few unused functions in `Unison.Hashing.V2.Patch` and `Unison.Hashing.V2.Reference` were deleted.
- Some boring function and constructor renamings so that all of `unison-hashing-v2` could be exported from a single module, `Unison.Hashing.V2`. I thought this was cleaner than having multiple exported modules e.g. `Unison.Hashing.V2.Term` because that would require, in certain cases, additional `.Internal` modules to hide details that should be visible from within `unison-hashing-v2`, but not via its external interface.


aryairani (approved):
Let's add some kind of [transcript?] test to blow up if the hashing function changes unexpectedly, and then merge this afterwards.

mitchellwrosen:
Sure, I've opened #3718

PR #3713, opened 2022-12-19
warn after pulling an empty namespace

### LINKED ISSUE #3648: `pull` command "succeeds" when pulling from a namespace that does not exist
Currently when you `pull` a library that doesn't exist on share, the UCM will "succeed" by creating a namespace with zero entities. 

```
.> pull stew.public.projects.json.releases.v3 lib.json

 Downloaded 0 entities.

  ✅
  
  ✅ Successfully pulled into newly created namespace lib.json.
```

This might be ultimately desirable but this came up because of our Exercism track integration so it does have some interesting downstream effects. We were pulling something that had been deleted but silently failing to setup the testing environment.

mitchellwrosen:
## Overview

Fixes #3648 

This PR makes `pull` output a warning message whenever it pulls a namespace that has history but no contents. I also got rid of the double checkmark output, since it looks weird.

Here's what this looks like now:

Screenshot from 2022-12-19 16-05-44


aryairani:
Great, thanks.
But don't we want to say something about having history? Or is that more confusing?
Like "Note: Although \<path> exists on \<server>, the latest version is empty."
cc @rlmark

mitchellwrosen:
@aryairani That sounds good to me; I was trying to reuse an existing output message type for consistency, and noticed we already had an `EmptyBranch` message, but we could make a new one for this purpose no problem. Can we nail down the exact wording first?

aryairani:
@mitchellwrosen How about "\<path> has some history, but is currently empty."

And then as a separate future effort, we can change how we deal with empty namespaces again, and/or how we refer to "history in messages".

aryairani (approved):

PR #3714, opened 2022-12-19
Fix a runtime segfault, plus debugging infrastructure

dolio:
This PR is the result of investigating a segfault that was occurring on some cloud jobs. The apparent culprit was the implementation of universal equality/compare for wrapped, boxed Haskell types. Somehow `unsafeCoerce` was causing bad behavior even though it didn't appear that the coercions were ill typed upon investigation.

The implementation of the comparisons has been modified to a form that doesn't seem to crash.

In addition, some debugging infrastructure that was written has been included in case it will be useful in the future.

Primitive array functions used in the runtime have been broken out into their own module, and bounds checking wrappers have been written for them. The checks are off by default, but can be turned on at build-time.

Functions for comparing intermediate code for alpha equality are now available. This was used to check that what we get back from a serialized standalone file is equivalent to the original code, and at least in this case, it appears to be so.

I'm not sure what percentage of the runtime this test case exercised, but all these checks (and more) turned up nothing actually wrong with the runtime, other than (most likely) an bad interaction of `unsafeCoerce` with the optimizer. So that's good news. I didn't include it in this PR, but at one point I had all uses of `unsafeCoerce` replaced with GHC Typeable, and got 0 runtime type errors. So everything was actually type correct in that respect.

### DIFF
--- a/parser-typechecker/package.yaml
+++ b/parser-typechecker/package.yaml
@@ -8,10 +8,15 @@ flags:
   optimized:
     manual: true
     default: true
+  arraychecks:
+    manual: true
+    default: false
 
 when:
   - condition: flag(optimized)
     ghc-options: -funbox-strict-fields -O2
+  - condition: flag(arraychecks)
+    cpp-options: -DARRAY_CHECK
 
 dependencies:
   - ListLike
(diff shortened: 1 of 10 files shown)


pchiusano (approved):
😎👍

PR #3716, opened 2022-12-20
prepend "test>" to edited tests

### LINKED ISSUE #3373: editing a test should add the test> prefix
if I edit a test, it should add the `test>` prefix to the function so the test gets run after editing.

I have found that we have inadvertently turned tests into regular functions when they show up on our `todo` and we save the function without manually re-adding the `test>` prefix :(

mitchellwrosen:
## Overview

Fixes #3373 

This PR puts `"test> "` before tests that are `edit`ed.

Notable implementation detail: we had a term binding printer that always put a type signature (if the term was annotated), e.g.:

```unison
mytest : [Test.Result]
mytest = []
```

However, I couldn't figure out a syntax for test watches that includes type signatures. This is a parse error, for example.

```unison
test>
  mytest : [Test.Result]
  mytest = []
```

I also figured type signatures were not very important to print for tests, since they're always type `[Test.Result]`. So, this PR puts tests in the scratch file without type signatures, e.g.:

```unison
test> mytest = []
```

### Testing

I tested this manually.

### DIFF
--- a/lib/unison-util-relation/src/Unison/Util/Relation3.hs
+++ b/lib/unison-util-relation/src/Unison/Util/Relation3.hs
@@ -94,6 +94,10 @@ mapD2Monotonic f Relation3 {d1, d2, d3} =
 member :: (Ord a, Ord b, Ord c) => a -> b -> c -> Relation3 a b c -> Bool
 member a b c = R.member b c . lookupD1 a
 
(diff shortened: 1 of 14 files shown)


aryairani (approved):
Rubber stamp ✅ 
Reminder to mention testing strategy in PR descriptions

Issue #481, opened 2019-05-04
Error message "I'm not sure what User "_200012" means"


atacratic:
I saved a .u file containing an error.  The error message talks about `User "_200012"`, but I don't know who that is...
```
master> 
I'm not sure what User "_200012" means at line 4, columns 12-13

    4 | fac n = if n = 0 

Whatever it is, it has a type that conforms to Boolean.
```

@aryairani said:
> I’m guessing that’s related to the Show instance for the new Var type; to accidentally Show-ing it in the message even now that it contains metadata

In subsequent attempts it seems to have stopped mentioning `User`.

aryairani:
@atacratic Can you still reproduce this or similar in `wip/branchless`?

atacratic:
Nope!

PR #3722, opened 2023-01-01
Fix build with cabal

jneira:
The pr updates cabal.project to adjust it to last structural changes.
Unfortunately the codebase now uses th referencing files with relative paths and stack and cabal sets a different cwd 😟 (https://github.com/haskell/haskell-language-server/issues/481)
Not sure how coud we solve that behaviour tbh. I believe atleats stack sets some C defs and we could add CPP conditions but i don't like it at all.

Another caveat: the haskeline commit 2944b11d19ee034c48276edc991736105c9d6143 does not exist anymore.
It is being referenced in stack.yaml but i am not sure how is even working. I've replicated the commit in my repo, but obviously it cant be the definitive solution.

The cabal build should be checked in ci to keep it sync (even with an optional job)

### DIFF
--- a/contrib/cabal.project
+++ b/contrib/cabal.project
@@ -3,20 +3,27 @@ packages:
   parser-typechecker
   unison-core
   unison-cli
+  unison-hashing-v2
   unison-share-api
+  unison-syntax
 
   codebase2/codebase
   codebase2/codebase-sqlite
+  codebase2/codebase-sqlite-hashing-v2
   codebase2/codebase-sync
   codebase2/core
-  codebase2/util
   codebase2/util-serialization
   codebase2/util-term
 
   lib/unison-prelude
   lib/unison-sqlite
   lib/unison-util-base32hex
+  lib/unison-util-base32hex-orphans-aeson
+  lib/unison-util-base32hex-orphans-sqlite
+  lib/unison-util-bytes
+  lib/unison-util-cache
   lib/unison-util-relation
+  lib/unison-util-rope
   lib/unison-pretty-printer
 
 source-repository-package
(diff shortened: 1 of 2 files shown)


jneira:
> Unfortunately the codebase now uses th referencing files with relative paths and stack and cabal sets a different cwd 😟 (https://github.com/haskell/haskell-language-server/issues/481)

I expected ci should fail due to changes in Query.hs 🤔

aryairani (changes requested):
Hi @jneira — we don't currently support building with cabal beyond what's in contrib/.

However, we might support cabal in the future, so I'm averse to adding cabal.project to .gitignore.

> Not sure how coud we solve that behaviour tbh.

We use symlinks (`sql/`) because stack and haskell-language-server already use different cwds. We use more symlinks (`unison/`) because we have another project that uses `unison` as a submodule under the `unison/` path.

Maybe adding more symlinks will work for you, but for the submodule reason above, we don't want to change `Query.hs`

If you want to merge just the `contrib/cabal.project` changes, possibly another symlink, and some `development.markdown` changes, that would be okay.

jneira:
Thanks for the feedback: sure, will revert the .gitignore, great to know cabal could be a first citizen. 😄  
hls support for cabal is better than for stack and that is a good argument to support cabal imo (the reason hls uses the same cwd as cabal is #481)

> Maybe adding more symlinks will work for you, but for the submodule reason above, we don't want to change Query.hs

Sure, my fault, i did not check lastest instructions for building in windows and i did not activate symlinking. `Query.hs`works with any change after activate them.

I am about to upload last versions with the needed changes

jneira:
@aryairani changes done, still left the issue with the haskeline commit used in stack.yaml

The commit does not belong to any branch: https://github.com/judah/haskeline/commit/d6c2643b0d5c19be7e440615c6f84d603d4bc648

And cabal throw an error when trying to fetch it:

```
fatal: Could not parse object 'd6c2643b0d5c19be7e440615c6f84d603d4bc648'.
```

(it works for non orphan commits)

Should be cherry-pick the commit to the unison haskeline fork? I think it would be good to not depend on an orphaned commit of a external repo

aryairani:
Hi @jneira sorry I missed your last messages. The haskeline issue should be resolved now if you want to try again 😬

dolio:
This looks good to me. It closely matches what I use locally.

Although, I don't even bother with a custom haskelline/configurator commit. I just use something from hackage and haven't noticed problems.

aryairani (approved):

PR #3725, opened 2023-01-05
Doc: add support for inline latex and svg

hojberg:
## Overview

Add new helpers for creating `Embed` and `EmbedInline` for `LaTeX` and `Svg`, similar to the form we use for `FrontMatter` and `Video`.

## Loose ends

* I feel like it'd be nice if we translated ```mermaid and ```latex as helpers like this as well. That might help with the asHtml bit. 
* Need to run the transcripts somehow...


pchiusano (approved):
Sorry, just seeing this. Seems good. Can you refresh to fix conflicts?

PR #3727, opened 2023-01-06
Update vim support. Comments are now correct, and syntax is closer to…

pragdave:
… current Unison

The current Vim language support was somewhat divorced from the current Unison: lots of Haskel stuff, comment and doc string syntax was wrong, etc.

This brings it closer to the current reality.

I made the choice to highlight watch clauses to make them stand out. 

## Test coverage

n/a

## Loose ends

It's missing UTF in identifiers: I need to research how that can be done in vim.


runarorama (approved):
I'd say this is unambiguously better! Thanks!

PR #3737, opened 2023-01-11
Add pattern match coverage checking

### LINKED ISSUE #913: Coverage checking for pattern matching

tstat:
## Overview

Closes #913

Add pattern match coverage checking to unison. 

Pattern match coverage checking is implemented following the algorithm described in [Lower Your Guards](https://simon.peytonjones.org/assets/pdfs/lower-your-guards.pdf). The goal of pattern match coverage checking is to identify the following problems that may arise in a pattern match:

* It is missing clauses (_i.e._ it is non-exhaustive)
* It contains redundant patterns (_i.e._ the case can be deleted without altering the program)
* It contains inaccessible patterns (_i.e_ the rhs can never be entered)

Furthermore, in the case of a non-exhaustive match, the goal to present the user with concrete values that do not match any of the existing patterns.

An inaccessible pattern in unison would be one that performs effects in a guard although the constraints are unsatisfiable. The match clause cannot be deleted without altering the program. This PR does not report inaccessible patterns (although most of the work to do so is done) as those should be emitted as a warning and we don't have compiler warnings.

The haddocks in `Unison.PatternMatchCoverage.hs` provide an overview and are a good entry point to reading the code.  The new transcript `pattern-match-coverage.md` has some examples to look over.

## Loose ends

Pattern match coverage checking for ability handlers has a few wrinkles that are not handled by this PR. My intention is to open a different PR with that addition after this is merged.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -161,6 +161,7 @@ import Unison.Symbol (Symbol)
 import Unison.Term (Term)
 import qualified Unison.Term as Term
 import Unison.Type (Type)
+import qualified Unison.Type as Type
 import Unison.Typechecker.TypeLookup (TypeLookup (TypeLookup))
 import qualified Unison.Typechecker.TypeLookup as TL
 import qualified Unison.UnisonFile as UF
@@ -338,6 +339,7 @@ lookupWatchCache codebase h = do
   maybe (getWatch codebase WK.TestWatch h) (pure . Just) m1
 
 typeLookupForDependencies ::
+  forall m a.
   (BuiltinAnnotation a) =>
   Codebase m Symbol a ->
   Set Reference ->
@@ -347,17 +349,34 @@ typeLookupForDependencies codebase s = do
   foldM go mempty s
   where
     go tl ref@(Reference.DerivedId id) =
-      fmap (tl <>) $
-        getTypeOfTerm codebase ref >>= \case
-          Just typ -> pure $ TypeLookup (Map.singleton ref typ) mempty mempty
-          Nothing ->
-            getTypeDeclaration codebase id >>= \case
-              Just (Left ed) ->
-                pure $ TypeLookup mempty mempty (Map.singleton ref ed)
-              Just (Right dd) ->
-                pure $ TypeLookup mempty (Map.singleton ref dd) mempty
-              Nothing -> pure mempty
+      getTypeOfTerm codebase ref >>= \case
+        Just typ -> pure $ tl <> TypeLookup (Map.singleton ref typ) mempty mempty
+        Nothing ->
+          getTypeDeclaration codebase id >>= \case
+            Just (Left ed) ->
+              pure $ tl <> TypeLookup mempty mempty (Map.singleton ref ed)
+            Just (Right dd) -> do
+              -- We need the transitive dependencies of data decls
+              -- that are scrutinized in a match expression for
+              -- pattern match coverage checking (specifically for
+              -- the inhabitation check). We ensure these are found
+              -- by collecting all type dependencies for all data
+              -- decls.
+              let constructorTypes :: [Type Symbol a]
+                  constructorTypes = snd <$> DD.constructors dd
+
+                  -- All references from constructorTypes that we
+                  -- have not already gathered.
+                  constructorRefs :: Set Reference
+                  constructorRefs = foldl' (\b a -> Set.filter (unseen tl) (Type.dependencies a) <> b) mempty constructorTypes
+
+              -- recursively call go for each constructor ref
+              let z = tl <> TypeLookup mempty (Map.singleton ref dd) mempty
+              foldM go z constructorRefs
+            Nothing -> pure tl
     go tl Reference.Builtin {} = pure tl -- codebase isn't consulted for builtins
+    unseen :: TL.TypeLookup Symbol a -> Reference -> Bool
+    unseen tl r = isNothing (Map.lookup r (TL.dataDecls tl))
 
 toCodeLookup :: (MonadIO m) => Codebase m Symbol Parser.Ann -> CL.CodeLookup Symbol m Parser.Ann
 toCodeLookup c =
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -161,7 +161,6 @@ import Unison.Symbol (Symbol)
 import Unison.Term (Term)
 import qualified Unison.Term as Term
 import Unison.Type (Type)
-import qualified Unison.Type as Type
 import Unison.Typechecker.TypeLookup (TypeLookup (TypeLookup))
 import qualified Unison.Typechecker.TypeLookup as TL
 import qualified Unison.UnisonFile as UF
@@ -362,13 +361,11 @@ typeLookupForDependencies codebase s = do
               -- the inhabitation check). We ensure these are found
               -- by collecting all type dependencies for all data
               -- decls.
-              let constructorTypes :: [Type Symbol a]
-                  constructorTypes = snd <$> DD.constructors dd
 
-                  -- All references from constructorTypes that we
-                  -- have not already gathered.
-                  constructorRefs :: Set Reference
-                  constructorRefs = foldl' (\b a -> Set.filter (unseen tl) (Type.dependencies a) <> b) mempty constructorTypes
+              -- All references from constructorTypes that we
+              -- have not already gathered.
+              let constructorRefs :: Set Reference
+                  constructorRefs = Set.filter (unseen tl) (DD.dependencies dd)
 
               -- recursively call go for each constructor ref
               let z = tl <> TypeLookup mempty (Map.singleton ref dd) mempty
--- a/parser-typechecker/src/Unison/PatternMatchCoverage.hs
+++ b/parser-typechecker/src/Unison/PatternMatchCoverage.hs
@@ -37,6 +37,7 @@ where
 
 import qualified Data.Set as Set
 import Debug.Trace
+import Unison.Debug
 import Unison.Pattern (Pattern)
 import Unison.PatternMatchCoverage.Class (Pmc (..))
 import Unison.PatternMatchCoverage.Desugar (desugarMatch)
@@ -75,8 +76,7 @@ checkMatch matchLocation scrutineeType cases = do
             P.hang "uncovered:" (NC.prettyDnf uncovered),
             P.hang "uncovered expanded:" (NC.prettyDnf (Set.fromList uncoveredExpanded))
           ]
-      shouldDebug = False
-      doDebug = case shouldDebug of
+      doDebug = case shouldDebug PatternCoverage of
         True -> trace (P.toPlainUnbroken debugOutput)
         False -> id
   doDebug (pure (redundant, inaccessible, sols))
--- a/parser-typechecker/src/Unison/PatternMatchCoverage.hs
+++ b/parser-typechecker/src/Unison/PatternMatchCoverage.hs
@@ -1,0 +1,45 @@
+-- Pattern match coverage checking following the algorithm described
+-- in "Lower Your Guards".
+-- https://simon.peytonjones.org/assets/pdfs/lower-your-guards.pdf
+module Unison.PatternMatchCoverage where
+
+import qualified Data.Set as Set
+import Debug.Trace
+import Unison.Pattern (Pattern)
+import Unison.PatternMatchCoverage.Class (Pmc (..))
+import Unison.PatternMatchCoverage.Desugar (desugarMatch)
+import Unison.PatternMatchCoverage.GrdTree (prettyGrdTree)
+import qualified Unison.PatternMatchCoverage.NormalizedConstraints as NC
+import Unison.PatternMatchCoverage.PmGrd (prettyPmGrd)
+import Unison.PatternMatchCoverage.Solve (classify, expand, expandSolution, uncoverAnnotate)
+import qualified Unison.Term as Term
+import qualified Unison.Type as Type
+import qualified Unison.Util.Pretty as P
+
+checkMatch ::
+  forall vt v loc m.
+  (Pmc vt v loc m) =>
+  loc ->
+  Type.Type vt loc ->
+  [Term.MatchCase loc (Term.Term' vt v loc)] ->
+  m ([loc], [loc], [Pattern ()])
+checkMatch matchLocation scrutineeType cases = do
+  v0 <- fresh
+  grdtree0 <- desugarMatch matchLocation scrutineeType v0 cases
+  (uncovered, grdtree1) <- uncoverAnnotate (Set.singleton (NC.declVar v0 scrutineeType id NC.emptyNormalizedConstraints)) grdtree0
+  uncoveredExpanded <- concat . fmap Set.toList <$> traverse (expandSolution v0) (Set.toList uncovered)
+  let sols = map (expand v0) uncoveredExpanded
+  let (_accessible, inaccessible, redundant) = classify grdtree1
+  let debugOutput =
+        P.sep
+          "\n"
+          [ P.hang "desugared:" (prettyGrdTree prettyPmGrd (\_ -> "<loc>") grdtree0),
+            P.hang "annotated:" (prettyGrdTree NC.prettyDnf (NC.prettyDnf . fst) grdtree1),
+            P.hang "uncovered:" (NC.prettyDnf uncovered),
+            P.hang "uncovered expanded:" (NC.prettyDnf (Set.fromList uncoveredExpanded))
+          ]
+      shouldDebug = False
+      doDebug = case shouldDebug of
+        True -> trace (P.toPlainUnbroken debugOutput)
+        False -> id
+  doDebug (pure (redundant, inaccessible, sols))
--- a/parser-typechecker/src/Unison/PatternMatchCoverage.hs
+++ b/parser-typechecker/src/Unison/PatternMatchCoverage.hs
@@ -1,7 +1,39 @@
--- Pattern match coverage checking following the algorithm described
--- in "Lower Your Guards".
--- https://simon.peytonjones.org/assets/pdfs/lower-your-guards.pdf
-module Unison.PatternMatchCoverage where
+-- | Pattern match coverage checking is implemented following the
+-- algorithm described in [Lower Your
+-- Guards](https://simon.peytonjones.org/assets/pdfs/lower-your-guards.pdf). The
+-- goal of pattern match coverage checking is to identify the
+-- following problems that may arise in a pattern match:
+--
+-- * It is missing clauses (/i.e./ it is non-exhaustive)
+-- * It contains redundant patterns (/i.e./ the case can be deleted without altering the program)
+-- * It contains inaccessible patterns (/i.e/ the rhs can never be entered)
+--
+-- Furthermore, in the case of a non-exhaustive match, the goal to
+-- present the user with concrete values that do not match any of the
+-- existing patterns.
+--
+-- /N.B./ An inaccessible pattern in unison would be one that performs
+-- effects in a guard although the constraints are unsatisfiable. Such
+-- a pattern cannot be deleted without altering the program.
+--
+-- == High-level algorithm overview
+--
+-- 1. [Desugar]("Unison.PatternMatchCoverage.Desugar") a match expression into a 'Unison.PatternMatchCoverage.GrdTree.GrdTree'.
+-- 2. Annotate the @GrdTree@ nodes with [refinement types]("Unison.PatternMatchCoverage.NormalizedConstraints")
+-- representing values that match this node. Redundant and inaccessible patterns are then identified by @GrdTree@ leaves
+-- with uninhabited refinement types. Inaccessible patterns are distinguished by an effect being performed between the
+-- @GrdTree@ root and the leaf.
+-- 3. Traverse the @GrdTree@ building up a refinement type representing uncovered values. If the resulting refinement type
+-- is inhabited then the match is missing clauses.
+-- 4. Find inhabitants of the uncovered refinement type to present to the user.
+--
+-- Step (1) is implemented by 'desugarMatch'. Steps (2) and (3) are
+-- implemented as a single traversal: 'uncoverAnnotate'. Step (4) is
+-- implemented by 'expandSolution'.
+module Unison.PatternMatchCoverage
+  ( checkMatch,
+  )
+where
 
 import qualified Data.Set as Set
 import Debug.Trace
@@ -16,12 +48,17 @@ import qualified Unison.Term as Term
 import qualified Unison.Type as Type
 import qualified Unison.Util.Pretty as P
 
+-- | Perform pattern match coverage checking on a match expression
 checkMatch ::
   forall vt v loc m.
   (Pmc vt v loc m) =>
+  -- | the match location
   loc ->
+  -- | scrutinee type
   Type.Type vt loc ->
+  -- | match cases
   [Term.MatchCase loc (Term.Term' vt v loc)] ->
+  -- | (redundant locations, inaccessible locations, inhabitants of uncovered refinement type)
   m ([loc], [loc], [Pattern ()])
 checkMatch matchLocation scrutineeType cases = do
   v0 <- fresh
--- a/parser-typechecker/src/Unison/PatternMatchCoverage.hs
+++ b/parser-typechecker/src/Unison/PatternMatchCoverage.hs
@@ -19,17 +19,17 @@
 -- == High-level algorithm overview
 --
 -- 1. [Desugar]("Unison.PatternMatchCoverage.Desugar") a match expression into a 'Unison.PatternMatchCoverage.GrdTree.GrdTree'.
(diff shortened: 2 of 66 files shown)


pchiusano (approved):
AMAZING!!! 

I enjoyed browsing [this file](https://github.com/unisonweb/unison/blob/7314d09623835160243811827e0dcd9d9c86a93b/unison-src/transcripts/pattern-match-coverage.output.md) which has a bunch of examples, with output error messages.

unison-src/transcripts-round-trip/main.output.md
```diff
@@ -241,40 +241,47 @@ f x = let
...
+    g : [()] -> ()
+    h : [[()]] -> ()
 
 .> edit f g
 
   ☝️
   
   I added these definitions to the top of
-  /Users/runar/work/unison/scratch.u
+  /home/traveler/code/haskell/unison/pattern-match-coverage/scratch.u
```
aryairani, line 274: Unrelated, but I wonder if we could get UCM to run in the directory with the transcript @ChrisPenner

unison-src/transcripts-round-trip/main.output.md
```diff
@@ -928,9 +936,10 @@ broken = cases
     broken : Optional Nat -> ()
     broken = cases
       Some
-        loooooooooooooooooooooooooooooooooooooooooooooooooooooooong | loooooooooooooooooooooooooooooooooooooooooooooooooooooooong
-        == 1 ->
+        loooooooooooooooooooooooooooooooooooooooooooooooooooooooong| loooooooooooooooooooooooooooooooooooooooooooooooooooooooong
+        == 1  ->
```
aryairani, line 940: I wonder why this space moved?

unison-src/transcripts-round-trip/main.output.md
```diff
@@ -875,7 +882,7 @@ broken tvar =
           (cases
             Some _ ->
               "oh boy isn't this a very very very very very very very long string?"
-            None   -> ""))
+            None -> ""))
```
aryairani, line 885: I wonder why these spaces went away?

unison-src/transcripts-using-base/fix2049.md
```diff
@@ -13,8 +13,6 @@ catcher act =
 
 tests _ =
   [ catcher do
-      match None with Some x -> x
```
aryairani, line 16: What was the deal here?

ChrisPenner:
It was pretty easy to wire up the diagnostic ranges, so I went ahead and added those so it works with LSP:

image

image

unison-src/transcripts/lambdacase.output.md
```diff
@@ -141,7 +141,7 @@ blorf = cases
     ⍟ These new definitions are ok to `add`:
     
       structural type B
-      blah  : B -> B -> Text
+      blah  : B -> hod1jusqau2 -> Text
```
aryairani: this is weird, right?

unison-src/transcripts/patternMatchTls.md
```diff
@@ -13,13 +13,18 @@ use builtin.io2.Tls newClient send handshake terminate
 
 frank: '{IO} ()
 frank = do 
-  (Right socket) = clientSocket.impl "example.com" "443"
```
aryairani, line 16: I didn't even know we had this syntax. I'm guessing it produces an error now? What kind?

unison-src/transcripts/lambdacase.output.md
aryairani: errr it's a universally quantified type that got named `hod1jusqau2`?

unison-src/transcripts/patternMatchTls.md
pchiusano: Destructuring binds all desugar to pattern matching in the obvious way, so it'll be a pattern coverage error.

unison-src/transcripts-round-trip/main.output.md
pchiusano: I think this file just shouldn't be checked in. CI just runs the transcript to make sure it passes, and that's sufficient.

unison-src/transcripts-round-trip/main.output.md
pchiusano: This output file shouldn't be checked in, and it's not updated consistently unless we remember to. So I'd guess the current checked in version is from some older version of the prettyprinter.

unison-src/transcripts-using-base/fix2049.md
pchiusano: It no longer typechecks - this was a test for exception catching. I think it's okay to delete.

unison-src/transcripts/lambdacase.output.md
pchiusano: This does seem odd. @tstat any idea what's going on here?

pchiusano: Oh, wait, I get it. Travis removed a case, which changed the type, exposing an existing quirk / bug that exists on trunk - it's just picking the synthetic variable name generated by `cases`

```
cases ...

converts to

synthetic19283 -> match synthetic19283 with ...
```

And when inferring a lambda, it picks type variables based on the the term variables. 

@tstat I'd update this example to force the second argument to be of type `B`, or else add a type signature.

unison-src/transcripts-using-base/fix2049.md
tstat: I tried to break these transcript changes into individual commits. It might be helpful to look at them that way and see the commit message. In this change, the commit message says that it removes a test for catching a missing case since this is handled at compile time now.

unison-src/transcripts/lambdacase.output.md
tstat: Oh whoops, I didn't notice this change. Thanks for catching it.

I added a commit that gives a type signature.

parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -347,17 +349,34 @@ typeLookupForDependencies codebase s = do
...
+              -- the inhabitation check). We ensure these are found
+              -- by collecting all type dependencies for all data
+              -- decls.
+              let constructorTypes :: [Type Symbol a]
+                  constructorTypes = snd <$> DD.constructors dd
+
+                  -- All references from constructorTypes that we
+                  -- have not already gathered.
+                  constructorRefs :: Set Reference
+                  constructorRefs = foldl' (\b a -> Set.filter (unseen tl) (Type.dependencies a) <> b) mempty constructorTypes
```
ChrisPenner: Could you use `Unison.DataDeclaration.dependencies` for this? I think it's doing the same thing and includes the constructor types.

parser-typechecker/src/Unison/PatternMatchCoverage.hs
```diff
@@ -0,0 +1,82 @@
...
+  let (_accessible, inaccessible, redundant) = classify grdtree1
+  let debugOutput =
+        P.sep
+          "\n"
+          [ P.hang "desugared:" (prettyGrdTree prettyPmGrd (\_ -> "<loc>") grdtree0),
+            P.hang "annotated:" (prettyGrdTree NC.prettyDnf (NC.prettyDnf . fst) grdtree1),
+            P.hang "uncovered:" (NC.prettyDnf uncovered),
+            P.hang "uncovered expanded:" (NC.prettyDnf (Set.fromList uncoveredExpanded))
+          ]
+      shouldDebug = False
```
ChrisPenner: BTW if you'd like to leave the debugging information in, you can add a new case to our [DebugFlags](https://github.com/unisonweb/unison/blob/8eaeeefb1959ed29c53f4ba82c2a436fd473ecbf/lib/unison-prelude/src/Unison/Debug.hs#L25) and use the debugging helpers there.


Then can display debugging information any time with `UNISON_DEBUG=PatternMatchChecker` or w/e

unison-src/transcripts/move-namespace.output.md
```diff
@@ -278,7 +278,7 @@ I should be able to move the root into a sub-namespace
   
   
   
-  □ 1. #hu6p22qbe4 (start of history)
+  □ 1. #j17hvtt1rm (start of history)
```
ChrisPenner, line 281: What might have caused this change? Did some of the builtins change hashes or something?

unison-src/tests/tictactoe2.u
```diff
@@ -18,36 +30,19 @@ isWin board =
     -- horizontal left/center/right
     -- diagonal rising/falling
     Board a b c
-          _ _ _
```
ChrisPenner, line 21: I love the implication that implementing a pattern match checker involves fixing a game of tic-tac-toe 🙃

parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -765,6 +783,27 @@ getEffectDeclaration r = do
...
 getDataConstructorType = getConstructorType' Data getDataDeclaration
 
+getDataConstructors :: (Var v) => Type v loc -> M v loc (EnumeratedConstructors (TypeVar v loc) v loc)
+getDataConstructors typ
+  | Type.Ref' r <- typ, r == Type.booleanRef = pure BooleanType
+  | Type.App' (Type.Ref' r) arg <- typ,
+    r == Type.listRef =
+      let xs =
+            [ (ListPat.Cons, [arg]),
+              -- (ListPat.Snoc, [typ, arg]),
```
ChrisPenner: What's this guy doing? Maybe leave a comment on it if it's for something in the future 😄

parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -173,6 +180,14 @@ instance Monad (Result v loc) where
   CompilerBug bug es is >>= _ = CompilerBug bug es is
   {-# INLINE (>>=) #-}
 
+instance MonadFix (Result v loc) where
+  mfix f =
+    let res = f theA
+        theA = case res of
+          Success _ a -> a
+          _ -> error "mfix Result: forced an unsuccessful value"
```
ChrisPenner, line 188: At a glance this seems like a potential issue, isn't it possible that monadfix could try to force something that has an error, and in  hitting this case would swallow the meaningful error with this one?

Maybe it's not possible right now because we only pass things that typecheck to the pattern-match-checker?

parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -959,3 +966,38 @@ syntax.docFormatConsole d =
...
+    (Either (UF.UnisonFile Symbol Ann) TFile)
+
+type EitherResult = Either String TFile
+
+showNotes :: (Foldable f) => String -> PrintError.Env -> f Note -> String
+showNotes source env =
+  intercalateMap "\n\n" $ PrintError.renderNoteAsANSI 60 env source Path.absoluteEmpty
+
+decodeResult ::
+  String -> SynthResult -> EitherResult --  String (UF.TypecheckedUnisonFile Symbol Ann)
```
ChrisPenner: Is this comment meant to be here?

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -1020,13 +1058,18 @@ renderContext env ctx@(C.Context es) =
       shortName v <> " : " <> renderType' env (C.apply ctx t)
     showElem _ (C.Marker v) = "|" <> shortName v <> "|"
 
-renderTerm :: (IsString s, Var v) => Env -> C.Term v loc -> s
+-- Term.Term' (TypeVar v loc) v loc
```
ChrisPenner: Stranded comment

parser-typechecker/src/Unison/PatternMatchCoverage/UFMap.hs
```diff
@@ -0,0 +1,234 @@
...
+    toMaybe :: ExceptT (UnionHaltReason k v) m r -> m (Maybe r)
+    toMaybe (ExceptT action) =
+      action <&> \case
+        Right m -> Just m
+        Left r -> case r of
+          KeyNotFound _k -> Nothing
+          MergeFailed _v0 _v1 -> Nothing
+
+toClasses :: forall k v. (Ord k) => UFMap k v -> [(k, Set k, v)]
+toClasses m0 =
```
ChrisPenner, line 223: I know some of these names probably come straight out of the paper, but if I ever have to patch something in here I'm definitely not going to know what the intended purpose of `phi` is haha.

Would you mind expanding some of the binding names and add a couple comments, type sigs and maybe doc-string?

unison-src/transcripts/move-namespace.output.md
tstat: I tried to keep these transcript changes cleanly separated by commit. Looks like a change in `unison-src/transcripts-using-base/base.u`.

https://github.com/unisonweb/unison/pull/3737/commits/bfe8364bb0f4b125e86f62993e1d12d8da3040a1

parser-typechecker/src/Unison/PatternMatchCoverage/PmLit.hs
```diff
@@ -0,0 +1,23 @@
...
+data PmLit
+  = Int Int64
+  | Nat Word64
+  | Boolean Bool
+  | Float Double
+  | Text Text
+  | Char Char
+  deriving stock (Show, Eq, Ord)
+
+prettyPmLit :: (IsString s) => PmLit -> Pretty s
```
ChrisPenner, line 15: Not sure where these pretty prints show up; if they're just in debug messages or w/e it doesn't matter,
but if they show up to end-users I  don't think this is quite right since Unison displays Nat's and Char's differently than Haskell does 😄

parser-typechecker/src/Unison/PatternMatchCoverage/PmGrd.hs
```diff
@@ -0,0 +1,64 @@
...
+  | -- | @PmLet x expr@ corresponds to a @let x = expr@ guard. This actually
+    -- /binds/ @x@.
+    PmLet v (Term' vt v loc) (Type vt loc)
+  deriving stock (Show)
+
+prettyPmGrd :: (Var vt, Var v) => PmGrd vt v loc -> Pretty ColorText
+prettyPmGrd = \case
+  PmCon var con convars ->
+    let xs = string (show con) : (formatConVar <$> convars) ++ ["<-", string (show var)]
+        formatConVar (v, t) = sep " " ["(", string (show v), "::", TypePrinter.pretty PPE.empty t, ")"]
```
ChrisPenner:
```suggestion
        formatConVar (v, t) = sep " " ["(", string (show v), ":", TypePrinter.pretty PPE.empty t, ")"]
```

This is a unison type sig right?

This one catches me all the time haha.

unison-src/tests/tictactoe2.u
tstat: Yeah, the first case is the only non-redundant case in this one. The author meant to be using `same` as a guard it looks like.

parser-typechecker/src/Unison/PatternMatchCoverage/GrdTree.hs
```diff
@@ -0,0 +1,86 @@
...
+             in [indent "  " (sep "\n" (go x1 xs))]
+
+pattern Leaf :: l -> GrdTree n l
+pattern Leaf x = Fix (LeafF x)
+
+pattern Grd :: n -> GrdTree n l -> GrdTree n l
+pattern Grd x rest = Fix (GrdF x rest)
+
+pattern Fork :: NonEmpty (GrdTree n l) -> GrdTree n l
+pattern Fork alts = Fix (ForkF alts)
```
ChrisPenner, line 86: A COMPLETE pragma may be helpful here.

parser-typechecker/src/Unison/PatternMatchCoverage/Fix.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Unison.PatternMatchCoverage.Fix where
+
+newtype Fix f = Fix {unFix :: f (Fix f)}
```
ChrisPenner, line 8: Just curious, any particular reason to add our own rather than use [`Data.Fix`](https://hackage.haskell.org/package/data-fix-0.3.2/docs/Data-Fix.html) (it's already in our transitive dependencies).

I can see us probably using more fix-points and recursion schemes in the future so I think I'd prefer not to have duplicate Fix types floating around if possible since that always ends up causing some confusion 😄

ChrisPenner (approved):
Won't claim to have understood everything I read, but skimmed it all and left a couple notes.

Looks great! Thanks for putting so much work into this, it's a huge benefit for Unison!

parser-typechecker/src/Unison/PatternMatchCoverage.hs
ChrisPenner: (Also it prints using a "pretty" trace which is always nice for complex types).

parser-typechecker/src/Unison/Typechecker/Context.hs
tstat: Oh, just need to delete that comment

parser-typechecker/src/Unison/Typechecker/Context.hs
tstat: It is true in general when using value recursion that you could force the the thunk before it is resolved and cause your program to hang, in this case forcing the thunk when you don't return a `Success` would be an error. This is also true in the `MonadFix` instances for `Maybe`, `Either`, and such.

The danger of doing so is in an `mfix` call, or recursive do block, as that's when you have this thunk that will be a bomb if the result isn't a `Success`. There isn't an `mfix` call that spans the typechecker and pattern match coverage checker or something like that though.

unison-src/transcripts/pattern-match-coverage.md
```diff
@@ -0,0 +1,190 @@
...
+test : [Boolean] -> ()
+test = cases
+  [a, b] ++ xs -> ()
+  [] -> ()
+  xs :+ false -> ()
+  true +: xs -> ()
+```
+
+```unison:error
+test : [Boolean] -> ()
```
dolio, line 252: Why is this one an error?

parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -1212,13 +1251,69 @@ synthesizeWanted e
       let outputType = existential' l B.Blank outputTypev
       appendContext [existential outputTypev]
       cwant <- checkCases scrutineeType outputType cases
+      ensurePatternCoverage e scrutinee scrutineeType cases
       want <- coalesceWanted cwant swant
```
pchiusano, line 1254: Kind of a basic thing, but I like that you kept the basic structure of the typechecker and just call out to this other function in this one place.

dolio:
Looks pretty good to me, too.

I'm a bit curious about the `MonadFix` stuff. It wasn't obvious to me what was actually recursive in the do block there. But I probably just need to work through it more.

I like the examples, too, since for some of them (the list ones) it's not even easy to work out what the problem is knowing that there is a problem. :)

parser-typechecker/src/Unison/Typechecker/Context.hs
tstat: This example might make what I'm saying clearer:

`goodExample` replaces two ints in a `Map` with a thunk that will resolve to their sum (avoiding a second pass of inserts). `forceLeftExample` and `loopExample` illustrate erroneous ways to use `mfix`.
```haskell
{-# LANGUAGE LambdaCase #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TupleSections #-}

import Control.Exception
import Control.Monad.Fix
import Data.Functor.Compose
import Data.Map (Map)

main :: IO ()
main = do
  print (goodExample (Map.fromList [(A, 2)]))
  print (goodExample (Map.fromList [(A, 2), (B, 3)]))
  print (forceLeftExample (Map.fromList [(A, 2), (B, 3)])) `catch` \(e :: SomeException) -> do
    putStrLn ("caught exception: " <> show e)
  print (loopExample (Map.fromList [(A, 2), (B, 3)]))

goodExample :: Map Key Int -> Either String (Map Key Int)
goodExample m = snd <$> mfix summedMap
  where
    summedMap :: (Int, Map Key Int) -> Either String (Int, Map Key Int)
    summedMap ~(theSum, _) = do
      (a, m1) <- replaceAndReturnOriginal A theSum m
      (b, m2) <- replaceAndReturnOriginal B theSum m1
      pure (a + b, m2)

replaceAndReturnOriginal :: forall k v. (Show k, Ord k) => k -> v -> Map k v -> Either String (v, Map k v)
replaceAndReturnOriginal k v n =
  let alterFunc :: Maybe v -> Compose (Either String) ((,) v) (Maybe v)
      alterFunc x = Compose $ (,Just v) <$> toEither x

      toEither :: Maybe v -> Either String v
      toEither = \case
        Nothing -> Left ("Missing key: " <> show k)
        Just x -> Right x
   in getCompose $ Map.alterF alterFunc k n

loopExample :: Map Key Int -> Either String (Map Key Int)
loopExample m = snd <$> mfix summedMap
  where
    summedMap :: (Int, Map Key Int) -> Either String (Int, Map Key Int)
    summedMap ~(theSum, _) = do
      (a, m1) <- replaceAndReturnOriginal A theSum m
      (b, m2) <- theSum `seq` replaceAndReturnOriginal B theSum m1 -- we force theSum while computing theSum
      pure (a + b, m2)

forceLeftExample :: Map Key Int -> Either String (Map Key Int)
forceLeftExample m = snd <$> mfix summedMap
  where
    summedMap :: (Int, Map Key Int) -> Either String (Int, Map Key Int)
    summedMap bomb = do
      Left (show bomb) -- we smuggled our thunk out in a Left, but this will error due to the MonadFix instance

data Key = A | B deriving (Eq, Ord, Show)
```

The output is:
```
Left "Missing key: B"
Right (fromList [(A,5),(B,5)])
caught exception: mfix Either: Left
test: <<loop>>
```

parser-typechecker/src/Unison/Runtime/IOSource.hs
tstat: No it isn't. But this is some code I forgot to clean up. It is copied from `parser-typechecker/tests/Unison/Test/UnisonSources.hs` because the `base.u` typechecking failures weren't printed initially (leaving me very confused about what was going wrong!).

So, this `decodeResult` stuff should be moved to some appropriate place where it can be imported by `IOSource` and the `UnisonSources` test module I pulled it out of.

parser-typechecker/src/Unison/PatternMatchCoverage/PmLit.hs
tstat: All of these pretty prints are only for debugging.

parser-typechecker/src/Unison/PatternMatchCoverage/Fix.hs
tstat: I used it because it was easier to define some instances with quantified constraints instead of adding `Show1`, `Eq1`, and friends to all the base functors.

The use of `Fix` should be purely internal to the `PatternMatchCoverage` subtree; it isn't exposed in the (very small) interface given in `Unison.PatternMatchCoverage`.

tstat:
@dolio laziness is used for some small optimizations in `UFMap`'s `alterF'` and `union` implementations: avoiding a second pass of inserts in both cases. When `UFMap` was hooked up to the constraint solver then I realized I didn't have a pure merge function, but needed to merge in some monad. So, I added the `MonadFix` instance and kept the lazy implementation.

PR #3740, opened 2023-01-12
Saving current echo state before running anything and then falling back to it if needed

unorsk:
## Overview

If you run the code below in the REPL, the echo state doesn't get reverted to where it was.

```
main : '{IO, Exception} ()
main = do
  setEcho stdIn false
```

It will result in a behaviour where you'd by typing commands and not see them printed out in the terminal.

Screenshot 2023-01-12 at 17 43 49

## Implementation notes

The desired behaviour would be to save the initial echo state (but I am guessing it could always be set to `false` which will make the implementation even shorter)
I made a very straightforward and somehow naïve implementation. So I am very open for any comments and feedback :)

No

## Test coverage

No

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -72,3 +72,4 @@ The format for this list: name, GitHub handle
 * Emil Hotkowski (@emilhotkowski)
 * Jesse Looney (@jesselooney)
 * Vlad Posmangiu Luchian (@cstml)
+* Andrii Uvarov (@unorsk)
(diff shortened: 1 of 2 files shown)


pchiusano (approved):
This seems good to me. Thanks! Could you add yourself to CONTRIBUTORS.markdown?

unorsk:
> This seems good to me. Thanks! Could you add yourself to CONTRIBUTORS.markdown?

Done!

pchiusano:
Nice! Merging.

PR #3745, opened 2023-01-13
Add builtins for Ref+Promise concurrency scheme

SystemFw:
## Overview

Adds Ref + Promise, two low level concurrency primitives that can be used to build a variety of concurrent abstractions. This should remove dependencies from STM for the time being, so that we have an easier time getting code onto the JIT 

## Implementation notes

Straightforward addition of builtins. There are some notes in the code about details of how Haskell treats compare-and-swap and memory fences.

They are documented in the code

## Test coverage

The test coverage is adequate for now, but this is concurrent code so building a richer suite over time doesn't hurt

## Loose ends

Various builtins like the MVar ops, `sleep`, etc have the wrong type in the sense that they are fallible, but those exceptions are async exceptions and should be represented in the type


pchiusano (approved):
LGTM.

pchiusano (approved):
Nice!

PR #3764, opened 2023-01-23
Add a Debug.toText builtin

dolio:
This reworks the `Debug.trace` machinery to allow getting just the text that would be shown for a value. Instead of carrying a tracing function in the runtime, a function that generates the (annotated) text is used, and the logic for printing it is moved to the runtime itself.

Also, a builtin operation `Debug.toText` is available for getting at the text. It has three possible results:

  * `None` - the tracer is completely off
  * `Some (Left txt)` - an 'ugly' textual representation
  * `Some (Right txt)` - a 'pretty' textual representation

@jaredly This lets you get text like `"termLink Map.insert"` for term links when running inside UCM.


pchiusano (approved):
Cool.

pchiusano:
I've added an alias to base for this as well.

PR #3778, opened 2023-01-27
Add nix flake for common development environment

tstat:


ceedubs:
@tstat wonderful thank you!

The first thing that I ran into is that this could use a `nix flake update`, because the flake-util in the lock file is an old version that doesn't have aarch64-darwin in the default systems.

I've changed that locally and I'll see how it goes.

ceedubs:
Unfortunately on aarch64-darwin this fails for me with:

```
<no location info>: error:
    Warning: Couldn't figure out LLVM version!
             Make sure you have installed LLVM between [9 and 13)
ghc: could not execute: opt
```

I don't get the same with https://github.com/ceedubs/unison-nix/blob/trunk/nix/dev-shell.nix . Does `my-stack` need to be modified to point to this ghc? I can poke around a bit, but it might be a few days before I get to it.

ceedubs (approved):
🎉 ❄️

PR #3780, opened 2023-01-28
Cryptographic Hashing! Adding some primops and crypto FOps

jaredly:
## Summary:
Because this introduces new runtime dynamic dependencies, I've made them "soft" dependencies for now: you only need to install libcrypto and libb2 if you're using the cryptographic hash functions. Otherwise, this shouldn't impact anyone else's workstream. I've also tried to provide verbose & helpful error messages.

## Test plan:
I'm using the following scratch file, and I've had to do some manual tweaking of the generated scheme 😅 but at least I've verified that the crypto.ss terms are working as expected!
```
sha = do
    printLine (Bytes.toHex (hashBytes Sha1 <| Text.toUtf8 ""))
    test.Result.Ok "Yo"
```

### DIFF
--- a/chez-libs/unison/primops.ss
+++ b/chez-libs/unison/primops.ss
@@ -101,12 +101,34 @@
     unison-POp-VALU
     unison-POp-VWLS
 
+    unison-POp-UPKB
+    unison-POp-ADDI
+    unison-POp-DIVI
+    unison-POp-EQLI
+    unison-POp-MODI
+    unison-POp-LEQI
+    unison-POp-POWN
+    unison-POp-VWRS
+
+    unison-FOp-crypto.HashAlgorithm.Sha1
+    unison-FOp-crypto.hashBytes
     )
 
   (import (chezscheme)
           (unison core)
           (unison string)
+          (unison crypto)
           (unison bytevector))
+
+  (define unison-POp-UPKB bytevector->u8-list)
+  (define unison-POp-ADDI +)
+  (define unison-POp-DIVI /)
+  (define (unison-POp-EQLI a b)
+    (if (= a b) 1 0)
+  )
+  (define unison-POp-MODI mod)
+  (define unison-POp-LEQI <=)
+  (define unison-POp-POWN expt)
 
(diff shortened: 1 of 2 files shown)


chez-libs/unison/primops.ss
```diff
@@ -145,11 +167,8 @@
   (define (unison-POp-FTOT f) (number->istring f))
   (define (unison-POp-IDXB n bs) (bytevector-u8-ref bs n))
   (define (unison-POp-IDXS n l)
-    (call/1cc
-      (lambda (k)
-        (with-exception-handler
-          (lambda (e) (list 0))
-          (lambda () (list-ref l n))))))
+    (guard (x [else (list 0)])
+      (list 1 (list-ref l n))))
```
jaredly, line 171: `guard` seemed like a simpler api to catch exceptions, and the previous version wasn't working for me 🙃

chez-libs/unison/primops.ss
```diff
@@ -101,13 +101,35 @@
...
     unison-POp-VWLS
 
+    unison-POp-UPKB
+    unison-POp-ADDI
+    unison-POp-DIVI
+    unison-POp-EQLI
+    unison-POp-MODI
+    unison-POp-LEQI
+    unison-POp-POWN
+    unison-POp-VWRS
```
jaredly, line 111: these POps aren't strictly necessary as part of this PR, but they were needed to run the tests I was doing.

chez-libs/unison/primops.ss
dolio: Yeah, somehow I missed `guard` when I was reading about exceptions (I guess because examples used `call/cc` instead for some reason). It's the portable solution, and exactly what we wanted anyway.

dolio (approved):
I think this looks good, from what I can understand.

FYI, I've been working a little on trying to split things into 'standard scheme' parts and implementation-specific parts. `primops` is one of the things that I think can just contain standard scheme, supported by more implementation specific modules like `crypto`.

PR #3781, opened 2023-01-30
Add a basic transcript testing chez generation

jaredly:
This runs `generateScheme` to generate some scheme! We don't yet have a ucm command to shell out, so you have to run `scheme` yourself.

It takes about a minute to download all of that code, which is kindof a pain 🙃


dolio (approved):
Looks good to me.

I guess this will only work on  unix (only Linux?), but I don't think there's much that can be done about that right now.

Edit: actually, is that last line not actually executed? In that case it's not doing anything platform specific, aside from advice.

pchiusano:
@jaredly we do have commands to shell out now, want to use those? https://github.com/unisonweb/unison/pull/3770 `IO.process.call` if you want to do a synchronous call to an external process.

jaredly:
Oh nice, yeah that sounds great

chez-libs/tests/basic.output.md
```diff
@@ -0,0 +1,93 @@
...
+  
+    ⍟ These new definitions are ok to `add`:
+    
+      test1 : '{IO, Exception} Nat
+
+```
+```ucm
+.> run test1
+
+  0
```
jaredly: there's something weird about the interaction of transcripts w/ IO.process; the stdout of the IO.Process.call doesn't get captured here, but it shows up fine if you `run` from `ucm`. @dolio do you know what might be happening here?

dolio: Yeah. Transcripts just don't capture standard output at all. This is also true if you just call prints in unison. I think if you want the output of the process to be displayed in the transcript, you're going to have to use the more complicated API with explicit handles.

The Haskell library has wrapper functions for these kind of things, like `Command -> [String] -> String -> IO String`, where the `String` argument is the input to be sent to the process's stdin, and the result string is what the process printed to stdout. I didn't wrap those, though, because I figured we could just write them in unison.

chez-libs/tests/basic.output.md
```diff
@@ -0,0 +1,77 @@
...
+  
+    ⍟ These new definitions are ok to `add`:
+    
+      test1 : '{IO, Exception} Text
+
+```
+```ucm
+.> run test1
+
+  ""
```
jaredly: so now it's differently weird, I'm reading the stdout from the process explicitly, and under transcript it's coming back empty, but under normal execution it's returning `"Hello\n"` like I expect

jaredly:
ok but why is github so confused about what commits are on this branch

scheme-libs/tests/basic.output.md
```diff
@@ -0,0 +1,39 @@
...
+    
+      test1 : '{IO, Exception} Text
+
+```
+```ucm
+.> run test1
+
+  "Hello\n"
+
+```
```
jaredly, line 39: it works!

PR #3793, opened 2023-02-02
Add a chez transcript to test `Bytes.toHex` to prepare for crypto tess

jaredly:
I had to make two changes to the scheme generation (illustrated in bytes.md)
- builtinLinks needed `Bytes.toList` and `Bytes.fromList` removed
- SchemeDefn.toIndentedText needed to be generating a different definition for Char.toNat and Char.fromNat. This probably isn't the right way to do it though 😉 `Char.fromNat` and `Char.toNat` are defined as a `cast` in `Builtin.hs`, which I assume is why they get generated as `identity` in the scheme output. Do we want to indicate there somehow that scheme will need it to be more than just a `cast`?

### DIFF
--- a/scheme-libs/tests/bytes.output.md
+++ b/scheme-libs/tests/bytes.output.md
@@ -1,14 +1,6 @@
 
 Note: This should be forked off of the codebase created by base.md
 
-```ucm
-.> pull dolio.public.internal.trunk.impls impls
-
-  ✅
-  
-  ✅ Successfully pulled into newly created namespace impls.
-
-```
 ```unison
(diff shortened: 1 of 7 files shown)


scheme-libs/tests/bytes.output.md
```diff
@@ -0,0 +1,95 @@
...
+  
+    ⍟ These new definitions are ok to `add`:
+    
+      bytes : '{IO, Exception} Text
+
+```
+```ucm
+.> run bytes
+
+  "480\n"
```
jaredly, line 93: So this is not the correct output still -- there's something else going sideways that I need to track down. But at least it's not throwing any exceptions!

PR #3800, opened 2023-02-03
Expect full scheme-libs structure for 'installed' libraries

dolio:
This allows just (recursively) copying the `scheme-libs` directory to the installed location, rather than requiring a merge of the directory structures depending on which implementation is to be used. We can even have a setting to toggle between implementations in the future.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
@@ -2766,7 +2766,10 @@ runScheme file args0 = do
   ensureSchemeExists
   gendir <- getSchemeGenLibDir
   statdir <- getSchemeStaticLibDir
-  let includes = gendir ++ ":" ++ statdir
+  let includes =
+        gendir ++ ":" ++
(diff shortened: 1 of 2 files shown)


pchiusano (approved):

unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2766,7 +2766,10 @@ runScheme file args0 = do
   ensureSchemeExists
   gendir <- getSchemeGenLibDir
   statdir <- getSchemeStaticLibDir
-  let includes = gendir ++ ":" ++ statdir
+  let includes =
+        gendir ++ ":" ++
+        (statdir </> "common") ++ ":" ++
+        (statdir </> "chez")
```
pchiusano, line 2772: @dolio should this be `"racket"` rather than `"chez"`?

dolio: Not until we're actually calling racket.

PR #3802, opened 2023-02-04
Add builtins test suite skeleton, which can be run using the JIT or interpreted

pchiusano:
This PR adds a test suite skeleton that can be run using the JIT or interpreted. For now we're just going to run the tests manually, but we'll incorporate some version of this into CI when we're further along.

The test suite isn't filled in yet but we can add to `unison-src/builtin-tests/tests.u`.

It builds on @jaredly's https://github.com/unisonweb/unison/pull/3781 PR which caches a codebase after downloading base and the needed scheme libs, but the transcripts use the `compile.native.fetch` and `run.native` commands rather than manually invoking the scheme generation code.

The nice thing about this setup is that we can now use any of the code in base to help us write our tests.

```bash
$ ./unison-src/builtin-tests/interpreter-tests.sh
```

image

Here's the Readme.md added to `unison-src/builtin-tests`:

Edit `tests.u` in this directory to add to the test suite. The same test suite can be run using the JIT or the interpreter, using either of the two scripts:

```bash
$ unison-src/builtin-tests/jit-tests.sh
```

```bash
$ unison-src/builtin-tests/interpreter-tests.sh
```

The scripts will fetch a copy of base and the scheme codegen library and cache it for subsequent runs.

At the moment, running with Chez fails with:

```
⚙️   Processing stanza 3 of 3.Exception: invalid primitive name $call-in-continuation at line 69, char 6 of /Users/pchiusano/.local/share/unisonlanguage/scheme-libs/chez/unison/cont.ss
```

But I believe this is just because the Chez implementation of abilities is busted / incomplete at the moment. @dolio is working on switching the abilities implementation to use Racket, so once that's done, hopefully this will all work.


jaredly (approved):
Love it! Very clean

SystemFw (approved):

PR #3806, opened 2023-02-07
Fix some bugs in racket/unison/core.ss

jaredly:
It's possible I'm just not running it right, but when I tried to evaluate `core.ss` with racket there were some undefined terms, which I solved by adding some requires, and creating the isolated-string-copy.rkt file.

Now `unison/core` is requirable in the racket runtime!


jaredly:
we can ignore this, dan's branch has this fixed

PR #3807, opened 2023-02-07
Add cryptographic hashing primitives for the racket backend

jaredly:
The `crypto-lib` library that I'm using is published under LGPL -- does that work for us? https://docs.racket-lang.org/crypto/index.html


pchiusano:
I'd really prefer not to use anything LGPL. It's a complicated and IMO unclear license that a lot of companies aren't sure what to make of (at many companies, "can we use LGPL" turns into a conversation with lawyers) and even if you know what you're doing it places a burden on our users - my understanding is that if our compilation process statically links against LGPL libraries, we'd have to be able to distribute an "unlinked" version that users can link against alternate versions of the LGPL library. Which sounds complicated for us, assuming we can even agree on what that means exactly in our case. 

Are there no other options?

jaredly:
oh yeah, if LGPL is a problem, we can definitely do the glue code ourselves.

pchiusano:
Okay, cool, let's do that. Kind of a shame that this library (which looks like it wraps some Apache-licensed C libraries like OpenSSL) went with a more copyleft license, but oh well.

jaredly:
@pchiusano ok, no more LGPL dependency! wasn't too painful.

jaredly:
huh so I added hmac functions, and the blake2 hmacs aren't matching what unison haskell runtime produces 🤔

pchiusano:
@jaredly nice! can you also add some tests to `unison-src/builtin-tests/tests.u`that exercise all these? (the tests don't successfully run at the moment, we think due to a bug in the abilities implementation in Racket, but @dolio should have a fix shortly)

Is it all the hmacs that are off, or just when using Blake2? If it's just Blake2, maybe you're using a different variant of it than the Haskell one? https://en.wikipedia.org/wiki/BLAKE_(hash_function)

jaredly:
Just Blake 2 - so I thought the "key" parameter the Blake functions accepted was for an hmac, and it's not; I'll just need to do the hmac by hand

pchiusano:
@jaredly there's Blake2, Blake2s, and Blake2b - we're using Blake2s (256 bits) and Blake2b (256 bits and 512 bits), is that also what you're using?

Also do the Blake2b and Blake2s hash values agree if you just hash some bytes (independent of using them in hmac?)

jaredly:
so the hashes agree, it's just the hmac that I was doing wrong. I've got it now :)

unison-src/builtin-tests/interpreter-tests.md
```diff
@@ -2,6 +2,7 @@
 Note: This should be forked off of the codebase created by base.md
 
 ```ucm
+.> builtins.merge
```
pchiusano: No need to do this. The transcript is already pulling base, which includes names for all builtins.

unison-src/builtin-tests/testlib.u
```diff
@@ -1,3 +1,4 @@
+use builtin.Universal ==
```
pchiusano: Use `===` for universal / structural equality, it's in base: https://share.unison-lang.org/@unison/p/code/latest/namespaces/public/base/latest/;/terms/Universal/%3D%3D%3D

unison-src/builtin-tests/tests.u
```diff
@@ -1,7 +1,27 @@
...
+  check "Sha2_512 hmacBytes" do hmacBytes Sha2_512 (toUtf8 "key") (toUtf8 "") == 0xs84fa5aa0279bbc473267d05a53ea03310a987cecc4c1535ff29b6d76b8f1444a728df3aadb89d4a9a6709e1998f373566e8f824a8ca93b1821f0b69bc2a2f65e
+  check "Sha3_256 hmacBytes" do hmacBytes Sha3_256 (toUtf8 "key") (toUtf8 "") == 0xs74f3c030ecc36a1835d04a333ebb7fce2688c0c78fb0bcf9592213331c884c75
+  check "Sha3_512 hmacBytes" do hmacBytes Sha3_512 (toUtf8 "key") (toUtf8 "") == 0xs7539119b6367aa902bdc6f558d20c906d6acbd4aba3fd344eb08b0200144a1fa453ff6e7919962358be53f6db2a320d1852c52a3dea3e907070775f7a91f1282
+  check "Blake2s_256 hmacBytes" do hmacBytes Blake2s_256 (toUtf8 "key") (toUtf8 "") == 0xs67148074efc0f6741b474ef81c4d98d266e880d372fe723d2569b1d414d234be
+  check "Blake2b_256 hmacBytes" do hmacBytes Blake2b_256 (toUtf8 "key") (toUtf8 "") == 0xs4224e1297e51239a642e21f756bde2785716f872298178180d7f3d1d36a5e4e4
+  check "Blake2b_512 hmacBytes" do hmacBytes Blake2b_512 (toUtf8 "key") (toUtf8 "") == 0xs019fe04bf010b8d72772e6b46897ecf74b4878c394ff2c4d5cfa0b7cc9bbefcb28c36de23cef03089db9c3d900468c89804f135e9fdef7ec9b3c7abe50ed33d3
```
pchiusano: I'd convert these to use `===` and merge with trunk, which has fixes to abilities implementation, so the test suite should actually be able to run.

unison-src/builtin-tests/jit-tests.sh
```diff
@@ -5,7 +5,7 @@ ucm=$(stack exec -- which unison)
 
 base_codebase=${XDG_CACHE_HOME:-"$HOME/.cache"}/unisonlanguage/base.unison
 
-if [ ! -d $base_dir ]; then
+if [ ! -d $base_codebase ]; then
```
pchiusano: This is fixed in trunk, probably just need to refresh.

pchiusano (approved):
Nice! Left a few comments for you, and I think you gotta merge trunk into your branch.

dolio (approved):
Looks good to me.

In case there was any doubt, I think it's fine to write the racket-specific libs using the racket 'language,' as was done here. The various surface languages can import libraries written in other languages.

Issue #3808, opened 2023-02-07
TLS example code (IO.net.Tls) doesn't work


jaredly:
https://share.unison-lang.org/@unison/p/code/latest/namespaces/public/base/latest/;/types/IO/net/Tls

The example given, "Connect to a TLS server" doesn't work
```
socket = Socket.client (HostName "google.com") (Port "443")
config = ClientConfig.default (HostName "google.com") ""
tls = newClient config socket
conn = handshake tls
TlsSocket.send conn (toUtf8 "GET / HTTP/1.0\r\n\r\n")
response = TlsSocket.receive conn
TlsSocket.close
response
```
for a couple of reasons, `TlsSocket.close` needs an argument, and then `TlsSocket.send` fails hard on evaluation, with the following runtime error:
```
  dumpData: bad closure: Foreign (Wrap ##Tls _)
  expected type: #00nv2
```

I did manage to get a TLS example running, based on [tls.md](https://github.com/unisonweb/unison/blob/trunk/unison-src/transcripts-using-base/tls.md), but I had to do some nonstandard things:
```ucm
.> builtins.mergeio
.> cd builtin
.> load unison-src/transcripts-using-base/base.u
.> add
```
and then I was able to get this to run:
```
getExample _ =
    socket = clientSocket "example.com" "443"
    config = ClientConfig.default "example.com" Bytes.empty
    tls = newClient config socket
    reraise (handshake.impl tls)
    send tls (toUtf8 "GET /index.html HTTP/1.0\r\n\r\n")
    fromUtf8 (receive tls)
```

runarorama:
This was a bug in the doc. I've changed it to be:

```unison
socket = Socket.client (HostName "google.com") (Port "443")
config = ClientConfig.default (HostName "google.com") ""
tls = newClient config socket
conn = handshake tls
send conn (toUtf8 "GET / HTTP/1.0\r\n\r\n")
response = receive conn
_ = close conn
fromUtf8 response
```

The issue was that the `close` function was missing an argument.

jaredly:
@runarorama it's true that the close function needed an argument, but it still crashes for me with that fixed 🤔

runarorama:
I'm not able to reproduce that. Tested with UCM version M4f as well as a trunk build as of this morning, both with `main` and `v1_1_0` of Base.

Which versions are you using?

PR #3811, opened 2023-02-07
Fix tls calling conventions

stew:
## Overview

Tls.send and Tls.terminate builtin functions had a problem with their calling convention. They should have been returning an `Either Failure ()` as they were declared, but they were actually retruning an `Either Failure Tls`.

This adds a transcript which would fail without this fix.

The test I added makes and actual TLS connection to example.com:443 which might not be Ideal. It is possible that we could write a test that spawns two threads on the same box which might be a little safer.

## Test coverage

see above


pchiusano (approved):

pchiusano:
cc @jaredly this sounds like the issue you were hitting in #3808

PR #3814, opened 2023-02-07
Builtins test suite no longer uses IO refs

pchiusano:
Also changed jit-tests.sh to fetch latest scheme generating library and fix a typo in the shell scripts.

With this, when I do `unison-src/builtin-tests/jit-tests.sh`, I now get as far as:

```
*** Test suite ***

result arity mismatch;
 expected number of values not received
  expected: 1
  received: 0
  context...:
   /Users/pchiusano/.local/share/unisonlanguage/scheme-libs/common/unison/boot.ss:86:27: fast-path
   [repeats 3 more times]
   body of "/Users/pchiusano/.cache/unisonlanguage/scheme-tmp/tests.scm"
```

The `*** Test suite ***` is actual Unison code being executed so that's a good sign - that's a message from the test library (`testlib.u`).

### DIFF
--- a/scheme-libs/racket/unison/crypto.ss
+++ b/scheme-libs/racket/unison/crypto.ss
@@ -1,0 +1,11 @@
+#!r6rs
+;; stubbed out just to avoid import error, replace with real thing
+(library (unison crypto)
+  (export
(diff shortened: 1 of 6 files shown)


dolio (approved):
Looks fine.

PR #3815, opened 2023-02-08
Add concurrency primitives for the JIT, based on Racket threads

SystemFw:
## Overview

Adds the foundational concurrency primitives to the JIT:
- mutable variables + CAS
- single-shot Promise
- sleep
- fork & kill threads

It also points UCM to the `unison.public.internal.trunk`  compiler lib,  fixes a bug in `killThread` that had gone undetected, and adds a `data` library to help constructs results that respect the calling convention.

## Implementation notes

The threading is built on Racket threading, which is based on single core green threads.
Most things are mapping and reexports, the most interesting thing is the implementation of Promise. 
Roughly speaking, it uses CAS on a box and a single empty semaphore to implement the Promise state machine. The key component here is that Racket exposes a `semaphore-peek-evt` which lets a thread wait on an empty semaphore _without_ decrementing the number of permits, which in turn means we can easily implement waiting for multiple readers.

Racket logs uncaught thread interruption exceptions, Haskell doesn't.  Should we mimic the Haskell behaviour or use Racket's?

I was tempted to improve the interruption model by changing `tryEval` to introduce interruption only for safe-points (`Promise.read`, `sleep`) , but then we decided not to.

## Test coverage

Tested via the new `tests.u` mechanism, specifically see `concurrency-tests.u`

## Loose ends

I've decided to tackle `tryEval` in a different PR.


pchiusano (approved):
Nice!

dolio (approved):
Looks fine to me.

PR #3819, opened 2023-02-10
Make scheme jit transcript work

dolio:
This PR includes work that allows the JIT test transcript to actually complete. It mainly includes fixes to the way abilities are implemented in scheme, but also includes a couple miscellaneous changes that I think are helpful when doing some debugging of the scheme output.


pchiusano:
NICE!!!!

pchiusano (approved):

PR #3820, opened 2023-02-10
Char.Class builtin type and supporting functions

runarorama:
## Overview

This adds a new builtin type `Char.Class` and supporting builtin functions, including integration with `Pattern Text`.

New built-in terms:

In `Char.Class` namespace:

```
  1.  alphanumeric (Class)
  2.  and          (Class -> Class -> Class)
  3.  any          (Class)
  4.  anyOf        ([Char] -> Class)
  5.  control      (Class)
  6.  is           (Class -> Char -> Boolean)
  7.  letter       (Class)
  8.  lower        (Class)
  9.  mark         (Class)
  10. not          (Class -> Class)
  11. number       (Class)
  12. or           (Class -> Class -> Class)
  13. printable    (Class)
  14. punctuation  (Class)
  15. range        (Char -> Char -> Class)
  16. separator    (Class)
  17. symbol       (Class)
  18. upper        (Class)
  19. whitespace   (Class)
``` 

In the `Text.patterns` namespace:

```
char         (Class -> Pattern Text)
```

## Implementation notes

This changes the `Unison.Util.Text.Pattern` module somewhat. Special-casing on character ranges and character sets is removed from the `Pattern` type, replaced with a `Char` constructor that takes a `CharacterPattern`:

```
data Pattern
  = Join [Pattern] -- sequencing of patterns
  | Or Pattern Pattern -- left-biased choice: tries second pattern only if first fails
  | Capture Pattern -- capture all the text consumed by the inner pattern, discarding its subcaptures
  | Many Pattern -- zero or more repetitions (at least 1 can be written: Join [p, Many p])
  | Replicate Int Int Pattern -- m to n occurrences of a pattern, optional = 0-1
  | Eof -- succeed if given the empty text, fail otherwise
  | Literal Text -- succeed if input starts with the given text, advance by that text
  | Char CharPattern -- succeed if input starts with a char matching the given pattern, advance by 1 char
  deriving (Show, Eq, Ord)
```

`CharacterPattern` looks like this:

```
data CharPattern
  = Any -- any char
  | Not CharPattern -- negation of the given pattern
  | Union CharPattern CharPattern -- match if either pattern matches
  | Intersect CharPattern CharPattern -- match if both patterns match
  | CharRange Char Char -- match if char is in the given range
  | CharSet [Char] -- match if char is in the given set
  | CharClass CharClass -- match if char is in the given class
  deriving (Show, Eq, Ord)

data CharClass
  = AlphaNum -- alphabetic or numeric characters
  | Upper -- uppercase alphabetic characters
  | Lower -- lowercase alphabetic characters
  | Whitespace -- whitespace characters (space, tab, newline, etc.)
  | Control -- non-printing control characters
  | Printable -- letters, numbers, punctuation, symbols, spaces
  | MarkChar -- accents, diacritics, etc.
  | Number -- numeric characters in any script
  | Punctuation -- connectors, brackets, quotes
  | Symbol -- symbols (math, currency, etc.)
  | Separator -- spaces, line separators, paragraph separators
  | Letter -- letters in any script
  deriving (Show, Eq, Ord)
```

This lets us do more general patterns with character classes, like:

* `many (char (and alphanumeric (not upper)))` matches a string of alphanumeric characters that are not uppercase letters.
* `join [letter, char (many (or symbol (or alphanumeric mark))]` to capture a valid XML element name.

I've left all the existing special-case pattern builtins but just changed their implementation.

## Test coverage

Tested by a Unison test suite in the Base library.


ceedubs:
@runarorama if you want to not special-case `Char` in `Pattern` then one option is to use a discrete interval encoding tree and support any discrete type. I've done this in the implementation of a regular expression and it worked pretty well. The potentially big downside that I can think of is that it might lead to extra boxing of chars and slow `Pattern` down? But I guess the current hyrbrid state feels a bit weird to me and I'd be inclined to either specialize `Pattern` to `Char` or make classes support multiple types.

pchiusano:
@ceedubs yeah, that's an interesting idea. No strong opinions on that, but I'd lean toward just having `Char.Class` be monomorphic as it is in this PR. We can always reuse code in the implementations on the implementation of the builtins, once we support more pattern types than just `Pattern Text`.

pchiusano (approved):
Nice!

runarorama:
I think the next one we're going to support is `Pattern Bytes`, and there's not really a coherent notion of a "bytes class", so I'm not sure how well that concept translates to non-text things.

PR #3824, opened 2023-02-13
Format codebase using Ormolu and fail PRs if not ormolu formatted

pchiusano:
Tagging a bunch of people on this PR so no one is surprised to see this coming. I'd like to just have CI enforce the formatting we want. Things have started to drift a bit. 

It's only a minor annoyance once per contributor to set up your editor and/or a pre-commit hook to format things correctly, but as a major bonus all subsequent PRs will not have needlessly large diffs and/or conflicts due to formatting or Ormolu version drift. And we can all avoid thinking any more about formatting.

Turns out there's already a handy GH action for this sort of thing, which CI invokes now, so this was super easy.

This seems pretty low risk. If people happen to have in-flight PRs that are behind this PR, they can just run this:

```
ormolu --mode inplace $(find . -name '*.hs')
```

One minor thing I noticed - Ormolu's parser barfs when fed files with CPP, so I wrapped the relevant sections in `{- ORMOLU_DISABLE -}` in this PR.

Instructions for this are added to [development.markdown](https://github.com/unisonweb/unison/blob/4bd4eff18460836d00e5ee88358e619810aad0ca/development.markdown)


dolio:
What do you want reviewed on this? I don't think I can review all 223 files. :)

pchiusano:
@dolio I just tagged core folks who work on the Haskell codebase just to make it more likely that you see this coming and aren't surprised. :)

.github/workflows/ci.yaml
```diff
@@ -17,9 +17,19 @@ on:
...
 
 jobs:
+
+  ormolu:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: unison
+      - uses: mrkkrp/ormolu-action@v10
```
aryairani:
```suggestion
      - uses: mrkkrp/ormolu-action@v10 # v10 uses ormolu 0.5.3.0
```

aryairani (approved):

dolio:
Okay. I thought that might be the case.

.github/workflows/ci.yaml
pchiusano: Good idea, thanks!

pchiusano: This version info is also noted in development.markdown btw.

PR #3826, opened 2023-02-14
Make compile.native use racket

dolio:
Not much to report on this one. Racket actually allows building self-contained executables rather than 'boot' files like chez.


pchiusano (approved):

pchiusano:
Nice!

PR #3827, opened 2023-02-15
Initial bare bones TLS implementation

jaredly:
Turns out racket's `openssl` lib did expose the things I needed 🎉

Just enough to make this example work:
```hs
getExample _ =
    socket = Socket.client (HostName "example.com") (Port "443")
    config = ClientConfig.default (HostName "example.com") ""
    tls = base.IO.net.Tls.newClient config socket
    conn = base.IO.net.Tls.handshake tls
    TlsSocket.send conn (toUtf8 "GET /index.html HTTP/1.0\r\n\r\n")
    response = receive tls
    TlsSocket.close conn
    builtin.printLine (base.Text.fromUtf8 response)
```

Open issues:
- [ ] I haven't handled any errors, need to add that
- [ ] I'm ignoring certificateSuffix for now, haven't figured out what that correlates to in racket's lib
- [ ] Text.fromUtf8 will be different, as we're now doing an impl of Text in scheme

### DIFF
--- a/scheme-libs/common/unison/boot.ss
+++ b/scheme-libs/common/unison/boot.ss
@@ -153,7 +153,10 @@
   (define-syntax request
     (syntax-rules ()
       [(request r t . args)
-       ((cdr (ref-mark (quote r))) (list (quote r) t . args))]))
+      (let ((ref-mark-result (ref-mark (quote r))))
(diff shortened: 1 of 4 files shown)


jaredly:
I'm splitting this out

PR #3828, opened 2023-02-15
Improve error message for unhandled ability

ceedubs:
In certain cases this will change the failure message from something
like `resolve: unhandled ability request: 1561` to `resolve: unhandled
ability request: #ifg08` where `#ifg08` is the has of the ability that
was unhandled. At this stage we don't have convenient access to the
actual constructor within the ability that was not handled.

I haven't added tests because I don't really know how to set them up.
But it's not really a big deal if there's a regression in this error
message, so I'm not sure whether they'd be worth the trouble.


dolio (approved):
Looks good.

PR #3831, opened 2023-02-17
Add tryEval to the JIT compiler

SystemFw:
## Overview

Adds tryEval, which catches native failures. The main practical use case is dealing with async exceptions.

## Implementation notes

Uses the native racket mechanism for exceptions.

Categorises exceptions when possible, using racket's predicates

## Test coverage

Tested the main use case with async exceptions, and tested `bug` as far as possible right now.

## Loose ends

- The JIT compiler doesn't have type links yet, so I'm making up a String tag. They will be replaced once type links are in.
- I didn't fully model the behaviour of `bug`, because we don't have universal toText yet.
-  I will open another PR to add a type link for ThreadKilled on the Haskell side, but I think it's better to merge this one now because it contains helpers to build exceptions that will be useful to add network/file builtins

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ dist-newstyle
 # GHC
 *.hie
 *.prof
+/libb2.dylib
--- a/scheme-libs/chez/unison/concurrent.ss
+++ b/scheme-libs/chez/unison/concurrent.ss
@@ -10,13 +10,13 @@
     promise-try-read
     fork
     kill
-    sleep)
+    sleep
+    try-eval)
 
 
   (define err "This operation is not supported on the pure Chez Scheme
 backend, use the Racket over Chez Scheme backend")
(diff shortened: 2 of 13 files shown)


ceedubs:
> The main practical use case is dealing with async exceptions.

Looking at the diff, it looks like this also catches `bug`, pattern match failures, etc like the Haskell runtime version does, right? If so, then another pretty important use case is making sure that one of these (sync) types of failures doesn't prevent resource cleanup.

SystemFw:
> If so, then another pretty important use case is making sure that one of these (sync) types of failures doesn't prevent resource cleanup.

it catches everything as far as I can tell yeah

pchiusano (approved):

dolio (approved):
I think this looks good.

PR #3832, opened 2023-02-18
Fix `request` for the case of unhandled abilities

jaredly:
Previously it would give an untraceable scheme error, "expected mpair, found #f".
This now just raises the ability request as an exception.


dolio:
So, this just raises with the list of arguments. Maybe it should include the `r` and `t` part, too?

Also, maybe we should define a [condition type](https://www.scheme.com/tspl4/exceptions.html#./exceptions:h2) for this so it's more obvious what's going on when it happens?

jaredly:
Ok, added the other args, and made it a condition!

jaredly:
@dolio can we merge this too?

dolio (approved):
Looks good.

PR #3833, opened 2023-02-18
Add new serialization builtins: `Code.serialize.small` and `Value.serialize.small`

pchiusano:
These work the same as `Code.serialize` and `Value.serialize` but assigns each `Reference` an integer id (starting from 0) and uses that instead of writing / reading 512 bit hashes every time. Furthermore, the ints are variable-length encoded, so in this new format a `Reference` that's been seen before will typically be encoded with 2 or 3 bytes (instead of 64 bytes with the old format).

Also uses variable-length encoding for constructor tags. Typically a data type has fewer than 127 constructors (!!) so this takes constructor tags down to 1 byte instead of 8 bytes.

There are no new functions for deserialization, instead, `Value.deserialize` and `Code.deserialize` have been modified to read this format as well, keying off the version number which starts off the serialized form. A version number of 4 is this new format, and a version number of 3 is the old format.

## Is it better? 

As a quick test, I compared lists like `[Some 1, Some 2, ...]` of various sizes. This is kind of a worst case for the existing format, since it's encoding the `Optional` hash over and over, and it's encoding the constructor tag with 8 bytes instead of 1. The uncompressed size is about 4x smaller in this PR.

However, compressed size with `zlib.compress` is not much different - ratio is about 12:15. Which kind of makes sense - as long as the hash (or any other substring) appears in the sliding window (by default 32k), it'll just insert a backreference in the output, same as this format is doing manually:

> An encoded match to an earlier string consists of an 8-bit length (3–258 bytes) and a 15-bit distance (1–32,768 bytes) to the beginning of the duplicate. Relative back-references can be made across any number of blocks, as long as the distance appears within the last 32 KiB of uncompressed data decoded (termed the sliding window).

zlib is very fast at this, too, so it's possible this format is actually slower, though I haven't benchmarked.

One benefit to this format is it recovers sharing of references - at least within a single serialized value. But I think (independently) that we want sharing of references across all values. (Like if I separately read two `[Optional Nat]`, I want the `Optional` reference to be shared) And this canonicalization could be done without changing the format at all.

So my feeling right now is this PR isn't necessarily worth it.

## Implementation notes

Encoding monad is now `MonadState EncodeState m, MonadPut m`, where `EncodeState` is a `Map Reference Int`. Decoding state is analogous (an `IntMap Reference`). The state is built up incrementally when encoding / decoding, but other than that there's no changes to the format.

I created serialization modules `*V3` and `*V4` - `V3` is the old format, `V4` is this "small" format.

@dolio I wasn't sure what `Interface.hs` should be using - is this just used for the `compile` command? I left it using `V3`.


dolio:
Yeah, I'm pretty sure that the import for `Interface` is so it can serialize the intermediate code stuff it needs for standalone files.

PR #3834, opened 2023-02-21
Racket implementations of chunked List/Text/Bytes

lexi-lambda:
This PR is a draft implementation of data structures that can be used to implement the `List`, `Text`, and `Bytes` types. The `vector-trie.rkt` module implements the basic underlying structure, and the `chunked-seq.rkt` module generates specialized variants on top. The specialized variants cache the first/last chunk in the root node, so `cons` and `snoc` are amortized constant time.

This PR is only a draft because some operations have not yet been finished, and some of them use a needlessly inefficient implementation. There are some comments that describe the latter.


lexi-lambda:
Closing this since the changes have been merged through other branches.

PR #3835, opened 2023-02-21
Add specific failure typeLink for async exceptions

SystemFw:
## Overview

Adds a `ThreadKilledFailure` that you can use with `tryEval` to act on async exceptions specifically.

## Implementation notes

Nothing of note

I'm treating all constructors of AsyncException uniformly (ThreadKilled, UserInterrupt, StackOverflow, HeapOverflow).

## Test coverage

Tested via the new tests.u machinery. Currently only tested on the interpreter (see below).

## Loose ends

- It can only be tested on the interpreter because the JIT has no typeLinks yet. `thread-killed-typeLinks-tests.u` should be merged back into `concurrency-tests.u` when the JIT can run it.
- Requires a `builtin.merge` to get `ThreadKilledFailure`, probably something to include in the next release of `base`


dolio (approved):
Looks good to me. I don't have a strong opinion about treating the other async exceptions the same. It seems fine, since they're effectively killing the thread in a way that should not be recovered from, just for slightly varying reasons.

pchiusano (approved):

PR #3838, opened 2023-02-23
JIT: use records rather than lists for encoding unison structures

dolio:
- Previously we were using lists, because Chez has built-in support for treating them as tagged records with case analysis.
- This instead introduces several Scheme record types for encoding the various sorts of data structures:
  * data records with a reference, tag and fields
  * sum records with a tag and fields
  * request records with an ability designation, tag and fields
  * pure records with a field (for pure abilty cases)
- Macros are defined to allow pattern matching on these records
- Various definitions have been tweaked to emit the right sort of record.
- Corresponding unison compiler code will emit wrappers around these new record types.
- The data record, for example, will be useful when eventually doing decompiling, since we need reference data to figure out what a data value actually is.

Corresponding unison code will be pushed to internal.


pchiusano (approved):
Cool. Did you (manually) run the `jit-tests.sh` script to make sure things still work?

Also you gotta sync with trunk, looks like.

SystemFw (approved):

jaredly (approved):
excellent

dolio:
I'm actually having some trouble running the JIT tests. It seems like racket is trying to open the wrong libcrypto shared library file on Linux. If someone else could try running the JIT tests against this branch, that'd be handy. I pushed the unison changes that I think are necessary to share.

Edit: Got it running locally by tweaking some stuff. Fixing some issues.

dolio:
Finally got the test suite running with this. Should be good to merge now.

dolio:
@jaredly I added some stuff to make the crypto stuff work on Linux. It just provides a list of possible versions to look for to `ffi-lib`. Can you make sure that it still works on Mac some time?

pchiusano:
I'm on mac and this worked fine

image

PR #3841, opened 2023-02-23
feat(respect-ucm-host): makes codebase server use the correct host

### LINKED ISSUE #2689: ui command doesn't use the correct host
if you have the UCM_HOST env variable set to force API server to bind to a specific, the `ui` command should use this host. Right now `ui` seems to always use `localhost`

mabasic:
Fixes #2689

## Overview

The environment variable `UCM_HOST` was already parsed, so I have just made it so that it tries to use the value of it or default to "http://127.0.0.1", instead of just defaulting to "http://127.0.0.1".

I've tested it and it works.

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -73,3 +73,4 @@ The format for this list: name, GitHub handle
 * Jesse Looney (@jesselooney)
 * Vlad Posmangiu Luchian (@cstml)
 * Andrii Uvarov (@unorsk)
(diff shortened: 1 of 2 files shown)


pchiusano (approved):
@mabasic thanks! Could you just add yourself to CONTRIBUTORS.markdown?

mabasic:
> @mabasic thanks! Could you just add yourself to CONTRIBUTORS.markdown?

Added.

PR #3846, opened 2023-02-28
Add TCP primops to racket

jaredly:
enough to connect to example.com!
I tried to add a test for "failure", but the `handle` plumbing doesn't seem to be all there -- it somehow swallowed the error.


dolio:
Should the `TODO` test be uncommented? Is the problem just that the thing in the handle is delayed like you mentioned yesterday?

jaredly:
oh lol forgot to push

jaredly:
Ok @dolio I think this is ready to go!

dolio (approved):
Okay, looks good.

PR #3848, opened 2023-03-03
Multiline raw text literals

pchiusano:
Pretty simple. Three or more `"` starts out a text literal, terminated by the same number of consecutive closing quotes. For instance:

```unison
lit1 = """
This is a raw text literal.
It can start with 3 or more ",
and is terminated by the same number of quotes.
Nothing is escaped. \n

The initial newline, if it exists, is ignored.
"""

> lit1
```

This is especially handy when testing out parsing libraries.

The column of the closing quote determines the starting column, and if all lines start with that many spaces, it's stripped from all the lines. This means you can do things like this:

```
foo = 
  """
  uno
  dos
  tres
  """
```

And it works as expected. Thanks to @ceedubs for the suggestion.

## Implementation notes

A small change to the lexer and pretty-printer

The pretty-printer opportunistically uses this syntax whenever the string contains `\n` and consists solely of printable characters, as long as it's at the top level. If the string is nested inside a function call, say, it won't use the syntax, since that could (I believe) potentially cause round-trip issues if the string ends on a weird column.

## Test coverage

Added a simple transcript to exercise.

### DIFF
--- a/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
@@ -227,8 +227,9 @@ pretty0
       --      metaprograms), then it needs to be able to print them (and then the
       --      parser ought to be able to parse them, to maintain symmetry.)
       Boolean' b -> pure . fmt S.BooleanLiteral $ if b then l "true" else l "false"
-      Text' s | Just quotes <- useRaw s -> 
-        pure . fmt S.TextLiteral $ PP.text quotes <> "\n" <> PP.text s <> PP.text quotes
+      Text' s
+        | Just quotes <- useRaw s ->
+            pure . fmt S.TextLiteral $ PP.text quotes <> "\n" <> PP.text s <> PP.text quotes
         where
           -- we only use this syntax if we're not wrapped in something else,
           -- to avoid possible round trip issues if the text ends at an odd column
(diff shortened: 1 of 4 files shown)


pchiusano:
@SystemFw suggested the following heuristic in the pretty-printer: if the text contains at least one newline, and no other escape characters, use this syntax. I think I want to try this but curious if anyone has strong feelings.

ceedubs:
Awesome! I've been wanting something like this 🎉.

One thing that can be awkward with these syntaxes is indentation. For example if this were nested in a function instead of a top-level definition, it would look like this, right?

```
someDefinition =
  doStuff = do
    t = """
This is a raw text literal.
It can start with 3 or more ",
and is terminated by the same number of quotes.
Nothing is escaped. \n

The initial newline, if it exists, is ignored.
"""
    printLine t
  !doStuff
```

The indentation gets a bit confusing (and we should maybe have a test that this doesn't mess anything up), but that's probably fine. The Dhall language [considers the indentation of the closing quotes](https://docs.dhall-lang.org/tutorials/Language-Tour.html#multi-line-text-literals) which provides some flexibility but at the cost of more complex rules.

A really nice thing about Unison is that this is always something that can be changed later without breaking code (except code on Stack Overflow and blog posts and whatnot 😬)!

pchiusano:
@ceedubs I did the same indentation stripping thing, keying off the column of the closing `"""`, thanks for the suggestion!

runarorama (approved):

Issue #3801, opened 2023-02-03
pretty-printer roundtrip error with multi-line lambda


ceedubs:
# Summary

Code that looks like the following compiles but pretty-prints as invalid code:

```
  foreach [1, 2, 3] let x -> let
    y = increment x
    ()
```

# Raw Transcript

<details>
<summary>Click to expand</summary>

````
```ucm
.> builtins.merge
```

```unison
foreach : [a] -> (a -> ()) -> ()
foreach xs f = match xs with
  [] -> ()
  h +: t ->
    f h
    foreach t f

foo =
  foreach [1, 2, 3] let x -> let
    y = increment x
    ()
```

```ucm
.> add foo
.> edit foo
.> load scratch.u
```


````

</details>

# Transcript Output

````
```ucm
.> builtins.merge

  Done.

```
```unison
foreach : [a] -> (a -> ()) -> ()
foreach xs f = match xs with
  [] -> ()
  h +: t ->
    f h
    foreach t f

foo =
  foreach [1, 2, 3] let x -> let
    y = increment x
    ()
```

```ucm

  I found and typechecked these definitions in scratch.u. If you
  do an `add` or `update`, here's how your codebase would
  change:
  
    ⍟ These new definitions are ok to `add`:
    
      foo     : ()
      foreach : [a] -> (a ->{g} ()) ->{g} ()

```
```ucm
.> add foo

  ⍟ I've added these definitions:
  
    foo     : ()
    foreach : [a] -> (a ->{g} ()) ->{g} ()

.> edit foo

  ☝️
  
  I added these definitions to the top of
  /home/cody/code/unison/scratch.u
  
    foo : ()
    foo =
      foreach
        [1, 2, 3] (x ->
          y = Nat.increment x
          ())
  
  You can edit them there, then do `update` to replace the
  definitions currently in this namespace.

.> load scratch.u

  offset=20:
  unexpected =
  expecting ,, :, and, bang, do, false, handle, if, infixApp, let, newline or semicolon, or, quote, termLink, true, tuple, or typeLink
      5 |       y = Nat.increment x
  

```



🛑

The transcript failed due to an error in the stanza above. The error is:


  offset=20:
  unexpected =
  expecting ,, :, and, bang, do, false, handle, if, infixApp, let, newline or semicolon, or, quote, termLink, true, tuple, or typeLink
      5 |       y = Nat.increment x

````

PR #3853, opened 2023-03-04
Fix roundtrip error with block lambdas

### LINKED ISSUE #3110: Roundtrip error for multi-line lambda syntax
The following code, with a multi-line lambda, can be added to the codebase: 

```
tmp =
  List.map (
    i -> let
      use Nat +
      x = i + 1
      y = x + 1
      z = y + 1
      z
  ) [1, 2, 3]

```

Upon saving and editing, the following, non-typechecking code is rendered back: 

```
tmp : [Nat]
tmp =
  use Nat +
  List.map
    (i ->
      x = i + 1
      y = x + 1
      z = y + 1
      z) [1, 2, 3]
```

In order to allow the expression to typecheck, folks are adding newlines or `let` expressions. 

Another example from the slack: 

```
-- this fails 
sum2 : List Nat -> Nat
sum2 = fromList >> mcata
  (f -> cases
    NilF -> 0 
    ConsF x xs -> x + (f xs))

-- but this typechecks
sum : List Nat -> Nat
sum = 
  fromList >> mcata
    (f -> cases
      NilF -> 0
      ConsF x xs -> x + (f xs))
```

runarorama:
This is an attempt to fix #3110 and #3801 by always forcing a `let` (or `do`) if the body of a lambda is a multi-line block.

There might be a better way of addressing this, but I suspect there's a parser issue that prevents a block being detected after the opening `->` of the lambda body. If always adding the `let` for a block lambda body is controversial, then I can look into that instead.

See transcript output diff for difference in output.

### DIFF
--- a/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
@@ -86,12 +86,12 @@ data AmbientContext = AmbientContext
   { -- The operator precedence of the enclosing context (a number from 0 to 11,
     -- or -1 to render without outer parentheses unconditionally).
     -- Function application has precedence 10.
-    precedence :: Int,
-    blockContext :: BlockContext,
-    infixContext :: InfixContext,
-    imports :: Imports,
-    docContext :: DocLiteralContext,
-    elideUnit :: Bool -- `True` if a `()` at the end of a block should be elided
+    precedence :: !Int,
+    blockContext :: !BlockContext,
+    infixContext :: !InfixContext,
+    imports :: !Imports,
+    docContext :: !DocLiteralContext,
+    elideUnit :: !Bool -- `True` if a `()` at the end of a block should be elided
   }
 
 -- Description of the position of this ABT node, when viewed in the
@@ -470,7 +470,7 @@ pretty0
                 paren (p >= 10) <$> do
                   lastArg' <- pretty0 (ac 10 Normal im doc) lastArg
                   booleanOps (fmt S.ControlKeyword "||") xs lastArg'
-              _ -> case (term, nonForcePred) of
+              _other -> case (term, nonForcePred) of
                 OverappliedBinaryAppPred' f a b r
                   | binaryOpsPred f ->
                       -- Special case for overapplied binary op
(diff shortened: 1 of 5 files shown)


parser-typechecker/src/Unison/Syntax/TermPrinter.hs
```diff
@@ -86,12 +86,12 @@ data AmbientContext = AmbientContext
...
-    infixContext :: InfixContext,
-    imports :: Imports,
-    docContext :: DocLiteralContext,
-    elideUnit :: Bool -- `True` if a `()` at the end of a block should be elided
+    precedence :: !Int,
+    blockContext :: !BlockContext,
+    infixContext :: !InfixContext,
+    imports :: !Imports,
+    docContext :: !DocLiteralContext,
+    elideUnit :: !Bool -- `True` if a `()` at the end of a block should be elided
```
runarorama, line 96: I took the opportunity to make this data structure strict, since it will always be used in a strict way anyway.

unison-src/transcripts-round-trip/main.output.md
```diff
@@ -875,7 +875,7 @@ broken tvar =
           (cases
             Some _ ->
               "oh boy isn't this a very very very very very very very long string?"
-            None   -> ""))
+            None -> ""))
```
runarorama: Changes in this file are unrelated. This transcript just hasn't been run for a while.

unison-src/transcripts/fix3265.output.md
```diff
@@ -33,19 +33,20 @@ are three cases that need to be 'fixed up.'
...
+                  match y with
+                    0 -> w + x
+                    n -> 1 + f0 (drop y 1)
+                f0 y =
+                  match y with
+                    0 -> x
+                    n -> 1 + f1 (drop y 1)
+                f2 x = f2 x
+                f3 x y = 1 + y + f2 x
+                g h = h 1 + x
```
runarorama, line 48: This is the difference in output. Note the introduction of `let` at the start of the block.

parser-typechecker/src/Unison/Syntax/TermPrinter.hs
```diff
@@ -484,16 +484,20 @@ pretty0
...
                     pbs <- printCase im doc branches
                     pure . paren (p >= 3) $
                       PP.group (fmt S.ControlKeyword "cases") `PP.hang` pbs
                   LamsNamedPred' vs body -> do
-                    prettyBody <- pretty0 (ac 2 Block im doc) body
+                    prettyBody <- pretty0 (ac 2 Normal im doc) body
+                    let hang = case body of
+                          Delay' (Lets' _ _) -> PP.softHang
+                          Lets' _ _ -> PP.softHang
+                          _ -> PP.hang
```
runarorama, line 517: This allows the `let` keyword to appear on the same line as the arrow.

parser-typechecker/src/Unison/Syntax/TermPrinter.hs
```diff
@@ -484,16 +484,20 @@ pretty0
...
                     pure $ f' `PP.hang` args'
-                _ -> case (term, \v -> nonUnitArgPred v && not (isDelay term)) of
+                _other -> case (term, \v -> nonUnitArgPred v && not (isDelay term)) of
                   (LamsNamedMatch' [] branches, _) -> do
                     pbs <- printCase im doc branches
                     pure . paren (p >= 3) $
                       PP.group (fmt S.ControlKeyword "cases") `PP.hang` pbs
                   LamsNamedPred' vs body -> do
-                    prettyBody <- pretty0 (ac 2 Block im doc) body
+                    prettyBody <- pretty0 (ac 2 Normal im doc) body
```
runarorama, line 513: Changing `Block` to `Normal` forces the `let` to show up in the body of a lambda.

ChrisPenner (approved):
Seems good 👍🏼 

It may be worth thinking through the ideal situation we want here though;
I think as an end user I'm likely to get confused about when I need a `let` vs not, so maybe it'd be best long-term if the pretty printer either ALWAYS used a let, or NEVER used a let (and the parser could handle it) 😄

PR #3863, opened 2023-03-13
Fix #3861

### LINKED ISSUE #3861: UnkownDecl error in record type
This code

```
unique type Foo = {
  i : Instant
}
```

Gives this error:

```
  UnknownDecl:
    data type
    reference = Instant
```

This works fine though:

```
unique type Foo = Foo Instant
```

As does this:

```
unique type Foo = {
  i : Text
}
```

tstat:
## Overview

There is another typechecking codepath that seeded the `TypeLookup` building function with `Term.dependencies` instead of `UnisonFile.dependencies`, so it didn't fetch data decl dependencies.

### DIFF
--- a/parser-typechecker/src/Unison/FileParsers.hs
+++ b/parser-typechecker/src/Unison/FileParsers.hs
@@ -96,7 +96,6 @@ resolveNames ::
     (Term v, TDNRMap v, TL.TypeLookup v Ann)
 resolveNames typeLookupf preexistingNames uf = do
   let tm = UF.typecheckingTerm uf
-      deps = Term.dependencies tm
       possibleDeps =
         [ (Name.toText name, Var.name v, r)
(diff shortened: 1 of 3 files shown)


mitchellwrosen (approved):

parser-typechecker/src/Unison/FileParsers.hs
```diff
@@ -96,17 +96,14 @@ resolveNames ::
     (Term v, TDNRMap v, TL.TypeLookup v Ann)
 resolveNames typeLookupf preexistingNames uf = do
   let tm = UF.typecheckingTerm uf
```
aryairani, line 98: Do we like maybe even moving `tm` into the list comprehension?

aryairani (approved):

pchiusano (approved):

PR #3865, opened 2023-03-13
Add `Md5` builtin to `crypto.HashAlgorithm`

chriskrycho:
## Overview

As discussed on Slack, expose Haskell's implementation the same way as the other existing crypto built-ins, and add it to the test transcripts. This is not a cryptographically secure algorithm, but is widely used in other contexts where it is still useful, e.g. checksums.

## Implementation notes

Adds a built-in `crypto.HashAlgorithm.Md5` using the existing `declareHashAlgorithm` machinery in `parser-typechecker`.

We talked about implementing this in Unison, and I actually started on that ([see here](https://share.unison-lang.org/@chriskrycho/p/code/latest/namespaces/public/md5); some of those pieces may still be worth generalizing in other ways, *especially* wrapping addition for 32-bit natural numbers, and the chunking-and-padding algorithms were fun to implement), but it seems worth having this as a built-in: it’s valuable for it to be correct (and fast!) and having the Haskell implementation covers that better than a Unison-native one.

## Test coverage

Added transcripts! So far this just uses the three base examples from Wikipedia, but that should be sufficient given that the "real" test coverage here comes from the fact that these are just Haskell implementations.

## Loose ends

The one thing that occurs to me is: Should Unison maybe just re-expose the majority of the implementations already present in Haskell?

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -74,3 +74,4 @@ The format for this list: name, GitHub handle
 * Vlad Posmangiu Luchian (@cstml)
 * Andrii Uvarov (@unorsk)
 * Mario Bašić (@mabasic)
+* Chris Krycho (@chriskrycho)
--- a/parser-typechecker/src/Unison/Builtin.hs
+++ b/parser-typechecker/src/Unison/Builtin.hs
@@ -762,7 +762,7 @@ hashBuiltins =
     B "crypto.hmac" $ forall1 "a" (\a -> hashAlgo --> bytes --> a --> bytes),
     B "crypto.hmacBytes" $ hashAlgo --> bytes --> bytes --> bytes
   ]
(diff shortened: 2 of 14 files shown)


pchiusano (approved):
Nice, thanks Chris! You'll just need to ormolu your contributions (CI is failing because of that - all code must be formatted using the same autoformatting). See `development.markdown` for details, but short summary is you can do:

```
ormolu --mode inplace $(find . -name '*.hs')
```

using 0.5.3.0 of ormolu.

chriskrycho:
Updated: I had done `stack install ormolu` and apparently that was my mistake. Got rid of that, replaced with the latest version via `brew`, and updated! 🚀

pchiusano:
Great, thanks again, Chris!

pchiusano:
@jaredly can you add this to the JIT as well? 

@runarorama created https://github.com/unisonweb/base/issues/146 to track adding this to base.

jaredly:
Sure thing

PR #3875, opened 2023-03-21
JIT: Implement TLS primops & add TLS tests

jaredly:
There are likely some edge cases I don't have covered, but this gets us off the ground.

jit-tests and interpreter-tests are passing


scheme-libs/common/unison/primops.ss
```diff
@@ -243,6 +257,8 @@
   (define (unison-POp-SUBN m n) (fx- m n))
   (define (unison-POp-TAKS n s) (list-head s n))
   (define (unison-POp-TAKT n t) (istring-take n t))
+  (define (unison-POp-DBTX x)
+    (format "~a" x))
```
jaredly, line 261: enables `Debug.toText`

scheme-libs/racket/unison/tcp.rkt
```diff
@@ -18,46 +20,54 @@
    socketAccept.impl.v3
    socketSend.impl.v3)))
 
-(define (input socket) (car socket))
-(define (output socket) (car (cdr socket)))
```
jaredly, line 22: switched to using a `struct` to represent the pair of input & output ports

dolio (approved):
I think this looks good.

On a side note, `prefix-out` and such look handy. I'll have to keep that in mind.

pchiusano (approved):
Nice!

pchiusano:
Important note: 

> You'll also need to install `x509-lib` with `raco pkg install x509-lib`

PR #3886, opened 2023-03-30
Initial implementation of projects

mitchellwrosen:
## Overview

This PR contains an initial implementation of projects consisting of five commands.

### Basic project factoids

- A project is a thing with a name and a collection of branches, each of which has a name and a place in the namespace (the branch contents).

- Example valid project names: `project`, `@username/project`.

- Example valid branch names: `branch`, `@username/branch`.

- Example project + branch specifier: `@unison/base/main`

### Local <-> remote branch mapping

Each branch has up to one associated remote branch per Share host. For example, I might have a local branch

```
project: <local project UUID>
branch: <local branch UUID>
```

related to a remote branch on Share

```
host: share.unison-lang.org
project: <remote project UUID>
branch: <remote branch UUID>
```

The mapping is just a convenience for users; for example, a no-argument `pull` will pull from the associated remote, if there is one. There's not any way to see or manually manipulate the mapping yet. We establish it automatically for you on first `push` or `pull`.

### Parent branches

Locally, each branch has a parent branch (which can be null, if it's the first branch in a project, or was cloned). 

A branch's parent is set to the branch that the user was _on_ when the new branch is created. For example, if I'm on branch `main` and create branch `topic`, then `topic`'s parent is `main`. 

This relationship is used to suggest a default merge target to Share when a branch is pushed for the first time. In this example, supposing `main` was associated with some branch `X` on Share, we'll suggest (in the Share UI) that new branch `Y` (created when pushing `topic`) targets `X`.

### Contributor branches

Contributor branches are a special concept in Unison projects, intended to streamline the DX of contributing a patch to a project.

They are implemented on the Share server as a check on the literal branch name: if a branch is named `@username/something`, then only user `@username` may push to it.

For example, the `@unison/base` project may have the following branches:

```
@unison/base/main
@unison/base/whatever
@unison/base/@arya/patch1
@unison/base/@runar/patch
@unison/base/@paul/topic
```

Here, we consider `main` and `whatever` long-lived branches, because they do not begin with usernames. The other branches are contributor branches that can be pushed to by their respective users _or_ anyone with write-access to the project.

Unknown (to me, at the time of jotting this down): do we allow a project maintainer to _create_ contributor branches on behalf of other users? I should think not. But in every other way, and maybe in this way, too, a contributor branch is just a branch with a particular name and authz rule implemented on Share. A project owner can delete these branches, rename them, etc.

A user may wish to create contributor branches attributed to themselves. For example, I might have a project like:

```
@mitchell/parser/main
@mitchell/parser/@mitchell/topic
```

Here, I've named a branch `@mitchell/topic`, perhaps so it will not appear alongside the long-lived branches of my project in the Share UI, or some other similar reason.

The `push` command facilitates the creation of contributor branches, as we expect most branches to be contributor branches. More on that below.

Hey, it's below. But even more below-below; this seems like the right place to mention a potential namespace collision that each user intending to contribute to a project will have to deal with.

If I, `@mitchell`, intend to contribute some `topic` code to your project, `@coolperson/coolproject`, then I have two reasonable options (assuming I've already cloned your project to a local project with the same name):

1. I can create a local branch called `topic`, leaving me with (say) two local branches:

    ```
    @coolperson/coolproject/main -- yours
    @coolperson/coolproject/topic -- mine
    ```
    
    This is less typing and thinking for me. `push` will still (if it's done its job) create a remote branch `@coolperson/coolproject/@mitchell/topic`. However, I do have to recall that `topic` is my contributor branch, and its name would clash with a long-lived branch called `@coolperson/coolproject/topic`, if one exists now or in the future.
 
2. I can create a local branch called `@mitchell/topic`, to remind myself that this branch is my contributor branch, and to disambiguate it from long-lived branches in `@coolperson/coolproject` that are not prefixed with usernames.

    ```
    @coolperson/coolproject/main -- yours
    @coolperson/coolproject/@mitchell/topic -- mine
    ```

In either case, the remote branches would be named like (2)    

### `project.create` (alias: `create`) (do we even want this alias?)

`project.create` creates a new, named project. It takes one required argument, the project name:

```
.> project.create myparser
```

If a project already exists by that name, it'll print some "project already exists" error. Otherwise, it will create a new project with a branch called `"main"` and cd the user into a special place in the namespace where that project is stored:

```
myparser/main>
```

This special rendering of the prompt is simply triggered on whether a user is exactly at the root of a branch of a project. If they (for example) `cd foo` from there, they'll see the hellish guts:

```
.__projects._6d26ae7b_1e61_467e_b300_dd05bd8a1328.branches._9d7ba3fc_4a74_46de_b32f_3db7ee4a8a27.foo>
```

The second and fourth namespace segments are the local UUIDs of the project and branch, respectively, with a leading underscore and dashes replaced by underscores.

We hope storing project contents in the root namespace does not last very long; it's a convenient shortcut to get some project functionality out the door earlier than otherwise. Rewriting large parts of UCM to not require the root namespace loaded into memory is an ongoing effort.

### `project.clone` (alias: `clone`)

`project.clone` is used to clone a remote branch. It takes one required argument, a project name with an optional branch.
If the branch name is omitted, we infer "main".

```
.> project.clone @unison/base
@unison/base/main>
```

If a local project/branch exists with the given name, we error. Otherwise, we download it from Share, create a local branch that maps to it (for subsequent pulls and pushes) and cd to it.

Not yet implemented: cloning a branch of the current project with the branch shorthand syntax (e.g. `/topic`)

Not yet implemented: an optional second argument to use as the destination project/branch. For example, to clone `@unison/base/main` into a local project `mybase`, branch `mymain`, you might write:

```
.> project.clone @unison/base/main mybase/mymain
```

### `project.switch` (alias: `switch`)

`project.switch` is like a combination of shell `cd`, `git switch`, and `git switch -c`: it can move between exiting projects and branches, and will create a new branch if necessary. It takes one required argument, which can take three forms:

1. Switch to a project, with no branch specified. This is equivalent to specifying the branch called `"main"`. In the future, we plan to track the last branch visited within a project.

```
.> project.switch myparser
myparser/main>
```

2. Switch to a project and branch. This requires the project to exist, but not the branch. If we create a branch, we say so.

```
-- Switch to existing "main" branch
.> project.switch myparser/main
myparser/main>

-- Switch to new "topic" branch
myparser/main> project.switch /topic
"Cool I just created a branch called topic!"
myparser/topic>
```

It's not possible to switch to a _new_ branch from outside that branch's project. This is because we want to track the new branch's parent. We can't default to using `"main"` because a local project need not have a `"main"` branch, or even a default branch (when we implement that).

```
-- Switch to new "topic2" branch from outside the project
@unison/base/main> project.switch myparser/topic2
"I can't do that! Switch to myparser first!"
@unison/base/main>
```

3. Switch to a branch, with no project specified. This requires the user to already be on some project branch. Otherwise, they'll get a "not on branch" error.

```
-- Can't switch branches when you're not on a branch already
.some.loose.code.path> project.switch /topic
"Not on a branch!"

myparser/main> project.switch /topic
myparser/topic>
```

When creating a new branch, the branch that the user switches _from_ determines the contents of the new branch, as well as its parent. There is currently no way to create a new, detached branch, without any history. (Do we want that? Probably...?)

### `push`

The existing `push` command has been upgraded to accept `project/branch` syntax. As above, we accept `project`, `project/branch`, and `/branch` variants.

Note: there is a syntactic ambiguity: a single-segment relative path (e.g. `foo`) can successfully parse as both a project name and "loose-code" (henceforth referred to as loose-code without quotes of questionable purpose) path. We parse these things as project names.

All four combinations of pushing a project branch or loose-code to a remote project branch or loose-code are supported:

- I can push local loose-code to a remote project branch
- I can push local loose-code to remote loose-code (as before)
- I can push a local project branch to remote loose-code (but why)
- I can push a local project branch to a remote project branch

`push` essentially tries to do something smart and reasonable when given no arguments, but accepts overrides for where to push (the optional first argument), and additionally what to push (the optional second argument).

Here are a few of the smart and reasonable features we've implemented:

- If the user is on a project branch, we push to its associated remote branch, if it has one.
- If there is no associated remote branch, we find the closest ancestor with an associated remote branch, and create a new branch in that project. This allows me to create a new branch `topic` from an already-pushed branch `main` and just push it, without having to manually specify that I want to push it to the same remote project as `main`.
- The _name_ that we give a new remote branch has the user's username prefixed to it (i.e. it's considered a contributor branch by Share). For example, if I create a local branch `topic` and push it, it'll be named `@mitchell/topic` on the server. If I _wanted_ to create a new long-lived branch, I'd have to explicitly name it with the `push` command's first argument (e.g. `push /topic`).
- Any time we end up pushing a branch somewhere, if that branch didn't already have a remote mapping, we set it to the branch we pushed to.

### `pull`

The existing `pull` command has similarly been upgraded to accept `project/branch` syntax. Not a lot to note about this implementation.

Not yet implemented: establishing a remote mapping, if pulling into a branch that doesn't have a remote mapping yet. Just a small oversight that can be fixed with one line of code somewhere.


codebase2/codebase-sqlite/sql/005-project-tables.sql
hojberg, line 1: Would it be possible to add `created_at` and `updated_at` fields to all these new tables? I've found that I've never regretted having them.

mitchellwrosen: They certainly aren't a bad idea, however at this time it's not super straightforward to bump the `updated_at` column of a project when its content changes. The `created_at` column could be added without difficulty, but I don't think I really ever ask "what date did I initialize this git repo?", and anyway we should be recording proper timestamps in commits, when we get to that.

hojberg: I think of these fields as very useful for debugging (for us) and in a few cases useful for end users. `updated_at` could be for when the fields of the database row changes rather than content changes (which I agree should be tracked in a different way).

pchiusano (approved):

PR #3889, opened 2023-03-31
Misc. projects-related cleanup and bugfixes

mitchellwrosen:
## Overview

- Fixed parsing of paths like `mitchell.public.whatever` in `push`
- Added output messages for when a remote project/branch is created
- Turned "Uploaded/Downloaded X entities" messages into a proper Output message, so they don't end without a newline
- Added output message for when we push to a remote branch that's already up-to-date
- Made a couple output messages that handle failures (like missing user on Share) more accurate


pchiusano (approved):

pchiusano:
@mitchellwrosen you can merge this whenever, seems very uncontroversial.

PR #3893, opened 2023-04-03
Add `branches` command that lists local branches and their remote mappings

mitchellwrosen:
## Overview

Screenshot from 2023-04-03 17-19-10

## Test coverage

I tested this manually at the repl.

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
@@ -112,6 +112,7 @@ module U.Codebase.Sqlite.Queries
     projectBranchExistsByName,
     loadProjectBranchByName,
     expectProjectBranch,
+    loadAllProjectBranchInfo,
     loadProjectAndBranchNames,
     insertProjectBranch,
     markProjectBranchChild,
@@ -353,6 +354,7 @@ import Unison.Sqlite
 import qualified Unison.Sqlite as Sqlite
 import qualified Unison.Util.Alternative as Alternative
 import qualified Unison.Util.Lens as Lens
+import qualified Unison.Util.Map as Map
 
 -- | A namespace rendered as a path, no leading '.'
 -- E.g. "base.data"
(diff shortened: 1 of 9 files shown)


hojberg:
This is awesome.
Would it be possible to alias "api.unison-lang.org" to "share.unison-lang.org" or just "Unison Share"?

hojberg:
Also the remote could be information we reserve for a separate command like `view.branch` or something like that. I feel like its information you rarely need and when its different than share, its rarer for branches within a project to have different remotes. Hmmm, perhaps in the case when its different than the `main` branch we have it show the remote in the list?

(sorry this is just a stream of random thoughts)

pchiusano:
This is great!

I'd do: 

```
   Branch  Remote branch URL
1. foo     share.unison-lang.org/@mitchellwrosen/blah/foo
2. bar     share.unison-lang.org/@mitchellwrosen/blah/bar
3. baz     
```

The most common use for this is "hey, what's the remote for this so I can go take a look at what's there or share a link with someone who wants to check out the work".

pchiusano (approved):

mitchellwrosen:
Sounds good; in order to render something other than the codeserver host (api.unison-lang.org) we'll have to hard-code a link in UCM. Is that ok? If so, I can make that change prior to merging. If we'd rather change the API somewhere so that Share provides UCM with links, then I think this could be merged now and improved later.

pchiusano:
@mitchellwrosen Yeah it's fine to hardcode for now. We can always improve it later.

hojberg:
@mitchellwrosen @pchiusano  the full url is going to be:

```
https://share.unison-lang.org/@mitchellwrosen/:projectSlug/branches/:branchRef
```

(`:branchRef` being either something like `main` or a contributor branch with a slash: `@hojberg/someTopicBranch`).

Also, I think it'd be nice to color remote url in a dim grey. This is very secondary information and I worry about information overload.

mitchellwrosen:
:shrug: This?

Screenshot from 2023-04-04 11-16-18

mitchellwrosen:
Note that the bright white in the first screenshot is how we show all other URIs in outputs.

pchiusano:
@mitchellwrosen looks great... doesn't look like you've pushed that change yet though?

hojberg:
@mitchellwrosen 👍

PR #3903, opened 2023-04-06
Anntoate tls & tcp functions to indicate where bytes, strings and lists are accepted or produced

jaredly:


SystemFw (approved):
This will speed me up, thanks

PR #3905, opened 2023-04-10
Chunked seq integration

SystemFw:
- Incorporates Alexis' data structures work
- Implements Unison `[a]` and `Text`via the above data structures
- Integrates Text Pattern api
- Implements a big chunk of Equality and comparison
- Fixes several bugs across the board

*Note*: it depends on merging `internal/text` onto `internal/trunk`, at which point I'll have to change the compiler library path back to `internal/trunk`

There is still some leftover work: various (small) TODOs, integrating Bytes, and fixing the TLS suite, but that can be done in follow-ups

Please review, but do not merge yet as I work out some git management stuff


pchiusano (approved):
Nice! It looks like this is still missing the character class support. Do we already have scheme implementations of that?

I don't think that needs to hold up this PR though.

SystemFw:
> It looks like this is still missing the character class support. Do we already have scheme implementations of that?

ah! we don't, although I need to check what's available in the Racket stdlib

dolio (approved):
Looks good to me.

PR #3912, opened 2023-04-13
add better sql quasi-quoter that allows variable interpolation

mitchellwrosen:
## Overview

This PR adds a SQL quasi-quoter that interpolates Haskell variables.

SQLite already has [a number of syntaxes](https://www.sqlite.org/c3ref/bind_blob.html) for named parameters. This quasi-quoter steals the `:colon` variant, which requires a `colon` variable in scope with a `ToField` instance.

We can now write queries like

```haskell
execute
  [sql|
    INSERT INTO project_branch_remote_mapping
      ( local_project_id
      , local_branch_id
      , remote_project_id
      , remote_branch_id
      , remote_host
      )
    VALUES (?, ?, ?, ?, ?)
    ON CONFLICT
      ( local_project_id,
      , local_branch_id,
      , remote_host)
      )
    DO NOTHING
  |]
  (pid, bid, rpid, rbid, host)
```

as

```haskell
execute2
  [sql2|
    INSERT INTO project_branch_remote_mapping
      ( local_project_id
      , local_branch_id
      , remote_project_id
      , remote_branch_id
      , remote_host
      )
    VALUES (:pid, :bid, :rpid, :rbid, :host)
    ON CONFLICT
      ( local_project_id,
      , local_branch_id,
      , remote_host)
      )
    DO NOTHING
  |]
```

Note that there is no need for any `?`s in query strings anymore, with the corresponding tuple of parameters provided as a second argument. This should also reduce our query API surface area by about half: no need for separate underscore-variants for queries that take no parameters.

The parser works by only understanding enough of the SQLite syntax to avoid accidentally interpreting variables inside quoted entities. That is, given a query like

```sql
SELECT "foo"
FROM `bar`
WHERE [baz] = :qux AND bazooka = 'boba'
```

our parser will only "think"

```sql
•••••• "foo"
•••• `bar`
••••• [baz] • :qux ••• ••••••• • 'boba'
```

All four string-like constructs (shown above) are supported, but I only suspect we will ever use `'strings'` and sometimes `"identifiers"`. Backticks and brackets are just a couple more weird ways that you can write an identifier that SQLite supports.

In SQLite, to write a `'` literal in a string, you write `''`; ditto for `"` and backtick, mutatis mutandis. These escape syntaxes are all handled properly.

There's one minor regression to how queries appear in debug logs with `UNISON_DEBUG=sqlite`. Because we use `recover-rtti` to make a best-effort display of query parameters (with the advantage that no type class instance is needed), logged queries look like this now:

```haskell
Query
    { sql = "SELECT time, from_root_causal_id, to_root_causal_id, reason FROM reflog ORDER BY time DESC LIMIT :numEntries"
    , params = "[:= <Fun> ":numEntries" 500]"
    , results = "[]"
    }
```

whereas prior to this PR, `params` would look more like a tuple. I think this is fine; the params are still readable, and these are just debug logs. We could also consider requiring `Show`, or a new `ShowInDebugLog` type class, if we really want something prettier here, but those solutions each have disadvantages as well.

## Test coverage

There's a test for the parser.

## Loose ends

I have not refactored many of the existing queries - just a couple, to show the idea. I figured it would be better to get the new functionality in first, then refactor over time, and eventually delete the unused old stuff.

That means we have some not-great names in the interim:

- a new type called `Sql2`, which deprecates `Sql`
- a new quasi-quoter called `sql2`, which deprecates `sql`
- new query functions `execute2`, `queryListRow2`, and `queryListCol2`, with all of the others to be filled out eventually

Additionally, this quasi-quoter does not yet feature query splicing. That would be useful for at least one query we've written, which is currently stitching together query fragments manually with `<>`.

### DIFF
--- a/lib/unison-sqlite/src/Unison/Sqlite/Sql2.hs
+++ b/lib/unison-sqlite/src/Unison/Sqlite/Sql2.hs
@@ -3,6 +3,9 @@
 module Unison.Sqlite.Sql2
   ( Sql2 (..),
     sql2,
+
+    -- * Exported for testing
+    internalParseSql,
   )
 where
 
@@ -44,7 +47,7 @@ data Sql2 = Sql2
 --
 -- @
 -- Sql2
---   { query = "SELECT foo\\nFROM bar\\nWHERE baz = :qux"
+--   { query = "SELECT foo FROM bar WHERE baz = :qux"
 --   , params = [":qux" := qux]
 --   }
 -- @
@@ -55,9 +58,9 @@ sql2 = TH.QuasiQuoter sql2QQ undefined undefined undefined
 
 sql2QQ :: String -> TH.Q TH.Exp
 sql2QQ input =
-  case runP (parser <* Megaparsec.eof) (Text.strip (Text.pack input)) of
-    Left err -> fail (Megaparsec.errorBundlePretty err)
-    Right ((), S {sql, variables}) -> do
+  case internalParseSql (Text.pack input) of
+    Left err -> fail err
+    Right (query, params0) -> do
       let params :: [TH.Q TH.Exp]
           params =
             map
@@ -66,11 +69,19 @@ sql2QQ input =
                     Nothing -> fail ("Not in scope: " ++ Text.unpack var)
                     Just name -> [|(Text.cons ':' var) Sqlite.Simple.:= $(TH.varE name)|]
               )
-              (variables [])
-      let query = Text.Builder.run sql
+              params0
       [|Sql2 query $(TH.listE params)|]
 
--- Parser state: the SQL parsed so far, and a difference list of variable names.
+-- | Parse a SQL string, and return the prettefied SQL string along with the named parameters it contains.
+--
+-- Exported only for testing.
+internalParseSql :: Text -> Either String (Text, [Text])
+internalParseSql input =
+  case runP (parser <* Megaparsec.eof) (Text.strip input) of
+    Left err -> Left (Megaparsec.errorBundlePretty err)
+    Right ((), S {sql, params}) -> Right (Text.Builder.run sql, params [])
+
+-- Parser state: the SQL parsed so far, and a difference list of parameter names.
 --
 -- For example, if we were partway through parsing the query
 --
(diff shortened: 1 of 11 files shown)


pchiusano (approved):
Looks great!

lib/unison-sqlite/src/Unison/Sqlite/Sql2.hs
```diff
@@ -0,0 +1,263 @@
...
+--     { sql = "SELECT foo FROM bar WHERE baz = :bonk AND "
+--     , params = ["bonk"]
+--     }
+--
+-- Why keep the SQL parsed so far: so we can make the query slightly prettier by replacing all runs of 1+ characters of
+-- whitespace with a single space. This lets us write vertically aligned SQL queries at arbitrary indentations in
+-- Haskell quasi-quoters, but not have to look at a bunch of "\n        " in debug logs and such.
+data S = S
+  { sql :: !Text.Builder,
+    params :: [Text] -> [Text]
```
tstat, line 105: Is there a reason to use a difference list if we are using named parameters?

lib/unison-sqlite/src/Unison/Sqlite/Sql2.hs
```diff
@@ -0,0 +1,263 @@
...
+import qualified Text.Builder
+import qualified Text.Builder as Text (Builder)
+import qualified Text.Megaparsec as Megaparsec
+import qualified Text.Megaparsec.Char as Megaparsec
+import Unison.Prelude
+
+-- | A SQL query.
+data Sql2 = Sql2
+  { query :: Text,
+    params :: [Sqlite.Simple.NamedParam]
```
tstat, line 29: I like the simplicity of using sqlite's named parameters. 

A couple of drawbacks that come to mind are: 

1. If we want to add splicing someday then a disadvantage compared to accumulating a `[SqlData]` is that you must be aware of names taken by other splices
2. We pay the small runtime cost of looking up the param index although we know it at compile time

tstat (approved):
LGTM

lib/unison-sqlite/src/Unison/Sqlite/Sql2.hs
mitchellwrosen: Oh, (1) is a good point, and that will make debug logs better as well. I'll think I'll address that in a separate PR, though. The use of named params here is just for convenience, but it's not that inconvenient to use `[SqlData]` instead.

PR #3913, opened 2023-04-13
Add support for chunked-bytes operations

SystemFw:
## Overview

Adds support for literals and operations on `Bytes` backed by `chunked-bytes`. It also sneaks in a fix to SPLL and SPLR

## Implementation notes

Bindings to the chunked-bytes data structure

A lot of places use Scheme/Racket libs which expect byte strings. This results in a bit of a conversion soup (as you can see from the diff).
Also, `Bytes.flatten` is not currently implementable, so it's a no-op.

## Test coverage

byte-tests.u

## Loose ends

There is no test case for SPLR, tried a few pattern matches but couldn't trigger it.


pchiusano (approved):
Nice!!

jaredly (approved):
looks good on my end!

dolio (approved):
Yep, looks good to me, too.

PR #3915, opened 2023-04-14
Add pattern match coverage checking for ability handlers

tstat:
## Overview

Adds pattern match coverage checking for ability handlers. 

Although all of the constructors of an ability might be uninhabited, the typechecker requires at least one be specified so that it can
determine which ability to discharge. So, the default pattern match coverage checking behavior of prohibiting covering any
of these (uninhabited) cases is problematic. Instead, the current hack is to require that at least one constructor be given, even if they are all uninhabited.

The messages in this case aren't the best, but I don't think uninhabited abilities will get handlers written for them often.

## Test coverage

See [this transcript](https://github.com/unisonweb/unison/blob/travis/ability-handler-pmcc/unison-src/transcripts/pattern-match-coverage.output.md#ability-handlers)

### DIFF
--- a/parser-typechecker/src/Unison/PatternMatchCoverage.hs
+++ b/parser-typechecker/src/Unison/PatternMatchCoverage.hs
@@ -62,23 +62,26 @@ checkMatch ::
   -- | (redundant locations, inaccessible locations, inhabitants of uncovered refinement type)
   m ([loc], [loc], [Pattern ()])
 checkMatch matchLocation scrutineeType cases = do
+  ppe <- getPrettyPrintEnv
   v0 <- fresh
   grdtree0 <- desugarMatch matchLocation scrutineeType v0 cases
+  doDebug (P.hang (title "desugared:") (prettyGrdTree (prettyPmGrd ppe) (\_ -> "<loc>") grdtree0)) (pure ())
   (uncovered, grdtree1) <- uncoverAnnotate (Set.singleton (NC.markDirty v0 $ NC.declVar v0 scrutineeType id NC.emptyNormalizedConstraints)) grdtree0
+  doDebug
+    ( P.sep
+        "\n"
+        [ P.hang (title "annotated:") (prettyGrdTree (NC.prettyDnf ppe) (NC.prettyDnf ppe . fst) grdtree1),
+          P.hang (title "uncovered:") (NC.prettyDnf ppe uncovered)
+        ]
+    )
+    (pure ())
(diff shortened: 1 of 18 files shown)


ceedubs:
@tstat I know this is still a draft PR so feel free to ignore this if you aren't ready for it, but I think that I found a test-case that can't be written with the current implementation.

Consider this function:

```
recurseIntoTheVoid : '{Abort} Void -> Void
recurseIntoTheVoid thunk = handle !thunk with cases
  { r } -> r
  { abort -> _ } -> recurseIntoTheVoid thunk
```

With the `{ r } -> r` case there you get this error:

```
  This case would be ignored because it's already covered by the preceding case(s):
        3 |   { r } -> r
```

Fair enough, though the word "preceding" is suspect since it's the first case. But if you take it out you get the following error:

```
  Pattern match doesn't cover all possible cases:
        2 | recurseIntoTheVoid thunk = handle !thunk with cases
        3 |   { abort -> _ } -> recurseIntoTheVoid thunk
    
  
  Patterns not matched:
   * { _ }
```

tstat:
@ceedubs I looked into your example with @dolio and the issue in your example is that without the `{ x } -> x` case the `match` infers a more general type `Request {Abort} r -> r`, but with it infers the type `Request {Abort} Void -> Void` where the pattern is redundant. If you wanted to write that infinite loop handler you could give it an explicit type:

```
recurseIntoTheVoid : '{Abort} Void -> Void
recurseIntoTheVoid thunk = 
  impl : Request {Abort} Void -> Void
  impl = cases
    { abort -> _ } -> recurseIntoTheVoid thunk
  handle !thunk with impl
```

Concerning:
> though the word "preceding" is suspect since it's the first case

Yes, this is a misleading message that is already in trunk. All we really know is that the pattern is uninhabited, but we attempted to come up with more user-friendly messages, and this one doesn't make sense for all uninhabited patterns.

pchiusano (approved):
@tstat awesome, can't wait for this to land. Could you add a test that uses a multihandler? For instance a handler of both `{Stream a, Abort}` simultaneously, which accumulates a list and returns the current list on `abort`.

ceedubs:
@tstat this is awesome thanks! And I imagine the infinite Void recursion is rare (but we do have one instance of this in nimbus!), so an explicit annotation is a totally fine workaround.

ChrisPenner (approved):
Only skimmed it but LGTM! I'm presuming the cache from last time still applies here, but maybe just double-check the time to parse IOSource on startup to ensure we don't have a regression there 👍🏼

PR #3917, opened 2023-04-14
Report original source location in define-unison macro

SystemFw:
## Overview

This is the patch sent by Alexis to report source locations. I've tested it, it's not 100% precise everywhere (this was expected) but it does work well.


pchiusano (approved):

dolio (approved):
Looking forward to being able to use this.

PR #3924, opened 2023-04-23
JIT implementations for `Bytes.{gzip,zlib}.{compress,decompress}` and `Bytes.{to,from}Base{16,32,64,64UrlUnpadded}`

jaredly:
zlib will be a bit more annoying, but should be doable.


unison-src/builtin-tests/jit-tests.output.md
```diff
@@ -9,4 +9,6 @@ TODO remove md5 alias when base is released
 ```ucm
 .> run.native tests
 
+  Scheme evaluation failed.
+
```
pchiusano: Can you check in a version that includes output from the tests? Guessing you got this running locally but maybe just didn't refresh this file.

pchiusano (approved):
Nice! Left one comment for you. 

@SystemFw can you take a quick look at this?

SystemFw (approved):

PR #3929, opened 2023-04-24
Relax valid branch names in UCM to allow releases/<semver> and releases/drafts/<semver>

mitchellwrosen:
## Overview

This PR relaxes what UCM considers a valid branch name.

Previously, we had:

```
branch-name
  = simple-branch-name
  | "@" user-name "/" simple-branch-name
```

Now, we have:

```
branch-name
  = simple-branch-name
  | "@" user-name "/" simple-branch-name
  | "releases/" semver
  | "releases/drafts/" semver
```

That is, we now consider strings like `"releases/1.2.3"` and `"releases/drafts/4.5.6"` to be valid branch names.

Previously, on `push`, if the remote branch name was left unspecified and there was no remote branch mapping, we derived the remote branch name as follows:

  - 1) If the local branch name has a user slug (i.e. is a contributor branch), leave it alone.
  - 3) Otherwise, if the local branch is called `"main"` and the logged-in user is pushing to a project whose name begins with that username (e.g. user `@mitchell` pushing a local branch `main` to a project called `@mitchell/stuff`, leave it alone.
  - 4) Otherwise, prefix the local branch name with the logged-in user.

The purpose of this derivation is to facilitate the creation of contributor branches.

Now, on `push`, things are mostly the same, except it doesn't make sense to prefix a username to a `releases/` or `releases/drafts/` branch, so there's an additional check between (i) and (iii):

  - 2) If the local branch name begins with `releases/` or `releases/drafts/`, leave it alone.

## Test coverage

No automated test coverage at the moment, as how exactly one creates such a branch may change. I did test these changes manually, though.

Update: I added some automated tests

## Loose ends

No restrictions were made to any of the UCM commands, so you are (for example) able to run `branch releases/1.2.3` successfully.


pchiusano (approved):

pchiusano:
@mitchellwrosen I approved and seems fine to merge whenever you're ready, but can you add "No restrictions were made to any of the UCM commands, so you are (for example) able to run branch releases/1.2.3 successfully." to your todo list?

mitchellwrosen:
@pchiusano Yup

mitchellwrosen:
I discovered a small bug in the parser and fixed it https://github.com/unisonweb/unison/pull/3929/commits/152b060cd0f530cb036b8a2432fabb8953067dde

There's also now an automated test suite
https://github.com/unisonweb/unison/blob/152b060cd0f530cb036b8a2432fabb8953067dde/unison-core/test/Main.hs

PR #3931, opened 2023-04-25
Move to Racket

SystemFw:
## Overview

Simplifies the scheme backend code to only support Racket, given that the Chez backend is starting to seriously bitrot.

## Implementation notes

I've simplified the directory structure, merged some files together, and removed unused code. Depends on some minor changes I have on a branch of the compiler lib.

I've pushed a `last-working-chez` tag on trunk on the latest commit that fully supports both backends

## Test coverage

Tested via the unison tests and the racket tests

## Loose ends

- We could/should also port all the r6rs code to Racket, I'm debating whether to do that in this PR or as we go along.
- If possible, it would be nice to port all the syntax and continuation related code from `core` to `boot`
- After that, I'd like to merge `data` and `core`


pchiusano (approved):

unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -2323,7 +2322,7 @@ doFetchCompiler username =
         { server = RemoteRepo.DefaultCodeserver,
           repo = ShareUserHandle (Text.pack username),
           path =
-            Path.fromList $ NameSegment <$> ["public", "internal", "trunk"]
+            Path.fromList $ NameSegment <$> ["public", "internal", "racket"]
```
dolio: This shouldn't be merged in, right?

dolio (dismissed):

unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
pchiusano: We just discussed this. It'll be changed back to "trunk" after the Unison repo is updated, just before merging this PR.

PR #3933, opened 2023-04-25
Rename `project.clone` to `branch.clone`; add back different `project.clone`

mitchellwrosen:
## Overview

This PR begins the work of refactoring the existing `project.clone` / `clone` command.

Prior to this PR, this command would accept any of the following:

```
project.clone @user/project
project.clone @user/project/branch
project.clone /branch
```

However, the issue we discovered is a contributor branch of the current project _must_ be preceded by a forward-slash:

```
project.clone /@runar/topic
```

We have a goal of (more or less) making this leading forward slash optional in most cases, as users are likely to forget it.

The proposed solution is to split the `project.clone` command into two commands:

1. `project.clone`, which works exactly like the existing `project.clone`, except it does not accept relative `/branch` syntax. You must either provide a project name (which means "the default branch" (which, today, means "main")), or a project name plus a branch name:

   ```
   project.clone @unison/base
   
   project.clone @arya/parser/topic
   
   -- parse error!
   project.clone /topic
   ```
   
2. `branch.clone`, which works mostly like the existing `project.clone`, except it always creates a branch in the current project, and never interprets a contributor branch as a project.

    ```
    branch.clone topic
    
    branch.clone /topic
    
    branch.clone @runar/topic
    
    branch.clone /@runar/topic
    ```
    
    We also imagine it will be possible to clone some _other_ project branch into the current project, though it will have to be given a name explicitly (which is not currently possible).
    
    ```
    branch.clone @user/some-other-project/some-cool-branch newname
    ```

This PR implements (1) above, and renames the old `project.clone` to `branch.clone` (with the existing alias `clone`).

## Test coverage

Update: added an automated test for the parser


pchiusano (approved):
Great!

PR #3934, opened 2023-04-26
Add `release.draft` command

mitchellwrosen:
- [x] Depends on #3933 

## Overview

This PR adds a `release.draft` command takes one required `x.y.z` argument, and creates a branch called `releases/drafts/x.y.z`:

image

It also makes `branch releases/x.y.z` invalid, with the following error message:

image

---

**Out-of-date info**

`branch releases/drafts/x.y.z` doesn't work either, but there's no error message:

image

This is because while tweaking the error messages, @aryairani and I realized the branch command could either:

- Fail with some error message
- Just behave as if you ran `release.draft` (i.e. `branch releases/drafts/1.2.3` = `release.draft 1.2.3`)

Wdyt, @pchiusano?

**End out-of-date info**

---

Resolution: `branch releases/drafts/1.2.3` has no special treatment, i.e. it's equivalent to `release.draft 1.2.3` but gives you a different output message.

## Test coverage

Update: there are transcripts.

## Loose ends

~Just the `branch releases/drafts/x.y.z` issue mentioned above.~


aryairani:
> Resolution: `branch releases/drafts/1.2.3` has no special treatment, i.e. it's equivalent to `release.draft 1.2.3` but gives you a different output message.

@mitchellwrosen What are the different output messages?

mitchellwrosen:
@aryairani The output message you see above for `release.draft`, and the normal one you get from `branch` (something like "tip: merge into main with this command", depending a little bit on the workdir from which you made the command)

unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1992,6 +1992,43 @@ notifyUser dir = \case
   DownloadedEntities n -> pure (P.wrap ("Downloaded" <> P.num n <> "entities."))
   UploadedEntities n -> pure (P.wrap ("Uploaded" <> P.num n <> "entities."))
   NotImplementedYet message -> pure (P.wrap ("Not implemented:" <> P.text message))
+  DraftingRelease branch ver ->
+    pure $
+      P.wrap ("😎 Great! I've created a draft release for you at " <> prettySlashProjectBranchName branch)
+        <> "."
+        <> P.newline
+        <> P.newline
+        <> P.wrap ("While preparing we suggest you create a releaseNotes Doc. It'll automatically show up on Unison Share when you publish.")
```
pchiusano, line 2001:
```suggestion
        <> P.wrap ("You can create a " <> P.group (P.backticked "ReleaseNotes : Doc") <> "in this branch to give an overview of the release. It'll automatically show up on Unison Share when you publish.")
```

pchiusano (approved):
LGTM! Made a suggestion on wording.

aryairani (approved):

unison-cli/src/Unison/CommandLine/OutputMessages.hs
aryairani: Oops, not sure how I missed this before merging

aryairani: And Github doesn't give me a way to move this forward to a new commit

PR #3942, opened 2023-04-28
fixes #3941

### LINKED ISSUE #3941: enable Tls client auth certificates
@alvaroc1 is looking for support for TlsClientCerts which we don't currently have support for.

looks like this is doing by supplying an "onCertificateRequest" method:

```haskell

    let clientHooks = def { onCertificateRequest = const . return $ Just (cert, key) }
    let tlsSettings = (defaultTlsSettings clientHooks caStore)
                      { clientSupported = def { supportedCiphers = ciphersuite_default } }
```

stew:
provide a way to use tls certs for client auth

## Overview

Provide a way to do TLS client auth so that users can use a certificate to prove their identity to a server.

## Test coverage

nope, maybe later, right not trying to rush this into @alvaroc1 's hands to test


aryairani:
@stew I'm assuming you tested by hand? Or is @alvaroc1 testing first here before we merge?

alvaroc1:
> @stew I'm assuming you tested by hand? Or is @alvaroc1 testing first here before we merge?

I have not tested, no... but looking forward to it being released :D

Issue #3742, opened 2023-01-12
All builtins needed for first release of Unison Cloud JIT


pchiusano:
Ultimately we want to have the JIT support the full language, but for a first milestone, we're trying to get it to compile all the things needed for Unison Cloud so we can run the JIT on all the cloud nodes. The cloud nodes don't allow arbitrary I/O so there's some builtins (particularly STM) that we can get away with not implementing for now.

Once we do a first release, we'll go back and implement STM and probably will do our own threading and async I/O implementations.

Here's primops that have not yet been implemented:

- [x] ABSF
- [x] ACOS
- [x] ACSH
- [x] ADDF
- [x] ADDI
- [x] ASIN
- [x] ASNH
- [x] ATAN
- [x] ATN2
- [x] ATNH
- [x] CACH
- [x] CATB
- [x] CEIL
- [x] CMPU
- [x] COSF
- [x] COSH
- [x] CVLD
- [x] DECN
- [x] DIVF
- [x] DIVI
- [x] DRPT
- [x] EQLF
- [x] EQLI
- [x] EQLU
- [x] EXPF
- [x] FLOR
- [x] FLTB
- [x] FORK
- [x] INCI
- [x] INCN
- [x] INFO
- [x] ITOF
- [x] LEQF
- [x] LEQI
- [x] LEQT
- [x] LKUP
- [x] LOAD
- [x] LOGB
- [x] LOGF
- [x] MAXF
- [x] MINF
- [x] MISS
- [x] MODI
- [x] MULF
- [x] MULI
- [x] NEGI
- [x] NTOF
- [x] PAKB
- [x] POPC
- [x] POWF
- [x] POWI
- [x] POWN
- [x] PRNT
- [x] RNDF
- [x] SDBX
- [x] SGNI
- [x] SINF
- [x] SINH
- [x] SIZB
- [x] SPLL
- [x] SPLR
- [x] SQRT
- [x] SUBF
- [x] SUBI
- [x] TAKB
- [x] TANF
- [x] TANH
- [x] TLTT
- [x] TRNF
- [x] TTOF
- [x] TTOI
- [x] TZRO
- [x] UCNS
- [x] UPKB
- [x] USNC
- [x] VWRS

Note, things like `Nat.+` are wrappers around primop instructions. The wrapper code gets compiled to scheme just like any other unison (intermediate) function, so the primops are what need to be implemented (unless the builtin corresponding to a primop is being replaced completely, but that is a separate mechanism).

[...]

PR #3946, opened 2023-05-01
Disable text.term.tests, which uses TermLinks that are not yet supported by the jit

jaredly:
This broke jit-tests


ceedubs (approved):

ceedubs:
Looks good. Is there a ticket tracking the `Link.Term.toText` builtin for the jit? If so it might be helpful to add a comment there to enable this test again once that builtin is supported.

jaredly:
I don't think it's slated for the initial release https://github.com/unisonweb/unison/issues/3742, not sure if it's tracked elsewhere.

We might also want to be able to add tests that only run on one or the other platform 🤔

PR #3947, opened 2023-05-01
Add jit-tests and interpreter-tests to github actions

jaredly:
See that if we introduce a failing test, the CI fails!

- [failing run](https://github.com/unisonweb/unison/actions/runs/4927779657/jobs/8805262454)
- [successful run](https://github.com/unisonweb/unison/actions/runs/4927933715/jobs/8805651101)

Only running on ubuntu at the moment, setting up for macos & windows will be more tricky.


.github/workflows/ci.yaml
```diff
@@ -41,8 +41,8 @@ jobs:
         os:
           # temporarily disable non-windows builds to speed up iteration time
           - ubuntu-20.04
-          - macOS-11.0
-          - windows-2019
+          # - macOS-11.0
+          # - windows-2019
```
ceedubs: We aren't going to run CI on macOS/windows any more?

Also if you change this can you remove the misleading comment?

jaredly: This is just temporary while I work it out! But yeah I can change the comment

ceedubs (approved):

pchiusano (approved):
Nice!

PR #3950, opened 2023-05-01
Fix branch.clone/project.clone reflog

mitchellwrosen:
## Overview

This PR fixes the reflog entry for `project.clone` and `branch.clone`.

Previously, the entry would just say `"project.clone"`. Now, it distinguishes between `"project.clone"` and `"branch.clone"` (since we have the two separate commands now, although `branch.clone` still needs to be refactored to do what we want), and it also includes the argument.

image

## Test coverage


aryairani (approved):
Good call

PR #3951, opened 2023-05-01
Don't require leading forward slash in `switch /branch`

mitchellwrosen:
## Overview

This PR begins the effort of making the input type of commands that can take a project or branch name more flexible, starting with `switch`.

Prior to this PR, `switch` would accept one of the following:

1. A project name, like `switch foo` (meaning `switch foo/main`)
2. A branch name with a leading forward slash, like `switch /topic`) (meaning switch to `topic` in the current project)
3. A project and branch name, like `switch @unison/base/main`

Now, the parser is more flexible, allowing (2) above without a leading forward slash:

Screenshot from 2023-05-01 13-37-39

It's still possible to switch to a project with the same syntax as before:

image

... unless there's also a branch in the current project with that name (I think this error message needs work):

image

There's also a new output message for when neither a project nor branch with the given name exist:

image

When switching from outside any project, though, `switch x` can only switch to project `x`, so we don't complain about no such branch in that case (as before).

image

## Test coverage

I wrote tests for the new parser and updated the `switch-command.md` transcript.

---

@tstat The `ProjectAndBranchNames` parser felt a little weird (https://github.com/unisonweb/unison/pull/3951/files#diff-a5d5ae8e194d9895050d15165d7d16024e13ccf523472cdc84df1d559f0534b2R294-R326). 

Perhaps you could take a look and suggest any improvements. Otherwise, there's not a lot of interesting stuff going on in this PR.

### DIFF
--- a/unison-cli/src/Unison/CommandLine/OutputMessages.hs
+++ b/unison-cli/src/Unison/CommandLine/OutputMessages.hs
@@ -457,6 +457,35 @@ notifyNumbered = \case
             prettyProjectAndBranchName (ProjectAndBranch remoteProject remoteBranch)
               <> " on "
               <> P.hiBlack (P.shown host)
+  BothLocalProjectAndProjectBranchExist project branch ->
+    ( P.wrap
+        ( "Project"
+            <> prettyProjectName project
+            <> "and branch"
+            <> prettySlashProjectBranchName branch
+            <> "both exist. Did you mean:"
+        )
+        <> P.newline
+        <> P.newline
+        <> P.numberedList
+          [ switch [prettySlashProjectBranchName branch],
+            switch [prettyProjectAndBranchName (ProjectAndBranch project (UnsafeProjectBranchName "main"))]
+          ]
+        <> P.newline
+        <> P.newline
+        <> tip
+          ( "use "
+              <> switch ["1"]
+              <> " or "
+              <> switch ["2"]
+              <> " to pick one of these."
+          ),
+      [ Text.unpack (Text.cons '/' (into @Text branch)),
+        Text.unpack (into @Text (ProjectAndBranch project (UnsafeProjectBranchName "main")))
+      ]
+    )
+    where
+      switch = IP.makeExample IP.projectSwitch
   where
     absPathToBranchId = Right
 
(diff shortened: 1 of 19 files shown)


pchiusano (approved):
Cool. For the message, maybe:

```
Project foo and branch /foo both exist. Did you mean:
  
  1. switch /foo
  2. switch foo/main

  Tip: use `switch 1` or `switch 2` to pick one of these.
```

mitchellwrosen:
@pchiusano sounds good

mitchellwrosen:
image

unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1919,6 +1947,13 @@ notifyUser dir = \case
   LocalProjectBranchDoesntExist projectAndBranch ->
     pure . P.wrap $
       prettyProjectAndBranchName projectAndBranch <> "does not exist."
+  LocalProjectNorProjectBranchExist project branch ->
+    pure . P.wrap $
+      "Neither project"
+        <> prettyProjectName project
+        <> "nor branch"
+        <> prettySlashProjectBranchName branch
+        <> "exist."
```
```suggestion
        <> "exists."
```

unison-src/transcripts/switch-command.output.md
```diff
@@ -56,3 +87,10 @@ Now, the demo.
   no-such-project/main does not exist.
 
 ```
+```ucm
+foo/main> switch no-such-project-or-branch
+
+  Neither project no-such-project-or-branch nor branch
+  /no-such-project-or-branch exist.
```
aryairani:
```suggestion
  /no-such-project-or-branch exists.
```

aryairani:
> @tstat The `ProjectAndBranchNames` parser felt a little weird ([#3951 (files)](https://github.com/unisonweb/unison/pull/3951/files#diff-a5d5ae8e194d9895050d15165d7d16024e13ccf523472cdc84df1d559f0534b2R294-R326)).
> 
> Perhaps you could take a look and suggest any improvements. Otherwise, there's not a lot of interesting stuff going on in this PR.

I marked ready-to-merge, but we can still apply any @tstat suggestions after.

PR #3956, opened 2023-05-02
project+branch parsing tweak: trailing forward slash means project

mitchellwrosen:
## Overview

This PR modifies the parser of the type used in some commands (currently just `switch`) that can take either a project name, branch name, or both.

Prior to this PR:

```
 -- switch to project "foo", or branch "foo"
 -- fails if "foo" is both a project name and branch name
 someproj/main> switch foo

-- switch to branch "foo"
someproj/main> switch /foo
```

Now, there's an additional form:

```
-- switch to project "foo"
someproj/main> switch foo/
```

I also modified the output of an ambiguous `switch` to suggest this syntax:

Screen Shot 2023-05-02 at 4 38 12 PM

## Test coverage

There's a transcript


mitchellwrosen:
I'm not sure I'm a huge fan of that output in the screenshot above 😓. It (and other commands that put the slash as a gentle reminder to use an unambiguous syntax) seems to imply the slash is part of the name, which it's not. Should we reconsider that idea?

Here's an alternative:

Screen Shot 2023-05-02 at 8 12 54 PM

Two independent changes here: 1) I removed a couple forward slashes, and 2) I removed the numbered args shortcut. I think if you are asked to type `switch 1` as a shorthand for `switch /bar`, then `1.` had better just say `/bar`.

mitchellwrosen:
Oh, to recover numbered args we could say something like:

Screen Shot 2023-05-02 at 8 31 55 PM

aryairani:
Oh boy this is kind of tricky huh. I think I like the last option the best, and although I started on a couple of alternative options, none of them turned out better than these.

We can revisit it after implementing menus/prompts

aryairani:
Another option:
```
foo/main> switch bar

  Project bar and branch bar both exist. Did you mean:

  1. foo/bar (/bar for short)
  2. bar/ (project bar, branch unspecified)
  
  Tip: use `switch 1` or `switch 2` to pick one of these.
```

Just an idea; if it's not worth it to provide the project name `foo` there, then don't sweat it.
Your last two screenshots also fine IMO.

I'm going to mark the PR as draft though, in case we are making these changes and/or the command consolidation we talked about. Feel free to promote it again whenever.

hojberg:
How about?

```
foo/main> switch banana

  There's both a Branch (in `foo`) and a Project called `banana`.
  Which one would you like to switch to?

  1. The Branch `/banana` in `foo`
  2. The Project `banana/`
  
  Tip: use `switch 1` or `switch 2` to pick one of these.

```

mitchellwrosen:
@hojberg One thing about that -

> I think if you are asked to type switch 1 as a shorthand for switch /bar, then 1. had better just say /bar.

That is, it's weird to type `switch 1`, where `1` gets substituted in for some other string (in your example `"/banana"`), even than the actual text you see next to `1.` is different.

So just tweaking your example, I think it should be something more like

```
foo/main> switch banana

  There's both a Branch (in `foo`) and a Project called `banana`.
  Which one would you like to switch to?

  1. `/banana` (the Branch `banana` in `foo`)
  2. `banana/` (the Project `banana`)
  
  Tip: use `switch 1` or `switch 2` to pick one of these.
```

That way, it's more clear that `1` stands in for the part of the line before the parenthesized explanation.

Also, the capitalization of Project and Branch - we haven't done that in other outputs. Should we do it everywhere?

mitchellwrosen:
I've combined both of your suggestions, or maybe used them to derive yet another suggestion.

Screen Shot 2023-05-03 at 5 30 52 PM

mitchellwrosen:
(Note: will be even more clear with https://github.com/unisonweb/unison/pull/3960)

aryairani (approved):
lgtm

aryairani:
I like @hojberg's suggestion too; it might be weird but fine to do that, despite the valid concern @mitchellwrosen mentioned. Or maybe we wait to do that until we have modal prompts? (And add option `3. Forget it.`)

mitchellwrosen:
sounds good, I will merge this for now and we can make another ticket to track additional rewordings

pchiusano:
Cool. Also this situation will come up very rarely. (One doesn't typically name branches the same way as projects). So further bikeshedding might be overkill for now. :)

PR #3959, opened 2023-05-03
implement tab-completion for `switch`

mitchellwrosen:
## Overview

This PR implements a rough (but usable) draft of tab-completion for the `switch` command.

`switch` can:

1. switch branches in the current project
2. switch to the default branch of a project
3. switch to a specific project/branch

This PR includes a tab-completer that only suggests entries for (1) and (2), but not (3). That is, if you type

```
> switch someproject/<TAB>
```

you will see no suggestions, though you would ideally see a list of all branches in `someproject`.

Branches and projects are rendered in different colors, and branches have a leading forward slash. The current project and branch (if any) are filtered out of the suggestions.

Screen Shot 2023-05-03 at 1 20 21 PM

It's not necessary to prefix branches with a forward slash before tab-completing. In the project above, typing

```
> switch to<TAB>
```

would tab-complete to

```
> switch /topic
```

(inserting the leading forward-slash and typing the longest common prefix to all matches). A subsequent `<TAB>` would show the options:

Screen Shot 2023-05-03 at 1 22 48 PM

If you do include a leading forward slash, though, you'll get only branch suggestions:

Screen Shot 2023-05-03 at 1 24 44 PM

If you're outside of a project, only projects are shown (obviously), since there are no other branches:

Screen Shot 2023-05-03 at 1 23 44 PM

There's an obnoxious corner-ish case of a name collision between branches and projects, as in the branch/project `"bar"` here:

Screen Shot 2023-05-03 at 1 25 53 PM

In the current implementation, if you tab complete with a prefix shared by both projects and branches, then you'll only get project suggestions:

Screen Shot 2023-05-03 at 1 27 36 PM

You might hope to see branches that begin with "b" in the screenshot above, too, but I just couldn't get that to work well with haskeline. See the big comment in the PR for details.

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -3,6 +3,7 @@
 -}
 module Unison.CommandLine.InputPatterns where
 
+import Control.Lens (preview, (^.))
 import qualified Control.Lens.Cons as Cons
 import Data.Bifunctor (Bifunctor (bimap), first)
 import Data.List (intercalate)
@@ -12,8 +13,14 @@ import Data.Proxy (Proxy (..))
 import qualified Data.Set as Set
 import qualified Data.Text as Text
 import Data.These (These (..))
+import qualified System.Console.ANSI as Ansi
 import System.Console.Haskeline.Completion (Completion (Completion))
+import qualified System.Console.Haskeline.Completion as Haskeline
 import qualified Text.Megaparsec as P
+import U.Codebase.Sqlite.DbId (ProjectBranchId, ProjectId)
+import qualified U.Codebase.Sqlite.Project as Sqlite
+import qualified U.Codebase.Sqlite.Queries as Queries
+import qualified Unison.Cli.ProjectUtils as ProjectUtils
 import qualified Unison.Codebase as Codebase
 import qualified Unison.Codebase.Branch as Branch
 import qualified Unison.Codebase.Branch.Merge as Branch
@@ -45,6 +52,7 @@ import Unison.Name (Name)
 import qualified Unison.NameSegment as NameSegment
 import Unison.Prelude
 import Unison.Project (ProjectAndBranch (..), ProjectAndBranchNames, ProjectBranchName, ProjectName, Semver)
+import qualified Unison.Sqlite as Sqlite
 import qualified Unison.Syntax.HashQualified as HQ (fromString)
 import qualified Unison.Syntax.Name as Name (fromText, unsafeFromString)
 import qualified Unison.Util.ColorText as CT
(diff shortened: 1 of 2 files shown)


mitchellwrosen:
Update: leading forward slashes on branches is now grayed out a bit

Screen Shot 2023-05-03 at 4 49 02 PM

aryairani:
> This PR includes a tab-completer that only suggests entries for (1) and (2), but not (3).

Are there some bigger challenges for (3) or is it just not here because (1) and (2) can work standalone?

aryairani (approved):

mitchellwrosen:
It's the latter

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2311,7 +2319,7 @@ projectClone =
     { patternName = "project.clone",
       aliases = [],
       visibility = I.Hidden,
-      argTypes = [(Required, projectAndBranchNamesArg)],
+      argTypes = [],
```
aryairani, line 2322: Don't we want an element for these? `noCompletionsArg` at least, or is it a fzf thing to leave it off?

mitchellwrosen: It turns out (I'm 95% sure) these things are only used for tab-completion. I went through and removed some in this PR because I discovered them to be inaccurate. This one, for example, shouldn't take the same `projectAndBranchNamesArg` as `switch` does because we don't want to populate `project.clone` tab completion with local projects and branches.

So, in summary, for commands that don't have any tab-completion yet, I think `argTypes` can be anything. I didn't know about `noCompletionsArg` when I set it to the empty list.

But do you happen to know if something like `[(Optional, noCompletionsArg)]` does anything different than `[]`? As far as I can tell there's no relevance to that `Optional`, `Required`, `ZeroPlus`, etc... all of that is just controlled by the actual parser. I could definitely have missed something, though.

tstat (approved):

unison-cli/src/Unison/CommandLine/InputPatterns.hs
aryairani: I kind of forget but I think there was (at some point) an element of counting args or knowing how many args a command took, but I could believe it's now vestigial. It may always have been unused.

aryairani:
Btw the new coloring in the autocomplete looks great

unison-cli/src/Unison/CommandLine/InputPatterns.hs
aryairani: > we don't want to populate project.clone tab completion with local projects and branches.

Not related to merging, but: We want to populate it with remote projects and branches right? (though we can't currently support that right? or can we already do IO when generating completions?)

mitchellwrosen: Yeah, that sounds right. And we can do database and http requests in the completer.

PR #3965, opened 2023-05-08
`ls` pretty print scope bug

tstat:
## Overview
`ls` unconditionally builds a pretty print environment from the namespace root, rather than the current path. This allows names outside of the current path to be referenced when running `ls`. This PR demonstrates this bug with a transcript added in the first commit then fixes the bug in the second commit.

## Test coverage

See [this transcript](https://github.com/unisonweb/unison/pull/3965/files#diff-12ff8046106780fa92dd9bd5021ede230b565a911109d2f3631ddc807b71a32aR1)

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
@@ -8,7 +8,6 @@ where
 
 -- TODO: Don't import backend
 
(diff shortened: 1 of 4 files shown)


aryairani:
> This PR demonstrates this bug with a transcript added in the first commit then fixes the bug in the second commit.

🚀

aryairani (approved):

mitchellwrosen (approved):

PR #3968, opened 2023-05-08
Add a new Pattern.captureAs builtin

stew:
## Overview

Adding a Pattern.captureAs which will act just like capture, but additionally takes an `a` which will be used as a replacement for the capture.

My motivation for adding it is that I want to match on escape sequence, but actually capture the unescaped character instead.

For example:

```
Pattern.captureAs "\n" (Pattern.literal "\\n")
```


dolio (approved):
Makes sense to me.

stew:
reopened as #4452

PR #3981, opened 2023-05-11
add `clone` command, delete `project.clone` and `branch.clone`

mitchellwrosen:
## Overview

This PR replaces the `project.clone` and `branch.clone` commands with one `clone` that has an unambiguous syntax for both "cloning a project" (really, just cloning the default branch of a project), and cloning a branch.

Each argument can be a `project/`, a `/branch`, a `project/branch`, or ambiguous `project-or-branch` (because it's a valid project name and a valid branch name, e.g. `foo` or `@bar/qux`).

---

Specify a project to clone with a trailing forward slash, which means clone the default branch (today that's hard-coded as "main")

Screen Shot 2023-05-16 at 5 00 38 AM

---

Specify a branch in the current project to clone with a leading forward slash (btw making this say "not on Unison Share" instead is somewhere on the TODO list):

Screen Shot 2023-05-16 at 5 02 29 AM

---

Or, specify a branch to clone including its project name.

Screen Shot 2023-05-16 at 5 17 58 AM

---

Sometimes an "ambiguous" project-or-branch argument isn't actually ambiguous. For example, when outside a project, the argument is always treated as a project name, never as a branch name, since there is no current project.

Screen Shot 2023-05-16 at 5 21 47 AM

---

Similarly, even when inside a project, an "ambiguous" project-or-branch that doesn't have a user slug is always treated as a branch, since we know Share only hosts projects whose names have user slugs.

Screen Shot 2023-05-16 at 5 23 23 AM

---

However, when inside a project, arguments that contain a user segment, and no leading nor trailing forward slash are ambiguous. In this PR, we simply try treating the argument as _both_ a project name and a branch name, and if exactly one exists, the clone will succeed.

In this example, the clone succeeds because the project `@mitchellwrosen/topic` does not exist, but the branch @mitchellwrosen/mycoolproject/@mitchellwrosen/topic` does.

Screen Shot 2023-05-16 at 5 25 37 AM

And this example succeeds because the branch `@mitchellwrosen/mycoolproject/@mitchellwrosen/test3` does not exist, but the project `@mitchellwrosen/test3` does.

Screen Shot 2023-05-16 at 5 26 51 AM

If both exist, though, we get an error.

Screen Shot 2023-05-16 at 5 29 11 AM

---

This PR also adds a second argument to `clone` that allows the user to give a different name to the cloned thing. (Its local name is copied over from the remote name, if not specified).

The local name can also be a `project/`, `branch/`, `project/branch` or ambiguous `project-or-branch`. If ambiguous, you might see:

Screen Shot 2023-05-16 at 5 34 17 AM

## Test coverage

I've tested these changes manually.

## Loose ends

1. When on a branch without a remote mapping, and without any ancestors with remote mappings, we fall back on a name-based clone. For example,

    ```
    .> project.create @foo/bar
    @foo/bar/main> clone /topic
    
        "There's no @foo/bar/topic on Unison Share"
    ```
    
    This is arguably weird behavior: maybe we should instead just fail:
    
    ```
    .> project.create @foo/bar
    @foo/bar/main> clone /topic
    
      "No remote mapping found, I don't know what project `/topic` refers to"
    ```
      
    This could additionally make `clone @something/ambiguous` unambiguous in such projects, because `clone /@something/ambiguous` would always fail.


aryairani:
Correct me if I'm wrong:

Summary:
- This PR supports `clone project/branch`, `clone project/`, and `clone /branch`
- `clone <ambiguous>` from outside of a project is treated as a project name, but inside a project prints an error.
- `clone <anything that doesn't start with @>` is treated as a branch name
- `clone <source> <ambiguous target>` is treated as a project when outside of a project, but inside a project prints an error

Some thoughts:
- [x] project/branch color in prompt
- [x] because UCM is doing some smart disambiguation, `clone` output should include the full name of the thing being cloned. e.g.
```
  Cloning @unison/json-core/main.
  
  Downloaded 11901 entities.
```
or
```
  Cloning @unison/json-core/main as bazinga/main.
  
  Downloaded 11901 entities.
```
- [ ] We want to change the disambiguation messages to something like:
```
  Which @foo/bar should I clone?

  1. @unison/json/@foo/bar
  2. @foo/bar/main
  
  Tip: Use `clone 1` or `clone 2` to choose one of these.
```

Re. loose ends:
1. The user requests to clone a branch (the project is left unspecified) but there's no remote mapping to implicitly use. Do we use the local project name instead? Or just give an error? 

I'm not sure, need a longer conversation.

2. The user says `clone topic` when on a branch. We don't know whether the user means `clone /topic topic/main` or `clone /topic /topic`. In this case it would be better to do the second thing. And in a certain way we could consider that the same as `clone X X` just needs to have some work happen first.
```
@mitchellwrosen/mycoolproject/main> clone topic

  Cloning @mitchellwrosen/mycoolproject/topic.
  
  Downloading blah blah
```

mitchellwrosen:
Screen Shot 2023-05-11 at 6 05 28 PM

Is the goal here to get the numbered args tip in? If so, I had left it out because it's not totally clear what to do for multiple argument clones. Could you show an example of that? (e.g. `clone <ambiguous> /newbranch`)

mitchellwrosen:
> The user requests to clone a branch (the project is left unspecified) but there's no remote mapping to implicitly use. Do we use the local project name instead? Or just give an error?

We use the local project name instead, but I'm wondering if we should (and leaning towards) giving an error instead.

mitchellwrosen:
> And in a certain way we could consider that the same as clone X X just needs to have some work happen first.

I agree, I think that resolving

- `clone topic`
- `clone /topic`
- `clone /topic /topic` 
- success

would be better than the current

- `clone topic`
- `clone topic topic`
- `clone /topic topic`
- fail

mitchellwrosen:
ok, loose end 2 is addressed in https://github.com/unisonweb/unison/pull/3981/commits/2e9896cefcf2b060a7a44a002e20ec02a5c8c2d9 and https://github.com/unisonweb/unison/pull/3981/commits/1b3575665886047dcab587395310e5e51d2a88c1. now, a `clone topic` as well as an ambiguous `clone @whats/this` within a project works (the latter, so long as either project `@whats/this` exists, or contributor branch `@whats/this` exists, but not both)

aryairani:
> Screen Shot 2023-05-11 at 6 05 28 PM
> Is the goal here to get the numbered args tip in? If so, I had left it out because it's not totally clear what to do for multiple argument clones. Could you show an example of that? (e.g. `clone <ambiguous> /newbranch`)

Oh good point. The goal was mainly to simplify the language and make it sound less like user error. I guess for now we could do this, though it's definitely pushing the limits of our pseudoprompting.

```
  Which @foo/bar should I clone?

  1. @unison/json/@foo/bar
  2. @foo/bar/main
  
  Tip: Use `clone 1 /newbranch` or `clone 2 /newbranch` to choose one of these.
```

mitchellwrosen:
@aryairani Ok. I'm not sure your latest suggestion is without issues unfortunately, though. Btw, is getting the wording right for this particular error a blocker for this PR, or could it be worked out separately?

A couple thoughts:

1. I had thought the idea was to suggest the unambiguous syntax wherever possible, but your output is missing that. Do we want the "tip" sections as well or did you intentionally omit them?
2. It seems to me that in order to suggest the user runs e.g. `clone 1 <thing>` we ought to validate that `<thing> ` is unambiguous, but it may not be. Overall I imagine it will require some pretty awkward extra code paths to get right in all cases, and we'll have to come up with an output message for when neither argument is unambiguous. I'm not sure it's really worth it, to be honest, especially since we ultimately suggest something rather odd looking.

I wonder if we could do something else, like only do numbered args suggestions for one-argument clones? Or perhaps we can just simplify the wording of the output message in a way that doesn't involve trying to leverage numbered args? (What is wrong with the output message, again?)

aryairani:
> Btw, is getting the wording right for this particular error a blocker for this PR, or could it be worked out separately?

Good question— No, it can be worked out separately, but would like to follow on right away.

> 1\. I had thought the idea was to suggest the unambiguous syntax wherever possible, but your output is missing that. 

Screen Shot 2023-05-11 at 6 05 28 PM

This is unambiguous syntax, isn't it? ^ Or you maybe mean the short (also unambiguous) syntax? 

> Do we want the "tip" sections as well or did you intentionally omit them?

I intentionally omitted them, I was thinking it was too busy, and Simon and I felt that it wasn't the right place to teach about the shorthand. Instead it's meant to clearly and succinctly telling you what you'd be getting with each option.

> 2\. 

Yeah I forgot clone could take two arguments when designing this.

> It seems to me that in order to suggest the user runs e.g. `clone 1 <thing>` we ought to validate that `<thing> ` is unambiguous, but it may not be. 

Remind me of the ways where `<thing>` can be ambiguous? Doesn't it always have to be a branch or project+branch?

I see, it can be a project name, and the branch name is inferred from the source branch 🤯. Do we need or want that, or can we just have it be branch or project+branch?

> I wonder if we could do something else, like only do numbered args suggestions for one-argument clones? Or perhaps we can just simplify the wording of the output message in a way that doesn't involve trying to leverage numbered args? (What is wrong with the output message, again?)

Yeah we can certainly do something different in the two-argument case.

aryairani (approved):
Awesome! We can merge even while we iterate on the output.

PR #3988, opened 2023-05-12
Remove project uuids from prompt and empty namespace output

tstat:
#### Before
before

#### After
after

## Implementation notes

`projectBranchPathPrism` and `getCurrentProjectBranch` now return the trailing path that may follow the branch path.

none

## Test coverage

Manually tested, see screenshots.

## Loose ends

I'm sure there are many other places where we leak project and branch uuids.


aryairani (approved):
Could you update the haddocks for `projectBranchPathPrism` and `getCurrentProjectBranch` and then 👍

PR #3991, opened 2023-05-16
Allow for jit-only and interpreter-only tests in our builtin-tests

jaredly:
There are some things that are only implemented on one side or the other


pchiusano (approved):
LGTM. Can you just resolve the conflicts? Then you can merge.

PR #3998, opened 2023-05-16
Add `delete.project` command

tstat:
## Overview

Adds the `delete.project` command that hard-deletes a project.

## Implementation notes

The handler deletes the project rows from sqlite then removes the project's namespace. If the user is in the project when deleting they are moved the the root namespace.

none

## Test coverage

There is a transcript in the PR.

## Loose ends

1. We want to support soft-deletes for projects and project branches. 
2. Project related stuff like this is not currently compatible with the reflog/undo. 
3. It would be best to delete the project rows and manipulate the namespace atomically, and I don't see a reason we couldn't do that right now. However, our api for modifying namespaces (`updateAt*`, `stepAt*`, `updateRoot`) doesn't give this control (i.e. doesn't accept a database connection or run in `Transaction`).

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/DeleteProject.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/DeleteProject.hs
@@ -1,0 +1,47 @@
+-- | @delete.project@ input handler
+module Unison.Codebase.Editor.HandleInput.DeleteProject
+  ( handleDeleteProject,
+  )
+where
+
+import Control.Lens (view, (^.))
+import Data.Function (on)
+import qualified U.Codebase.Sqlite.Queries as Queries
+import Unison.Cli.Monad (Cli)
+import qualified Unison.Cli.Monad as Cli
+import qualified Unison.Cli.MonadUtils as Cli
+import qualified Unison.Cli.ProjectUtils as ProjectUtils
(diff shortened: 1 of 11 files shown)


unison-cli/src/Unison/Codebase/Editor/HandleInput/DeleteProject.hs
```diff
@@ -0,0 +1,47 @@
...
+import qualified Unison.Codebase.Editor.Output as Output
+import qualified Unison.Codebase.Path as Path
+import Unison.Prelude
+import Unison.Project (ProjectAndBranch (..), ProjectName)
+
+-- | Delete a project branch.
+--
+-- Currently, deleting a branch means deleting its `project_branch` row, then deleting its contents from the namespace.
+-- Its children branches, if any, are reparented to their grandparent, if any. You may delete the only branch in a
+-- project.
```
aryairani: Are these docs accurate, or copy/pasta?

aryairani (approved):
Awesome!

unison-cli/src/Unison/Codebase/Editor/HandleInput/DeleteProject.hs
tstat: Oh, whoops. No, they are not accurate.

PR #4010, opened 2023-05-22
Add jit primops for Bytes.{encode,decode}Nat{16,32,64}

jaredly:


pchiusano (approved):
LGTM. For performance I might unroll the loops or do it with a macro, though we could look into that later once everything's working. Here it may not matter since it's doing some sequence operations as well, but for like the ByteArray equivalents, where there's nothing else happening other than the encoding/decoding, I wouldn't use a loop.

PR #4012, opened 2023-05-22
Improve Debug.toText when using run.compiled

### LINKED ISSUE #3980: improve run.compiled tracing support for term and type links
If you are using `run.compiled` and call `Debug.toText` or `Debug.trace` with a term or type link you'll get something like `Foreign (Wrap ##Link.Type _)` or `Foreign (Wrap ##Link.Term _)`. It would be more helpful if these messages included the hash of the term/type.

There is already a `Link.Term.toText` builtin that does about what we would want for terms. But I don't see the equivalent for types.

ceedubs:
Previously when using `run.compiled`, `Debug.toText` was not attempting to decompile closures, which generally led to pretty terrible output. This changes the prety-printing for `run.compiled` to work similarly to the pretty-printer for `run`  but with empty name mappings for types and terms.

Resolves #3980.

Testing and example output
==========================

I couldn't figure out how to test this change. I don't see any examples of calling `run.compiled` and capturing output, and `Debug.toText` inside a transcript acts as the pretty-printer in `run` as opposed to `run.compiled`. But I did some manual testing with the following `main` method:

```unison
main = do
  Debug.trace "output" (1, "foo", -1, -3.2, 0xsabba, ["one", "two"], [-2.1, +3.4], [0xsabba, 0xsdeadbeef])
```

Previous output
---------------

```
trace: output
DataB2 #2lg4a 3276800 (DataU1 ##Nat 1310720 1) (DataB2 #2lg4a 3276800 (Foreign (Wrap ##Text "foo")) (DataB2 #2lg4a 3276800 (DataU1 ##Int 983040 (-1)) (DataB2 #2lg4a 3276800 (DataU1 ##Float 589824 (-4608983858650965606)) (DataB2 #2lg4a 3276800 (Foreign (Wrap ##Bytes _)) (DataB2 #2lg4a 3276800 (Foreign (Wrap ##Sequence _)) (DataB2 #2lg4a 3276800 (Foreign (Wrap ##Sequence _)) (DataB2 #2lg4a 3276800 (Foreign (Wrap ##Sequence _)) (Enum #00nv2 3342336))))))))
```

New output
----------

```
(1,
"foo",
-1,
-3.2,
0xsabba,
["one", "two"],
[-2.1, 3.4],
[0xsabba, 0xsdeadbeef])
```

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Interface.hs
+++ b/parser-typechecker/src/Unison/Runtime/Interface.hs
@@ -369,12 +369,8 @@ evalInContext ppe ctx activeThreads w = do
       prettyError (BU tr nm c) = either id (bugMsg ppe tr nm) $ decom c
 
       debugText fancy c = case decom c of
-        Right dv -> SimpleTrace . fmt $ pretty ppe dv
+        Right dv -> SimpleTrace . (debugTextFormat fancy) $ pretty ppe dv
         Left _ -> MsgTrace ("Couldn't decompile value") (show c)
-        where
-          fmt
-            | fancy = toANSI 50
-            | otherwise = toPlain 50
 
   result <-
     traverse (const $ readIORef r)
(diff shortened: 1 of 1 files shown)


ceedubs:
This is what the output looks like if there is a constructor without a name:

```
#r3afapqua2#0
  1
  "foo"
  -1
  -3.2
  0xsabba
  ["one", "two"]
  [-2.1, 3.4]
  [0xsabba, 0xsdeadbeef]
```

dolio (approved):
Looks good.

PR #4014, opened 2023-05-23
Dynamic code loading in scheme

dolio:
This PR is the Haskell part of the initial support for dynamic code loading in scheme. Here some some highlights.

- Files are now pregenerated with the necessary primop wrappers, instead of the wrapper code needing to be generated as part of the transitive dependencies of what we want to compile. This is necessary, because there isn't an easy way to get access to the wrappers during scheme runtime.
  * Not all the wrappers are generated at the moment, because not all the primops exist.
  * This also means we aren't forced to use any wrappers; we could just omit them from the list of generated code and implement the corresponding builtin directly. We just need a file for the latter sort of thing.
  * Some builtins are easier as wrappers, though, because the right constructor numbering has to be generated
- There is also a new checked-in scheme module `primops-generated`, which contains primop implementations that are based on code generated _from_ unison, which is in the `boot-generated` module. Right now this is the `CACH` and `LOAD` instructions, but it could expand in the future.
- Some macros have been rewritten to be more informative, and have optional debugging.
- Various primop implementations are included because they were necessary to run the code generation (which is written in unison)
- `describe-value` has been fleshed out a bit
- Tests have been added for loading and running the serialized test cases in the existing unison test suite
  * At the moment, only one is enabled. Two others require some minor builtin tweaks to get running, but I thought I'd get the PR underway first. The fourth test case requires figuring out how to deserialize and load a continuation from the other runtime, which might be a bit more involved.
  * The listing of the serialized test cases is also currently hard-coded, because I decided it wasn't worth spending more time trying to figure out why `directoryContents` wasn't working correctly. 
- There are also some minor other tweaks. E.G. some of the wrapper code in `Unison.Runtime.Builtin` was technically broken, but probably only in a way that showed up when generating code for the Scheme runtime.

There's also a corresponding merge that has to happen to the code on unison share, but I haven't pushed anything there yet (except to my own user's repository). I would guess the test suite won't pass on this PR even once I get it properly merging, because the new tests won't work with the old share code. We might have to coordinate that in a slightly tricky way.


pchiusano (approved):

dolio:
One thing I should note about this PR and its associated stuff on share.

I've made it so that the compiler no longer traces all the way to builtin wrappers. Instead, a couple files are pre-generated with all the wrappers for the term links in a list (in the unison repo). Then the wrapper files are imported in the files that need them. This prevents duplication of all the wrappers. But, it also means that if you implement a primop, you need to add the wrapper to the list, or else generated scheme code won't find the right builtin.

But, this also gives us flexibility to do things like directly implementing builtins in scheme, instead of being forced to use the Haskell calling convention, as mentioned in the original writeup. I don't have a file for those, but we could add one.

PR #4032, opened 2023-05-25
Update nix flake for GCH 9.2.7

### LINKED ISSUE #4026: update nix flake to ghc 9.2.7
#4016 updated to GHC 9.2.7 (along with some other updates), so the nix flake currently doesn't work. I'll give this a try but might need to recruit some help from @tstat.

ceedubs:
Resolves #4026

I was able to remove some overrides that we were using to get ormolu
0.5.0.1 to build with GHC 8.10.7. It's possible that more cruft can be
removed, but I don't understand it well enough to feel confident in
messing with it. We are no longer explicitly asking for ormolu 0.5.0.1
because I couldn't figure out how to do that, but 0.5.0.1 is the
version in the nixpkgs hash that we are currently locked on.

On my linux box both `stack build && stack test` and HLS support work.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1667395993,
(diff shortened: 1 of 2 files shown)


tstat (approved):

ceedubs:
There was an unrelated CI failure, but this change isn't actually exercised by CI, so I'm just going to merge.

PR #4034, opened 2023-05-25
replace executeFile with executeStatements, which doesn't have to be careful about semicolons

mitchellwrosen:
## Overview

This PR replaces our use of `executeFile`, which split strings on `;` and required the strings to carefully omit semicolons in comments, strings, and at the end of the file, with an `executeStatements` wrapper around [`sqlite3_exec()`](https://www.sqlite.org/c3ref/exec.html), which doesn't have this limitation.

## Test coverage

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,13 +24,15 @@ jobs:
       - uses: actions/checkout@v2
         with:
           path: unison
-      - uses: mrkkrp/ormolu-action@v10 # v10 uses ormolu 0.5.3.0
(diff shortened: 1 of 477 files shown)


mitchellwrosen:
This looks ready to go except the jit tests are failing, I'll merge trunk in after #4036 / #4038

aryairani:
Woof one more paper cut down.

aryairani:
Now we can add the `;` on the last statement of the file, which we couldn't support before.  (Is that true that we can?)

aryairani (approved):

PR #4038, opened 2023-05-25
key haskell cache by resolver

aryairani:
* For `~/.stack`, the initial cache will be keyed by: the commit hash, `stack.yaml`, stackage resolver (major+minor, major only), os name; in that order until something matches
* Same for `**/.stack-work`
* For `base.unison`: the commit hash, `base.md` (the PR splits the execution of base.md into a separate step for timing)
* For Racket: os name + manual version number 
* For apt pkgs: managed by https://github.com/awalsh128/cache-apt-pkgs-action

Also adds some output from `stack-clean-old` to CI to let us know how big the stackage caches are getting.

## Interesting decisions
Why bother with the commit hash in the key? It should always miss; but is it useful to know which commit wrote the cache last?

## Loose ends:
~`base.md` still doesn't pin a version of `base`, but this PR adds a note about the importance of doing so :P~ done in #4040

### DIFF
--- a/unison-src/builtin-tests/base.md
+++ b/unison-src/builtin-tests/base.md
@@ -1,6 +1,9 @@
 When this file is modified, CI will create a new codebase and re-run this;
 otherwise it may reuse a previously cached codebase.
 
+Thus, make sure the contents of this file define the contents of the cache
+(e.g. don't pull `latest`.)
+
 ```ucm
 .> pull unison.public.base.releases.v1_1_1 .base
 .> compile.native.fetch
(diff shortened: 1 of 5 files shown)


pchiusano (approved):

unison-src/builtin-tests/base.md
```diff
@@ -1,3 +1,8 @@
+When this file is modified, CI will create a new codebase and re-run this;
+otherwise it may reuse a previously cached codebase.
+
+Thus, make sure the contents of this file define the contents of the cache
```
mitchellwrosen, line 4: The contents of this file seem to violate this warning, no?

aryairani: Good catch. I don't know what to change the file to that works though yet, so I reverted my initial change, left the warning, which is still valid even though the file violates the rule, and plan to address in a later PR.

PR #4043, opened 2023-05-26
Finish the move to a better SQL quasi-quoter and delete the old one

mitchellwrosen:
## Overview

This PR:

- Deletes `executeMany` and rewrites all uses of it as `traverse execute2`.

    There isn't a great way to get an `executeMany`-like API with the new quasi-quoter that interpolates variables. We had a couple options to deal with this:

    1. Leave these queries alone, i.e. with `?`s and parameters passed in a tuple. The main disadvantage here is we don't have one unified way to write queries.
      
    2. Rewrite using `execute` so long as it doesn't tank performance, with the idea to speed up later if necessary with a prepared query cache.
      
    We went with the second. I did measure the raw timings of:

    1. 1,000,000 (`prepare` + `bind` + `step` + `finalize`) (i.e. `traverse execute`)
    2. `prepare` + 1,000,000 (`bind` + `step`) + `finalize` (i.e. `executeMany`)

    and observed the first to be about 6x slower on my machine. However, the difference is only 130ms for 1,000,000 rows inserted - not something we are likely to notice. 

- Introduces a new `VALUES :name` form to our quasi-quoter, which interpolates `name :: FromRow a => List.NonEmpty a` into a query as a list of row literals. For example,

  ```haskell
  let name :: List.NonEmpty (Only Int, Only Int) -- anything with a FromRow instance
      name =
          (Only 1, Only 2) :| [(Only 3, Only 4)]
  in 
    [sql| VALUES :name |]
  ```

  corresponds to the SQL
  
  ```sql
  VALUES (1, 2), (3, 4)
  ```

  This functionality is merely a port of our `Values` newtype, now deleted, that accomplished the same thing.

- Rewrites the remaining queries using the old quasi-quoter with the new one.

- Deletes the old quasi-quoter and its API.

- Renames all the 2 variants (e.g. `execute2` to `execute`, etc)

## Test coverage

There are some new tests for the `VALUES :name` syntax, and I manually messed around at the repl a bit (`push`, `pull`, etc.). Things seem to be working the same. There is a small risk here that I accidentally introduced a typo in a query when porting it over to the new style. Please review carefully accordingly.


codebase2/codebase-sqlite/U/Codebase/Sqlite/NamedRef.hs
```diff
@@ -55,12 +55,15 @@ instance (FromRow ref) => FromRow (NamedRef ref) where
...
+-- This type has a ToRow instance of the form:
 -- [reversedName, namespace, lastNameSegment] <> ref fields...
-namedRefToScopedRow :: (ToRow ref) => NamedRef ref -> [SQLData]
-namedRefToScopedRow (NamedRef {reversedSegments = revSegments, ref}) =
-  toRow $ (SQLText reversedName, SQLText namespace, SQLText lastNameSegment) :. ref
-  where
-    reversedName = (Text.intercalate "." . toList $ revSegments) <> "."
-    namespace = (Text.intercalate "." . reverse . NEL.tail $ revSegments) <> "."
-    lastNameSegment = NEL.head revSegments
+newtype ScopedRow ref
```
mitchellwrosen, line 60: @ChrisPenner Heads up, this helper went away. I replaced `namedRefToScopedRow` with a newtype that emits the same `[SQLData]` in its `toRow` instance, because it's a little more straightforward to work into our QQ (actually -- heh -- I just realized you can interp a `[SQLData]` just fine -- but a newtype to house the instance seems a little cleaner)

aryairani (approved):
Great

aryairani:
We should add a test for anything we later discover to have been ported incorrectly.

PR #4047, opened 2023-05-30
Add `getLine` and `seekHandle` primitives to the jit

jaredly:
I couldn't determine a way to implement the `SeekFromEnd` variant of `seekHandle`, so that one just raises a Failure at the moment.

### DIFF
--- a/scheme-libs/racket/unison/io-handles.rkt
+++ b/scheme-libs/racket/unison/io-handles.rkt
@@ -37,23 +37,16 @@
     getLine.impl.v1
     ))
 
+; Still to implement:
 ;    getBuffering.impl.v3
-
-;    handlePosition.impl.v3 ; hTell
-    ; (port-position port)
-
-    ; set-port-position!
+;    handlePosition.impl.v3
 ;    isSeekable.impl.v3
-    ; port-has-set-port-position!?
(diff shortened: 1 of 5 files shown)


pchiusano (approved):

dolio:
I think you know, but the bytes you've included by hand aren't stable under certain changes, like hashing algorithm. This is normally something that would be generated from unsion. So, here's an idea how we could go about it:

I think we could already generate it. In the unison `internal` repo on share, the `bootSpec` used to generate `boot-generated` includes a list of `SchemeDefn` values to generate. I just use that to dump out constructor tags for certain things right now, but I think byte arrays could also be generated.

However, I think `boot-generated` is too late, because it depends on `primops`, since it also contains code generated from unison, which is used in other primops built on that generated code. So maybe we need and even more primitive generated file that just has some auto-generated scheme values, like tags and `Reference` byte arrays for builtin types.

If we did this, maybe most of the primop/builtin distinction could just disappear, because the point of that (at the moment) is to get wrapper code to be generated that contains the right mapping to unison data types. But we could instead just generate the right underlying information to write the builtins directly. So, instead of generating wrappers on top of (Haskell implementation alike) primops like I am now, generate the underlying data and implement the unison builtins.

I guess we can just merge this now and I'll take care of auto-generating the hard coded parts some time in the future.

dolio (approved):

PR #4054, opened 2023-05-31
fix bug in halting condition of loadRemoteProjectBranchGen

tstat:
## Overview

The previous halting condition omits the final parent. This change includes it.

## Test coverage

manual testing


mitchellwrosen (approved):

aryairani:
Thanks for the quick fix

PR #4055, opened 2023-05-31
Fill out array primops, with tests

dolio:
Adds several missing array primops, and adds tests for them to the jit tests.

Requires an associated unison share push, but I think that won't be as much of a factor with the jit tests temporarily out of CI.


dolio:
Guess JIT tests aren't disabled yet/anymore?

pchiusano (approved):
Cool. LGTM. If the tests pass (given the push to Share), then I'd just merge this and then update Share.

PR #4059, opened 2023-06-01
add "IN :param" form to sql quasi-quoter

mitchellwrosen:
## Overview

This PR adds an `IN :param` form to our SQL quasi-quoter.

From the [SQLite docs](https://www.sqlite.org/lang_expr.html#the_in_and_not_in_operators) on the `IN` expression,

> If the right operand of an IN or NOT IN operator is a list of values, each of those values must be scalars

Thus, `:param` must be a `ToField a => [a]`, and the form expands to `IN (?, ?, ?)` (with one `?` per element in `:param`, of course).

## Test coverage

I added a test for the parser, and also manually experimented at the repl:

```
> let foo = [1::Int,2,3] in [sql| IN :foo |]
Sql {query = "IN (?, ?, ?)", params = [SQLInteger 1,SQLInteger 2,SQLInteger 3]}
```

### DIFF
--- a/lib/unison-sqlite/src/Unison/Sqlite/Sql.hs
+++ b/lib/unison-sqlite/src/Unison/Sqlite/Sql.hs
@@ -69,11 +69,12 @@ params__ (Sql _ x) = x
 --
 -- which, of course, will require a @qux@ with a 'Sqlite.Simple.ToField' instance in scope.
 --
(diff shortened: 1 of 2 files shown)


aryairani (approved):

aryairani:
cc @ChrisPenner for when you want it

PR #4064, opened 2023-06-01
Add `branch.rename` command

mitchellwrosen:
## Overview

This PR adds a one-argument `branch.rename` command that renames the current branch.

I also generalized the branch name parser to allow optional leading forward slashes, so `branch.rename foo` and `branch.rename /foo` both work.

You can't rename a branch to `releases/x.y.z`.

---

- Renaming `foo` to `bar` looks like this

Screen Shot 2023-06-02 at 10 48 53 AM

---

- Renaming `foo` to `foo` looks like this

Screen Shot 2023-06-02 at 10 49 55 AM

---

- Renaming `foo` to `releases/1.2.3` looks like this

Screen Shot 2023-06-02 at 10 51 39 AM

## Test coverage

There are tests for the relaxed branch parser, but no tests for branch rename. I tested the command manually.


unison-core/src/Unison/Project.hs
```diff
@@ -104,11 +104,11 @@ instance From ProjectBranchName Text
 
 instance TryFrom Text ProjectBranchName where
   tryFrom =
-    maybeTryFrom (Megaparsec.parseMaybe projectBranchNameParser)
+    maybeTryFrom (Megaparsec.parseMaybe (projectBranchNameParser True))
```
aryairani, line 107: Some Boolean blindness here; wdyt about a new type for AllowLeadingSlash / AllowTrailingSlash

unison-core/src/Unison/Project.hs
```diff
@@ -386,8 +387,8 @@ instance TryFrom Text (ProjectAndBranch ProjectName (Maybe ProjectBranchName)) w
 --   3. project/branch
 projectWithOptionalBranchParser :: Megaparsec.Parsec Void Text (ProjectAndBranch ProjectName (Maybe ProjectBranchName))
 projectWithOptionalBranchParser = do
-  (project, hasTrailingForwardSlash) <- projectNameParser
-  branch <- if hasTrailingForwardSlash then optional projectBranchNameParser else pure Nothing
+  (project, hasTrailingSlash) <- projectNameParser
+  branch <- if hasTrailingSlash then optional (projectBranchNameParser False) else pure Nothing
```
aryairani, line 391: Maybe my suggestion would be weird here if you have to convert between HasTrailingSlash and AllowTrailingSlash

aryairani (approved):
Left some suggestions but approve overall; feel free to update/merge or we can merge as-is on Tuesday or whatever

unison-core/src/Unison/Project.hs
mitchellwrosen: I think it's ok as-is, the boolean arguments are private to this module

PR #4065, opened 2023-06-02
add Bytes.indexOf and Text.indexOf builtins

### LINKED ISSUE #4060: `Bytes.indexOf : Bytes -> Bytes -> Optional Nat`
And similarly `Text.indexOf` builtin.

This should use the efficient Boyer-Moore or KMP-style substring matching, so it doesn't do unnecessary work.

stew:
fixes #4060


pchiusano (approved):
I'll let @dolio take a look, but seems good to me.

@lexi-lambda is there a similarly easy way to implement this for the rope type used in the JIT for Text and Bytes? For the Haskell implementation, @stew is just converting to a lazy text or bytestring and then calling the existing search function.

dolio (approved):
Looks good to me.

I don't think it should be too difficult to implement one of the string search algorithms directly on the chunked sequences. KMP is probably easiest, because it operates on elements in order. But that's not what e.g. `Text` in Haskell is using (I'm not sure what it is; maybe Boyer-Moore, since it mentions `O(m*n)` uncommon cases). I'm not really up on what is considered the best string search algorithm these days.

pchiusano:
Cool, we're going to hold off merging this until @SystemFw can get a barebones implementation in Racket, likely early next week.

pchiusano:
@SystemFw I added these as todos to the [must have](https://github.com/unisonweb/unison/issues/3742#must-have) list here, we'll address in a follow up. @stew was wanting these ASAP for some other work and merged it.

PR #4075, opened 2023-06-06
Implement some pieces of code/value serialization in Scheme

dolio:
This PR provides infrastructure for the `(de)serialize` functions for `Code` and `Value` in Scheme. The actual functions are generated from unison, because the Scheme representations are secretly unison data structures. But the functions require some builtins that were missing.

In addition to the missing builtins, I took the liberty of moving some arithmetic functions to directly implemented builtins rather than involving (vacuous) wrappers. I also fixed a couple miscellaneous bugs I noticed when I was testing this all out.

Note, this doesn't completely implement the capability of sending arbitrary unison values over the wire, because part of that is the step that goes `a -> Value` or `Link.Term -> Code`, which is still not implemented in Scheme (and is considerably more complicated in some cases). This is just the part where you can move between these reified representations and the equivalent `Bytes`.

Some tests are included, piggybacking on the serialized test cases for code loading. The tests now try to re-serialize the `Code` and `Values` to see if they match.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Foreign.hs
+++ b/parser-typechecker/src/Unison/Runtime/Foreign.hs
@@ -120,6 +120,10 @@ charClassCmp :: CharPattern -> CharPattern -> Ordering
 charClassCmp = compare
 {-# NOINLINE charClassCmp #-}
 
+codeEq :: SuperGroup Symbol -> SuperGroup Symbol -> Bool
+codeEq sg1 sg2 = sg1 == sg2
+{-# NOINLINE codeEq #-}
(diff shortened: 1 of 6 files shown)


pchiusano (approved):
Nice!

dolio:
I don't understand the Mac failure. Does the output of that transcript depend on exactly how fast it runs?

pchiusano:
You mean the one that shows the reflog timings? Yeah that transcript needs to be made deterministic. If that's the only failure then you can merge. You can also rerun the job if you want.

PR #4080, opened 2023-06-08
tweak parsing of pull target to be more like merge

### LINKED ISSUE #3953: Pulling to a relative path now gives a confusing error message
```
.tmp> pull unison.public.base.latest base
 
  base/main does not exist.
```

It works fine if I change it to `.tmp> pull unison.public.base.latest .tmp.base`

mitchellwrosen:
## Overview

Fixes #3953 

This PR tweaks the parsing logic of the second argument of `pull` to be like `merge`.

Slack context: https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1686154477422069

Screenshot from 2023-06-08 13-50-22

---

Prior to this PR, the second argument of `pull` could be:

1. A project name `project/`
2. A branch name `/branch`
3. A project and branch name `project/branch`
4. An absoute path `.loose.code`
5. A relative multi-segment path `loose.code`
6. A ambiguous `foo` thing that parses as a valid one-segment relative path, project name, and branch name.

We would treat (6) above as the same as (1).

---

Now, the second argument of `pull` can be:

1. ~~A project name `project/`~~
2. A branch name `/branch`
3. A project and branch name `project/branch`
4. An absoute path `.loose.code`
5. A relative multi-segment path `loose.code`
6. A ambiguous `foo` thing that parses as a valid one-segment relative path, project name, and branch name.

We treat (6) above as the same as (5), because there is no valid syntax for a relative path that disambiguates it from a project branch. Thus, arguments like (2) must have a leading forward slash, which is unfortunate. That's how `merge` currently works, though.

## Test coverage

I've tested this change manually.

### DIFF
--- a/unison-cli/src/Unison/Cli/Pretty.hs
+++ b/unison-cli/src/Unison/Cli/Pretty.hs
@@ -18,12 +18,11 @@ module Unison.Cli.Pretty
     prettyHumanReadableTime,
     prettyLabeledDependencies,
     prettyPath',
-    prettyPathOrProjectAndBranchName,
     prettyProjectAndBranchName,
     prettyProjectBranchName,
     prettyProjectName,
     prettyProjectNameSlash,
-    prettyPullTarget,
+    prettyNamespaceKey,
     prettyReadGitRepo,
     prettyReadRemoteNamespace,
     prettyReadRemoteNamespaceWith,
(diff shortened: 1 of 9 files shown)


aryairani:
> ## Test coverage
> I've tested this change manually.

@mitchellwrosen What do you recommend for regression-testing this sort of thing going forward?

aryairani (approved):

PR #4090, opened 2023-06-12
use breakOn instead of indices which for Text.indexOf

stew:
This reimplements Text.indexOf to use breakOn which will work correctly with text containing multibyte characters.

I added a text to ensure its working correctly


pchiusano (approved):
Nice!

PR #4107, opened 2023-06-17
non-mergeable PR for inspecting the CI build environment

aryairani:
Uses the `mxschmitt/action-tmate@v3` action into CI to get a `tmate` shell on the build VMs. With it, I saw that the Windows VM has about 14 GB in the build directory, which doesn't seem tiny enough to explain the out-of-space errors we were seeing. I'll leave this as a template for doing more debugging later, and we can inspect again the next time we actually get out-of-space.


aryairani:
I guess you can only edit an open PR :P

PR #4111, opened 2023-06-20
Implement setBuffering and getBuffering in racket

jaredly:


scheme-libs/racket/unison/io-handles.rkt
```diff
@@ -83,3 +82,34 @@
...
+
+(define-unison (setBuffering.impl.v3 handle mode)
+    (data-case mode
+        (0 ()
+            (file-stream-buffer-mode handle 'none)
+            (Right none))
+        (1 ()
+            (file-stream-buffer-mode handle 'line)
+            (Right none))
+        (2 ()
```
dolio, line 109: Are these two identical because there isn't the extra block buffering mode that Haskell has?

dolio: Oh, but on that note, doesn't one come with an argument? So it needs to have one, even if it ignores it.

jaredly: oh idk, yeah racket doesn't support the `SizedBlockBuffering`. I wasn't sure if I should raise an exception, or just pretend it's the same as normal block

dolio (approved):

Issue #4083, opened 2023-06-09
updated welcome flow


aryairani:
* [ ] Remove old startup flow (of automatically downloading base)
* [ ] Reword startup message to mention `project.create` (no args)
* [ ] Might be some obsolete command flags that can be ripped out (`--no-base`)
* [x] `project.create` with no args:
  * Picks a random name
  * Does `pull @unison/base/releases/latest lib.base` (don't print out anything except progress indication, then DONE)
  * Gives nice help message given below
* [ ] `help-topic projects` with copy below.

```
I created a new codebase for you at /Users/hojberg/throwaway

  Now starting the Unison Codebase Manager (UCM)...

   _____     _
  |  |  |___|_|___ ___ ___
  |  |  |   | |_ -| . |   |
  |_____|_|_|_|___|___|_|_|

  👋 Welcome to Unison!

  You are running version: release/M5

  📚 Read the official docs at https://www.unison-lang.org/learn/

  Type 'project.create' to get started.

.> project.create

  🎉 I've created the project with the randomly chosen
  name 'illuminated-llama' (use `project.rename <new-name>`
  to change it).
  
  I'll now fetch the latest version of the base Unison library...
  
  ✅ DONE
        
  🎨 Type `ui` to explore this project's code in your browser.
  🌏 Discover libraries at https://share.unison-lang.org
  📖 Use `help-topic projects` to learn more about projects.

  Write your first Unison code with UCM:
  
    1. Open /Users/hojberg/scratch.u.
    2. Write some Unison code and save the file.
    3. In UCM, type `add` to save it to your new project.

  🎉 🥳 Happy coding!
  
illuminated-llama/main> help-topic projects

  A project is a versioned collection of code that can 
  be edited, published, and depended on in other projects.
  Unison projects are analogous to Git repositories. 
    
  Here's some common commands you'll use when working
  with projects:
    
  1. project.create  create a new project
  2. projects        list all your projects
  3. branch          create a new workstream 
  4. branches        list all your branches
  5. merge           merge one branch into another
  6. switch          switch to a project or branch
  7. push            upload your changes to Unison Share
  8. pull            download code(/changes/updates) from Unison Share
  9. clone           download a Unison Share project or branch for contribution

  Tip: use `help project.create` or `help 1` to learn more.
  
  For full documentation, see https://unison-lang.org/learn/projects
```

hojberg:
Depends on https://github.com/unisonweb/unison/issues/4081

PR #4117, opened 2023-06-21
download latest version of base on project.create

mitchellwrosen:
## Overview

This PR makes `project.create` download the latest release of `base`, as described in #4083. Although not explicitly stated, I assumed we'd want to go down a similar code path whether or not the user provided a project name to `project.create` - that is, (try to) pull base into `lib.base`, and display a "happy coding!" blurb.

Screen Shot 2023-06-21 at 1 46 21 PM

I also added a `project.create-empty` command to replace to old `project.create` that didn't download base. This is necessary to use in transcripts, otherwise we'll get an unstable "num entities downloaded" output message.

A couple additional miscellaneous things accomplished along the way:

- Previously, we'd print `Something went wrong with the connection. Try again?` whenever an unexpected HTTP connection error occurred. Now, we parse it a little bit, and print a different message if it seems the user if offline: `You appear to be offline.`
- The pull logic was tweaked so we no longer print `Downloaded X entities...` unless `X` is at least 1. I did this because otherwise every single `project.create` after the first would show `Downloaded 0 entities.`, which was annoying.

And a loose end:

- Currently, in the case that the user already has the latest release of `base` locally, there is no visible distinction in the output messages to a user in the case that we did or did not successfully pull the latest release of `base` into `lib.base`, such as when the user is offline. In the latter case, the user will simply end up in an empty project without a `lib.base` namespace.

## Test coverage

I've tested this change locally, and there are transcripts that capture the output messages.

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Queries.hs
@@ -75,6 +75,7 @@ module U.Codebase.Sqlite.Queries
     -- ** causal table
     saveCausal,
     isCausalHash,
+    causalExistsByHash32,
     expectCausal,
     loadCausalHashIdByCausalHash,
     expectCausalHashIdByCausalHash,
@@ -1212,6 +1213,19 @@ isCausalHash hash =
         SELECT 1
         FROM causal
         WHERE self_hash_id = :hash
+      )
+    |]
+
+-- | Return whether or not a causal exists with the given hash32.
+causalExistsByHash32 :: Hash32 -> Transaction Bool
+causalExistsByHash32 hash =
+  queryOneCol
+    [sql|
+      SELECT EXISTS (
+        SELECT 1
+        FROM causal
+        JOIN hash ON causal.self_hash_id = hash.id
+        WHERE hash.base32 = :hash
       )
     |]
 
--- a/unison-cli/package.yaml
+++ b/unison-cli/package.yaml
@@ -189,6 +189,7 @@ default-extensions:
   - InstanceSigs
   - LambdaCase
   - MultiParamTypeClasses
+  - MultiWayIf
   - NamedFieldPuns
   - NumericUnderscores
   - OverloadedLabels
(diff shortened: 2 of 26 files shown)


aryairani:
We can merge this, but how hard would it be to at least print a message about failing to grab `base` if we fail to grab `base`?

mitchellwrosen:
@aryairani not hard, we just need to come up what to say. any suggestions?

aryairani:
> I tried to add the latest version of `@unison/base` to your project, but failed

or if you have the info: 
> I tried to add `@unison/base/releases/x.y.z` to your project, but failed 

and then one of:
* because I couldn't connect to the Unison Share server.
* \<whatever other or more specific reason if there are multiple error paths\>
* for some reason that I don't know how to explain. 😬 

and later, once we have a command that they could use to retry this:
> You can try again with \<command\>.

mitchellwrosen:
> I tried to add the latest version of @unison/base to your project, but failed

I like that; I like trying to distinguish between different ways we might've failed less, since there are so many.

How about:

> I tried to add the latest version of @unison/base to your project, but couldn't. Perhaps you are offline?

Agree it would be nice to say

> Try running `pull @unison/base/releases/latest lib.base`.

if only you could ;|

aryairani:
How about: 

> I tried to add the latest version of @unison/base to your project, but something went wrong.
> (One way this can happen is if you are offline.)

mitchellwrosen:
Screen Shot 2023-06-22 at 4 45 07 PM

@aryairani I made that change, but before I push it, I have one last objection: it's a little odd to me that we say

"I'll now fetch **the latest version of the base Unison library**..."

but then just below

"I tried to add **the latest version of @unison/base** to your project, but something went wrong."

I feel it would make more sense to use the same wording in these two places, no?

aryairani:
@mitchellwrosen I see. In context, this seems better:

> I'll now fetch the latest version of the base Unison library...
> 
> Sorry, something went wrong while fetching the library, but I've created the project anyway without it.

aryairani (approved):
@mitchellwrosen Go ahead and merge it once satisfied with the message.

mitchellwrosen:
@aryairani ok, will do. I'm going to drop this part

> but I've created the project anyway without it.

because it already says "I've created the project" before fetching base

PR #4118, opened 2023-06-22
Add time zone builtin

### LINKED ISSUE #4114: Add a builtin for getting the local time zone
We can use https://hackage.haskell.org/package/time-1.12.2/docs/src/Data.Time.LocalTime.Internal.TimeZone.html#getTimeZoneCTime

runarorama:
## Overview

This adds a `systemTimeZone` builtin which returns the configured time zone of the operating environment, for the given time. This allows the Base library to include basic functionality like getting the current local time.

Fixes #4114

There is a transcript test that checks that the time zone returned has the correct type.


dolio (approved):
Looks good to me.

PR #4136, opened 2023-06-27
Add `getEcho` and `setEcho` to the jit

jaredly:
I'm only supporting setting & getting echo on stdin; I'm not sure what it would mean to turn on/off echoing for stdout or stderr


jaredly:
Huh so because the CI isn't a `tty`, `setEcho` and `getEcho` aren't working 😬 it doesn't look like unison has a function for checking whether we're in a `tty` 🤔.
Is there a primitive for executing a system cli command? That would be one workaround. (we could gate the tests on a `system()` call to check if we're in a tty)

pchiusano:
@jaredly check out https://share.unison-lang.org/@unison/p/code/latest/namespaces/public/base/latest/;/types/IO/Process

which you might also need to implement :)

pchiusano (approved):
Okay to merge after CI passes... we discussed adding more robust checking for terminal type so this can work inside CI.

Later on, we'll want a version which works on Windows as well, but that's not MVP for now.

PR #4142, opened 2023-06-27
refactor: make Codebase.typeLookupForDependencies add type lookup for builtins itself

mitchellwrosen:
## Overview

Everywhere we use the `TypeLookup` computed by `Codebase.typeLookupForDependencies`, we merge in a `TypeLookup` for builtins before passing it down into the typechecking code.

I think you always want these builtins, so I moved the merging of builtins into the definition of `Codebase.typeLookupForDependencies` and deleted it from all the call sites.

I think it would make sense to push the builtins down further, to the `FileParsers` helper, but that would be a slightly larger refactor, and it would require monomorphizing all of those `v` to `Symbol`. Still, doing that makes sense to me.


aryairani (approved):

aryairani:
I guess this is fine, but seems a little weird to have things that are not dependencies in the `typeLookupForDependencies`. Then again, it's free to include the builtins, so idk.

PR #4149, opened 2023-06-29
Use a project and versioning for in-unison compiler code on share

dolio:
This modifies some of the ucm commands to look for the in-unison compiler code in a project on share. It still downloads the compiler to the same 'loose code' location, so that I didn't have to mess with the execution part of compiling code. But we should now be able to use projects to manage development of the compiler.

`compile.native.fetch` now takes up to two arguments. The first argument is an alternate username (default `unison`). The second argument is an alternate branch name (default for the `unison` user is a specific released version; default for everyone else is `main`). The project name is always `internal`. So, you can work on your own copy of the code with your own fork if you want.

Since the default is to fetch `@unison/internal/releases/X.Y.Z`, there should no longer be a race condition with submitting PRs. Just cut the `X.Y.Z` release on share, update `X.Y.Z` in the github PR, and everything should work out any temporary bad states. We also shouldn't have to worry about whether the `main` branch on share introduces breaking changes relative to the code on github.

I think this should allow @aryairani to do better caching he was wanting. I'm not sure how to signal when the cache should be invalidated, though. I just hard coded the `X.Y.Z` right now, and presumably you should change caches whenever that changes. Let me know if you need me to do anything to facilitate that.

### DIFF
--- a/unison-cli/src/Unison/JitInfo.hs
+++ b/unison-cli/src/Unison/JitInfo.hs
@@ -1,4 +1,3 @@
-
 module Unison.JitInfo (currentRelease) where
 
 currentRelease :: String
--- a/unison-cli/src/Unison/JitInfo.hs
+++ b/unison-cli/src/Unison/JitInfo.hs
@@ -1,0 +1,5 @@
+
+module Unison.JitInfo (currentRelease) where
+
+currentRelease :: String
+currentRelease = "releases/0.0.1"
--- a/scheme-libs/racket/unison/primops.ss
+++ b/scheme-libs/racket/unison/primops.ss
@@ -36,6 +36,7 @@
     builtin-Int.signum
     builtin-Nat.increment
     builtin-Nat.toFloat
+    builtin-Text.indexOf
 
     unison-FOp-internal.dataTag
     unison-FOp-Char.toText
@@ -350,6 +351,8 @@
                  exn:fail:contract?
                  file-stream-buffer-mode
                  with-handlers
+                 match
+                 regexp-match-positions
                  sequence-ref
                  vector-copy!
                  bytes-copy!)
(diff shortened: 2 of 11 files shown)


pchiusano (approved):
This looks good to me. @aryairani would you like to review?

Needs ormolufication. See `development.markdown` for instructions on how to install correct version.

Also I wish we had a bot that would detect a PR in need of ormolu-fication and just add a commit for you (or open a PR against your PR).

dolio:
Okay, fixed all the problems I think.

I screwed up a bit, and the version of base that got pulled in when I created the project is new enough that it has the `indexOf` primop, and I guess that gets used in the jit tests. So I wrote a temporary version based on regexes that works well enough for the test suite. Just don't try to search for any strings with special regex characters until the real version is implemented.

aryairani:
@dolio Could we have it so that you specify the compiler version in some dedicated file? Originally I was suggesting it could go into `base.md`, like:
```
.> compile.native.fetch 0.0.1
```
and CI looks at the hash of this file as part of the cache key.

but we could also just have a dedicated Haskell file `JitVersion.hs` or whatever, with
```
release = "releases/0.0.1"
```
and import that into `InputPatterns.hs` or wherever. Any other parameters that would affect the contents of the startup codebase could go in here too, and then we can hash that file as part of the cache key too to signal an invalidation?

And actually if `@unison/internal/releases/X.Y.Z` includes the copy of base that the compiler needs, then maybe we don't need that pull command in `base.md` anyway.

aryairani:
> Also I wish we had a bot that would detect a PR in need of ormolu-fication and just add a commit for you (or open a PR against your PR).

@pchiusano I wonder how hard that would be to write

dolio:
Yeah, I can do that. Does `Unison.JitInfo` in `unison-cli` sound good?

unison-cli/src/Unison/JitInfo.hs
```diff
@@ -0,0 +1,5 @@
+
```
aryairani:
```suggestion
```

aryairani (approved):
Thanks!

PR #4154, opened 2023-06-30
autoformat PRs with ormolu

aryairani:
## Overview

Auto-formats new PRs with ormolu, and checks the result into the branch before running CI.

## Implementation notes

It uses a fork of https://github.com/haskell-actions/run-ormolu that which allows us to set `mode: inplace`, and then uses https://github.com/stefanzweifel/git-auto-commit-action to commit those changes to the branch before building and testing.

## Testing / Loose ends

I tested it on one of my own PRs, but I haven't tested what happens when an external contributor opens a PR.

I haven't tested with mergify, but it looks like it will be okay, because mergify doesn't care about the ormolu step per se.

https://github.com/haskell-actions/run-ormolu still fails, which is kind of a bummer, but I ignore the failure and still commit the results. Maybe this could be improved, not failing if the only problem is that formatting has occurred. It would probably still be good if it failed on a parse error or something weird.


pchiusano (approved):
Wow nice! Thanks for figuring this out.

PR #4156, opened 2023-07-01
put JitInfo into base.md cache key-2

aryairani:
Complementing #4149, this PR incorporates `JitInfo.hs` into the cache key for `base.unison` along with `base.md`.

# Loose ends

I got rid of `**/` now that we've removed the `unison/` subdir; seeing if that works.


aryairani:
This depended on #4158

PR #4157, opened 2023-07-01
put JitInfo into base.md cache key

aryairani:
Complementing #4149, this PR incorporates `JitInfo.hs` into the cache key for `base.unison` along with `base.md`.

## Loose ends
I think we want this change, but I am not sure why the jit tests are still so much slower than on my local machine, (8-10 mins vs <1 min).


aryairani:
This depended on #4158

aryairani:
I think we want this, but I am not sure why the jit tests are still so much slower than on my local machine, (8-10 mins vs <1 min).

pchiusano (approved):

PR #4159, opened 2023-07-02
Add a secure random generator builtin

runarorama:
## Overview

This adds a cryptographically secure source of randomness to Unison, for applications like generating passwords and keys:

```
IO.randomBytes : Nat ->{IO} Bytes
```

## Implementation notes

This simply calls `MonadRandom.getRandomBytes`.

## Test coverage

There is a transcript that checks that the requested number of bytes are generated.


runarorama:
@dolio We should add this builtin also to the Scheme backend, and maybe you could show me how.

dolio (approved):
Looks good to me.

PR #4163, opened 2023-07-03
Bump max queued connections on builtin Socket to 2048

runarorama:
The `SYS.listenSock` call in `Runtime.Builtin` was setting the max number of queued connections to 2. I'm assuming that's a typo as it seems absurdly low. The docs for `listenSock` suggest using 2048 as a sensible default, so I've set it to that.


pchiusano:
Can you rebase this to not include the randomBytes PR? Otherwise seems good!

aryairani:
Could you also rebase this on `trunk` which has a little more diagnostic output about the Windows failure hopefully.

pchiusano (approved):

PR #4167, opened 2023-07-05
project.create: add more adjectives and animals

hojberg:
## Overview

Add more adjectives and animal names for the `project.create` project name generator.
I tried picking fun and cute ones. Let me know if they sound ok.

## Loose ends

* I wonder if we should move these somewhere different than a Haskell file?
* A cool thing we could do could be to try and do alliteration when possible, so we'd get fun names like "spectacular-shrew".


aryairani:
Nice! Yeah we could move them to a plain text file or something, though having it in a Haskell file means ghc parses it instead of us.

aryairani (approved):

unison-cli/src/Unison/Codebase/Editor/HandleInput/ProjectCreate.hs
```diff
@@ -183,53 +183,151 @@ generateRandomProjectNames = do
...
+          "tiger",
+          "toad",
+          "toucan",
           "turkey",
+          "turtle",
           "urchin",
           "vole",
           "walrus",
+          "wallaby",
+          "wallaroo",
```
aryairani, line 326: These seem a little too similar to each other, but otherwise LGTM.

aryairani (approved):

unison-cli/src/Unison/Codebase/Editor/HandleInput/ProjectCreate.hs
hojberg: I think it's ok since it's rare they that they will appear at the same time

Issue #4150, opened 2023-06-30
argument type error truncates function name too much


ceedubs:
Consider the following error:

```
  The 1st argument to `lib.distributed.Remo...`
  
            has type:  Thread
      but I expected:  Unit
```

Below it showed a block of code that called functions about four different functions with the prefix `Remote...`. Can we increase the length we show before truncation? Also if we are going to truncate we might want to knock path segments off of the front of the namespace instead of removing characters from the function name itself.

PR #4174, opened 2023-07-08
[#4150] truncate argument type error from the front

HatimKhambati26:
## Overview

This PR fixes the inconvenience caused while displaying the argument type error.

**[Before]** The type error would be truncate from the end, causing the user to not know exactly which function threw the error.

**[After]** The function will be truncate from the front. Providing readability for user to fix the error more efficiently.

## Implementation notes

Using `drop` instead of `take` and discarding the excess characters from the beginning. 

Value for the `renderTermMaxLength` could have been increased but didn't see the need of it. 
(let me know if it needs to be updated to an appropriate value)

This PR is raised for the following issue : https://github.com/unisonweb/unison/issues/4150


aryairani:
Great, thanks @HatimKhambati26!

cc @ceedubs

aryairani (approved):

ceedubs:
thank you!

PR #4182, opened 2023-07-11
Allow Debug.toText usage in tests

### LINKED ISSUE #4172: Allow `Debug.toText` in `test` command evaluation

pchiusano:
Previously, the `test` command couldn't use `Debug.toText` due to that being a sandboxed operation. Which was not great, since the main test runner in base uses `Debug.toText` to print failed results. Thus, you could end up in the situation where you update some code, but can't rerun affected tests due to a failed sandbox check!

`Debug.toText` is safe to call here as it's not doing any I/O. Thus, this PR allows `Debug.toText` in the `test` command and adds a regression transcript.

Aside: We could potentially use the helper function I wrote (`HandleInput.evalPureUnison`) for `display` too. Currently `display` also doesn't allow use of `Debug.toText`, but I was more nervous about touching that codepath since it's used for doc rendering on Share. I think it might be fine, but would want an audit from someone else before changing that.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
@@ -3178,16 +3178,19 @@ evalUnisonFile sandbox ppe unisonFile args = do
         Cli.runTransaction (Codebase.putWatch kind hash value')
     pure rs
 
-evalPureUnison :: 
+evalPureUnison ::
   PPE.PrettyPrintEnv ->
-  Bool -> 
+  Bool ->
   Term Symbol Ann ->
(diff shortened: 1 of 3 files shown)


aryairani:
> I was more nervous about touching that codepath since it's used for doc rendering on Share. I think it might be fine, but would want an audit from someone else before changing that.

@pchiusano What are we looking for here?

aryairani:
@pchiusano Allowing `Debug.toText` in tests sounds good to me but I don't know how the sandboxing works or how this PR works. We can walk through it together tomorrow or something, or maybe someone else should evaluate the PR, or both?

aryairani (approved):

PR #4183, opened 2023-07-11
jit: Add Text.indexOf and Bytes.indexOf primops

lexi-lambda:
As it says on the tin: this PR adds implementations of the `Text.indexOf` and `Bytes.indexOf` primops for the JIT. It includes some tests, though I haven’t done any benchmarking.


pchiusano (approved):
Awesome. And I appreciate all the comments and docs on the implementation.

PR #4187, opened 2023-07-11
Add DECN and EXPF to the jit

jaredly:
not sure actually how to generate `EXPF` -- I though `Float.exp` would do it, but turns out that's also a builtin.


dolio (approved):
Looks good to me.

PR #4188, opened 2023-07-11
Add pattern match coverage check switch to synthesize

### LINKED ISSUE #4100: record types with fields of user-defined type don't round-trip
e.g. In this example:
```unison
unique type Foo = { text : Text }
unique type Foo2 = { text2 : Foo }
```

`Foo` will round-trip as a record, but `Foo2` won't.

Here's what happens after adding these definitions:
```
innocent-reindeer/main> view Foo Foo2

  unique type Foo = { text : Text }
  
  unique type Foo2 = Foo2 Foo

innocent-reindeer/main> 
```

Note that `Foo` renders normally, while `Foo2` renders brokenly.

There's already a transcript that should change when this is fixed, where
https://github.com/unisonweb/unison/blob/17fc0db0d876e2e8ce8522a3c4cb56429208c8b3/unison-src/transcripts/records.output.md?plain=1#L73
incorrectly becomes
https://github.com/unisonweb/unison/blob/17fc0db0d876e2e8ce8522a3c4cb56429208c8b3/unison-src/transcripts/records.output.md?plain=1#L81-L82

Possible duplicate of #2905 #2971

tstat:
fixes #4100

## Overview

To determine if a term should be printed as a record or not we [call fieldNames](https://github.com/unisonweb/unison/blob/ed45b8c1b9a9e5ada96bca4402be2b25c2290715/parser-typechecker/src/Unison/Syntax/DeclPrinter.hs#L117-L127). This seems to generate the accessors, infer their types, then do some sort of check that the hashes are all in the ppe. `fieldNames` returns `Nothing` if any of the following are true:

1. there is more than one constructor
2. the generated accessors fail to have their types inferred
3. the ppe doesn't contain the hashes

For records with user defined types we fail at (2) since the provided typeLookup doesn't contain the necessary decls. This is normally populated by `typeLookupForDependencies`, but we don't want to be passing db connections around the pretty printer so we can pull stuff out and ensure the pattern coverage is complete for these machine generated accessors. So, we work around this for now by allowing the pattern match coverage check to be disabled.

Hopefully in the future we will improve record support and won't have to jump through hoops to figure out if something is a record in the pretty printer.

## Test coverage

See the `records` transcript

### DIFF
--- a/parser-typechecker/src/Unison/Syntax/DeclPrinter.hs
+++ b/parser-typechecker/src/Unison/Syntax/DeclPrinter.hs
@@ -75,7 +75,7 @@ prettyGADT env ctorType r name dd =
     constructor (n, (_, _, t)) =
       prettyPattern env ctorType name (ConstructorReference r n)
         <> fmt S.TypeAscriptionColon " :"
-        `P.hang` TypePrinter.prettySyntax env t
+          `P.hang` TypePrinter.prettySyntax env t
     header = prettyEffectHeader name (DD.EffectDeclaration dd) <> fmt S.ControlKeyword " where"
 
 prettyPattern ::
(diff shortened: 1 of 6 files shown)


aryairani (approved):
🎉

Issue #4185, opened 2023-07-11
autocomplete for delete.project


ceedubs:
It should fill in project names, similarly to `switch` (but would only offer project names and not branch names).

PR #4191, opened 2023-07-11
#4185 | Add autocomplete for `delete.project` command

HatimKhambati26:
Autocompletes the project name for `delete.project` command.

[Before] There were no suggestions provided to the user while using the `delete.project`.

Issue: https://github.com/unisonweb/unison/issues/4185


mitchellwrosen (approved):
Great, thanks! Did you happen to test this manually? If so, good to go.

HatimKhambati26:
> 

Yes, I did test it manually on the following test case:

**Autocomplete:**
1. No project name in input -> return all the projects available
2. Partial project name ->  return all the projects, whose prefix matched with the given input
3. Wrong project name (such as branch name) -> return no suggestions

Also, tested the `delete.project` command

PR #4195, opened 2023-07-13
tests: Add simple test cases for Text.indexOf and Bytes.indexOf

lexi-lambda:
This is a follow-up to #4183 which just adds some simple Unison test cases for `Text.indexOf` and `Bytes.indexOf`, since those didn’t already exist.


pchiusano (approved):
Great, thanks!

PR #4197, opened 2023-07-13
Fix various issues with structural find and replace

pchiusano:
This PR fixes all currently known issues with the new structural find and replace feature. See #4198 for details. It also adds docs and regression tests to the existing transcript.

Together these fixes make rewrite rules usable even when they refer to definitions (terms or type declarations) that exist only in the scratch file. There's some subtle logic to avoid accidental variable capture in these situations, while still allowing variable capture when that's the user's intent. The regression tests I added exercise this and the other fixes.

See the comment on the function `UnisonFile.prepareRewrite` if you're curious about the implementation.

### DIFF
--- a/parser-typechecker/src/Unison/Syntax/TermParser.hs
+++ b/parser-typechecker/src/Unison/Syntax/TermParser.hs
@@ -95,7 +95,7 @@ rewriteBlock = do
     rewriteCase = rewriteTermlike "case" DD.rewriteCase
     rewriteType = do
       kw <- quasikeyword "signature"
-      vs <- fromMaybe [] <$> optional (some prefixDefinitionName <* symbolyQuasikeyword ".")
+      vs <- P.try (some prefixDefinitionName <* symbolyQuasikeyword ".") <|> pure []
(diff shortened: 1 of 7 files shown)


parser-typechecker/src/Unison/Syntax/TermParser.hs
```diff
@@ -95,7 +95,7 @@ rewriteBlock = do
     rewriteCase = rewriteTermlike "case" DD.rewriteCase
     rewriteType = do
       kw <- quasikeyword "signature"
-      vs <- fromMaybe [] <$> optional (some prefixDefinitionName <* symbolyQuasikeyword ".")
+      vs <- P.try (some prefixDefinitionName <* symbolyQuasikeyword ".") <|> pure []
```
pchiusano, line 98: This one line fix makes it optional to introduce variables in a `signature` rule. Previously, you'd have to write `signature unused . Optional ==> Optional2` even if the rule didn't need any type variables.

aryairani (approved):

PR #4206, opened 2023-07-17
Allow `single backticks` for inline monospaced text in docs

### LINKED ISSUE #3872: Can Doc literals use backticks for fixed-width spans?
I've been creating `Doc` literals for a year or so now and I still use a backtick (`` ` ``) instead of a double-single quote `''` _every_ single time. And looking through other docs, I often see other people make the same mistake. Can we use the standard markdown convention of backticks for fixed-width/code spans in `Doc` literals?

I believe that there is or was an issue with backticks being used for symbolic identifiers or something, but I couldn't easily find a historical discussion about this.

pchiusano:
This is a very small lexer change to allow a single <code>`</code> to start and end inline monospaced text in a doc, which people are very accustomed to from markdown. Tested in the doc parsing / pretty-printing transcript.

```
d = {{ This is `monospaced` text. }}
```

Previously, single backticks were used for infix function application for an alphanumeric operator, as in:

```haskell
foo `append` bar
```

However we long ago got rid of that syntax since the pretty-printer had no way to know when to use it. 

Thus, the single backtick syntax has been available for a while.

I left the pretty-printer alone in this PR, so it still prefers to use the `''` syntax when pretty-printing. Potentially we could have the pretty-printer use backticks preferentially, that'd be easy to do in a subsequent PR. Note that the `''` syntax is a bit more flexible re: escaping - you can start and end with any number of `''` (2 or more, and the end will require a matching number).

Thus, I see this PR as mostly being about avoiding accidental bad rendering from markdown muscle memory.

## Summarizing various doc syntax

```
''monospaced'' or `monospaced` or '''monospaced text'''
``1 + 1`` -- an inline code example (two backticks, it will be typechecked, and hyperlinked)
```

And then there's fenced code blocks, starting with at least three backticks, ending with the same number of backticks.

### DIFF
--- a/unison-src/transcripts-using-base/doc.output.md
+++ b/unison-src/transcripts-using-base/doc.output.md
@@ -114,7 +114,7 @@ and the rendered output using `display`:
       section elements.
       
       Text can be **bold**, __italicized__, ~~strikethrough~~,
-      or ''monospaced''.
+      or ''monospaced'' (or ''monospaced'').
       
       You can link to Unison terms, types, and external URLs:
       
@@ -141,7 +141,7 @@ and the rendered output using `display`:
     elements.
   
     Text can be bold, *italicized*, ~~strikethrough~~, or
-    `monospaced`.
+    `monospaced` (or `monospaced`).
   
     You can link to Unison terms, types, and external URLs:
   
@@ -581,7 +581,7 @@ Lastly, it's common to build longer documents including subdocuments via `{{ sub
       section elements.
     
       Text can be bold, *italicized*, ~~strikethrough~~, or
-      `monospaced`.
+      `monospaced` (or `monospaced`).
     
       You can link to Unison terms, types, and external URLs:
     
--- a/unison-src/transcripts/api-doc-rendering.output.md
+++ b/unison-src/transcripts/api-doc-rendering.output.md
@@ -41,9 +41,7 @@ Inline code:
 
 `` 1 + 2 ``
 
-`
-"doesn't typecheck" + 1
-`
+`"doesn't typecheck" + 1`
 
 [Link](https://unison-lang.org)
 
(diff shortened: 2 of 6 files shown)


unison-src/transcripts-using-base/doc.output.md
```diff
@@ -114,7 +114,7 @@ and the rendered output using `display`:
       section elements.
       
       Text can be **bold**, __italicized__, ~~strikethrough~~,
-      or ''monospaced''.
+      or ''monospaced'' (or ''monospaced'').
```
aryairani, line 117: This is no big deal but also kind of weird. Do we actually want two syntaxes, and if so is this the one that we want shown after a round trip? Note that the rendered markdown below shows it as `` `monospaced` ``.

unison-src/transcripts/api-doc-rendering.output.md
```diff
@@ -606,28 +604,11 @@ GET /api/getDefinition?names=term
...
-                                                        "tag": "Word"
-                                                    },
-                                                    {
-                                                        "contents": "`",
-                                                        "tag": "Word"
+                                                        "contents": {
+                                                            "contents": "\"doesn't typecheck\" + 1",
+                                                            "tag": "Word"
+                                                        },
+                                                        "tag": "Code"
```
aryairani, line 611: is this change intended or ok? (The array length is smaller, with all words in a single element)

aryairani (approved):

aryairani:
We might look at simplifying the Doc syntax later? Some of the markup still seems kind of weird to me.

unison-src/transcripts/api-doc-rendering.output.md
pchiusano: yeah, it's fine. There was a weird example in the api-docs, it looked like this:

```
`
"doesn't typecheck" + 1
`
```

with a single backtick on each line like that, which would previously parse as the literal backtick.

PR #4209, opened 2023-07-18
feature: fix unique type churn

### LINKED ISSUE #2196: Fix `unique` type update churn
A unique type definition not deleted from a scratch file will always be treated as an update (as opposed to an existing/known definition) after an `add`.  See #2195.

mitchellwrosen:
## Overview

Fixes #2196 

This PR implements a solution to the "unique type churn" problem: every time a `unique type` is saved, it gets a new GUID, which means that you often need to delete unique types from scratch files so you don't up with accidental duplicate copies.

The solution proposed in #2196 was attempted but discarded for the following alternative solution:

- The parser monad has been generalized over a monad `m`, and the parsing environment contains a new lookup function `uniqeTypeGuid :: Name -> m (Maybe Text)`.
- Whenever we parse a `unique type` declaration, we pass the type's name to `uniqueTypeGuid`, and if returns back a `Just`, we use that text as the new type's GUID rather than generate a new one.
- The file parser is run given a callback function that looks up all unique types with the given name in the codebase, and if there are any, returns an arbitrary one's GUID for reuse.

In summary / to give an example, whenever a Unison file containing

```unison
unique type Some.Name = ...
```

is saved, then if there is any

```unison
unique type Some.Name = ...
```

that already exists in the codebase (at the current path in ucm), regardless of its definition, its GUID will be reused in the new type being added.

Thus, if the new and old structures match, so too will their hashes, and if their structures don't match, nor will their hashes (though they will happen to have the same GUID tumbled in). 

Note that this PR contains cleanup and refactoring related to attempting to implement the original proposed solution before pivoting. I still think it's good and helpful, though it makes the PR larger than necessary. Grab me during review if you would like a more real-time walkthrough about what's going on and why.

## Test coverage

I've added a transcript that asserts re-saving unique types results in a "these types already exist in the codebase" output, and I've confirmed the transcript fails on trunk.

## Loose ends

In the current implementation, each unique type in a Unison file being saved will induce a separate lookup into the database for unique types that share a name.

That means we may perform many more branch lookups than necessary.

For example, say we are at the path `foo.bar.baz` and save a Unison file with two unique types named `A` and `B`.

We'll first fetch branch object `foo.bar.baz` from the database (which itself is many lookups, involving a looking up and decoding causals and branches at each segment along the way), then check to see if it has any unique types named `A`. Then, we'll repeat that whole process for `B`.

I elected to hit the database rather than use the in-memory branch structure because we want to move towards a less stateful architecture in general. However, I did not yet implement any sort of branch trie cache. It seems simple enough to add, and users are unlikely to experience significant slowdowns due to the lack of a cache here, since Unison files don't tend to have hundreds or thousands of unique types to parse and type check (which would be slow for other reasons, probably).


pchiusano:
Sweet!!!

aryairani:
> That means we may perform many more branch lookups than necessary.

How many are "necessary"? Is it because it's O(number of unique types in the file) as opposed to something cleverer?

mitchellwrosen:
> How many are "necessary"? Is it because it's O(number of unique types in the file) as opposed to something cleverer?

I think in the context of a parsing a single Unison file, say with absolutely no in-memory state, we ought not to need to _re-fetch_ each namespace data structure more than once.

As-written, if typechecking a file inside `.__projects.GUID.branches.GUID` with N unique types in the file, then we'll end up fully fetching and deserializing the `.__projects.GUID.branches.GUID` namespace twice (which is itself only reachable by fetching and deserializing `.`, then `.__projects`, and so on) N times, rather than once.

aryairani (approved):

aryairani:
@mitchellwrosen Ok, I see. Will it be easy to know in the future whether this is slow? i.e. How will we know when we want to do something to speed it up?  It'll just be "when saving a file with N unique types is slow, but N structural types is fast"?

PR #4213, opened 2023-07-18
Fix name shadowing / capture issue in pretty-printer

### LINKED ISSUE #525: Prettyprinter should do something to avoid conflicts between local names and names chosen for any refs
Paul will do a longer writeup of this issue.

pchiusano:
## Overview

Fixes #525 

Previously, in code like this:

```
qux.quaffle = 23
   
example : Text -> Nat
example quaffle = qux.quaffle + 1
```

... the pretty-printer would pick the name `quaffle` instead of using `qux.quaffle` to distinguish it from the local `quaffle` introduced as a function parameter. 

This PR fixes that.

## Implementation notes

I left some detailed comments in the code, but the general idea is to use the qualified name rather than the suffixed name in cases where there's a conflict with a local variable. The fact that the `PrettyPrintEnv` returns both the qualified and suffixed name made this easier than expected.

## Test coverage

Added a regression test to the pretty-print round trip. It checks for the fix for both term and type references.

## Possible future work

See code comments. There's a less conservative algorithm we could use that would avoid using qualified names in more places, though I'm not sure how much it actually matters and it would require folding into the pretty-printer logic itself, whereas this can be done as a preprocessing step.


unison-src/transcripts-round-trip/main.md
```diff
@@ -7,23 +7,22 @@ This transcript verifies that the pretty-printer produces code that can be succe
 
 ## How to use this transcript: checking round-trip for inline definitions
 
-```unison:hide
+```unison:hide roundtrip.u
```
pchiusano, line 10: I did some cleanup in this file, sending the updates / edits to `roundtrip.u` to avoid clobbering your local `scratch.u` if you're running this transcript locally.

unison-src/transcripts-round-trip/main.md
```diff
@@ -588,5 +582,43 @@ d5 x = match x with
...
+structural type Id a = Id a
+
+exampleType : forall qualifiedName . Id qualifiedName -> Id Fully.qualifiedName
+exampleType z = Id (Dontcare () 19)
+```
+
+```ucm:hide
+.> add
+```
+
```
aryairani:
```suggestion
We'd get a parse error here if `exampleTerm` or `exampleType` didn't round-trip, but it parses okay! 🎉 
```

aryairani (approved):

PR #4217, opened 2023-07-19
Numerous pretty-printer fixes and enhancements

### LINKED ISSUE #3239: Weird treatment of abilities with multiple name segments in operation name
```
structural ability z.Zoink where
  yay.there : Nat -> Nat
  nay : Nat -> Nat
```

Now `view Zoink`

```
  structural ability z.Zoink where
    there : Nat ->{z.Zoink} Nat
    nay : Nat ->{z.Zoink} Nat
```

What happened to `yay`? 

It's just a pretty-printing bug it seems:

```
.scratch> view Zoink

  structural ability z.Zoink where
    there : Nat ->{z.Zoink} Nat
    nay : Nat ->{z.Zoink} Nat

.scratch> ls z.Zoink

  1. nay  (Nat ->{Zoink} Nat)
  2. yay/ (1 definition)

.scratch> ls 2

  1. there (Nat ->{Zoink} Nat)

.scratch> 
```

### LINKED ISSUE #3883: pretty-printer regression with delay and capture
In the version of ucm that is currently deployed to Share, the following code roundtrips fine, but on a trunk build it does not:

```
unique type UUID = UUID Nat Nat

UUID.random : '{Random} UUID
UUID.random = do UUID nat! nat!

UUID.randomUUIDBytes : '{Random} Bytes
UUID.randomUUIDBytes = do
  (UUID a b) = !UUID.random
  (encodeNat64be a) ++ (encodeNat64be b)
```

On trunk `randomUUIDBytes` pretty-prints as:

```
  UUID.randomUUIDBytes : '{Random} Bytes
  UUID.randomUUIDBytes = use Bytes ++
  do
    (UUID.UUID a b) = !UUID.random
    encodeNat64be a ++ encodeNat64be b
```

This doesn't compile.

### LINKED ISSUE #3936: Use nested `do` blocks more aggressively in pretty-printer
The following watch expression renders quite nicely:

```
> do
  x = 1
  y = 2
  x + y
  
          ⧩
          do
            use Nat +
            x = 1
            y = 2
            x + y
```

However, if you add another layer of delay, it renders with (arguably) awkward multi-line parentheses:

```
> do
    do
      x = 1
      y = 2
      x + y
      
                ⧩
          '(do
              use Nat +
              x = 1
              y = 2
              x + y)
```

It's probably rare to nest a `do` block directly within a `do` block like this, but it comes up in more common scenarios like this:

```
main = do
  catchAll do
    x = 1
    y = 2
    x + y
```

```
.> view main
 
  main : '{IO} Either Failure Nat
  main =
    '(catchAll do
        use Nat +
        x = 1
        y = 2
        x + y)
```

The simplest heuristic that comes to mind is to use `'(foo)` if `foo` renders on a single line and `do` with a nested `foo` block if `foo` does not render on a single line. But I'm not sure how well that fits into the way the pretty-printer works.

### LINKED ISSUE #3973: Round Trip bug: multiline strings get extra leading \n
When I store a string with a \n in it, it becomes a multiline string that has had an extra \n prepended

Transcript:
````
```unison

a = "a\nb"

```

```ucm
.> add
.> view a
```
````

Output:
````
```unison
a = "a\nb"

```

```ucm

  change:
  
      a : ##Text

```
```ucm
.> add

    a : ##Text

.> view a

  a : ##Text
  a =
    """
    a
    b"""

```
````

### LINKED ISSUE #4214: `use` clause insertion can sometimes cause variable capture and roundtrip issues
```haskell
Foo.bar.quaffle = 32

example2 : Int -> Nat
example2 oo =
  quaffle = "hi"
  Foo.bar.quaffle 
    + Foo.bar.quaffle 
    + Foo.bar.quaffle + 1
```

This prints as:

```haskell
example2 : Int -> Nat
example2 oo =
  use Foo.bar quaffle
  quaffle = "hi"
  quaffle + quaffle + quaffle + 1
```

Here's why this is happening - the `use` clause insertion happens after name selection. So while `quaffle` correctly gets `Foo.bar.quaffle` as its chosen name, since it's referenced multiple times in the block, the `use` clause insertion pulls out the common prefix.

Not sure the best fix yet.

pchiusano:
In terms of fixes, this PR fixes #3883 and fixes #4214 and fixes #3936 and fixes #3973 and fixes #3239. It also fixes a performance bug that caused quadratic performance slowdown for deeply nested terms. Also fixed a quadratic performance bug when doing things like `edit 1-100`, which previously did repeated linear time lookups into a list.

It also makes improvements in a number of areas.

@SystemFw has been dogfooding this for the past week and @stew has tried it out as well and been very helpful reporting various bugs. I think it's good to go!

## `use` clause insertion is better

We still try to push down `use` clauses to the smallest scope possible, but we only consider scopes which are a `let` or `let rec`. This avoids silliness like:

```Haskell
x = 1 + 1
-- > printing as
x = 
  use Nat +
  1 + 1
```

Instead, the `use Nat +` will go in the parent scope.

## More usage of "soft hangs"

Here's some examples of "soft hangs", where a block element appears as the last function argument or after a `with`:  

```Haskell
List.foreach xs cases
  None -> 0
  Some x -> x + 1

forkAt somewhere do
  x = 1
  y = 2
  x + y

handle something with cases
  { a } -> a
  { abort -> _ } -> 0
```

These now roundtrip as above, rather than doing something not so nice like:

```
List.foreach 
  xs
  (cases 
      None -> 0
      Some x -> x + 1)
```

## Examples

Here's a side by side:

Screenshot 2023-07-24 at 15 59 54

After:

CleanShot 2023-07-24 at 11 01 52@2x


parser-typechecker/src/Unison/Runtime/IOSource.hs
```diff
@@ -607,7 +607,7 @@ unique[ae4e05d8bede04825145db1a6a2222fdf2d890b3044d86fd4368f53b265de7f9] type Do
   = Svg Text
 
 -- ex: Doc2.term 'List.map
-Doc2.term : 'a -> Doc2.Term
+Doc2.term : ∀ a g1 g. '{g} a ->{g1} Doc2.Term
```
pchiusano, line 610: Added a manual type signature here so it has an identical hash to the current definition. This also shows that inference is leaving around some stray ability variables.

parser-typechecker/src/Unison/Syntax/DeclPrinter.hs
```diff
@@ -73,9 +74,9 @@ prettyGADT env ctorType r name dd =
         (DD.constructors' dd)
   where
     constructor (n, (_, _, t)) =
-      prettyPattern env ctorType name (ConstructorReference r n)
+      prettyPattern (PPED.unsuffixifiedPPE env) ctorType name (ConstructorReference r n)
```
pchiusano, line 77: Passing in a `PrettyPrintEnvDecl` (which has both suffixed and unsuffixed PPEs) was part of the fix for #3239.

parser-typechecker/src/Unison/Syntax/TermPrinter.hs
pchiusano, line 1: Extensive changes to this file all over the place to implement the pretty-printing improvements.

parser-typechecker/src/Unison/Builtin/Decls.hs
```diff
@@ -760,7 +763,7 @@ forceTerm :: (Var v) => a -> a -> Term v a -> Term v a
 forceTerm a au e = Term.app a e (unitTerm au)
 
 delayTerm :: (Var v) => a -> Term v a -> Term v a
-delayTerm a = Term.lam a $ Var.named "()"
+delayTerm a = Term.lam a $ Var.typed Var.Delay
```
pchiusano, line 766: We now use a `Var.typed Var.Delay` to for variable names that correspond to the argument to a thunk.

parser-typechecker/src/Unison/Syntax/TypeParser.hs
```diff
@@ -55,12 +55,15 @@ type2a = delayed <|> type2
...
+  t <- effect <|> (pt <$> type2a)
   pure $
     Type.arrow
       (Ann (L.start q) (end $ ann t))
-      (DD.unitType (ann q))
+      (DD.thunkArgType (ann q))
       t
+  where
+    -- if no abilities listed on 't, assume '{} t
+    pt t = Type.effect (ann t) [] t
```
pchiusano, line 66: This is just limiting the places where we do ability inference, since that causes round trip issues with type hashes changing. I think it's weird to use `'Nat` when you really are intending to allow effects there.

parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -1249,6 +1249,10 @@ synthesizeWanted e
           et = existential' l B.Blank e
       appendContext $
         [existential i, existential e, existential o, Ann arg it]
+      when (Var.typeOf i == Var.Delay) $ do
+        -- '(1 + 1) turns into a lambda with an arg variable of type Var.Delay
+        -- here's where the typechecker assumes this must be of type 'thunkArgType'
+        subtype it (DDB.thunkArgType l)
```
pchiusano, line 1255: Previously, we were ending up with inferred thunks that had types like `forall a . a -> Blah` since they didn't use their argument. Which is actually inconsistent, since if you write out the signature of the thunk, it will get a type of `Unit -> ...`. This causes some round trip issues (you'd infer the thunk, but then when pretty-printing it out, it'd get a "different" type because the signature had been added).

parser-typechecker/tests/Unison/Test/Syntax/TermPrinter.hs
pchiusano, line 1: This file's gone. I decided the tests are overly brittle, and instead, we now have a nice test harness in `reparses.u` and `reparses-with-same-hash.u` for testing that things round trip. And we can add more cases there wherever we want.

Another thing that's not great about the unit tests in this file is that when they failed, it's actually pretty hard to tell which one failed. With the new harness, you get line numbers and/or a definition name.

parser-typechecker/tests/Unison/Test/Syntax/TypePrinter.hs
pchiusano, line 1: There's a number of changes in this file due to the new parsing of `'Nat` as `'{} Nat`.

unison-cli/src/Unison/Codebase/Editor/Output.hs
```diff
@@ -623,6 +624,6 @@ isNumberedFailure = \case
   ShowDiffAfterModifyBranch {} -> False
   ShowDiffAfterPull {} -> False
   ShowDiffAfterUndo {} -> False
-  ShowDiffNamespace {} -> False
+  ShowDiffNamespace _ _ _ bd -> BD.isEmpty bd
```
pchiusano, line 627: I added this so that we can do round trip testing in the following way:

- Add a bunch of definitions to one namespace
- Edit those added definitions
- Add the pretty-printed definitions to another namespace
- Diff the two namespaces. They should be identical if the hashes are all the same. And if not, you'll be told which names have different hashes.

See the `transcripts-round-trip/main.md` file which uses this strategy.

unison-cli/src/Unison/CommandLine.hs
```diff
@@ -148,12 +150,12 @@ parseInput getRoot currentPath numberedArgs patterns segments = runExceptT do
...
-  maybe
-    [s]
-    (map (\i -> fromMaybe (show i) . atMay numberedArgs $ i - 1))
-    expandedNumber
+expandNumber numberedArgs s = case expandedNumber of
+  Nothing -> [s]
+  Just nums ->
+    [s | i <- nums, Just s <- [vargs Vector.!? (i - 1)]]
   where
+    vargs = Vector.fromList numberedArgs
```
pchiusano, line 158: A couple things:

1. The old code had a quadratic performance bug, where it was asking for the 1st index of a list, then the second, then the third, and so on. But list indexing takes linear time! So I've converted the args to a vector.
2. This new code allows you to ask for more args than are available, and it'll just ignore any numbers that don't exist. This is handy for saying `edit 1-1000` and not having to worry about whether the list has exactly 1000 elements. I use this for the round trip transcripts so I don't have to keep changing the number passed to the `edit` command.

Aside, the syntax `edit 1-` could be cool, it fills in the ending number for you. But I didn't trust myself to edit the regex code here. :)

unison-core/src/Unison/Term.hs
```diff
@@ -624,13 +624,13 @@ pattern Lam' ::
...
-pattern Delay' :: (Ord v) => Term2 vt at ap v a -> Term2 vt at ap v a
+pattern Delay' :: (Var v) => Term2 vt at ap v a -> Term2 vt at ap v a
 pattern Delay' body <- (unDelay -> Just body)
 
-unDelay :: (Ord v) => Term2 vt at ap v a -> Maybe (Term2 vt at ap v a)
+unDelay :: (Var v) => Term2 vt at ap v a -> Maybe (Term2 vt at ap v a)
 unDelay tm = case ABT.out tm of
   ABT.Tm (Lam (ABT.Term _ _ (ABT.Abs v body)))
-    | Set.notMember v (ABT.freeVars body) ->
+    | Var.typeOf v == Var.Delay || Var.typeOf v == Var.User "()" ->
```
pchiusano, line 633: This was overly broad. Any lambda that ignores its argument is a thunk? That's not correct and causes round trip issues as discussed in an earlier comment.

unison-src/transcripts-round-trip/main.md
pchiusano, line 1: This file has been totally revamped. All the unison code is now loaded from `reparses.u` or `reparses-with-same-hash.u`, so it's now trivial to add more round trip tests without a bunch of boilerplate.

I ported all the existing tests over to this new scheme.

unison-src/transcripts-round-trip/main.output.md
pchiusano, line 1: I think this can be checked in and diff'd as part of CI. It writes to /tmp/roundtrip.u though, so maybe just run it on linux.

unison-src/transcripts-round-trip/nested.u
pchiusano, line 1: Moved into `reparses-with-same-hash.u`.

unison-src/transcripts-using-base/doc.output.md
pchiusano, line 1: here we see some of the pretty-printing changes

unison-src/transcripts/bug-strange-closure.output.md
pchiusano, line 1: These changes are all fine.

unison-src/transcripts/constructor-applied-to-unit.output.md
```diff
@@ -26,27 +26,29 @@ structural type Zoink a b c = Zoink a b c
...
+              , 4
+              , 4
+              , 4
+              , 4
+              , 4
+              , 4
+              , 4
+              , 4
+              , 3
+              ]
```
pchiusano, line 50: New rendering of lists (and tuples get same treatment). Easier to find the extent of the list - the `,` acts as a margin. Closing brace goes on its own line.

unison-src/transcripts/destructuring-binds.output.md
```diff
@@ -43,10 +43,7 @@ ex1 tup =
     c + d
   
   ex1 : (a, b, (Nat, Nat)) -> Nat
-  ex1 = cases
-    (a, b, (c, d)) ->
-      use Nat +
-      c + d
+  ex1 = cases (a, b, (c, d)) -> c Nat.+ d
```
pchiusano, line 46: So much better! Not worth making this multiline just to avoid a `Nat.+`.

unison-src/transcripts/fix2053.output.md
```diff
@@ -3,12 +3,10 @@
...
-      go i as acc =
-        match List.at i as with
-          None   -> acc
-          Some a ->
-            use Nat +
-            go (i + 1) as (acc :+ f a)
+      use Nat +
+      go i as acc = match List.at i as with
+        None   -> acc
+        Some a -> go (i + 1) as (acc :+ f a)
```
pchiusano, line 9: Here we see use of soft hang - the first line of the inner function is a "soft hangable" element (a `match`, but also `do` or `cases` would work), so we swing it up right after the `=`.

unison-src/transcripts/fix2254.output.md
```diff
@@ -98,9 +98,7 @@ Let's do the update now, and verify that the definitions all look good and there
   
   f3 : NeedsA Nat Nat -> Nat
   f3 = cases
-    NeedsA a ->
-      use Nat +
-      f a + 20
+    NeedsA a -> f a Nat.+ 20
     _        -> 0
```
pchiusano, line 102: Shows tweaks to use clause insertion.

unison-src/transcripts/fix2693.output.md
pchiusano, line 1: huge diff due to list rendering changes :)

unison-src/transcripts/fix3171.output.md
```diff
@@ -28,6 +28,6 @@ f x y z _ = x + y * z
   
     5 | > f 1 2 3
           ⧩
-          '(1 Nat.+ 2 Nat.* 3)
+          _ -> 1 Nat.+ 2 Nat.* 3
```
pchiusano, line 31: This was an example where it was being rendered as a delay, but it was really just a lambda that ignored its argument.

unison-src/transcripts/fix3265.output.md
```diff
@@ -35,14 +35,12 @@ are three cases that need to be 'fixed up.'
...
-                f0 y =
-                  match y with
-                    0 -> x
-                    n -> 1 + f1 (drop y 1)
+                f1 y = match y with
+                  0 -> w + x
+                  n -> 1 + f0 (drop y 1)
+                f0 y = match y with
+                  0 -> x
+                  n -> 1 + f1 (drop y 1)
```
pchiusano, line 43: More use of soft hang, swinging the `match` up onto the starting line of the function.

unison-src/transcripts/merge.output.md
```diff
@@ -328,7 +328,7 @@ Now merging `c1b` into `c1a` should result in the updated version of `a` and `f`
   a = "hello world!"
   
   f : Text
-  f = '(y -> y) a "woot!"
+  f = (x y -> y) a "woot!"
```
pchiusano, line 331: This is correct / expected with new stricter thunk recognition logic.

unison-src/transcripts/pattern-match-coverage.output.md
```diff
@@ -858,7 +858,7 @@ result f = handle !f with cases
   Patterns not matched:
   
     * { _ }
-    * {abortWithMessage _ -> _}
+    * { abortWithMessage _ -> _ }
```
pchiusano, line 861: See: https://github.com/unisonweb/unison/pull/4217/files#diff-83b826ee888dba10b0eba9eba2d00923a2dfb9e4578f8b404227628c27ae424c

unison-src/transcripts/pattern-pretty-print-2345.output.md
```diff
@@ -183,8 +183,8 @@ doc = cases
 
   throaty : Request {g, Ab} x -> ()
   throaty = cases
-    {Ab.a a -> k} -> ()
-    { _ }         -> ()
+    { Ab.a a -> k } -> ()
+    { _ }           -> ()
```
pchiusano, line 187: For some strange reason, the pure case of request matching was done with spaces after the braces, but the operations left out the spaces. WTH?

Now both use a space after the braces. I find it a bit more readable to give the symbol some breathing room.

unison-src/transcripts/text-literals.output.md
pchiusano, line 1: @stew you can take a look at this file, it shows the new parsing of raw text literals to fix #3973. Raw strings now all round trip, but note that a leading newline after the `"""` is ignored, as is the trailing newline. You can use an extra blank line at the start or end if you want a newline in either of those places.

unison-syntax/src/Unison/Syntax/Lexer.hs
```diff
@@ -793,12 +793,16 @@ lexemes' eof =
...
-      let lines' s = lines s <> (if take 1 (reverse s) == "\n" then [""] else [])
-      pure $ case lines' s of
+      -- a last line that's equal to `leading` is ignored, since leading
+      -- spaces up to `col` are not considered part of the string
+      let tweak l = case reverse l of
+            last : rest
+              | col > 1 && last == leading -> reverse rest
+              | otherwise -> l
+            [] -> []
+      pure $ case tweak (lines s) of
```
pchiusano, line 805: Lexer fix for #3973

pchiusano (commented):
Self review.

aryairani:
> More usage of "soft hangs"

I wish ormolu did this

unison-core/src/Unison/Term.hs
aryairani: Looking forward to getting rid of the second case

unison-core/src/Unison/Var.hs
```diff
@@ -87,6 +87,7 @@ rawName typ = case typ of
   Irrelevant -> "_irrelevant"
   UnnamedReference ref -> Reference.idToText ref
   UnnamedWatch k guid -> fromString k <> "." <> guid
+  Delay -> "()"
```
aryairani, line 90: Is this weird?

unison-src/transcripts/ability-term-conflicts-on-update.md
```diff
@@ -29,7 +29,7 @@ unique ability Channels where
 Channels.send : a -> ()
 Channels.send a = ()
 
-thing : '()
+thing : '{Channels} ()
```
aryairani, line 32: Whoa, haha. Kind of a big difference.

unison-src/transcripts/ability-term-conflicts-on-update.output.md
```diff
@@ -77,13 +77,14 @@ These should fail with a term/ctor conflict since we exclude the ability from th
 
 .ns> update patch thing
 
-  x These definitions failed:
+  ⍟ I've added these definitions:
```
aryairani, line 80: Is this expected?

unison-src/transcripts/bug-strange-closure.output.md
aryairani: And super boring, spare yourself from looking

unison-src/transcripts/contrabilities.md
```diff
@@ -3,6 +3,6 @@
 ```
 
 ```unison
-f : 'a -> Nat
```
aryairani, line 6: @pchiusano Why this change?

unison-src/transcripts/fix1926.output.md
```diff
@@ -25,7 +25,7 @@ sq = 2934892384
 
     1 | > 'sq
           ⧩
-          'sq
+          do sq
```
aryairani, line 28: I guess this is fine?

unison-src/transcripts/fix2268.md
```diff
@@ -13,7 +13,7 @@ unique ability A where
 unique ability B where
   a : Char
 
-test : 'Nat
+test : () -> Nat
```
aryairani, line 16: Why this change?

aryairani (approved):

parser-typechecker/src/Unison/Runtime/IOSource.hs
aryairani: @pchiusano What, if anything, do we need to do on this line after tweaking the inference (or whatever you and @tstat  settled on doing, if not that)?

aryairani:
Great! I left some short questions

parser-typechecker/src/Unison/Runtime/IOSource.hs
pchiusano: Nothing, we can leave it alone. The pretty-printer will still elide these "useless" ability variables since there's a lot of code in the wild that has them (including this definition - the current inferred type is what I added here).

unison-src/transcripts/ability-term-conflicts-on-update.md
pchiusano: This one I added - it was previously relying on ability inference (you can see it calls operations of `Channels` in the body of `thing`), but now that `'blah` is assumed pure, you have to mention what abilities you're allowing if you use the `'` syntax.

unison-src/transcripts/contrabilities.md
pchiusano: Hmm, I'm honestly not sure the purpose of this transcript and was just mechanically translating, since `'blah` doesn't allow effects and the existing output did allow effects. But looking at it now, its purpose might be to see if arrows in contravariant position get inferred as ability polymorphic? I switched it to use `(() -> a)` here, and it infers as `(() ->{g} a)`.

unison-src/transcripts/fix2268.md
pchiusano: `'Nat` is now equivalent to `'{} Nat`, but I believe this example relies on inference of abilities on that arrow.

unison-core/src/Unison/Var.hs
pchiusano: It's what we did before, though it happened during parsing - anytime we parsed a thunk, we'd parse it as a lambda with a variable called `()`.

I think ideally we'd just store the variable type in the database, but right now we convert everything to strings and this PR doesn't change that. So we need to pick a name that the user can't use and that the pretty-printer can reverse to decide when to use the thunk / `do` syntax.

unison-src/transcripts/fix1926.output.md
pchiusano: I guess.

What's happening is that the surrounding precedence is 0, so it's allowed to use `do`. It's also allowed to use `do` as the last argument to a function call. In cases where you could get away with both, it's easy to tweak to prefer more or less `do`. Going to leave it for now though.

PR #4218, opened 2023-07-21
Implement value reflection in scheme

dolio:
This PR is part of the implementation of value reflection in scheme, as well as many other modifications to the way things work that were implemented as part of that effort. Following is a summary of some of the changes.

- A new `unison/data-info` module is generated from unison data. It contains hash and constructor tag information and in-scheme constructor functions for various datatypes that are used in builtin functions. This depends on little other functionality (mostly `unison/data`), so it can be very early in the dependency graph, and things like `unison/primops` can use it to directly implement builtins without needing wrappers.
- The `LOAD` and `CACH` primops now check the dependency information for the objects they're called with, so that they shouldn't try to evaluate scheme code that won't actually work, and instead report the necessary additional dependencies in such cases.
- Universal compare/eq works for more runtime objects.
- `describe-value` has been expanded for better debug/trace messages covering more cases
- `unison/data` `unison/core` and `unison/boot` are now written in the racket language instead of r6rs. Mixing the two causes quite a few headaches
- Racket structs are now used for various builtin types, like term/type links. One thing this avoid is a sort of circularity that resulted from references being represented as unison data (which would have a `Reference` reference in the representation, which was a data, which had a `Reference` reference ...). But, this is also necessary for proper reflection, because otherwise it is unclear what unison type a data value is supposed to correspond to. There must be some kind of runtime tag corresponding to the type to properly implement reflection.
- One of the structs is for partial application of unison functions. A property has been specified that allows these structs to act like ordinary scheme procedures. But we can also detect them and get their partial argument list for reflection purposes.
- Some pseudo-builtins have been added to replace (un)wrappers for these structs. The pure unison code goes through serialization to mediate between the two representations, but there are much more efficient scheme functions that accomplish the same thing, so the latter are swapped in for the former in the scheme pipeline.

The serialized test cases have been updated to check that reflecting and re-serializing the deserialized test case results in the same bytes. This should exercise the new reflection machinery somewhat.

I merged in trunk at some point during development, but I seem to be out of date again. I think I also have some hard-coding to replace still in light of some of the above changes. I also need to do the share end of this and cut a new `internal` release to point to. But otherwise I think this is pretty close to working.


pchiusano (approved):
Awesome, Dan! Can't wait. You can go ahead and merge this once you fix conflicts and CI passes.

PR #4220, opened 2023-07-24
Add RNDF SQRT TANH TANF TZRO POPC to the jit

jaredly:
## Summary:
While testing TZRO I discovered that racket's fixnums (at least in the version I've got,
v8.7 [cs]) are a lot smaller than than unison's `Int`s.
`(greatest-fixnum)` is `1152921504606846975`
but `Int.maxInt` is `9223372036854775807`

### DIFF
--- a/scheme-libs/racket/unison/math.rkt
+++ b/scheme-libs/racket/unison/math.rkt
@@ -127,13 +127,3 @@
         (if (eq? -1 bit)
             64
             bit)))
-
-; (do ((result 0 (+ result 1))
-;      (bits (if (fxnegative? fx)
-;                (fxnot fx)
-;                fx)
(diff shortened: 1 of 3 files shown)


pchiusano:
@jaredly does that matter for correctness or is it just that your code might switch to using boxed arb precision numbers earlier than in the Haskell runtime?

jaredly:
@pchiusano So the problem I uncovered will have to be handled in a bunch of places, figured we should do it all at once. Currently using an Integer larger than `1152921504606846975` produces runtime exceptions

jaredly:
fwiw we can land this branch independent of fixing the fixnum issue

dolio (approved):
Looks good to me.

PR #4224, opened 2023-07-25
Ensure `TVar` values can be compared for equality

### LINKED ISSUE #4221: Structural comparison of `TVar`s fails at runtime
Comparing two `TVar` values (or one to itself) with `Universal.===` results in a runtime exception:

```haskell
var = atomically TVar.new
var === var
```

```
Encountered exception:
Attempting to check equality of two values of different types:
    ( ##TVar
    , ##TVar
    )
    error
```

dfreeman:
## Overview

This change fixes #4221, following the precedent set by `MVar` and others of deferring to the default `Eq` implementation and just checking identity.

## Test coverage

I haven't added any test coverage as part of this change. I don't see a lot in the way of comparable coverage for other types, and this doesn't seem like something that's suuuper likely to regress now that it's fixed, but I'm happy to go add a test somewhere if it's desired.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Foreign.hs
+++ b/parser-typechecker/src/Unison/Runtime/Foreign.hs
@@ -20,6 +20,7 @@ module Unison.Runtime.Foreign
 where
 
 import Control.Concurrent (MVar, ThreadId)
+import Control.Concurrent.STM (TVar)
 import Crypto.Hash qualified as Hash
 import Data.IORef (IORef)
 import Data.Primitive (ByteArray, MutableArray, MutableByteArray)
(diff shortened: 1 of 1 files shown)


dolio (approved):
The code changes look fine to me.

Someone else should probably weigh on on the test thing, because my attitude towards that is more lax.

pchiusano:
Nice! Thanks @dfreeman. Going to merge as is.

One other one to check is universal ordering: https://share.unison-lang.org/@unison/base/code/releases/2.3.0/latest/terms/Universal/ordering I wonder if that also triggers a similar bug.

dfreeman:
It looks like it does trigger a similar bug, though for some reason `MVar` doesn't, despite not being special-cased for ordering like it is for equality 🤔

dolio:
Neither `MVar` nor `TVar` have ordering. I'm not sure why `MVar` wouldn't cause an error if you tried to ask about ordering. That's pretty odd if so.

dfreeman:
Oh scratch that, I had an unrelated watch expression with a type issue and wasn't paying attention, so my `run seeWhatHappens` was running old code instead of `compare mvar mvar`.

Confirmed now that `MVar` has the same problem when you try to get an ordering.

```
Attempting to compare two values of different types:
    ( ##MVar
    , ##MVar
    )
```

dolio:
I guess I should make an issue to improve the error message for cases like that, because it doesn't make much sense.

PR #4229, opened 2023-07-26
Fix performance problems with `branch.delete`

### LINKED ISSUE #4072: Why is branch.delete slow?
When I run branch.delete it regularly takes 10+ seconds. What takes it so long? I would expect it to essentially be the deletion of a name -> hash mapping. Is it doing a vacuum/gc of newly unreachable state?

I think that it should either be made faster or should provide a message about what it's doing.

tstat:
Fixes #4072

## Overview

See [this comment](https://github.com/unisonweb/unison/issues/4072#issuecomment-1650562835) for an overview.

There was back-and-forth discussion on slack as to whether we should be consing empty `Branch`es onto child causals that exist in the base `Branch` but not the head `Branch`, or if we should instead just leave them deleted in the head. 

We initially planned to leave them deleted only for project branch deletions, but then decided to leave them deleted everywhere.

## Test coverage

Tested manually, and tested by everything that calls some `updateAt` variant.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Branch.hs
+++ b/parser-typechecker/src/Unison/Codebase/Branch.hs
@@ -747,18 +747,18 @@ consBranchSnapshot headBranch baseBranch =
           (head headBranch & children .~ combinedChildren)
           (_history baseBranch)
   where
-    combineChildren :: These (Branch m) (Branch m) -> Branch m
+    combineChildren :: These (Branch m) (Branch m) -> Maybe (Branch m)
     combineChildren = \case
       -- If we have a matching child in both base and head, squash the child head onto the
       -- child base recursively.
-      (These base head) -> head `consBranchSnapshot` base
-      -- This child has been deleted, recursively replace children with an empty branch.
-      (This base) -> empty `consBranchSnapshot` base
+      (These base head) -> Just (head `consBranchSnapshot` base)
+      -- This child has been deleted, let it be
+      (This _) -> Nothing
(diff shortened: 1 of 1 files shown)


aryairani:
I think any weirdness this change may cause is likely best fixed in the merge algorithm, which we're currently reworking anyway.

parser-typechecker/src/Unison/Codebase/Branch.hs
```diff
@@ -747,18 +747,18 @@ consBranchSnapshot headBranch baseBranch =
...
-    combineChildren :: These (Branch m) (Branch m) -> Branch m
+    combineChildren :: These (Branch m) (Branch m) -> Maybe (Branch m)
     combineChildren = \case
       -- If we have a matching child in both base and head, squash the child head onto the
       -- child base recursively.
-      (These base head) -> head `consBranchSnapshot` base
-      -- This child has been deleted, recursively replace children with an empty branch.
-      (This base) -> empty `consBranchSnapshot` base
+      (These base head) -> Just (head `consBranchSnapshot` base)
+      -- This child has been deleted, let it be
```
aryairani, line 755: rest in peace

aryairani (approved):

ChrisPenner (approved):
Nice!
🤞🏼🤞🏼 🤞🏼

ceedubs:
Thanks! branch deletion went from ~30 seconds to ~40ms for me!

PR #4235, opened 2023-07-30
Improve error message for missing result of block

### LINKED ISSUE #4077: bad error message if a block ends in a variable declaration
In the following code, the `foo` function ends with a declaration of the variable `x` (`x = 3`), but we don't actually use or return `x`. The compiler catches this as an error (which is good), but it gives a bad error message:

```
foo =
  x = 3
```

```
  I couldn't find any definitions matching the name _100003 inside the namespace .__projects._90b3d793_b355_4859_b6e2_6ba8b33c55a5.branches._e56b737a_033a_4866_9883_6691ba53715b
  
      2 |   x = 3
  
  Some common causes of this error include:
    * Your current namespace is too deep to contain the definition in its subtree
    * The definition is part of a library which hasn't been added to this project
  
  To add a library to this project use the command: `fork <.path.to.lib> .__projects._90b3d793_b355_4859_b6e2_6ba8b33c55a5.branches._e56b737a_033a_4866_9883_6691ba53715b.lib.<libname>`
  
```

## Notes

- This output also includes an issue with pretty-printing of a project namespace, but that's a separate issue.
- This has happened for a long time so I figured there would be an issue for it, but I couldn't find one.

runarorama:
This PR introduces changes to how the Unison typechecker handles and reports the error case where the user forgets to add an expression to the end of a block.

Fixes #4077 

## File Changes

`parser-typechecker/src/Unison/PrintError.hs`
Improved the error messages that are returned when the last statement of a block is a definition rather than an expression. The message now includes the term causing the error and suggestions on what type it should be.

`parser-typechecker/src/Unison/Typechecker.hs`
Introduced the variable `v` in the Resolution data structure for improved error reporting. The `suggest` and `resolveNote` functions are updated accordingly to handle the new field.

`unison-core/src/Unison/Blank.hs`
Added a new `MissingResultPlaceholder` type to the `Recorded` data structure.

`unison-core/src/Unison/Term.hs`
Included `missingResult` function to generate missing result placeholders and adjusted `prepareTDNR` to account for missing results.

`parser-typechecker/src/Unison/Typechecker/Context.hs`
Adjusted the synthesizeWanted function to consider the possibility of a `MissingResultPlaceholder`.

`unison-hashing-v2/src/Unison/Hashing/V2/Term.hs`
Updated the hashing functionality to consider the new `MissingResultPlaceholder`.

## Justification

These changes improve the clarity of error messages and allow the typechecker to handle cases where the last statement of a block is a definition instead of an expression. This PR improves the user experience by providing more detailed and helpful error messages.

## Tests

No new tests, but a transcript exercises this codepath and will highlight any regression here.

## Documentation

Added comments in the modified sections of the code.

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -609,6 +609,21 @@ renderTypeError e env src curPath = case e of
     Pr.hang
       "This case would be ignored because it's already covered by the preceding case(s):"
       (annotatedAsErrorSite src loc)
+  UnknownTerm {..}
+    | Var.typeOf unknownTermV == Var.MissingResult ->
+        mconcat
+          [ "The last statement of a block must be an expression, ",
+            "but this is a definition:\n\n",
+            annotatedAsErrorSite src termSite,
+            "\nI don't know what the result of this block should be.\n",
+            "Did you forget to add an expression at the end of the block?\n",
+            case expectedType of
+              Type.Var' (TypeVar.Existential {}) -> "I don't know what type it should be either."
+              _ ->
+                "It should be of type "
+                  <> style Type1 (renderType' env expectedType)
+                  <> "."
+          ]
   UnknownTerm {..} ->
     let (correct, wrongTypes, wrongNames) =
           foldr sep id suggestions ([], [], [])
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -611,16 +611,16 @@ renderTypeError e env src curPath = case e of
       (annotatedAsErrorSite src loc)
   UnknownTerm {..}
     | Var.typeOf unknownTermV == Var.MissingResult ->
-    Pr.lines [ 
-      Pr.wrap "The last element of a block must be an expression, but this is a definition:", 
-      "",
-      annotatedAsErrorSite src termSite,
-      Pr.wrap $ "Try adding an expression at the end of the block." <> msg
-      ]
-      where
-        msg = case expectedType of
-          Type.Var' (TypeVar.Existential {}) -> mempty
-          _ -> Pr.wrap $ "It should be of type " <> Pr.group (style Type1 (renderType' env expectedType) <> ".")
+        Pr.lines
+          [ Pr.wrap "The last element of a block must be an expression, but this is a definition:",
+            "",
+            annotatedAsErrorSite src termSite,
+            Pr.wrap $ "Try adding an expression at the end of the block." <> msg
+          ]
+    where
+      msg = case expectedType of
+        Type.Var' (TypeVar.Existential {}) -> mempty
+        _ -> Pr.wrap $ "It should be of type " <> Pr.group (style Type1 (renderType' env expectedType) <> ".")
   UnknownTerm {..} ->
     let (correct, wrongTypes, wrongNames) =
           foldr sep id suggestions ([], [], [])
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -611,19 +611,16 @@ renderTypeError e env src curPath = case e of
       (annotatedAsErrorSite src loc)
   UnknownTerm {..}
     | Var.typeOf unknownTermV == Var.MissingResult ->
-        mconcat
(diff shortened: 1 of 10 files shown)


unison-hashing-v2/src/Unison/Hashing/V2/Term.hs
```diff
@@ -162,6 +162,8 @@ instance (Var v) => Hashable1 (TermF v a p) where
                         [tag 1, Hashable.Text (Text.pack s)]
                       B.Recorded (B.Resolve _ s) ->
                         [tag 2, Hashable.Text (Text.pack s)]
+                      B.Recorded (B.MissingResultPlaceholder _) ->
+                        [tag 3, Hashable.Text (Text.pack "_")]
```
pchiusano: @runarorama This looks like a problem maybe. It's reusing a tag that's already used for another case (see below).

runarorama: oh I see. Subtle!

parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -1222,7 +1222,10 @@ synthesizeWanted e
   | Term.TermLink' _ <- e = pure (Type.termLink l, [])
   | Term.TypeLink' _ <- e = pure (Type.typeLink l, [])
   | Term.Blank' blank <- e = do
-      v <- freshenVar Var.blank
```
pchiusano, line 1225: Is this change needed? This is just a new type of blank variable, so freshening should work the same, no?

unison-hashing-v2/src/Unison/Hashing/V2/Term.hs
aryairani: You mean `TermApp`? I would think it's ok because `MissingResultPlaceholder` would produce `1 : 3 : ...`, whereas `TermApp` would just produce `3 : ...`

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -609,6 +609,21 @@ renderTypeError e env src curPath = case e of
...
+            annotatedAsErrorSite src termSite,
+            "\nI don't know what the result of this block should be.\n",
+            "Did you forget to add an expression at the end of the block?\n",
+            case expectedType of
+              Type.Var' (TypeVar.Existential {}) -> "I don't know what type it should be either."
+              _ ->
+                "It should be of type "
+                  <> style Type1 (renderType' env expectedType)
+                  <> "."
+          ]
```
pchiusano: Suggestion to use `Pr.lines` and/or `Pr.wrap` as in the other cases. Don't use hard line breaks except when breaking paragraphs. This deals with terminal being too short for "The last element of a block must be an expression, but this is a definition."

Also simplifying language a bit, wdyt? 

```suggestion
    Pr.lines [ 
      Pr.wrap "The last element of a block must be an expression, but this is a definition:", 
      "",
      annotatedAsErrorSite src termSite,
      Pr.wrap $ "Try adding an expression at the end of the block." <> msg
      ]
      where
        msg = case expectedType of
          Type.Var' (TypeVar.Existential {}) -> mempty
          _ -> Pr.wrap $ "It should be of type " <> Pr.group (style Type1 (renderType' env expectedType) <> ".")
```

pchiusano (approved):

parser-typechecker/src/Unison/Typechecker/Context.hs
runarorama: Yeah, this is needed to preserve the `typeOf` the variable. `Var.blank` sets the type to `Blank` instead of `MissingResult`. Preserving it lets the error printer know that this is specifically a missing result type of blank.

parser-typechecker/src/Unison/PrintError.hs
pchiusano: My take is that "I don't know what type it should be either" sounds weird / somewhat confusing, especially for someone who doesn't already know that there's an alternate form of the message. It also doesn't really add any information. If there's no constraints, just don't tell the user about any constraints.

aryairani:
Cool, thanks for working this out.

Do we have access to the thing where we can show what line of source the type we're expecting came from? If so, maybe we could add that to the output too wherever we say it should be of type \<whatever>.

PR #4244, opened 2023-08-03
Add github action to populate nix development cache

tstat:
## Overview

It's all in the title I think. Users of nix won't have to build things that CI has built.

## Loose ends

There aren't github action runners for aarch64-darwin right now, so only x86_64-linux and x86_64-darwin are built.

We should consider moving more github actions into a nix workflow in the future so that we don't duplicate building effort and we can simplify our action caching.

### DIFF
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -1,0 +1,33 @@
+name: Nix development cache
+
+on:
+  # Build on every pull request (and new PR commit)
+  pull_request:
+  # Build on new pushes to trunk (E.g. Merge commits)
+  # Without the branch filter, each commit on a branch with a PR is triggered twice.
+  # See: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662
+  push:
+    branches:
+      - trunk
+
+jobs:
+  nix:
+    name: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      # Run each build to completion, regardless of if any have failed
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-20.04
+          - macOS-12
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v22
+    - uses: cachix/cachix-action@v12
+      with:
+        name: unison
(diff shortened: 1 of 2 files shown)


aryairani:
Is it workable to cross-build aarch64-darwin?

ceedubs:
Do we really want to have every PR (and specifically non-maintainer PRs) push to the cachix cache? My inclination would be to only have releases and merges to `trunk` populate the cache. It would mean some more cache misses, but hopefully enough of the build would be stable that most of it would be reused? I guess that I don't quite know the granularity of the objects that haskell.nix puts in the nix store.

tstat:
> Is it workable to cross-build aarch64-darwin?

Cross compilation won't help us with a development cache as the cross-compiled products have different dependencies than the natively compiled products, and thus have different hashes. So, someone using aarch64-darwin would get all cache misses. Cross compilation would be fine for something like creating a release binary.

We would instead need to emulate aarch64-darwin on x86_64-darwin or something like that. I don't know how to arrange this on darwin, or if apple supports it.

tstat:
> Do we really want to have every PR (and specifically non-maintainer PRs) push to the cachix cache? My inclination would be to only have releases and merges to trunk populate the cache. It would mean some more cache misses, but hopefully enough of the build would be stable that most of it would be reused? I guess that I don't quite know the granularity of the objects that haskell.nix puts in the nix store.

Each cabal component is a derivation with haskell.nix, so that is the granularity. 

I think one motivating example of pushing to this cache for PRs is being able to test out other contributor's PRs without compiling it yourself. For example, if I put up some PR then you can check it out and execute `nix run`, which will download the ucm binary from cachix instead of doing any compilation.

Note that workflows triggered via `pull_request` opened from a different repository (as all non-maintainer PRs must be) do not have access to action secrets and thus cannot (and should not) be able to push to cachix. It would be best to not execute this workflow for those PRs, but I'm not sure how to add a restriction on the base repo.

ceedubs (approved):
Ah @tstat that makes sense that this won't push to cachix for forks from other repos. 👍 LGTM

aryairani:
@tstat Possibly relevant: https://michaelheap.com/github-actions-check-permission/
We could check either `author_association` (not sure if the groups overlap or are disjoint) or look for `write` permission before loading the cachix key.

Issue #3773, opened 2023-01-26
Allow multiple (ignored) bindings called `_` in a block


pchiusano:
Due to a quirk of the parser, this isn't allowed:

```haskell
foo = 
  _ = 1
  _ = 3
  1 + 1
```

The `_` is parsed as a regular variable name! And you can't have multiple bindings of the same name in the same block.

One easy fix without shaking things up:

`_ = blah` parses as `_ignore_<digits> = blah` 

(Where `<digits>` is just some random location-dependent digits, guaranteed unique in the file, which we use for other things in the parser.)

That is, it's just a regular binding with a special name. In the typechecker, for bindings whose name has this form, we can switch off the check that it be of type `()`. The prettyprinter can then print variables of this form just as `_`.

Alternately we could leave the typechecker alone and have it parse as `(_ -> ()) blah`, and then the pretty-printer elides this lambda if the binding name has the right form. I don't like this as much.

ceedubs:
It's nice that `_ = blah` isn't many characters to type, but it feels a bit clunky to me. This adds an exception to the rule of what variable assignments do, and for a not-obvious effect. `ignore blah` is more characters but avoids special cases and seems more obvious to users. And if you see it on Share you can click on `ignore` to see what it does which isn't true for `_ = blah`.

Or we could name it `discard` or whatever.

PR #4247, opened 2023-08-06
Term declarations should not allow hashes in the term name (typesig or term definition)

### LINKED ISSUE #4233: term declarations appear to allow hashes in the term names
```
(Numeric.>=##Nat) : Int -> Int -> Boolean
(Numeric.>=##Nat) = todo "implement"
```

UCM does not complain about this. Have been told on Slack this should be a syntax error.

The issue appears to be the use of `prefixDefinitionName` in the `typedecl` parser in `TermParser.hs`:

```
typedecl =
  (,)
    <$> P.try (prefixDefinitionName <* reserved ":")
    <*> TypeParser.valueType
    <* semi

binding = 
  -- [snip]
  let prefixLhs = do
        v <- prefixDefinitionName
        vs <- many prefixDefinitionName
        pure (ann v, v, vs)
```

and in `Parser.hs`:

```
prefixDefinitionName = wordyDefinitionName <|> parenthesize symbolyDefinitionName

wordyDefinitionName = queryToken $ \case
  L.WordyId s _ -> Just $ Var.nameds s
  L.Blank s -> Just $ Var.nameds ("_" <> s)
  _ -> Nothing

symbolyDefinitionName = queryToken $ \case
  L.SymbolyId s _ -> Just $ Var.nameds s
  _ -> Nothing
```

I'm guessing instead of `prefixDefinitionName` there should be a separate function in `Parser.hs`,

```
prefixTermName = 
  wordyIdString <|> parenthesize symbolyIdString -- should fail on the presence of a hash
```

and then in `TermParser.hs`:

```
typedecl =
  (,)
    <$> P.try (prefixTermName <* reserved ":")
    <*> TypeParser.valueType
    <* semi

binding = 
  -- [snip]
  let prefixLhs = do
        v <- prefixTermName
        vs <- many prefixTermName -- iffy about this one, maybe it should just be wordyId <|> symbolyId?
        pure (ann v, v, vs)
```

kylegoetz:
## Overview

This PR fixes issue #4233. Without this, UCM/Unison did not have a problem with code that featured hashes in the names of type signatures and term definition. This PR errors when the same code is attempted.

There is also now a transcript test that verifies this code now results in an error:

```
x##Nat : Int -> Int -> Boolean
x##Nat = 5
```

previously resulted in no error (in, say, `release/M5a`). Now this transcript test fails if you replace `unison:hide:all:error` with just `unison` (which does not hide errors). Previously it would "fail" with a no-error success as no error would have occurred.

This fixes #4233.

## Implementation notes

In `Parser.hs` I created a new function, `prefixTermName :: (Var v) => P v (L.Token v)`, which errors if there is a hash. Then in `TermParser.hs`, I swapped out use of `prefixDefinitionName` for `prefixTermName`. The former merely discarded a hash rather than erroring out on one.

I could have leveraged the existing `wordyIdString :: (Ord v) => P v m (L.Token String)` and `symbolyIdString :: (Ord v) => P v m (L.Token String)`, which also reject if there's a hash.

However, the type signature of these is `(Ord v) => P v m (L.Token String)`, while the function being replaced in code was type signature `(Var v) => P v (L.Token v)`. Not knowing downstream effects of changing the type signature here, I did not use either of the existing functions in my new `prefixTermName`, which internally uses new `wordyTermName` and `symbolyTermName`.

## Test coverage

I added a transcript test, located at `unison-src/transcripts/no-hash-in-term-declaration.md` plus the corresponding output `md` file.

I think this is adequate. Consider it a regression test, and if code changes to once again allow hashes in this place by mistake, `stack exec transcripts` should fail.

### DIFF
--- a/unison-syntax/src/Unison/Syntax/Parser.hs
+++ b/unison-syntax/src/Unison/Syntax/Parser.hs
@@ -31,6 +31,7 @@ module Unison.Syntax.Parser
     peekAny,
     positionalVar,
     prefixDefinitionName,
+    prefixTermName,
     queryToken,
     reserved,
     root,
@@ -299,7 +300,7 @@ prefixDefinitionName =
 
 -- Parse a prefix identifier e.g. Foo or (+), rejecting any hash
 -- This is useful for term declarations, where type signatures and term names should not have hashes.
-prefixTermName :: (Var v) => P v (L.Token v)
+prefixTermName :: (Var v) => P v m (L.Token v)
 prefixTermName = wordyTermName <|> parenthesize symbolyTermName
   where
     wordyTermName = queryToken $ \case
--- a/unison-syntax/src/Unison/Syntax/Parser.hs
+++ b/unison-syntax/src/Unison/Syntax/Parser.hs
@@ -297,6 +297,19 @@ prefixDefinitionName :: (Var v) => P v m (L.Token v)
 prefixDefinitionName =
   wordyDefinitionName <|> parenthesize symbolyDefinitionName
 
+-- Parse a prefix identifier e.g. Foo or (+), rejecting any hash
+-- This is useful for term declarations, where type signatures and term names should not have hashes.
+prefixTermName :: (Var v) => P v (L.Token v)
+prefixTermName = wordyTermName <|> parenthesize symbolyTermName
+  where
+    wordyTermName = queryToken $ \case
+      L.WordyId s Nothing -> Just $ Var.nameds s
+      L.Blank s -> Just $ Var.nameds ("_" <> s)
+      _ -> Nothing
+    symbolyTermName = queryToken $ \case
+      L.SymbolyId s Nothing -> Just $ Var.nameds s
+      _ -> Nothing
+
 -- Parse a wordy identifier e.g. Foo, discarding any hash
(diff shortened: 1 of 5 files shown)


unison-syntax/src/Unison/Syntax/Parser.hs
```diff
@@ -297,6 +298,19 @@ prefixDefinitionName :: (Var v) => P v m (L.Token v)
...
   wordyDefinitionName <|> parenthesize symbolyDefinitionName
 
+-- Parse a prefix identifier e.g. Foo or (+), rejecting any hash
+-- This is useful for term declarations, where type signatures and term names should not have hashes.
+prefixTermName :: (Var v) => P v m (L.Token v)
+prefixTermName = wordyTermName <|> parenthesize symbolyTermName
+  where
+    wordyTermName = queryToken $ \case
+      L.WordyId s Nothing -> Just $ Var.nameds s
+      L.Blank s -> Just $ Var.nameds ("_" <> s)
```
pchiusano, line 308: Making a note for myself that when fixing https://github.com/unisonweb/unison/issues/3773 to try parsing this differently, rather than as a regular variable.

pchiusano (approved):
LGTM, thanks @kylegoetz !

PR #4255, opened 2023-08-10
Change runtime term link representation to allow code validation

dolio:
This PR changes the way various compiler/runtime functionality works to allow us to perform some checks when loading code.

1. The codebase->intermediate compilation has been changed. Previously we were just pulling hashed terms out of the codebase, performing some transformations on them, and hashing new terms that were generated from the transformations. However, this introduces situations where mutually recursive definitions needn't have a common base hash, which is a useful property for subsequent steps. So, instead we undo some of the hashing, perform our transformations, hash again using the codebase hash, and keep track of the mapping between the two.
2. After the code above has been compiled into intermediate code, we again rehash it using an algorithm that operates directly on the intermediate code. This takes advantage of the fact that mutually recursive definitions will share a common hash to calculate serialized representations of the intermediate terms that are independent of the hash used to refer to the mutually recursive group. Then those serialized representations are hashed to calculate a new group hash.
3. Since the above process only depends on some ordering of the definitions within a group, the hash generation for a group should be idempotent. So, before loading code, we can rerun the hashing algorithm on the code to be loaded to make sure we get the same hash, meaning that it makes sense to load the given code with the provided hashes.

After merging this PR, all term links at runtime will contain the intermediate hashes. However, I store the various mappings in the evaluation context, so that decompilation and such can continue to report sensible information. However, compiled and dynamically loaded code won't have this information at the moment, so hashes printed in those scenarios will be harder to match up to codebase values. Perhaps some kind of 'debugging' file/info could be generated for those situations in the future.

I haven't written any tests yet, but I thought I'd open the PR so that people could start looking it over in the mean time. Maybe check out if the verification function's API looks good. The idea is that it's a function:

    validateLinks : [(Link.Term,Code)] ->{Exception} Either [Link.Term] [Link.Term]

If you get back a left list, then you are missing parts of a binding group. If you get back a right list, it is a list of the input links that do not match the hash we calculated for the associated code. Exceptions result from feeding in nonsense, like builtin or constructor term links.


pchiusano (approved):
Looks good. This actually is getting some test coverage from the deserialization tests which I see you refreshed.

Can you explain the component missing a binding error? Is that like if you try to verify code but not the other code in the same cycle? That would also fail to load, but I guess you're just saying that we also can't verify it without having the full cycle.

Other question: this rejiggering that you do to support this, that is just happening once when you compile the term, right? It's not happening every time you hash something or verify Code, is it?

I'd say try to think of some obvious tests (testing the failure paths seems good), then get CI passing then you can merge. This will get lots of ambient testing in cloud.

ceedubs:
Exciting! How will this fail if you try to deserialize code that was serialized in the old format? Will it error with something understandable like "Unexpected serialization version: 1"?

Also based on my understanding of your description of the type signature `validateLinks : [(Link.Term,Code)] ->{Exception} Either [Link.Term] [Link.Term]`, in the common case of successful validation you should get a `Right []`, right?

dolio:
@pchiusano Yeah. In order to calculate the hash of mutually recursive functions, you need all the functions involved in the mutual recursion. So, if some are missing, it returns the links of the missing pieces.

I think the answer to your other question is that yes, the unhashing only happens when loading code from the codebase and compiling it to intermediate code. Once it's in the intermediate store the rehashing is purely based on generating flat representations, taking advantage of the already-calculated indexing of mutually recursive references.

@ceedubs I think this should actually be backwards compatible, and forwards compatible _as long as_ you don't try to verify the hashes. The runtime doesn't really care what references are used to refer to things. They're just unique blobs for referring to code objects as far as it's concerned. So, if a new version receives code from an old version of unison, it should still be able to load and run it. It's just that if you try to check the hashes, they won't match what is expected. And similarly, if you send these new hashes to old versions of unision, they should load and run (and it can't possibly be checking the hashes). You might just have the same code associated with two different hashes in your cache.

I actually don't think there can even be hash collisions between the two, because the new hashes have a prepended version byte, so they're a different length.

And yes, `Right []` is the successful return value. `Left []` should probably never happen.

dolio:
These new tests check some of the failure conditions of verification, as well as adding more tests of the verifier.

I'm not sure what other tests to add. I could try coming up with more elaborate examples to check for verification, but I'm not sure it'd exercise anything that isn't already being done by the ones I've already added.

pchiusano:
@dolio cool, LGTM, merging!

PR #4267, opened 2023-08-17
Round tripping code now yields same hash after reparsing in all known cases

pchiusano:
The new test harness in #4217 revealed that there were numerous definitions that didn't round trip with the same hash. Which isn't ideal as it leads to spurious updates if you happen to bring such a definition into your scratch file but don't update it.

This PR fixes all known instances of this and adds regression tests. If you find a case where a definition round trips with a different hash, that's a bug (and it can be added as a pending example to `reparses.u`)

## Implementation notes

This works by normalizing type signatures in a couple ways:

* Ability variables in strictly positive position are eliminated. So `a ->{g} b ->{g2} c` becomes `a ->{} b ->{} c`. This was being done previously, but there was a bug where the `{}` were being eliminated by one code path, but not others.
* Order of quantifiers in `forall` is normalized based on where the variables appear. So `forall b a . a -> b -> (a,b)` becomes the signature `forall a b . a -> b -> (a,b)` (since `a` appears first in the traversal of the body of the forall). Not doing this caused some code paths (inference vs partial inference vs checking) to get different quantifier order, which affects the hash.

Empty ability lists `{}` are elided during pretty-printing of types. So you can write `{}` in a signature, and the typechecker will enforce it, but it won't show up in any printed type. This is what we've done previously, but there were some code paths I don't fully understand where you could get the `{}` to show up in a printed type. In any case, with this PR, we ignore `{}` uniformly during type printing. Adding the implied `{}` everywhere is noisy, and eliding it also makes it easy to do a refactoring where you add effects to a function.

There are still many definitions out in the wild which work fine, but if you bring them into your scratch file, they'll typecheck with a different hash. But after this PR, if you do that once, the next time you do it, it'll get the same hash.

## Test coverage

The round trip tests are pretty good.

## Loose ends

This work has made me realize that our type-based search index is much less useful due to the sorts of irrelevant differences that this PR normalizes away. But from #4046 and seeing how fast it is to scan the namespace, I suspect we could just get rid of the type index and instead do a linear scan over the namespace (and perhaps 1 level of `lib`), doing a check for equivalence and/or a subtype check.


parser-typechecker/src/Unison/Typechecker/Context.hs
```diff
@@ -544,6 +544,23 @@ debugShow :: (Show a) => a -> Bool
...
+
+debugType :: Var v => String -> Type.Type v a -> Bool
+debugType tag ty
+  | debugEnabled = debugTrace $ "(" <> show tag <> "," <> showType ty <> ")"
+  | otherwise = False
+
+debugTypes :: Var v => String -> Type.Type v a -> Type.Type v a -> Bool
+debugTypes tag t1 t2
+  | debugEnabled = debugTrace $ "(" <> show tag <> ",\n  " <> showType t1 <> ",\n  " <> showType t2 <> ")"
+  | otherwise = False
```
pchiusano, line 562: @dolio these might be handy next time you're debugging something. `TypePrinter.prettyStr` which is used here does no normalization, so it just prettily prints exactly what's there.

unison-src/transcripts-round-trip/main.md
```diff
@@ -72,17 +72,20 @@ x = ()
...
 .a3> delete.namespace.force lib.builtin
+.a3_old> delete.namespace.force lib.builtin
 ```
 
 These are currently all expected to have different hashes on round trip, though we'd prefer if they round tripped with the same hash.
 
-```ucm
-.> diff.namespace a1 a3
+NOTE, since we don't currently have anything that round trips with a different hash, this fails. If you find an example that reparses with a different hash, add it to `reparses.u` and change this stanza to `ucm` rather than `ucm:error`.
+
```
pchiusano, line 86: 🎉

unison-src/transcripts-round-trip/reparses-with-same-hash.u
```diff
@@ -258,3 +255,208 @@ raw_d = """
...
+  a = "asldkfj"
+  b = "asdflkjasdf"
+  do do run runrun do do runrun
+
+-- Regression test for https://github.com/unisonweb/unison/issues/4239
+-- `n` was replaced by `error` but should not be. Instead, render as if
+-- a second param, _, had been provided in the definition. 
+(>>>>) : Nat -> Nat -> ()
+(>>>>) n = cases
+  _ -> bug ""
```
pchiusano, line 462: Moved over from `reparses.u` as these definitions now reparse with the same hash.

unison-src/transcripts-using-base/serial-test-00.output.md
```diff
@@ -80,7 +80,7 @@ mkTestCase = do
                    -> (Tree Text -> Text)
                    -> (Tree Nat, Tree Nat, Tree Nat, Tree Text)
                    -> Text
-      foldMap    : ∀ a r g2 g1 g.
+      foldMap    : ∀ r g2 g1 a g.
                      r
                      -> (r ->{g2} r ->{g1} r)
                      -> (a ->{g} r)
```
pchiusano, line 86: Examples here showing the forall order normalization. Note that the order doesn't matter most of the time and won't even be shown unless an explicit `forall` is needed during pretty-printing.

unison-src/transcripts/alias-many.output.md
```diff
@@ -701,7 +701,7 @@ Let's try it!
   511. Universal.>= : a -> a -> Boolean
   512. Universal.compare : a -> a -> Int
   513. Universal.murmurHash : a -> Nat
-  514. unsafe.coerceAbilities : (a ->{e1} b) -> a ->{e2} b
+  514. unsafe.coerceAbilities : (a ->{e1} b) -> a -> b
```
pchiusano, line 704: Interesting quirk. The `{e2}` is there, but since it's in strictly positive position, it's elided during printing.

unison-src/transcripts/fix1063.output.md
```diff
@@ -16,7 +16,7 @@ noop = not . not
   
     ⍟ These new definitions are ok to `add`:
     
-      .    : ∀ o g1 i1 g i.
+      .    : ∀ i1 g1 o i g.
```
pchiusano, line 19: Another example showing quantifier normalization.

pchiusano (commented):
Self review

pchiusano:
Note: CI is actually passing save for the Nix caching targets, which @tstat will look at when he's back.

parser-typechecker/src/Unison/Typechecker/Context.hs
dolio: Nice. That does sound handy.

dolio (approved):
Looks sensible to me.

parser-typechecker/src/Unison/Typechecker/Context.hs
aryairani: cc @unisonweb/ucm

parser-typechecker/src/Unison/PrintError.hs
```diff
@@ -281,7 +281,7 @@ renderTypeError e env src curPath = case e of
         debugSummary note
       ]
   FunctionApplication {..} ->
-    let fte = Type.removePureEffects ft
+    let fte = Type.removePureEffects False ft
```
aryairani, line 284: Boolean blindness here; recommend a newtype

aryairani (approved):

parser-typechecker/src/Unison/PrintError.hs
pchiusano: I didn't entirely agree on this one, so leaving it as is. :) My thinking is: if you saw `DontKeepEmptied` here instead of `False`, are you going to know what that means without consulting the docs on `removePureEffects`? Adding a newtype which is literally only used in one function like this feels overkill, and it's another level of indirection to follow (to understand it, you now need to look at the docs on the type, and then see how it's used by the function - easier to just look at the docs on the function).

The places I'd newtype a boolean are if the concept crops up again and again, or perhaps if I've got a function that takes two or more booleans and it'd otherwise be error prone keeping them straight. Or perhaps if it's a very public facing function (though I'd probably introduce a pair of functions with a common implementation before newtyping a boolean for just the one case).

PR #4275, opened 2023-08-21
bump haskell.nix to support new lsp

tstat:
add iohk cache but remove macos due to codesigning bug

### DIFF
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -20,7 +20,6 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
-          - macOS-12
     steps:
     - uses: actions/checkout@v3
     - uses: cachix/install-nix-action@v22
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -23,6 +23,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: cachix/install-nix-action@v22
+      with:
+        extra_nix_config: |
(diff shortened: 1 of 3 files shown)


ceedubs (approved):
Thanks!

If there is a code signing bug, then should iohk be removed from the flake instead of just preventing this GitHub Action from running on macos? Isn't this going to be an issue if devs build locally with macOS?

tstat:
> If there is a code signing bug, then should iohk be removed from the flake instead of just preventing this GitHub Action from running on macos? Isn't this going to be an issue if devs build locally with macOS?

The IOHK cache was removed from the flake on trunk when @aryairani first observed the codesigning bug, and this PR doesn't change that, but the github action can't really survive without the cache when we bump the haskellNix input since the action runner doesn't have sufficient time or disk space to compile ghc and all of our other dependencies when we bump the haskell.nix flake input. 

We ran into this last week because a version of lsp newer than our haskell.nix flake pin was added to the stack.yaml. Unfortunately, bumping the haskell.nix flake input included changes we don't care about like haskell.nix bumping their nixpkgs input, which (unfortunately, I think) tracks nixpkgs-unstable, so things like coreutils changed which means we no longer got a cache hit for ghc and had to recompile ghc-9.2.8 instead of downloading the IOHK cached ghc.

So, this PR adds the IOHK cache _to the github action workflow_, but not the nix flake, and removes intel mac from the workflow. Linux workflows should be fixed, but intel and aarch64 mac don't have a cachix populated by CI. So, the flake does not add the IOHK cache for any local builds, but CI should mirror the required IOHK deps to the unison cachix for x86_64-linux.

PR #4287, opened 2023-08-28
Implement verified hashing for racket backend, with assorted changes

dolio:
This PR implements the non-unison changes necessary to have verified hashing in the racket backend. This includes:

- Rewriting and fleshing out some primop stuff. Some stuff that was previously defined in a monolithic macro has been broken apart and rewritten using racket's `match` for example. A few new switch-out pseudo-builtins have been implemented that new unison code relies on, too. The dynamic code loading was also not generating all the information expected for subsequently reflecting values.
- Updating test cases to base 2.5.0. Some function names have changed and such. I needed the new version for some graph functionality it has.
- Adding some test cases for the hashing. The dynamic code loading tests now call the `validateLinks` function to check that the code being loaded makes sense. I also added in the hash part of the tests for the serialized code, because the builtins for hashing arbitrary unison values are hooked up now.

I think even though there aren't a lot of test cases, they're actually covering the relevant issues, because I had to fix a bug with sorting the SCCs while working on this (one of the serialized cases has a non-trivial SCC).

### DIFF
--- a/scheme-libs/racket/unison/boot.ss
+++ b/scheme-libs/racket/unison/boot.ss
@@ -41,6 +41,12 @@
   describe-value
   decode-value
 
+  reference->termlink
+  reference->typelink
+  referent->termlink
+  typelink->reference
+  termlink->referent
+
   unison-tuple->list)
 
 (require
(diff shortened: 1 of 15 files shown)


pchiusano (approved):
Nice!

PR #4299, opened 2023-08-30
refactor: eliminate duplicate ShortHash type

mitchellwrosen:
## Overview

This PR merges together `unison-codebase/U.Codebase.ShortHash` and `unison-core1/Unison.ShortHash` into `unison-core/Unison.ShortHash`.

The two types were not quite the same: in the older one, component and constructor ids were modeled with `Maybe Text`, and in the newer, `Maybe Word64`. So, this PR also includes a little bit of cleanup and eliminating `error` calls that were related to code assuming those `Text`s contained `Word64`s.

The old short hashes encoded whether or not a constructor is part of an effect or normal data type in the actual suffix (e.g. `a3` or `d0`), but that information is not available in the new short hash type.

## Test coverage

No additional tests, I just tweaked the ones that were failing due to having component ids like `y10` (now `10`), and constructor ids like `a3` (now `3`). The existing tests cover this stuff pretty well, I think.


codebase2/codebase/U/Codebase/Reference.hs
```diff
@@ -3,11 +3,11 @@ module U.Codebase.Reference where
 import Control.Lens (Lens, Prism, Traversal, lens, prism)
 import Data.Bifoldable (Bifoldable (..))
 import Data.Bitraversable (Bitraversable (..))
-import U.Codebase.ShortHash (ShortHash)
-import U.Codebase.ShortHash qualified as SH
 import Unison.Hash (Hash)
 import Unison.Hash qualified as Hash
 import Unison.Prelude
+import Unison.ShortHash (ShortHash)
+import Unison.ShortHash qualified as SH
```
aryairani, line 10: If we move Reference to unison-core, we should move these helpers to the ShortHash module

unison-share-api/src/Unison/Server/Doc.hs
```diff
@@ -238,7 +237,7 @@ renderDoc pped doc = renderSpecial <$> doc
     renderSrc srcs =
       srcs & foldMap \case
         EvaluatedSrcDecl srcDecl -> case srcDecl of
-          MissingDecl r -> [(Type (Reference.toText r, DO.MissingObject (SH.unsafeFromText $ Reference.toText r)))]
```
aryairani, line 241: yikes

unison-syntax/test/Main.hs
```diff
@@ -214,3 +220,6 @@ t s expected =
             note $ "expected: " ++ show expected
             note $ "actual  : " ++ show actual
             crash "actual != expected"
+
+instance IsString ShortHash where
```
aryairani, line 224: nice

codebase2/codebase/U/Codebase/Reference.hs
mitchellwrosen: Which helpers?

aryairani: I was thinking like switching `Reference.fromShortHash` to `ShortHash.fromReference`, wdyt

aryairani (approved):

Issue #4194, opened 2023-07-12
delete help should mention varargs


ceedubs:
`delete.term`, `delete.type`, and `delete` all seem to support varargs like `foo bar`, `1 2`, or `1-2`, but their help output doesn't mention it:

```
cloud/cloudability2_cody> help delete.term

  delete.term
  `delete.term foo` removes the term name `foo` from the namespace.
```

This should be documented.

sixfourtwelve:
Happy to take, forgot to mention before creating a PR :)

sixfourtwelve:
PR Merged, my first and hopefully of many ✨

Probably worth closing the ticket @ceedubs ?

PR #4302, opened 2023-09-05
Add extra help line for delete command

sixfourtwelve:
## Overview

Provides one extra help info for the delete command

Before:
```bash
.> help delete

  delete
  `delete foo` removes the term or type name `foo` from the namespace.
```

After: 
```bash
.> help delete

  delete
  `delete foo` removes the term or type name `foo` from the namespace.
  `delete foo bar` removes the term or type name `foo` and `bar` from the namespace
```

## Implementation notes

## Test coverage
Manual testing

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -697,8 +697,9 @@ deleteGen suffix target mkTarget =
           (P.sep
             " "
             [ backtick (P.sep " " [P.string cmd, "foo bar"]),
+            "removes the",
             P.string target,
-            "remotes the name `foo` and `bar` from the namespace" ], "")]
+            "name `foo` and `bar` from the namespace" ], "")]
       warn =
         P.sep
           " "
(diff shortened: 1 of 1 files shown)


sixfourtwelve:
Fixes https://github.com/unisonweb/unison/issues/4194

ceedubs (approved):
Thank you! 🎉

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -686,13 +686,20 @@ deleteGen :: Maybe String -> String -> ([Path.HQSplit'] -> DeleteTarget) -> Inpu
...
             P.string target,
             "name `foo` from the namespace."
-          ]
+          ], ""),
+          (P.sep
+            " "
+            [ backtick (P.sep " " [P.string cmd, "foo bar"]),
+            "removes the",
+            P.string target,
+            "name `foo` and `bar` from the namespace" ], "")]
```
aryairani:
```suggestion
            "name `foo` and `bar` from the namespace." ], "")]
```

PR #4304, opened 2023-09-05
Add logic to flip welcome hint message

### LINKED ISSUE #4266: UCM Welcome message: don't show "project.create" message when there are projects
When a user already has projects, we shouldn't show this message in the welcome flow:

CleanShot 2023-07-16 at 12 39 48@2x.

Bonus: we could instead show something like `Hint: type 'projects' to list all your projects`—perhaps a hint cycling thing.

sixfourtwelve:
## Overview
Added some logic to check whether any projects are present in UCM, if so flip a switch and show a hint

Before:
```bash
  👋 Welcome to Unison!

  You are running version: e7b891914


  📚 Read the official docs at https://www.unison-lang.org/learn/

  Type 'project.create' to get started.
```

After:

```bash
  👋 Welcome to Unison!

  You are running version: e7b891914


  📚 Read the official docs at https://www.unison-lang.org/learn/

  Hint: type 'projects' to list all your projects
```

## Test Coverage
Manual testing

Fixes: https://github.com/unisonweb/unison/issues/4266

### DIFF
--- a/unison-cli/src/Unison/CommandLine/Welcome.hs
+++ b/unison-cli/src/Unison/CommandLine/Welcome.hs
@@ -15,7 +15,7 @@ import Prelude hiding (readFile, writeFile)
 data Welcome = Welcome
   { onboarding :: Onboarding, -- Onboarding States
     unisonVersion :: Text,
-    welcomeHint :: Bool
+    showWelcomeHint :: Bool
   }
 
 -- Previously Created is different from Previously Onboarded because a user can
@@ -36,11 +36,11 @@ data Onboarding
   deriving (Show, Eq)
 
 welcome :: CodebaseInitStatus -> Text -> Bool -> Welcome
-welcome initStatus unisonVersion welcomeHint =
-  Welcome (Init initStatus) unisonVersion welcomeHint
+welcome initStatus unisonVersion showWelcomeHint =
+  Welcome (Init initStatus) unisonVersion showWelcomeHint
 
 run :: Welcome -> [Either Event Input]
-run Welcome {onboarding = onboarding, unisonVersion = version, welcomeHint = welcomeHint} = do
+run Welcome {onboarding = onboarding, unisonVersion = version, showWelcomeHint = showWelcomeHint} = do
   go onboarding []
   where
     go :: Onboarding -> [Either Event Input] -> [Either Event Input]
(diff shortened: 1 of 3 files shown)


unison-cli/src/Unison/CommandLine/Welcome.hs
```diff
@@ -14,7 +14,8 @@ import Prelude hiding (readFile, writeFile)
 
 data Welcome = Welcome
   { onboarding :: Onboarding, -- Onboarding States
-    unisonVersion :: Text
+    unisonVersion :: Text,
+    showWelcomeHint :: Bool
```
sixfourtwelve, line 18: I'm 50/50 on this being a `Bool` and not a `Text` and in `launch` we just pass in the appropriate hint message?

unison-cli/src/Unison/CommandLine/Welcome.hs
```diff
@@ -110,9 +111,12 @@ authorSuggestion =
...
-getStarted :: P.Pretty P.ColorText
-getStarted =
+getStarted :: Bool -> P.Pretty P.ColorText
+getStarted showWelcomeHint =
   P.wrap "📚 Read the official docs at https://www.unison-lang.org/learn/"
     <> P.newline
     <> P.newline
-    <> P.wrap "Type 'project.create' to get started."
+    <> P.wrap (if showWelcomeHint
+              then "Hint: type 'projects' to list all your projects."
```
aryairani: What do you think about:
```suggestion
              then "Hint: Type 'projects' to list all your projects, or 'project.create' to start something new."
```

sixfourtwelve: Yeah, I prefer that. Probably should've thought of that first 😁. Thanks Arya

unison-cli/src/Unison/CommandLine/Welcome.hs
aryairani: Yeah, we can always change it later if we have some reason to skew off of 50/50.  It could also be a multi-valued thing to choose between instead of the two Bool values.  I guess, at a point, `Text` is the extreme version of multi-valued ;-)

aryairani (approved):

aryairani:
Thanks @sixfourtwelve!

PR #4310, opened 2023-09-07
refactor: one reference type

mitchellwrosen:
## Overview

This PR makes the `Reference` type defined in `Unison.Reference` just a type alias for the `Reference` defined in `U.Codebase.Reference`.

One snag I encountered: the deleted `Reference` type had a naughty hand-written `Show` instance, whereas the other has a derived `Show` instance. I tried to read over wherever we were using the old `Show` instance, and couldn't find any places that relied on it being prettified - it's just occasionally used in debug error messages (in which the derived instance is fine). All the existing tests and transcripts pass.


aryairani:
@mitchellwrosen Can you doublecheck how enlil reacts to this

aryairani (approved):

mitchellwrosen:
@aryairani I think it should be fine per slack conversation from last week, but I can try building. If it fails, it'll be due to a direct pattern match on `Reference` data constructors, which are in `Reference'` types with slightly different names. Was there anything specific you think we should verify?

mitchellwrosen:
@aryairani Ok, I confirmed enlil builds without issue

PR #4312, opened 2023-09-07
Prefer triple backtick syntax during doc source printing, even for old docs

pchiusano:
There are old docs in the wild that use a lambda like `(_ -> 1 + 1)` for the delayed block inside triple backticks. Due to an oversight from #4217 (I missed one spot that should be using the `DDelay` pattern), this is currently resulting in rendering like this when viewing old docs:

CleanShot 2023-09-07 at 17 15 57@2x

(Note: this only affects their source, not their rendered form, which still looks fine)

This PR is a 1-character change in the pretty-printer plus a regression test. Thanks to @runarorama for reporting.

### DIFF
--- a/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
@@ -1931,7 +1931,7 @@ toDocVerbatim ppe (App' (Ref' r) (toDocWord ppe -> Just txt))
 toDocVerbatim _ _ = Nothing
 
 toDocEval :: (Var v) => PrettyPrintEnv -> Term3 v PrintAnnotation -> Maybe (Term3 v PrintAnnotation)
-toDocEval ppe (App' (Ref' r) (Delay' tm))
+toDocEval ppe (App' (Ref' r) (DDelay' tm))
   | nameEndsWith ppe ".docEval" r = Just tm
   | r == _oldDocEval = Just tm
 toDocEval _ _ = Nothing
--- a/unison-src/transcripts-round-trip/main.output.md
+++ b/unison-src/transcripts-round-trip/main.output.md
@@ -641,19 +641,34 @@ x = ()
       Put definitions in here that are expected to
       parse with a different hash after pretty-printing.
       """
+    
+    sloppyDocEval : Doc2
+    sloppyDocEval =
+      use Nat +
+      {{
+      Here's an example of an eval block that's technically a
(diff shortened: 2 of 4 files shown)


parser-typechecker/src/Unison/Syntax/TermPrinter.hs
```diff
@@ -1931,7 +1931,7 @@ toDocVerbatim ppe (App' (Ref' r) (toDocWord ppe -> Just txt))
 toDocVerbatim _ _ = Nothing
 
 toDocEval :: (Var v) => PrettyPrintEnv -> Term3 v PrintAnnotation -> Maybe (Term3 v PrintAnnotation)
-toDocEval ppe (App' (Ref' r) (Delay' tm))
+toDocEval ppe (App' (Ref' r) (DDelay' tm))
```
pchiusano, line 1934: This is the fix.

unison-src/transcripts-round-trip/main.output.md
```diff
@@ -641,19 +641,34 @@ x = ()
...
+      use Nat +
+      {{
+      Here's an example of an eval block that's technically a
+      lambda but should print as a backticked block (since old
+      docs in the wild still use this format).
+      
+      ```
+      1 + 1
+      ```
+      }}
```
pchiusano, line 656: Here we see it using the backtick syntax even though the input doc had a lambda with a `_` parameter.

unison-src/transcripts-round-trip/main.output.md
```diff
@@ -641,19 +641,34 @@ x = ()
...
 
 ```ucm
 .> diff.namespace a3 a3_old
 
-  The namespaces are identical.
+  Updates:
+  
+    1. sloppyDocEval : Doc2
+       ↓
+    2. sloppyDocEval : Doc2
```
pchiusano, line 671: @runarorama FYI, we can now add pretty-print round trip tests by adding regular Unison code to `reparses.u` or `reparases-with-same-hash.u`. 

Everything in `reparses-with-same-hash.u` has to round trip with the same hash or the test fails, and everything in `reparses.u` has to at least parse and typecheck on round trip, but can have a different hash and still pass.

No transcript changes needed, just run `stack exec unison -- transcript unison-src/transcripts-round-trip/main.md` and it will pick up all the stuff in those `.u` files.

aryairani (approved):
The normal representation for triple backticked is a lambda I guess, and then it gets forced when docs are rendered?

PR #4318, opened 2023-09-09
update delete cli commands

### LINKED ISSUE #4315: Update delete completion for types, terms and their verbose variants
Similar to `deleteProject`, its probably worth while removing deleteGen and having type, term and verbose commands be their own functions.

This should negate the issue in https://github.com/unisonweb/unison/issues/4190 which is a issue for  the delete commands that use deleteGen

sixfourtwelve:
## Overview

Updating the delete commands from being a single deleteGen that has caused issues with completion by adding extra characters such as a `.` after pressing <tab>. Hopefully this draft PR, will provide a solution not to just negate that, but to also separate the individual concerns into their own functions same as other delete functions are done

## Test coverage

Manual & Transcripts, the latter I am yet too write :)

Fixes: https://github.com/unisonweb/unison/issues/4315

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -726,7 +726,7 @@ delete =
     { patternName = "delete",
       aliases = [],
       visibility = I.Visible,
-      argTypes = [(Required, exactDefinitionTermQueryArg)],
+      argTypes = [(Required, exactDefinitionArg)],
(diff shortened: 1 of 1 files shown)


sixfourtwelve:
Next Steps: Fix issue with help not showing on empty commands, and add transcript tests

sixfourtwelve:
Miss click, not actually ready in for review

sixfourtwelve (commented):
.

PR #4323, opened 2023-09-13
Fix compile error in benchmarks

pchiusano:
Also tweaked some of the constants so some of the benchmarks take longer.

### DIFF
--- a/unison-src/transcripts-manual/benchmarks/array-sort.u
+++ b/unison-src/transcripts-manual/benchmarks/array-sort.u
@@ -10,6 +10,7 @@ cmp m n = match compare m n with
   -1 -> Less
   +0 -> Equal
   +1 -> Greater
+  _ -> Less
 
 main = do
   list = map shuffle <| range 1000 2000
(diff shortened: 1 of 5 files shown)


unison-src/transcripts-manual/benchmarks/array-sort.u
```diff
@@ -10,6 +10,7 @@ cmp m n = match compare m n with
   -1 -> Less
   +0 -> Equal
   +1 -> Greater
+  _ -> Less
```
pchiusano, line 13: with pattern coverage checking, this case is now needed (though it will never be hit in practice)

unison-src/transcripts-manual/benchmarks/tmap.u
```diff
@@ -13,12 +13,10 @@ main = do
   _a7 = timeit "TVar Map concurrent (4)" '(gos_TVar_Map 4 (atomically '(TVar.new Map.empty)) N)
   _a8 = timeit "TVar Map concurrent (8)" '(gos_TVar_Map 8 (atomically '(TVar.new Map.empty)) N)
   _a9 = timeit "TVar Map concurrent (16)" '(gos_TVar_Map 16 (atomically '(TVar.new Map.empty)) N)
-  _a10 = timeit "TVar Map concurrent (32)" '(gos_TVar_Map 32 (atomically '(TVar.new Map.empty)) N)
```
pchiusano, line 16: this benchmark was taking too long and we weren't seeing any nonlinear slowdown with more concurrency.

dolio (approved):
Looks fine to me.

PR #4324, opened 2023-09-13
Allow multiple ignored bindings in a block

### LINKED ISSUE #3773: Allow multiple (ignored) bindings called `_` in a block

pchiusano:
Previously, you couldn't do:

```Haskell
foo = 
  _ = 1
  _ = 22
  42
```

It would complain about duplicate bindings.

Now you can. This PR allows duplicate bindings as long as they're called `_`. There's no syntax for referring to a variable called `_` so this seems to work fine in the typechecker and runtime. Added a regression test.

@dolio tagging you in case you can think of issues with this approach. Like will the bytecode compiler complain or somesuch?

Addendum: also did some manual testing with the `compile` command:

```
foo = do
  _ = 
    printLine "hi"
    1
  _ = 
    printLine "there"
    2
  999 
```

I did a `compile` and then `unison run.compiled foo.uc` and it worked fine. (I thought I'd double check this)

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Components.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Components.hs
@@ -12,6 +12,7 @@ import Unison.Prelude
 import Unison.Term (Term')
 import Unison.Term qualified as Term
 import Unison.Var (Var)
+import Unison.Var qualified as Var
 
 unordered :: (Var v) => [(v, Term' vt v a)] -> [[(v, Term' vt v a)]]
 unordered = ABT.components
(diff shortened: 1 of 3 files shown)


dolio (approved):
I can't think of a reason it'd cause problems. It seems like it'd just cause some shadowing, and if the variables aren't even used, that doesn't really matter.

PR #4326, opened 2023-09-19
Add missing dependency crawling for MCode

dolio:
The changes to the the intermediate code formats added additional branching constructs to MCode. But, the code that figures out the dependencies of MCode wasn't updated to take these into account, so any branches using these were treated as having no dependencies (due to the default case).

I added some structure to the integration test for generating a compiled binary to catch this case.

If @stew could test this branch with his cloud code that'd be appreciated. For some reason I'm unable to actually run his compiled binary locally, and I'm not sure why.


pchiusano (approved):
@stew confirmed that this fixes things for him

PR #4327, opened 2023-09-19
refactor: replace runEitherTransaction with more ergonomic runTransactionWithRollback

mitchellwrosen:
## Overview

This PR replaces the somewhat clunky

```haskell
runEitherTransaction :: Transaction (Either Output a) -> Cli a
```

with a more ergonomic

```haskell
runTransactionWithRollback
  :: (forall void. Output -> Transaction void)
  -> Transaction a
  -> Cli a
```

so that you can use a nicer short-circuiting syntax within transactions, rather than reach for `ExceptT`/`lift`.

### DIFF
--- a/lib/unison-sqlite/src/Unison/Sqlite/Transaction.hs
+++ b/lib/unison-sqlite/src/Unison/Sqlite/Transaction.hs
@@ -83,7 +83,7 @@ runTransaction conn (Transaction f) = liftIO do
         pure result
 {-# SPECIALIZE runTransaction :: Connection -> Transaction a -> IO a #-}
 
--- An internal exception type that allows powers `runTransactionWithRollback`
+-- An internal exception type that allows `runTransactionWithRollback`
 data RollingBack
   = forall a. RollingBack !Unique !a
   deriving anyclass (Exception)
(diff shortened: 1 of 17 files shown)


aryairani:
I'm guessing with `Cli` this doesn't affect Enlil; or does it?

lib/unison-sqlite/src/Unison/Sqlite/Transaction.hs
```diff
@@ -80,6 +83,30 @@ runTransaction conn (Transaction f) = liftIO do
         pure result
 {-# SPECIALIZE runTransaction :: Connection -> Transaction a -> IO a #-}
 
+-- An internal exception type that allows powers `runTransactionWithRollback`
```
aryairani:
```suggestion
-- An internal exception type that allows `runTransactionWithRollback`
```
```suggestion
-- An internal exception type that powers `runTransactionWithRollback`
```

aryairani (approved):

mitchellwrosen:
> I'm guessing with Cli this doesn't affect Enlil; or does it?

It does not

PR #4331, opened 2023-09-22
Add kind inference

tstat:
## Overview

Implements Haskell 98 style kind inference for Unison.

## Implementation notes

All kind inference code is at or under `Unison.KindInference`. `Unison.KindInference` gives an introduction and contains the exposed interface. `Unison.KindInference.Error.Pretty` contains the user facing pretty printer for errors.

## Test coverage

There is a new `kind-inference.md` transcript, and all of the other transcripts are kind-checked now of course.

## Loose ends

Lower-case ability declarations are accepted but generate broken constructor types. Details in this commit message: https://github.com/unisonweb/unison/pull/4331/commits/591d210bdd52ad6b501db3d77af10ab4ecf529be. I removed these examples from a test since they are broken on trunk, but they are broken in a way that fails kind checking. When this is fixed we can revert that commit.

Additionally, `@rewrite signature foo ==> bar` generates an application to a builtin decl constructor of type `forall a b. (a -> b -> ())-> RewriteSignature a b`, so it is often not well-kinded (when foo or bar are not of kind `*`). I don't know why this constructor takes an argument, it seems like it should just require two phantom type variables. Even if this is fixed though, `a` and  `b` are meant to be polykinded, something like `data RewriteSignature (a :: k) (b :: k) = RewriteSignature` in haskell. So, handling it correctly will require some special casing or a more powerful kind system. Right now the generated annotation is removed before kind checking, so `foo` and `bar` in the above example are not kind checked. (https://github.com/unisonweb/unison/pull/4331/commits/d0bd100ad71383e06007c409f1bdaa68f1c25088)


pchiusano (approved):
Awesome!

@dolio should probably have a look at this.

Also, this is still marked as a draft, is that intentional (like you're still working on things) or is it ready for review?

tstat:
I realized I forgot to update the LSP machinery.

unison-src/transcripts/kind-inference.output.md
```diff
@@ -0,0 +1,289 @@
...
+unique type Pong = Pong (Ping Optional)
+```
+
+```ucm
+
+  Kind mismatch arising from
+        1 | unique type Ping a = Ping a Pong
+    
+    The arrow type (->) expects arguments of kind *; however, it
+    is applied to a which has kind: * -> *.
```
aryairani: In a future ticket we should improve this message which mentions arrow though we don't see any.

aryairani (approved):
Looks awesome! Could you open tickets for the loose ends you mentioned. Maybe one generally for beginner-friendlying the kind error messages too.

dolio (approved):
I think this looks pretty good. I didn't notice any serious problems browsing through it.

One general question I have is: do we want the kind of types to be called `*`? That's traditional, but maybe `Type` would be better, since the other base kind is `Effect`. And GHC (for instance) is sort of moving toward alphabetical names as well.

tstat:
Good question @dolio. `Type` probably would be less confusing for people unfamiliar with kinds. I'll defer to @pchiusano on that. 

I also named the kind of abilities `Effect` without much thought, but maybe calling that kind `Ability` would be easier.

pchiusano:
@tstat Yeah, I'd rename `*` to `Type` and use `Ability` as the kind instead of `Effect`. Other than that, I'd go ahead and merge when ready! Thanks for putting this together.

aryairani:
🎉

PR #4333, opened 2023-09-26
Add new BranchV3 type

mitchellwrosen:
## Overview

This PR adds a new `BranchV3` type (in `U.Codebase.BranchV3`) that fits the shape of the new low-level branches, post-merge. (This work is part of the new merge algorithm, but was extracted to a smaller PR).

The old branch type supports conflicted terms and decls, metadata on each naming of each term and decl, and has patches:

```haskell
data Branch m = Branch
  { terms :: !(Map NameSegment (Map Referent (m MdValues))),
    types :: !(Map NameSegment (Map Reference (m MdValues))),
    patches :: !(Map NameSegment (PatchHash, m Patch)),
    children :: !(Map NameSegment (CausalBranch m))
  }
  
type CausalBranch m = Causal m CausalHash BranchHash (Branch m) (Branch m)
```

The new branch type just has one term/decl per name, no metadata, and no patches:

```haskell
data BranchV3 m = BranchV3
  { children :: !(Map NameSegment (CausalBranchV3 m)),
    decls :: !(Map NameSegment TypeReference),
    terms :: !(Map NameSegment Referent)
  }

type CausalBranchV3 m = Causal m CausalHash BranchHash (Branch m) (BranchV3 m)
```

I also renamed `types` to `decls` here (though I don't think that matters much, and would be happy to change it back for consistency), as I believe `decls` is the right name for these things. Yet, we call them "types" in a lot of code, in except code that is concerned with actual types (i.e. `Unison.Type`) _and_ decls, in which case we usually call them "decls" (but also sometimes, confusingly, "types").

This PR also contains some minor hashing and database support (write-path only) for new V3 branches. We still don't store these things any differently on disk, so you'll always read a V2 branch out. That can come another time, or never.

Additionally, the `Causal` type was generalized slightly to support the shape of both V2 branches and V3 branches, since they're different:

- A V2 causal has a V2 branch at the head of the history, and V2 causal ancestors.
- A V3 causal has a V3 branch at the head of the history, and V2 causal ancestors (some of which may actually be V3, but again, we don't know or care about that, because V3 branches are not different once stored).

### DIFF
--- a/codebase2/codebase-sqlite-hashing-v2/package.yaml
+++ b/codebase2/codebase-sqlite-hashing-v2/package.yaml
@@ -9,6 +9,7 @@ dependencies:
   - bytes
   - bytestring
   - containers
+  - generic-lens
   - lens
   - text
   - unison-codebase
@@ -49,6 +50,7 @@ default-extensions:
   - LambdaCase
   - MultiParamTypeClasses
   - NamedFieldPuns
+  - OverloadedLabels
   - OverloadedStrings
   - PatternSynonyms
   - RecordWildCards
--- a/codebase2/codebase-sqlite-hashing-v2/src/U/Codebase/Branch/Hashing.hs
+++ b/codebase2/codebase-sqlite-hashing-v2/src/U/Codebase/Branch/Hashing.hs
@@ -1,10 +1,20 @@
-module U.Codebase.Branch.Hashing (hashBranch) where
+module U.Codebase.Branch.Hashing
+  ( hashBranch,
+    hashBranchV3,
+  )
+where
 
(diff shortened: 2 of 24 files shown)


aryairani:
> renamed types to decls

tempting to call it decls, but what do we call it when the TypeReference is a type reference that's not a decl, e.g. ##Nat?

mitchellwrosen:
> tempting to call it decls, but what do we call it when the TypeReference is a type reference that's not a decl, e.g. ##Nat?

Oh, yeah. Good call. Hmm... not seeing a great alternative... I'll change that back.

aryairani (approved):

Issue #4335, opened 2023-09-27
Change decompilation strategy to allow partial decompilation


pchiusano:
We do decompilation for various things like printing watch expressions, trace statements, and exceptions that bubble to top level.

Right now, decompilation can fail (it returns an `Either Error Term`) which is not a great experience if you're printing out a big value and one of its sub-expressions is something we don't know how to decompile (like a `TVar`, say).

We'd like to change this to allow partial decompilation, so decompilation would return a `(Set Error, Term)` and anything that couldn't be decompiled would be added to the errors and then the decompiled result for that sub-term can just be a string literal like `"<continuation>"` or `"<mvar>"` or `"<recursive-function>"` or whatever.

For now the output can be unchanged, but we may want to later indicate to the user that the decompilation was partial.

PR #4336, opened 2023-09-29
Restructure decompilation

dolio:
The Runtime API has been changed so that 'partial' decompilation is possible. Even if a term result is returned, a set of errors may be yielded indicating that the entire result value could not be decompiled.

In cases with errors, the result will not be cached, although it allows displaying a proper result.

The decompiler in the runtime now takes advantage of this, in spots that it doesn't know how to decompile with calls to `bug`. These should always be valid terms, but obviously won't work correctly, hence the cache avoidance.

I added a test case in the watch expressions test that builds an array. Since there are no array literals, this would have failed before, but now it doesn't cause a hard error, just some warnings.


pchiusano (approved):
Nice!

dolio:
Fixes #4335

PR #4355, opened 2023-10-24
Fill out more scheme builtins

dolio:
This PR includes various builtin additions, mainly for scheme, the most significant of which are sandboxing related.

- `Value.dependencies` and `Code.dependencies` will be included, as they are auto-generated from unison, and will appear when the share version is incremented
- `Code.isMissing` has been implemented in scheme
- A system mirroring the sandbox checks has been implemented in scheme. This keeps track of sensitive builtins for each defined function, and looks them up to check if a value's evaluation should be allowed
- A couple new sandbox related builtins have been added:
    * `sandboxLinks` looks up the list of sensitive builtins associated with a termlink. This is used when statically generating code, to just look up the sandboxing information in the current runtime and add it to the file.
    * `Value.validateSandboxed` is a revised version of the sandboxing check. It allows checking the sandboxing status of a `Value` *before* it is loaded into an arbitrary unison value. This is a bit easier to check on the scheme side. It also has a different API, which can report all the builtins that caused the value to fail the sandbox check, rather than just a boolean.

Some tests are included, but I'm going to add a bit more, as I think there might be at least one corner case I missed on the scheme end.


parser-typechecker/src/Unison/Runtime/Machine.hs
```diff
@@ -388,17 +393,36 @@ exec !env !denv !_activeThreads !ustk !bstk !k _ (BPrim1 DBTX i)
...
-  s <- peekOffS bstk i
-  c <- peekOff bstk j
-  l <- decodeSandboxArgument s
-  b <- checkSandboxing env l c
-  ustk <- bump ustk
-  poke ustk $ if b then 1 else 0
-  pure (denv, ustk, bstk, k)
+exec !env !denv !_activeThreads !ustk !bstk !k _ (BPrim2 SDBX i j)
+  | sandboxed env =
+      die "attempted to use sandboxed operation: sandboxLinks"
```
pchiusano, line 398: Should the error message say "sandboxLinks" here or something else?

scheme-libs/racket/unison/sandbox.rkt
```diff
@@ -0,0 +1,149 @@
...
+(sandbox-builtin "Tls.encodeCert")
+(sandbox-builtin "Tls.decodePrivateKey")
+(sandbox-builtin "Tls.encodePrivateKey")
+(sandbox-builtin "Tls.receive.impl.v3")
+(sandbox-builtin "Tls.terminate.impl.v3")
+(sandbox-builtin "IO.randomBytes")
+(sandbox-builtin "IO.array")
+(sandbox-builtin "IO.arrayOf")
+(sandbox-builtin "IO.bytearray")
+(sandbox-builtin "IO.bytearrayOf")
```
pchiusano, line 166: I went ahead and double checked these to make sure they're consistent with what's in `Builtin.hs`.

pchiusano (approved):
LGTM, left one very minor comment. I'd say merge whenever you're ready.

parser-typechecker/src/Unison/Runtime/Machine.hs
dolio: The error message is correct, but I'd put it in the wrong case there. It was meant for `SDBL`.

PR #4358, opened 2023-10-26
improve rendering of uncaught exceptions

mitchellwrosen:
## Overview

This PR fixes this annoying output of uncaught errors:

```
Prelude.undefined
    error
```

The call stack in cases like this got stripped because we were using a `pShow` combinator from a pretty-printing library whose purpose is to nicely render derived Show instances, and I guess it doesn't handle stuff that doesn't look like derived Show instances very well.

The solution implemented here is to just use `displayException` instead.


ChrisPenner (approved):
Great find!

That's disappointing about pShow, but the callstack is WAAAAY more important in this case anyways.

aryairani (approved):

PR #4359, opened 2023-10-27
Delay kind defaulting

tstat:
## Overview

When inferring the kinds of data decls, unconstrained vars are defaulted to kind `Type`. This defaulting normally occurs after processing each strongly connected decl component. 

We have since learned that haskell 98 kinds are not sufficient for the `distributed` package, which expects some unconstrained phantom types to have kind `Ability`. Polymorphic kinds would be an ideal way to accommodate this. A kind signature is another possibility. We don't want to invest the time on polymorphic kinds right now, and we don't want to change the type by adding a kind signature, so this PR provides the desired behavior by delaying kind defaulting until all decls have been processed.

This change means that the kind of a decl with unconstrained vars may change depending on what dependents are in scope. This is ok for now, especially because we don't persist any kind information at this time. For a more permanent solution, we want to introduce polymorphic kinds.

## Test coverage

There is a new example in the kind-inference transcript.

### DIFF
--- a/unison-src/transcripts/kind-inference.md
+++ b/unison-src/transcripts/kind-inference.md
@@ -54,6 +54,13 @@ unique type T a = T a
 unique type S = S (T Nat)
 ```
 
+Demonstrate kind defaulting by component
+```unison:error
+unique type T a = T
+
+unique type S = S (T Optional)
+```
+
 Catch invalid instantiation of `T`'s `a` parameter in `S`
 ```unison:error
 unique type T a = T a
--- a/unison-src/transcripts/kind-inference.md
+++ b/unison-src/transcripts/kind-inference.md
@@ -54,7 +54,9 @@ unique type T a = T a
 unique type S = S (T Nat)
 ```
 
-Delay kind defaulting until all components are processed
+Delay kind defaulting until all components are processed. Here `S`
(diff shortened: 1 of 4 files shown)


unison-src/transcripts/kind-inference.md
```diff
@@ -54,6 +54,13 @@ unique type T a = T a
 unique type S = S (T Nat)
 ```
 
+Delay kind defaulting until all components are processed
+```unison:error
+unique type T a = T
+
+unique type S = S (T Optional)
+```
+
```
aryairani: Could you add a little extra explanation about what this stanza is showing?
I'm also confused because although it's marked as error, I don't see an error in the output?

pchiusano:
Nice @tstat looks good!

unison-src/transcripts/kind-inference.md
tstat: Sure, the `unison:error` was accidentally left from the preceding commit, when this was an error. I removed the `error` and added a comment.

aryairani (approved):
Great!

unison-src/transcripts/kind-inference.md
aryairani: Got it, thanks!

aryairani:
cc @stew

PR #4361, opened 2023-10-28
Add missing spaces to `pull` explanation

JohanWinther:
## Overview

Adds missing spaces to the explanation of all the `pull` commands.

Before:
```ucm
@unison/base/main> ? pull

  pull (or pull.silent)
  The`pull`command merges a remote namespace into a local namespacewithout listing the merged entities
```
After:
image

## Test coverage
No changes to tests needed.

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -78,3 +78,4 @@ The format for this list: name, GitHub handle
 * Hatim Khambati (@hatimkhambati26)
 * Kyle Goetz (@kylegoetz)
 * Ethan Morgan (@sixfourtwelve)
+* Johan Winther (@JohanWinther)
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -1210,11 +1210,11 @@ pullImpl name aliases verbosity pullMode addendum = do
           argTypes = [(Optional, remoteNamespaceArg), (Optional, namespaceArg)],
           help =
             P.lines
-              [ P.wrap
-                  "The "
-                  <> makeExample' self
-                  <> " command merges a remote namespace into a local namespace "
-                  <> addendum,
+              [ P.wrap $
+                  "The"
+                    <> makeExample' self
+                    <> "command merges a remote namespace into a local namespace"
+                    <> addendum,
                 "",
                 P.wrapColumn2
                   [ ( makeExample self ["@unison/base/main"],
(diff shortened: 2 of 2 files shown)


aryairani:
Hey @JohanWinther, thanks for noticing this and putting together a PR!

> Probably no changes to tests needed.

For this sort of thing, a screenshot or paste like what you have in the description is sufficient for testing.

Screenshot is better though — _because_ I think if you doublecheck you will find that this PR _doesn't_ produce exactly the output you have in the "After" section.

The bug you found is pretty subtle:
`P.wrap` is not supposed to require spaces, so it was odd that you needed them. In fact, it deletes spaces and adds its own, so it was odd that adding spaces would help! What's actually happening is that `P.wrap` is only being applied to the first argument, `"The"`, and so it does in fact delete the space that you tried to insert there, but fails to set up the spaces through the rest of the message.

The correct fix is change `P.wrap` to `P.wrap $`, or else put all of the text in parens. And then remove the extra spaces like it was originally. Do you wanna do this?

JohanWinther:
> Hey @JohanWinther, thanks for noticing this and putting together a PR!
> 
> > Probably no changes to tests needed.
> 
> For this sort of thing, a screenshot or paste like what you have in the description is sufficient for testing.
> 
> Screenshot is better though — _because_ I think if you doublecheck you will find that this PR _doesn't_ produce exactly the output you have in the "After" section.
> 

Yes, screenshot is better because I didn't actually paste in what the new version would output, I added what *should* be the result. I was too lazy to actually pull down the repo and figure out how to compile it and test it out, since I thought the change was trivial. Thanks for being diligent :)

> The bug you found is pretty subtle: `P.wrap` is not supposed to require spaces, so it was odd that you needed them. In fact, it deletes spaces and adds its own, so it was odd that adding spaces would help! What's actually happening is that `P.wrap` is only being applied to the first argument, `"The"`, and so it does in fact delete the space that you tried to insert there, but fails to set up the spaces through the rest of the message.
> 

I sort of understood that `P.wrap` should do this from the other uses of the function in the file, but couldn't really figure it out so I kind of ignored it.

> The correct fix is change `P.wrap` to `P.wrap $`, or else put all of the text in parens. And then remove the extra spaces like it was originally. Do you wanna do this?

I will try to do these changes and actually compile and test locally. Thanks for the explanation :+1:

JohanWinther:
@aryairani I've pushed the changes and updated the description to include a screenshot

aryairani:
Awesome!

aryairani:
Oh, @JohanWinther would you also add yourself to `CONTRIBUTORS.markdown` to acknowledge the license.

JohanWinther:
@aryairani Done! Sorry for the long wait, I missed the notification.

aryairani:
No worries, thanks!

aryairani:
Oh... I thought I had queued it to merge.

aryairani (approved):

PR #4365, opened 2023-10-31
Implement Murmurhash in racket

jaredly:
There are probably some perf optimizations that could be done if we need it.


dolio (approved):
Looks fine to me.

pchiusano (approved):
Nice. @dolio do you want to add to this PR the integration with the runtime? Or do you think we should merge as is and then do the integration with a separate PR?

dolio:
I'm just going to merge this, because I think the only associated change for the rest of the integration is going to be bumping the share code version.

dolio:
BTW, while I was hooking this up, I was a little skeptical of the implementation, because it turns every value into bytes, and hashes individual bytes. However, it actually matches what we do in Haskell, apparently. We're just using the `ByteString` instance after converting to bytes, and that's how it behaves, rather than using 64bit chunks.

Issue #4362, opened 2023-10-30
Exception when updating `data.List.foldRight` in @unison/base/main


jskripsky:
I'm using UCM version eb83404. 

# Exception
```
Encountered exception:
[malformed type] IntroOuter not in lexical scope of matching Forall
CallStack (from HasCallStack):
  error, called at src/Unison/KindInference/Generate.hs:145:18 in unison-parser-typechecker-0.0.0-6MGtlMk5aXuJWsQZ16HFt2:Unison.KindInference.Generate
```

# Unison Code
```Haskell
data.List.foldRight : (a ->{𝕖} b ->{𝕖} b) -> b -> [a] ->{𝕖} b
data.List.foldRight f b =
  go b = cases
    as :+ a -> go (f a b) as
    [] -> b
  go b
```

# How to reproduce
- `clone @unison/base`
- paste the code above into scratch file
- `update`

Adding the same function with a different name works, only updating causes this exception.

ceedubs:
This looks related to #4354. At least, the error message looks similar. That this one is reported to only happen on update is surprising to me.

jskripsky:
Downgrading UCM to release/M5g "fixes" the problem, just like in #4354.

PR #4366, opened 2023-10-31
Fix kind inference context bug

### LINKED ISSUE #4354: weird malformed type error related to >>
I can't make any sense of this typechecker crash that I'm seeing on a trunk build at d5e809b87f . It does not seem to happen on release M5g, so I'm guessing this is related to the recently-added kind checking.

The following code (which depends on the `>>` from base) crashes the typechecker:

```haskell
unique ability Return a where
  return : a ->{Return a} any

unique type Fold.Impl g a b x = Impl (x -> a ->{g, Return b} x) ('{g} x) (x ->{g} b)

Return.map : (a -> {g1} b) -> (i -> {g2, Return a} r) -> i -> {g1, g2, Return b} r
Return.map f =
  go = cases
    { r } -> r
    { return a -> _ } ->
      return (f a)
  run -> input -> handle run input with go

Fold.Impl.dimap : (a -> {g1} b) -> (c -> {g2} d) -> (Fold.Impl {g} b c x) -> (Fold.Impl {g, g1, g2} a d x)
Fold.Impl.dimap pre post = cases Fold.Impl.Impl step init extract ->
  step' : x -> a ->{g, g1, g2, Return d} x
  step' x = Return.map post (pre >> step x)
  extract' : x ->{g, g1, g2} d
  extract' x = post (extract x)
  f : Fold.Impl {g, g1, g2} a d x
  f = Fold.Impl.Impl step' init extract'
  f
```

```
Encountered exception:                                                                                
[ malformed type ] IntroOuter not in lexical scope of matching Forall                                 
    error
```

What's really strange is that if I copy/paste the definition of `>>` from base into my scratch file (so it is now totally self contained and doesn't depend on base), the error goes away:

```haskell
(Function.>>) : (a ->{𝕖} b) -> (b ->{𝕖} c) -> a ->{𝕖} c
(Function.>>) f g x = g (f x)
```

```
  I found and typechecked these definitions in ~/code/unison/foldl.u. If you do an `add` or
  
      unique type Fold.Impl g a b x
      unique ability Return a
      Fold.Impl.dimap : (a ->{g1} b) -> (c ->{g2} d) -> Impl {g} b c x -> Impl {g, g1, g2} a d x
      Function.>>     : (a ->{𝕖} b) -> (b ->{𝕖} c) -> a ->{𝕖} c
      Return.map      : (a ->{g1} b) -> (i ->{g2, Return a} r) -> i ->{g1, g2, Return b} r
```

tstat:
fixes #4354, #4362

## Overview

The kind inference context was erroneously just a Map, rather than some sort of ordered context. So, shadowed type variables were problematic.

## Test coverage

Added a new transcript that demonstrates the fix.

### DIFF
--- a/parser-typechecker/src/Unison/KindInference/Generate.hs
+++ b/parser-typechecker/src/Unison/KindInference/Generate.hs
@@ -20,7 +20,7 @@ import Unison.KindInference.Constraint.Context (ConstraintContext (..))
 import Unison.KindInference.Constraint.Provenance (Provenance (..))
 import Unison.KindInference.Constraint.Provenance qualified as Provenance
 import Unison.KindInference.Constraint.Unsolved (Constraint (..))
-import Unison.KindInference.Generate.Monad (Gen, GeneratedConstraint, freshVar, insertType, lookupType, scopedType)
+import Unison.KindInference.Generate.Monad (Gen, GeneratedConstraint, freshVar, pushType, lookupType, scopedType)
 import Unison.KindInference.UVar (UVar)
 import Unison.Prelude
 import Unison.Reference (Reference)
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
👍

dolio (approved):
I see. So in shadowing situations, exiting the shadowing scope would completely remove the necessary information from the enclosing scope? Makes sense.

PR #4374, opened 2023-11-08
print/parse `update`

aryairani:
## Overview

This PR introduces a significantly simplified `update` process:
- Given your already typechecked scratch file, UCM attempts to substitute the new definitions everywhere the old were referenced in your namespace and other definitions, propagating recursively.
- If successful, you are done!
- If unsuccessful, UCM will load the dependent definitions into the scratch file for you to edit until they typecheck.

The old update process involved `patch`, `todo`, and due to outstanding bugs, often resulted in incomplete update propagation and definitions with hashes still showing. The new process doesn't involve any of that. The old `update` command is still available as `update.old`.

### Future enhancements (apart from bugfixes):
- The dependents will be listed in the scratch file in the order you should work on them.

### Limitations / Known issues:
- **Record field removal:** To remove a field from a record type, you must first delete the three field accessors per deleted field before you can successfully `update`.
    e.g. To go from
    ```
    unique type A { x : Nat, y : Nat }
    ```
    to
    ```
    unique type A { x : Nat }
    ```
    you should:
   ```
    > delete.term A.y
    > delete.term A.y.set
    > delete.term A.y.modify
    > update
    ```
- **Unnamed dependents:** If any of your dependent definitions have dependencies without names, this could cause the propagation step to end early. All of your definitions and dependencies must have names. The good news is that this new `update` process is expected to no longer create situations where you have missing names.
- **Misplaced constructors:** If you've moved `type` or `ability` constructors out from where they initially were placed by UCM, `update` may fail. Make sure that all of your constructors have names somewhere under the namespace corresponding to the `type` or `ability`.
- **Some aliases become non-aliases:** If one of your updates' dependents is aliased and part of a cycle, aliases may be altered from being an actual alias to being a similar definition that is still a dependent but not part of the cycle. e.g. `x,y <-> z` ( i.e. `x` and `y` are aliases, in a cycle with `z`) may become `x <-> z <- y` (i.e. `y` is no longer an alias of `x` and no longer in a cycle with `z`).

### Other workflow changes:
- **Aliases are updated independently.** It used to be that if you had two aliases for a definition and updated just one of them, the other would be updated implicitly as well. That's no longer true. Now only the one you updated will change, and the other will remain unchanged.
- **Patches are no longer part of `update`.** Whereas we used to try to save partially updated code to the codebase, and you'd use `todo` to find non-updated code, that's no longer the case. The non-updated code will all be in your scratch file, and `todo` is not part of the `update` process. However, we haven't made any changes to the `todo` command.  If `todo` tells you have stuff to upgrade, it's basing this on some patch that resulted from using the old `update` command (now `update.old`). This new `update` command doesn't use patches in any way.

## Implementation notes

This implementation avoids limitations in our parsing/typechecking architecture by treating it as a black box. We render the dependents into a new scratch file as a string, then parse and typecheck it like normal, which turned out to be much less work than previously attempted approaches.

TBD, there were many.

## Test coverage

There are several new transcript tests.

## Loose ends

TBD, there are many.


aryairani:
Successfully enabling record printing (via "strategy 2") broke all the record-related transcripts because the record syntax and explicit accessor are both present in the scratch file. I'm going to make another time boxed attempt at suppressing the explicit accessor definitions; otherwise probably switch the transcripts to `ucm:error`. The alternative is to re-break record syntax printing. Each approach causes different extra work for the user.

pchiusano (approved):
Awesome, congrats on getting this across the finish line, looking forward to trying it out!!

PR #4387, opened 2023-11-14
update `update` info messages

aryairani:
Per https://unisonlanguage.slack.com/archives/CLG4P0GSD/p1699714157423269?thread_ts=1699713010.873129&cid=CLG4P0GSD

Later I think it would be nice if we could print fewer messages for small updates.  Like maybe just print extra info if the steps are taking more than 1 second each, otherwise something more concise.


pchiusano (approved):
Cool. Yeah we can always tweak this later, but this is an improvement.

PR #4388, opened 2023-11-14
make findCtorNames return an Output instead of error

aryairani:
## Implementation notes

`findCtorNames` now returns `Either Output` and we thread a transaction abort through various places in order to catch the error and terminate the `update`/`upgrade`.

We considered doing it as a precondition check, but didn't have time before releasing today.

## Test coverage

See updated transcripts.


mitchellwrosen (approved):

aryairani:
Something is stuck on the mac workflow, preventing it from starting? Canceling the workflow to restart it also doesn't seem to be working. I'm going to merge it anyway and let `trunk` build try again.

PR #4394, opened 2023-11-16
Add build-tools only devShell that doesn't depend on haskell.nix

tstat:
## Overview

Most of the team uses nix to provide build tools only and haskell.nix doesn't add much value for them. Haskell.nix does pull in many dependencies and using the associated cache is required to avoid very long build times when we bump our haskell.nix dep. This is exacerbated by the bad signature issue for the aarch64-darwin iohk cache.

To avoid these issues, this PR adds a new devShell that provides build tools from nixpkgs, using the official nixos cache. Specifically the new shell provides: ghc, stack, hpack, ormolu, and hls with our company-wide pinned versions. 

This new devShell is the default, so `nix develop` will no longer rely on haskell.nix.

### DIFF
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -20,6 +20,7 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
+          - macOS-12
     steps:
     - uses: actions/checkout@v3
     - uses: cachix/install-nix-action@v22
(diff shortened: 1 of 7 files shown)


aryairani (approved):
Seems to work for me!

ceedubs:
I tried this out and it doesn't work on a mac if you don't have the `Xcode` command line tools installed. At some point there was a version of the nix-based `stack` flake that worked for me. When I find the time I'll try to find the version that worked for me and compare what they do.

PR #4413, opened 2023-11-21
fix for test watches in `update`

### LINKED ISSUE #4405: new update adds double type annotation to tests
Adding a test via `update` results in a weird term that seems to have 2 type signatures. The same issue does not happen when using `add` instead of `update`. Note that this happens _even_ if the test doesn't exist before the `update`.

This is based on a trunk build at ba27a9479efecc5a63d7efd9816cd90e923f5f23.

# Transcript reproduction

````
```ucm
.> builtins.merge
```

```unison
foo n = n + 1

test> mynamespace.foo.test =
  n = 1
  if (foo n) == 2 then [ Ok "passed" ] else [ Fail "wat" ]
```

```ucm
.> update
.> view mynamespace.foo.test
.> edit mynamespace.foo.test
.> load
```
````

# Failing transcript output

*Note that `view` shows two type signatures for the term. Trying to `edit` it will also result in an error.*

````
```ucm
.> builtins.merge

  Done.

```
```unison
foo n = n + 1

test> mynamespace.foo.test =
  n = 1
  if (foo n) == 2 then [ Ok "passed" ] else [ Fail "wat" ]
```

```ucm

  change:
  
      foo                  : Nat -> Nat
      mynamespace.foo.test : [Result]
  
  `>`)... Ctrl+C cancels.

    4 |   n = 1
    
    ✅ Passed passed

```
```ucm
.> update

  Okay, I'm searching the branch for code that needs to be
  updated...

  That's done. Now I'm making sure everything typechecks...

  Everything typechecks, so I'm saving the results...

  Done.

.> view mynamespace.foo.test

  mynamespace.foo.test : [Result]
  mynamespace.foo.test : [Result]
  mynamespace.foo.test =
    n = 1
    if foo n == 2 then [Ok "passed"] else [Fail "wat"]

.> edit mynamespace.foo.test

  ☝️
  
  I added these definitions to the top of
  /Users/cody/code/work/unison/scratch.u
  
    mynamespace.foo.test : [Result]
    mynamespace.foo.test : [Result]
    mynamespace.foo.test =
      n = 1
      if foo n == 2 then [Ok "passed"] else [Fail "wat"]
  
  You can edit them there, then do `update` to replace the
  definitions currently in this namespace.

.> load

  offset=8:
  unexpected :
      2 | mynamespace.foo.test : [Result]
  

```



🛑

  offset=8:
  unexpected :
      2 | mynamespace.foo.test : [Result]
  
````

### LINKED ISSUE #4410: test watches don't round-trip under `rewrite`
specifically, each rewrite adds a type signature until it doesn't parse anymore

Same underlying issue as #4405

### PR DESCRIPTION
**Pull request by aryairani**
A few changes to fix #4405:
- Change `update` to render tests (terms having type `[Test.Result]`) as test watches when recreating a scratch file.
- Hide the test type signature when `prettyUnisonFile`-ing those tests, since it is added automatically by the parser.

Also fixes #4410.

## Overview

Changes the round-trip under `update` of
```
test> mynamespace.foo.test = if (foo 2) == 2 then [ Ok "passed" ] else [ Fail "wat" ]
```
from
```
mynamespace.foo.test : [Result]
mynamespace.foo.test = if foo 2 == 2 then [Ok "passed"] else [Fail "wat"]
```
or worse,
```
mynamespace.foo.test : [Result]
mynamespace.foo.test = if foo 2 == 2 then [Ok "passed"] else [Fail "wat"]

test> mynamespace.foo.test = if (foo 2) == 2 then [ Ok "passed" ] else [ Fail "wat" ]
```
to
```
test> mynamespace.foo.test = if (foo 2) == 2 then [ Ok "passed" ] else [ Fail "wat" ]
```

## Test coverage

Adds a couple of transcripts.

## Loose ends

- [ ] #4417 (unison-src/transcripts/update-test-to-non-test.md)

### DIFF
--- a/unison-src/transcripts/add-test-watch-roundtrip.md
+++ b/unison-src/transcripts/add-test-watch-roundtrip.md
@@ -7,8 +7,7 @@ test> foo : [Test.Result]
 foo = []
 ```
 
-Apparently when we add a test watch, we add a type annotation to it, even if it already has one.
-We don't want this to happen though!
+Apparently when we add a test watch, we add a type annotation to it, even if it already has one. We don't want this to happen though!
 
 ```ucm
 .> add
--- a/unison-src/transcripts/add-test-watch-roundtrip.md
+++ b/unison-src/transcripts/add-test-watch-roundtrip.md
@@ -1,0 +1,16 @@
+```ucm:hide
+.> builtins.mergeio
+```
+
+```unison:hide
+test> foo : [Test.Result]
+foo = []
+```
+
+Apparently when we add a test watch, we add a type annotation to it, even if it already has one.
+We don't want this to happen though!
+
+```ucm
+.> add
+.> view foo
+```
--- a/unison-src/transcripts/add-test-watch-roundtrip.output.md
+++ b/unison-src/transcripts/add-test-watch-roundtrip.output.md
@@ -3,7 +3,7 @@ test> foo : [Test.Result]
 foo = []
 ```
 
-Apparently when we add a test watch, we add a type annotation to it, even if it already has one.
+Apparently when we add a test watch, we add a type annotation to it, even if it already has one. We don't want this to happen though!
 
 ```ucm
 .> add
--- a/unison-src/transcripts/add-test-watch-roundtrip.output.md
+++ b/unison-src/transcripts/add-test-watch-roundtrip.output.md
@@ -9,7 +9,7 @@ Apparently when we add a test watch, we add a type annotation to it, even if it 
 .> add
 
   ⍟ I've added these definitions:
-
+  
     foo : [Result]
 
 .> view foo
--- a/unison-src/transcripts/add-test-watch-roundtrip.output.md
+++ b/unison-src/transcripts/add-test-watch-roundtrip.output.md
@@ -1,0 +1,21 @@
+```unison
+test> foo : [Test.Result]
+foo = []
+```
+
+Apparently when we add a test watch, we add a type annotation to it, even if it already has one.
+
+```ucm
+.> add
+
+  ⍟ I've added these definitions:
+  
+    foo : [Result]
+
+.> view foo
+
+  foo : [Result]
+  foo : [Result]
+  foo = []
+
+```
--- a/unison-src/transcripts/add-test-watch-roundtrip.output.md
+++ b/unison-src/transcripts/add-test-watch-roundtrip.output.md
@@ -9,7 +9,7 @@ Apparently when we add a test watch, we add a type annotation to it, even if it 
 .> add
 
   ⍟ I've added these definitions:
-  
+
     foo : [Result]
 
 .> view foo
--- a/parser-typechecker/src/Unison/Codebase/MainTerm.hs
+++ b/parser-typechecker/src/Unison/Codebase/MainTerm.hs
@@ -67,12 +67,8 @@ builtinMainWithResultType a res = Type.arrow a (Type.ref a DD.unitRef) io
   where
     io = Type.effect a [Type.builtinIO a, DD.exceptionType a] res
 
--- [Result]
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: unison-src/transcripts/add-test-watch-roundtrip.md [hunk 1]
```diff
@@ -0,0 +1,15 @@
+```ucm:hide
+.> builtins.mergeio
```
ceedubs, line 2: Off topic: why do both `builtins.merge` and `builtins.mergeio` exist, and is there a reason to prefer `mergeio` here?

unison-src/transcripts/add-test-watch-roundtrip.output.md
```diff
@@ -0,0 +1,21 @@
...
+.> add
+
+  ⍟ I've added these definitions:
+  
+    foo : [Result]
+
+.> view foo
+
+  foo : [Result]
+  foo : [Result]
```
ceedubs, line 18: This is still showing the incorrect output, isn't it?

ceedubs (commented):
I see at least one place where your transcript reports that there is a current bug. But I see another where it seems to be adding the double type annotation that I thought should be fixed b this PR. So I'm confused.

unison-src/transcripts/add-test-watch-roundtrip.output.md
aryairani: Confusing; sorry.
Yes, but that transcript doesn't involve `update`. So it's tracked by loose end #4417.

ceedubs: Oh I see the "we don't want this to happen" is saying that this is a bug but the transcript is just capturing that it currently happens, right?

ceedubs (approved):
Looks to be an improvement, though tests will probably continue to be not very usable until #4417 is completed.

unison-src/transcripts/add-test-watch-roundtrip.md
aryairani: Good question. `builtins.merge` just includes "builtins" (from the point of view of the runtime). So that's primitive types like `Int`, `Nat`, primitive terms like `Nat.+`.

`builtins.mergeio`, includes all the "builtins" but also all the user-defined types that UCM commands rely on. So `Doc`, `Test.Result`, `Author`, etc. It is defined with actual Unison code embedded in ucm, and parsed/typechecked when you call the command.

`IO` used to be a user-defined ability, hence the name "builtins.mergeio", and the rest of the Unison-based definitions were lumped in there as well for convenience. Today, `IO` is a primitive ability, so you actually get it from `builtins.merge`, and not from `builtins.mergeio` 😓

aryairani: `builtins.merge` takes less time to execute, so I prefer that when I can get away with it, since 2-3 seconds per transcript (or whatever it takes to run the parser/typechecker like `builtins.mergeio` must do) gets multiplied by 100 transcripts or whatever.

Kind of like:
`builtins.merge` fastest
`builtins.mergeio` fast-ish
`project.create` clone @unison/base 😑

(Actually skipping all of them is the fastest.)

unison-src/transcripts/add-test-watch-roundtrip.output.md
aryairani: Right. I don't think you will actually run into #4417 unless you are manually putting type signatures on your tests as in that transcript though. Let me know if you find otherwise!

unison-src/transcripts/add-test-watch-roundtrip.md
aryairani: I was mistaken, `Test.Result` is in the smaller `builtins.merge`.

PR #4414, opened 2023-11-21
Fix bug in scratch file updates

tstat:
## Overview
The system temp directory could be a different device (like a tmpfs), which would cause renameFile to fail since we are attempting to rename across device boundaries. To avoid this issue, create the temp file in the same directory as the unison scratch file that we will update.

## Test coverage

manual


aryairani:
Ah good catch.

aryairani (approved):

Issue #4415, opened 2023-11-21
Unable to resolve type name


pchiusano:
Just put this in a scratch file:

```
unique type A = C Int
unique type sub.A = D Int

--- results in

    ❓
    
    I couldn't resolve any of these symbols:
    
        1 | unique type A = C Int
    
    
    Symbol   Suggestions
             
    A        No matches
 
```

@stew reported this, it was introduced in M5g.

stew:
```c4920426ed58bdd046a0858abd74936fe9aee0f5 is the first bad commit
commit c4920426ed58bdd046a0858abd74936fe9aee0f5
Author: Paul Chiusano <paul.chiusano@gmail.com>
Date:   Thu Sep 14 08:51:53 2023 -0500

    fixes type signature resolution issue

 parser-typechecker/src/Unison/UnisonFile/Names.hs | 31 +++++++++++++++++++++--
 unison-core/src/Unison/DataDeclaration/Names.hs   | 13 +++++++---
 unison-src/transcripts/fix3759.md                 | 22 ++++++++++++++++
 unison-src/transcripts/fix3759.output.md          | 28 ++++++++++++++++++++
 4 files changed, 89 insertions(+), 5 deletions(-)
 create mode 100644 unison-src/transcripts/fix3759.md
 create mode 100644 unison-src/transcripts/fix3759.output.md

```

aryairani:
@pchiusano When the name in the scratch file also exists in the codebase, we get an even crazier situation, where the type of the result of the constructor call comes from the codebase, not from the scratch file (reminiscent of #4325 where the bisected commit came from), leading to an invalid decl (and a crash)

e.g. `scratch.C : Int -> codebase.C` instead of `scratch.C : Int -> scratch.C`

Can you give any insights  re. #4325?

tstat:
Looks to me like the issue arises from here:
https://github.com/unisonweb/unison/blob/bc0788ba53abc77613674916cf263b21a0eb4b89/parser-typechecker/src/Unison/UnisonFile/Names.hs#L111
where we drop all mutimap keys that have non-singleton values. One of these keys could be the FQN for a type or term (e.g. `Foo => [Foo, sub.Foo]`), then we drop it and have no mapping for `Foo`.

I don't know the context of this `variableCanonicalizer` function, but I assume we want to keep some name for every input?

pchiusano:
[...]

PR #4419, opened 2023-11-22
Fix name resolution bug in the parser

tstat:
## Overview

Fixes a bug that arose if the FQN of a decl was a suffix of some other FQN. This would cause the parser would crash, failing to resolve the decl's FQN.

## Implementation notes

See #4415 

## Test coverage

Added a new transcript `fix4415.md` that demonstrates the issue, then followed with a commit that fixes it.


pchiusano (approved):
Nice!

aryairani (approved):

aryairani:
````
```ucm
.> clone @unison/httpclient
@unison/httpclient/main> edit Config proxy.Config
@unison/httpclient/main> load
```
````
@stew ^ This seems to work now

PR #4428, opened 2023-11-25
don't use `deepTermMetadata` for `test`

### LINKED ISSUE #4406: update `test` command to check for `[Test.Result]` type, not `IsTest` metadata

aryairani:
## Overview

Use the type index rather than deepTermMetadata to determine what is a test, both for `test` and for `edit`.

Closes #4406.

## Implementation notes

Adds a sql query and codebase functions that filters a set of `TermReferenceId` according to the type index.
Also adds an untested query that filters a set of `Referent.Id` according to the type index, but the former is used in the `test` command, because constructors won't be tests.

I thought about trying to remove metadata altogether, but it would've been too big. There is some work in https://github.com/unisonweb/unison/tree/topic/remove-deep-metadata.

I believe the only place `deepTermMetadata`/`deepTypeMetadata` are still used now is for 
* printing namespace diffs where metadata has changed
* [ ] metadata-based checks for docs (in `HandleInput.docsI`), can be removed

## Test coverage

I modified `unison-src/transcripts/test-command.md` to show examples of running tests that don't have `IsTest` metadata.

### DIFF
--- a/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
+++ b/codebase2/codebase-sqlite/U/Codebase/Sqlite/Operations.hs
@@ -70,6 +70,8 @@ module U.Codebase.Sqlite.Operations
     -- ** type index
     Q.addTypeToIndexForTerm,
     termsHavingType,
+    filterTermsByReferenceHavingType,
+    filterTermsByReferentHavingType,
 
     -- ** type mentions index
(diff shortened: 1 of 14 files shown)


aryairani:
cc @ChrisPenner in case we want to do something similar for `getTermTag`; checking for 3 concrete types in the index, instead of passing in a `Maybe (Type v a)` at all.

pchiusano (approved):

PR #4431, opened 2023-11-27
Add support for the `controls` attribute in Video

hojberg:
## Overview

For Unison Doc videos, we want to allow users to provide the `controls` attribute and have that be passed through to the generated HTML.

cc @alvaroc1


hojberg:
Also updated ui-core (and thus Share) to support this.

pchiusano (approved):

PR #4434, opened 2023-11-28
fix unsigned int error in update

### LINKED ISSUE #4425: `update` fails to add a new record type
I created https://github.com/unisonweb/unison/tree/fix/update-new-record with this failing transcript to kick things off.

```unison
unique type Foo = { bar : Nat }
```

```ucm

  change:
  
      unique type Foo
      Foo.bar        : Foo -> Nat
      Foo.bar.modify : (Nat ->{g} Nat) -> Foo ->{g} Foo
      Foo.bar.set    : Nat -> Foo -> Foo

```
This shouldn't be an error.

```ucm
.> update

  Okay, I'm searching the branch for code that needs to be
  updated...

  That's done. Now I'm making sure everything typechecks...

  Everything typechecks, so I'm saving the results...

  I couldn't complete the update because I couldn't find 0
  constructor(s) for Foo where I expected to. I found: []
  
  You can use `view Foo` and
  `alias.term <hash> Foo.<ConstructorName>` to give names to
  each constructor, and then try again.

.> view Foo

  ⚠️
  
    Foo

```

aryairani:
This subtle bug was bad.  This PR includes the failure case as a transcript.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/Update2.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/Update2.hs
@@ -381,9 +381,9 @@ findCtorNames names forwardCtorNames ctorCount n =
       insertShortest m _ = m
       m = foldl' insertShortest mempty (Foldable.toList center)
       ctorCountGuess = fromMaybe (Map.size m) ctorCount
-   in if Map.size m == ctorCountGuess && all (isJust . flip Map.lookup m) [0 .. fromIntegral ctorCountGuess - 1]
+   in if Map.size m == ctorCountGuess && all (isJust . flip Map.lookup m . fromIntegral) [0 .. ctorCountGuess - 1]
         then Right $ Map.elems m
-        else Left $ Output.UpdateIncompleteConstructorSet n m ctorCountGuess
+        else Left $ Output.UpdateIncompleteConstructorSet n m ctorCount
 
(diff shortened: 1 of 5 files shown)


unison-cli/src/Unison/Codebase/Editor/HandleInput/Update2.hs
```diff
@@ -381,9 +381,9 @@ findCtorNames names forwardCtorNames ctorCount n =
       insertShortest m _ = m
       m = foldl' insertShortest mempty (Foldable.toList center)
       ctorCountGuess = fromMaybe (Map.size m) ctorCount
-   in if Map.size m == ctorCountGuess && all (isJust . flip Map.lookup m) [0 .. fromIntegral ctorCountGuess - 1]
+   in if Map.size m == ctorCountGuess && all (isJust . flip Map.lookup m . fromIntegral) [0 .. ctorCountGuess - 1]
```
aryairani, line 384: This was somehow counting up from `[0 .. unsigned(-1)]` instead of `[0 .. -1]` and then quietly failing the `all`.

ceedubs (approved):
🎉 this fixed the issue that I was running into. Thank you!

aryairani:
~~I'm actually confused as to how any type updates were working with this bug.~~
Oh, the result for type updates with more than 0 existing constructors I guess would not have wrapped around with the `-1`. 👍

PR #4447, opened 2023-11-29
Unique type churn transcript: add case of revert

ceedubs:
The unique type churn transcript now captures that if you revert a type
back to its original form (while keeping the same name), you'll end up
with the original hash again.

This behavior surprised me, but I can't see a reason it's bad, and it
might occasionally be useful. If nothing else we should capture the
current behavior so it doesn't change unintentionally.


aryairani (approved):
Thanks!

PR #4458, opened 2023-12-01
Allow layout blocks within Doc transclusion

### LINKED ISSUE #4384: Parse error parsing blocks in doc quasiquote
Parses fine:

```
foo = docExampleBlock 0 do 2
```

Also fine:

```
foo = {{ {{ docExampleBlock 0 '2 }} }}
```

Doesn't parse:

```
foo = {{ {{ docExampleBlock 0 do 2 }} }}
```

The error is:

```
  offset=24:
  unexpected end of input
  expecting newline or semicolon
```

pchiusano:
Fixes #4384 

Previously, code like this would not parse:

```
fix_4384a = {{ {{ docExampleBlock 0 do 1  }} }}

fix_4384b = {{ {{ docExampleBlock 0 do
  x = 1
  y = 2
  x + y
  }} }}
```

Now it parses fine. The fix was to re-enable layout within a transclusion, and to have `}}` pop layout up to the enclosing `{{`. Prior to this change, layout was incorrectly being entirely disabled within a doc block, even within a nested transclusion which can contain arbitrary Unison code that uses layout.

The diff is actually very small, just one function (see self-review), but I had to move some helper functions out of a where clause to make them available for this change, so the actual diff looks bigger.

Nothing.

## Test coverage

Added regression tests, including some weird examples I could think of, and existing transcripts cover the doc syntax. I feel pretty confident this can only improve parsing and round trip issues, but after merging to trunk, I'd recommend dogfooding it for a couple days just to make sure there aren't any regressions in weird corner cases I haven't thought of.

cc @stew @runarorama @ceedubs who I know have been affected by this.

### DIFF
--- a/unison-syntax/src/Unison/Syntax/Lexer.hs
+++ b/unison-syntax/src/Unison/Syntax/Lexer.hs
@@ -525,10 +525,7 @@ lexemes' eof =
               ex <- CP.space *> lexemes' end
               pure ex
 
-        -- general idea is these can push onto layout stack
-        -- and pop, 
         docClose = [] <$ lit "}}"
-        -- and this can pop from layout stack
         docOpen = [] <$ lit "{{"
         
         link =
--- a/unison-syntax/src/Unison/Syntax/Lexer.hs
+++ b/unison-syntax/src/Unison/Syntax/Lexer.hs
@@ -547,7 +547,7 @@ lexemes' eof =
                 -- pop items off the layout stack up to the nearest enclosing
                 -- syntax.docTransclude.
                 ts <- lexemes' (P.lookAhead ([] <$ lit "}}"));
-                S.modify (\env -> env { inLayout = inLayout env });
+                S.modify (\env -> env { inLayout = inLayout env0 });
                 pure ts
               }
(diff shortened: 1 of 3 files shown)


unison-syntax/src/Unison/Syntax/Lexer.hs
```diff
@@ -544,8 +536,21 @@ lexemes' eof =
...
+                S.put (env0 { inLayout = True });
+                -- Note: this P.lookAhead ensures the }} isn't consumed,
+                -- so it can be consumed below by the `close` which will
+                -- pop items off the layout stack up to the nearest enclosing
+                -- syntax.docTransclude.
+                ts <- lexemes' (P.lookAhead ([] <$ lit "}}"));
+                S.modify (\env -> env { inLayout = inLayout env0 });
+                pure ts
+              }
+              <+> close ["syntax.docTransclude"] (lit "}}")
```
pchiusano, line 553: This is the actual fix / change. The rest is just shuffling things around.

pchiusano (commented):
Self review.

aryairani (approved):
Thanks!

PR #4471, opened 2023-12-05
bugfix: tweak upgrade to ignore aliases of old things

mitchellwrosen:
## Overview

This PR fixes an issue that caused `upgrade` to behave unintuitively.

The algorithm essentially works as follows (before and after this PR):

- When upgrading library dependency `old` to `new`, render out to a Unison file all dependents of all terms in `lib.old` (e.g. `lib.old.thingy#oldthingy`), but with the `old` segment swapped out for `new`, then parse and typecheck that.

  For example, consider the namespace:

  ```
  lib.old.thingy#oldthingy = ...
  lib.new.thingy#newthingy = ...
  myfoo = #oldthingy + 100
  ```

  On `upgrade old new`, we render, parse, and typecheck a Unison file like

  ```
  myfoo = lib.new.thingy + 100
  ```

The issue was to do with aliases. Prior to this PR, additional aliases of things in `lib.old.*` would essentially prevent upgrading as follows.

Consider the namespace:

```
lib.old.thingy#oldthingy = ...
lib.new.thingy#newthingy = ...
mythingy#oldthingy = ...
myfoo = #oldthingy + 100
```

The old implementation would render, parse, and typecheck a Unison file like

```
myfoo = mythingy + 100
```

as the "old" reference `#oldthingy` still had a set of names `{ lib.new.thingy, mythingy }` from which we make the pretty-print environment.

This behavior is especially bad considering dependencies' dependencies are actually in scope for developers (something we neglected to think about when implementing the algorithm the first time around).

While in the example above, the user's project code itself has a name for `#oldthingy`, it's probably much more common for `#oldthingy` to be referenced by some _dependency_ that still depends on `old` (which the user is trying to upgrade to `new):

```
lib.old.thingy#oldthingy = ...
lib.someOtherDependency.lib.old.thingy#oldthingy = ...
```

This would also keep `#oldthingy` alive / prevent an upgrade to `#newthingy` in the exact same way.

So, the solution implemented here is simply to pretty-print hashes that are _directly_ referenced by the old dependency, i.e. things at the top level of `lib.old.*` (no matter what's in `lib.old.lib.*`), as `lib.new.*` (appropriately suffixified), no matter if there are any other aliases.

### DIFF
--- a/parser-typechecker/src/Unison/PrettyPrintEnv/Names.hs
+++ b/parser-typechecker/src/Unison/PrettyPrintEnv/Names.hs
@@ -1,6 +1,12 @@
 {-# LANGUAGE OverloadedStrings #-}
 
-module Unison.PrettyPrintEnv.Names (fromNames, fromSuffixNames) where
+module Unison.PrettyPrintEnv.Names
+  ( fromNames,
+    fromSuffixNames,
+    prioritize,
+    shortestUniqueSuffixes,
+  )
+where
 
 import Data.Set qualified as Set
 import Unison.HashQualified' qualified as HQ'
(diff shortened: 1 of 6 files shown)


aryairani:
@mitchellwrosen Looks good! Could you update the transcript to `view bar` also after the upgrade?

mitchellwrosen:
@aryairani sure, done

aryairani:
We found another bug where `upgrade` doesn't work if you haven't loaded any scratch file yet in your ucm session. @mitchellwrosen is working on a fix for that now 🙏

aryairani (approved):

PR #4477, opened 2023-12-06
Add `move` command that moves existing namespaces,terms, and types

### LINKED ISSUE #3767: implement `move` command that moves term, type, and child namespace
This would be a pretty nice quality of life improvement.

How many times have you done `move.type Foo Bar` and `move.namespace Foo Bar` (to get the constructors and helper functions?) Why not just do both in the same command? (If you really meant to only move the type, which I'm pretty sure I have never wanted, you can always just `move.namespace Bar Foo` the the namespace back to its original location)

Likewise for `move.term` (move the terms underneath, including docs). 

Implementation notes, for anyone who wants to take this on:

* If there's no child namespace, it can just noop for the `move.namespace` part. 
* And you can look at the `move.namespace` logic and call into that in `move.term/type`.
* This logic all lives in `HandleInput`

tstat:
## Overview

Add `move` command which moves namespaces, terms, and types from src to dest.

fixes #3767 

## Implementation notes

Break existing move* code up a bit to be callable from the new move code.

## Test coverage

There is a new transcript named `move-all.md`.

## Loose ends

Moving the namespace and moving the term/type is done in two separate causal steps, where it would preferably be done in one. The central issue is that we want to move the namespace's history from src to dest (as [`modifyAtM`](https://github.com/unisonweb/unison/blob/cbc9d9a6e27d17fda326aca1b3acaec78ce63d5b/parser-typechecker/src/Unison/Codebase/Branch.hs#L594-L609) allows), but moving the term and type is accomplished with [`stepManyAt`](https://github.com/unisonweb/unison/blob/cbc9d9a6e27d17fda326aca1b3acaec78ce63d5b/parser-typechecker/src/Unison/Codebase/Branch.hs#L519-L526) which calls [`consBranchSnapshot`](https://github.com/unisonweb/unison/blob/cbc9d9a6e27d17fda326aca1b3acaec78ce63d5b/parser-typechecker/src/Unison/Codebase/Branch.hs#L744-L780) to manage causal stepping. `consBranchSnapshot` discards child history for namespaces that appear in the updated branch, but not the original branch.

### DIFF
--- a/unison-cli/src/Unison/Cli/MonadUtils.hs
+++ b/unison-cli/src/Unison/Cli/MonadUtils.hs
@@ -51,6 +51,7 @@ module Unison.Cli.MonadUtils
     updateRoot,
     updateAtM,
     updateAt,
+    updateAndStepAt,
 
     -- * Terms
     getTermsAt,
@@ -87,6 +88,7 @@ import Control.Monad.Reader (ask)
 import Control.Monad.State
 import Data.Configurator qualified as Configurator
 import Data.Configurator.Types qualified as Configurator
+import Data.Foldable
 import Data.Set qualified as Set
 import U.Codebase.Branch qualified as V2 (Branch)
 import U.Codebase.Branch qualified as V2Branch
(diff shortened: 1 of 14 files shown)


pchiusano:
Nice! Question: If there isn't a term or type, does it still create three separate steps in the history?

tstat:
No, If neither the term nor the type exist then no move term/type step is made, if at least one exists then there is one step to move the term and/or type. Similarly, a namespace move step is made iff the src namespace exists.

aryairani (approved):

SystemFw:
Thanks for this!

PR #4479, opened 2023-12-06
bugfix: make `upgrade` not require a latest-touched file (default to "scratch.u")

mitchellwrosen:
## Overview

`upgrade` would previously accidentally require the user to have touched some `.u` file before running it, and if they hadn't, it would confusingly make a little progress (including creating a new project branch) before dumping a misleading error message (which is the fault of the output message of `NoUnisonFile` saying something surprising...)


aryairani:
Do we still need to change something about `NoUnisonFile` saying something surprising?

aryairani:
I'm guessing the error doesn't occur in transcript mode?

aryairani (approved):

mitchellwrosen:
> Do we still need to change something about NoUnisonFile saying something surprising?

Yeah, that could be a new ticket

PR #4486, opened 2023-12-07
Don't typecheck dependents of unchanged defns in upgrade

### LINKED ISSUE #4482: `upgrade old new` shouldn't pull in dependents of `lib.old.lib.*`, just `lib.old.*`

### LINKED ISSUE #4483: `upgrade old new` shouldn't care about dependents of `lib.old.foo#foo` if `lib.new.foo#foo` (same hash) exists

tstat:
## Overview

fixes #4483 
fixes #4482 

We don't handle conflicted aliases during upgrade on trunk; instead picking a mapping at random (via the PPE). This is wrong, and there isn't really a way for us to handle this correctly. With patches we would refuse to apply such an update.

The logic for determining if the defn is changed leans into the behavior on trunk a bit though: we consider something unchanged if the intersection of the names of a ref on the old dep and the new dep is nonempty. 

## Test coverage

There is a transcript that demonstrates pulling in dependents for transitive deps of "old dep" if they only occur in "old dep" and giving them a name that will fail to resolve.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/Upgrade.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/Upgrade.hs
@@ -72,7 +72,7 @@ handleUpgrade oldDepName newDepName = do
 
   currentV1Branch <- Cli.getBranch0At projectPath
   let currentV1BranchWithoutOldDep = deleteLibdep oldDepName currentV1Branch
-  oldDepWithoutItsDeps <- over Branch.children (Map.delete Name.libSegment) <$> Cli.expectBranch0AtPath' oldDepPath
+  oldDep <- Cli.expectBranch0AtPath' oldDepPath
 
   newDepV1Branch <- Cli.expectBranch0AtPath' newDepPath
 
@@ -141,8 +141,8 @@ handleUpgrade oldDepName newDepName = do
         dependents <-
           Operations.dependentsWithinScope
             (Names.referenceIds namesExcludingLibdeps)
-            ( filterUnchangedTerms (Branch.deepTerms oldDepWithoutItsDeps)
-                <> filterUnchangedTypes (Branch.deepTypes oldDepWithoutItsDeps)
+            ( filterUnchangedTerms (Branch.deepTerms oldDep)
+                <> filterUnchangedTypes (Branch.deepTypes oldDep)
             )
         addDefinitionsToUnisonFile
           abort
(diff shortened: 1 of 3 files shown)


pchiusano:
@tstat I think we want the logic to be that it pulls in a transitive dependency if that's providing the only name for a hash (I know we talked about not bothering with this logic for V1, but after discussion with @stew I'm pretty convinced it's important given existing usage). Does this PR do that or no?

tstat:
No, this PR does not pull in local dependents of dependencies of old. I can adjust #4482 and update this PR to align with the new goal if that is the behavior we want.

However, I do think it is unexpected and undesirable that symbols from transitive dependencies are automatically in scope to begin with.

aryairani:
Did I get this right:

The scenario is: my project has a function `func` depends on `lib.old.lib.foo.bar`
I upgrade `old` to `new`, and there may or may not be a `lib.new.lib.foo.bar`, and if there is one, it may have changed. 

Thus @pchiusano is saying we want to re-typecheck `func` when doing the upgrade. This makes sense to me, and @tstat updated the PR to reflect this.

@tstat says it is weird to accidentally be able to reference stuff in the transitive dependencies (e.g. suffixified).

I'm assuming/extrapolating here: no objection to have them available in a less-automatic way, e.g. requiring a `use` directive to access them?

error (implicit use of transitive dependency):
```
func1 = bar
```
> I can't find `bar` in your branch or in its dependencies, but there's a `lib.old.lib.foo.bar`. You can do \<whatever> to use it.

ok (explicit use of transitive dependency):
```
func2 = lib.old.lib.foo.bar

func3 = 
  use lib.old.lib.foo
  bar
```

mitchellwrosen:
> The scenario is: my project has a function func depends on lib.old.lib.foo.bar
> I upgrade old to new, and there may or may not be a lib.new.lib.foo.bar, and if there is one, it may have changed.

I think in this case (well I need to add a hash to fully state what's going on) - considering `lib.old.lib.foo.bar#whatever` - if the only source of names for `#whatever` are from `lib.old.lib.*`, then we will render them (unsuffixified) with name `lib.old.lib.*`, not `lib.new.lib.*`

And that reminds me: I think we decided we ought to do that same thing for deleted definitions outside of transitive deps, for consistency.

For example, if `lib.old.foo` exists but `lib.new.foo` doesn't, render the has with name `lib.old.foo`, not `lib.new.foo` as we do currently (including in this PR). I think we should go ahead in this PR @tstat, we just forgot.

aryairani:
> we will render them (unsuffixified) with name lib.old.lib.*, not lib.new.lib.*

Ok that sounds good

> do that same thing for deleted definitions outside of transitive deps

Ok that sounds good

aryairani:
Could you add a golden transcript for those two cases as well

aryairani (approved):

PR #4505, opened 2023-12-13
fix "branch already exists" error message to show project, too

### LINKED ISSUE #4494: error message for cloning an existing repository can be wrong
I made the mistake of clone a repo that I had already cloned. But the error message should have `switch @unison/advent-of-code-client/main` instead of just `switch /main`.

Ideally the fix would still omit the project name if it's within the same project.

```
nimbus/health-check> clone @unison/advent-of-code-client/main
 
  @unison/advent-of-code-client/main already exists. You can switch to it with `switch /main`.
```

mitchellwrosen:
## Overview

Fixes #4494

### DIFF
--- a/unison-cli/src/Unison/CommandLine/OutputMessages.hs
+++ b/unison-cli/src/Unison/CommandLine/OutputMessages.hs
@@ -1941,7 +1941,7 @@ notifyUser dir = \case
       prettyProjectAndBranchName projectAndBranch
         <> "already exists."
         <> "You can switch to it with "
-        <> IP.makeExampleEOS IP.projectSwitch [prettyBranchName projectAndBranch]
+        <> IP.makeExampleEOS IP.projectSwitch [prettyProjectAndBranchName projectAndBranch]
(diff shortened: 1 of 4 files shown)


unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1941,7 +1941,7 @@ notifyUser dir = \case
       prettyProjectAndBranchName projectAndBranch
         <> "already exists."
         <> "You can switch to it with "
-        <> IP.makeExampleEOS IP.projectSwitch [prettyBranchName projectAndBranch]
+        <> IP.makeExampleEOS IP.projectSwitch [prettyProjectAndBranchName projectAndBranch]
```
mitchellwrosen, line 1944: This is the tweak; the rest is refactoring

mitchellwrosen:
Just forgot to move this from draft last week after it passed CI

aryairani (approved):

PR #4508, opened 2023-12-13
update the name of the pre-release build

aryairani:
per https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1702495233966029 which will be accessible until mid-March

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -162,7 +162,7 @@ jobs:
       - uses: "marvinpinto/action-automatic-releases@latest"
         with:
           repo_token: "${{ secrets.GITHUB_TOKEN }}"
-          automatic_release_tag: "latest"
+          automatic_release_tag: "trunk"
           prerelease: true
           title: "Development Build"
           files: |


ceedubs (approved):
Looks good to me. Is there anywhere else that we'd need to update documentation?

aryairani:
> Looks good to me. Is there anywhere else that we'd need to update documentation?
Good question — @rlmark Do you know?

This PR is about changing the nomenclature and url for UCM binaries that are built off of `trunk` 
from "latest" / `https://github.com/unisonweb/unison/archive/refs/tags/latest.zip` / `https://github.com/unisonweb/unison/releases/download/latest/ucm-windows.zip` / etc.

to "trunk" / `https://github.com/unisonweb/unison/archive/refs/tags/trunk.zip` / `https://github.com/unisonweb/unison/releases/download/trunk/ucm-windows.zip` / etc.

aryairani:
this may require a corresponding change to the homebrew formula for —head. is there one for nix too?

ceedubs:
> this may require a corresponding change to the homebrew formula for —head. is there one for nix too?

No, not with the way that the derivation is currently set up.

rlmark:
You should be good! We don't link to the pre-release builds for the installs page, only the official latest ones!

rlmark (approved):

PR #4511, opened 2023-12-14
Partial implementation for native execution/compilation via a standalone executable

dolio:
This PR implements infrastructure for evaluation/compilation via a standalone executable produced with `raco exe/distribute`. It's as yet incomplete, and doesn't really change the existing behavior of commands, but I wanted to get something submitted before winter break (and I was already planning on taking today off).

Here are the notable changes:

- `compile.native.genlibs` can be given a target directory, so that the generated files go somewhere other than the 'cache' directories we've been using. This will be used to generate a unison 'package' that is used for building the standalone racket executable.
- `run.native` now just accepts a string that it tries to make use of inside the command. This makes it uniform with respect to the `run` command, even though they still use different implementations at the moment.
- A native version of the `Runtime` record has been implemented that calls out to a standalone executable.
  * An implementation of that executable is in the `scheme-libs/racket/runner.rkt` file. I realize that isn't a good name, but it can be changed later.
- The native runtime has been plumbed around in various places, although it isn't currently used
- Various fixes to the actual loading and evaluation of code at runtime have been implemented, so that the jit tests can be run completely via the executable that works completely via that mechanism. They still aren't in CI, though, as that would require figuring out the steps needed to build and call the executable in CI, which I haven't completely worked out yet.

I was hoping to get compilation done, too, but I realized yesterday that I hadn't actually tested running the jit tests via the new method, and it took me until quite late yesterday to figure out what all the problems were. So I'll have to leave that to another day.

I haven't yet bumped the share code for generating scheme, so I'll get to that shortly after this PR is open. The effects of this PR should be minimal, though, as right now this is just new infrastructure that isn't used. I just wanted to get it somewhere besides my own machine/branch.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Decompile.hs
+++ b/parser-typechecker/src/Unison/Runtime/Decompile.hs
@@ -6,6 +6,7 @@
 
 module Unison.Runtime.Decompile
   ( decompile
+  , DecompResult
   , DecompError (..)
   , renderDecompError
   ) where
--- a/parser-typechecker/src/Unison/Runtime/Decompile.hs
+++ b/parser-typechecker/src/Unison/Runtime/Decompile.hs
@@ -5,14 +5,14 @@
 {-# LANGUAGE ViewPatterns #-}
 
 module Unison.Runtime.Decompile
-  ( decompile
-  , DecompResult
-  , DecompError (..)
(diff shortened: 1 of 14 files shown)


pchiusano (approved):
nice, feel free to merge this @dolio

PR #4529, opened 2023-12-19
Fix bug in `Bytes.fromBase16` when chunks have an odd size

### LINKED ISSUE #4527: Nat.fromHex fails for some input when it is a substring of larger Text
This code fails to parse valid hex code for the second last line in the input: 

```
process : Text -> Nat
process text =
  text
    |> Text.drop 6
    |> Text.take 5
    |> (x -> Nat.fromHex x |> getOrBug ("Failed to parse '" ++ x ++ "'"))

> List.map process (Text.lines input)

input = """
      3e6b8--
      2b1de--
      1465b--
      549d2--
      aa999--
      27e7f--
      aa999--
      5a393--
      1465b--
      23ade--
      49b5b--
      39ad5--
      54c4f--
      49151--
      4f596--
      357be--
      4f214--
      258b4--
      4f957--
      1ea37--
      7be6c--
      67bd6--
      a3124--
      6bf7d--
      a3124--
      3464a--
      7be6c--
      33f09--
      5c053--
      4c4e1--
      611e9--
      12823--
      5ecef--
      73a8c--
      73a8b--
      11111--
      11112--
      11113--"""
```
The error message in ucm is 
```


  💔💥
  
  I've encountered a call to builtin.bug with the following value:
  
    "Failed to parse '11112'"
  
  
  Stack trace:
    bug
    #qgnfurir8d
    #keb2ngg77c
```
Removing earlier lines in the input makes the error disappear.

ucm version: 7778bdc (built on 2023-11-30)

runarorama:
Fixes https://github.com/unisonweb/unison/issues/4527

The implementation was making the assumption that the chunks of the rope backing the `Bytes` were valid input to `ByteArray.convertFromBase`, which they only are if the chunk size is an even number of bytes.

Replaced with an implementation that first flattens the rope and converts the whole array at once.

### DIFF
--- a/lib/unison-util-bytes/src/Unison/Util/Bytes.hs
+++ b/lib/unison-util-bytes/src/Unison/Util/Bytes.hs
@@ -370,13 +370,7 @@ arrayToChunk bs = case BA.convert bs :: Block Word8 of
 chunkFromArray = arrayToChunk
 
 fromBase16 :: Bytes -> Either Text.Text Bytes
-fromBase16 bs = case traverse convert (chunks bs) of
-  Left e -> Left (Text.pack e)
-  Right bs -> Right (fromChunks bs)
-  where
-    convert b =
-      BE.convertFromBase BE.Base16 (chunkToArray @BA.Bytes b)
-        <&> arrayToChunk @BA.Bytes
+fromBase16 = fromBase BE.Base16
 
 toBase32, toBase64, toBase64UrlUnpadded :: Bytes -> Bytes
 toBase32 = toBase BE.Base32


ceedubs (approved):

aryairani:
Sweet! Do we want some kind of regression test? Or what's the right strategy here.

pchiusano:
@runarorama Thanks for spotting and fixing! Our policy for bugfixes is to include a regression test. If that's not possible or is extra difficult / not worth it for some reason it should be noted in the PR.

Issue #4528, opened 2023-12-18
add.run `CallStack` error issue on 0.5.11 


rlmark:
When writing cloud services, you might choose to `add.run` the last deployment to add a `ServiceHash` into the project for others to call. Currently this causes the UCM to exit with the message 

```
   ( Id "cpkb5ammhcpec7ju8fc4qfshq1lvdq7elgcn79u25khg3jjmu6fbj7rbvt64pacemhfhp2vftngfk61nlotqqivg064f5p9ruhf7t60" 0 )
  CallStack ( from HasCallStack ):
    error

```

A transcript reproducing the issue is attached here. 

[addRunIssue.md](https://github.com/unisonweb/unison/files/13708944/addRunIssue.md)

aryairani:
@mitchellwrosen 🙏

rlmark:
I noticed @mitchellwrosen had mentioned issues with the `run` command in the PR. In case it helps, here's a markdown file with a failing `run` case. Running where the return type should be a unique type returns a failure. 
[addRunIssueMini.md](https://github.com/unisonweb/unison/files/13721587/addRunIssueMini.md)

PR #4530, opened 2023-12-19
bugtweak: force result of `synthesizeForce` earlier

mitchellwrosen:
## Overview

In investigating #4528, I discovered that the actual error occurs earlier (in `run`), but is only observed later (on `add.run`).

This PR contains the following slight behavior changes:

- We now force the result of a `synthesizeForce` call earlier, so that `run` fails instead of `add.run` in certain cases (see #4528 for one such case).
- Similarly, we now force a `Branch` to WHNF before putting it in the root branch TVar. Previously, an impure exception was being put in here, which caused a background thread to crash, which caused the whole application to crash. This was the cause of the same error being printed twice, for the transcript in #4528
- We now have a better uncaught exception handler that explains "you should never see this" along with a `show` and (if different) a `displayException` rendering of the exception. 

And the following refactorings:

- The `load`, `run` and `addRun` handlers were all factored out into separate files.

The diff is best viewed commit-by-commit. Here are the commits that contain changes that aren't just moving code around:

- https://github.com/unisonweb/unison/pull/4530/commits/68a208eec937ef845e7603529b4ace88dfaaf4a6
- https://github.com/unisonweb/unison/pull/4530/commits/0631bfecccc6df16d6c72a668bfaa01b45ca6c57
- https://github.com/unisonweb/unison/pull/4530/commits/00a5f53532bb521265651513c8f8453b8caf5ebc

### DIFF
--- a/unison-cli/src/Unison/Cli/MonadUtils.hs
+++ b/unison-cli/src/Unison/Cli/MonadUtils.hs
@@ -250,7 +250,7 @@ modifyRootBranch f = do
   rootVar <- use #root
   atomically do
     root <- takeTMVar rootVar
-    let newRoot = f root
+    let !newRoot = f root
     putTMVar rootVar newRoot
     pure newRoot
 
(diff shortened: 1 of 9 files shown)


aryairani (approved):
Nice, thanks.

Issue #4532, opened 2023-12-20
Feedback when a scratch file change is picked up


ceedubs:
Right now I have just a single function in my scratch file, but when I save my scratch file it takes > 6 seconds for output to show up in ucm (`UnisonFileChanged: 6.40 s (cpu), 6.40 s (system)`). I'm not sure how much of this time is spent parsing/typechecking/pretty-printing the error message. But sometimes I think that ucm didn't notice that I saved my scratch file, so I try saving it again.

Hopefully at some point this feedback becomes much more instantaneous. But for now, I think that ucm should immediately print something like `Loading changes detected in scratch.u ...` whenever it picks up on a scratch save event.

ceedubs:
I think that [this](https://github.com/unisonweb/unison/blob/715cf108b60102ffac9ffb9e71c9b5b53f11e37a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs#L222) might be the relevant bit of the ucm code. This might be a good first PR for a contributor?

ceedubs:
For what it's worth, I think that the slowness might be in pretty-printing certain error messages. I made a slight change to the code and while it still has a compile error, it only takes 240ms for changes to render in ucm.

unorsk:
@ceedubs I just made a [PR](https://github.com/unisonweb/unison/pull/4540) :)
One tiny concern that I have is that it'll also show the message when loading files with `load` from `ucm` But I think this is reasonable – since the chunk 'I found and typechecked these definitions' gets also shown.

PR #4540, opened 2023-12-22
Adding a message for when ucm started reloading changes

unorsk:
## Overview

Addressing this [issue](https://github.com/unisonweb/unison/issues/4532) 

## Implementation notes

The original issue refers to [this bit of ucm code](https://github.com/unisonweb/unison/blob/715cf108b60102ffac9ffb9e71c9b5b53f11e37a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs#L222) and it is indeed relevant, but for what I found it could be probably best to add the message one level deeper, so we could skip printing the message if the event was programmatically generated.

The change is trivial

## Test coverage

* Updated the transcripts
* Tested manually. Screenshot 2023-12-22 at 14 07 49

## Loose ends

None

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/Load.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/Load.hs
@@ -63,7 +63,7 @@ handleLoad maybePath = do
 
 loadUnisonFile :: Text -> Text -> Cli ()
 loadUnisonFile sourceName text = do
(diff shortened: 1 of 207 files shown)


ceedubs (approved):
Nice thank you!

unorsk:
> Nice thank you!

I forgot to set it as 'ready for review' 🙈

PR #4558, opened 2024-01-03
Fix TDNR bug

### LINKED ISSUE #4543: Baffling error messages when using / instead of Float./
To reproduce with UCM 0.5.12, I have to be inside a project that clones `yap`:
```
clone @dfreeman/yap bug4543/
```
Then I put the following in my scratch file:
```
test : Float -> Float
test x = x * (1.0 / x)

> test 1.0
```
which provokes this error:
```
dumpDataNoTag: bad closure: PAp (CIx (ReferenceDerived (Id "01j2nvlku4s0sgpbhstjegb1t5t9d8m8l1nfph194h019i252i70o" 0)) 609 0) [] (fromListN 0 [])
  expected type: ReferenceDerived (Id "nk9jfsoidsc5h3nhcf1p6528t6c5hqui3hridbvaqnruel4jns3qo6plgups2sgi82c9jgt9ba1qlkum1bdjdgp75h7si2thbo7tcfg" 0)
```

If I remove the explicit signature, it does not type-check at all:
```
The 2nd argument to `(Float./)`
  
            has type:  Bag a
      but I expected:  Float
  
     45 |   x * (1.0 / x)
```
(no idea where `Bag` comes from!)

Everything works if I change the `/` operator to `Float./`, or add a `use Float /` line, but without the qualified name, the error messages are just very surprising to me.

tstat:
fixes #4543

## Overview

We want to rank the suggestions by lib depth, so we need to retain the shallowest lib depth name.

## Test coverage

Manually tested with the example from #4543.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker.hs
+++ b/parser-typechecker/src/Unison/Typechecker.hs
@@ -328,7 +328,7 @@ typeDirectedNameResolution ppe oldNotes oldType env = do
       Context.InfoNote v loc ->
       Result (Notes v loc) (Maybe (Resolution v loc))
     resolveNote env (Context.SolvedBlank (B.Resolve loc n) v it) =
-      fmap (Just . Resolution (Text.pack n) it loc v . dedupe . join)
+      fmap (Just . Resolution (Text.pack n) it loc v . join)
         . traverse (resolve it)
         . join
         . maybeToList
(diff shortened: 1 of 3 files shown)


aryairani:
@tstat Sorry I didn't catch it in December but we can't use `project.create` in the normal transcripts unfortunately; mainly because we don't want to wait to pull `base` during CI, but also because `base` can change (like we're seeing in the CI failure here) and we don't want it to unless we are doing an integration test.

mitchellwrosen (approved):

PR #4571, opened 2024-01-03
feat: mention the name of the file we put failed upgrades in

### LINKED ISSUE #4473: upgrade command should tell you what file to edit
when an `upgrade` fails and dumps a bunch of code into a scratch buffer, it should tell you the name of the file like `edit` does.

mitchellwrosen:
## Overview

This PR makes `upgrade` tell you what file it puts busted stuff in, if it fails.

Fixes #4473

### DIFF
--- a/unison-cli/src/Unison/CommandLine/OutputMessages.hs
+++ b/unison-cli/src/Unison/CommandLine/OutputMessages.hs
@@ -2221,12 +2221,20 @@ notifyUser dir = \case
                   <> operationName
                   <> "again."
             ]
-  UpgradeFailure old new ->
-    pure . P.wrap $
-      "I couldn't automatically upgrade"
-        <> P.text (NameSegment.toText old)
-        <> "to"
-        <> P.group (P.text (NameSegment.toText new) <> ".")
+  UpgradeFailure path old new ->
+    pure $
+      P.wrap
+        ( "I couldn't automatically upgrade"
+            <> P.text (NameSegment.toText old)
+            <> "to"
+            <> P.group (P.text (NameSegment.toText new) <> ".")
+        )
+        <> P.newline
+        <> P.newline
+        <> P.wrap
+          ( "I added the problematic definitions to the top of"
+              <> P.group (prettyFilePath path <> ".")
+          )
   UpgradeSuccess old new ->
     pure . P.wrap $
       "I upgraded"
--- a/unison-cli/src/Unison/CommandLine/OutputMessages.hs
+++ b/unison-cli/src/Unison/CommandLine/OutputMessages.hs
@@ -2222,19 +2222,13 @@ notifyUser dir = \case
                   <> "again."
             ]
   UpgradeFailure path old new ->
-    pure $
-      P.wrap
-        ( "I couldn't automatically upgrade"
-            <> P.text (NameSegment.toText old)
-            <> "to"
-            <> P.group (P.text (NameSegment.toText new) <> ".")
-        )
-        <> P.newline
-        <> P.newline
-        <> P.wrap
-          ( "I added the problematic definitions to the top of"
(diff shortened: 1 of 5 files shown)


unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -2221,12 +2221,20 @@ notifyUser dir = \case
...
+      P.wrap
+        ( "I couldn't automatically upgrade"
+            <> P.text (NameSegment.toText old)
+            <> "to"
+            <> P.group (P.text (NameSegment.toText new) <> ".")
+        )
+        <> P.newline
+        <> P.newline
+        <> P.wrap
+          ( "I added the problematic definitions to the top of"
```
aryairani: How about "I added some definitions that I need your help with to the top of"

aryairani: It would be nice if we had a button on Github to re-run transcripts :)

mitchellwrosen: Yeah; hmm... I will make that change (in a bit, so I don't have to do it again; more on that below) but I do think we rather often go through this series of steps with this kind of issue:

1. Someone makes a ticket about a bad error message, and would probably accept any improvement over the status-quo, because the error message is bad
2. A PR goes up some time later with a wording reluctantly written by the PR author (whose main job seems to be just identifying where the strings should go and doing the grunt work of re-running transcripts, with no strong opinion about what the strings are)
3. A change request comes in, and now the PR author worries that more might follow; addressing any one in particular is tedious so it's best to wait until there is unanimous agreement (but it's also unclear who might ultimately have an opinion, and whether they have had an opportunity to see the PR)

So, I wonder if we'd save some time in the long run by ironing out error message wording before any work begins. I certainly could have done that here.

aryairani: As a general rule I think we've said we want to iron things out before work begins (or be comfortable changing it after the fact) so that sounds fine.

Wrt worries that more feedback might follow, unclear who might have an opinion: first review pass for ucm stuff will be me, plus anyone else you want; I'll tag whoever else I want to wait for.

As far as initial wording of messages, we are going for first person "I tried to blah", helpful/friendly, active voice, positive sentiment. So basically avoid the word "problematic". Beyond that, I would imagine not caring too much.

For fun, here's what ChatGPT 4 came up with:

> Can you help me verify or rewrite the following message into one that follows this guideline: first person, polite, friendly, active voice, complete sentences, positive sentiment.
> 
> ```
> "I couldn't automatically upgrade"
>             <> P.text (NameSegment.toText old)
>             <> "to"
>             <> P.group (P.text (NameSegment.toText new) <> ".")
>         )
>         <> P.newline
>         <> P.newline
>         <> P.wrap
>           ( "I added the problematic definitions to the top of"
>               <> P.group (prettyFilePath path <> ".")
> ```
> 

> Certainly! To rewrite your message following the guidelines of being in the first person, polite, friendly, using active voice, complete sentences, and maintaining a positive sentiment, consider the following version:
> 
> "I wanted to let you know that I was not able to automatically upgrade [old name]. However, I've carefully added the definitions that needed attention to the top of [file path]. I'm looking forward to resolving this and ensuring everything is up-to-date."
> 
> This revision maintains a friendly and positive tone, uses active voice, and is structured in complete sentences. Remember to replace "[old name]" and "[file path]" with the specific names and paths relevant to your message.

Lol. And then maybe edited to:
> I was not able to automatically upgrade [old name] to [new name]. However, I've added the definitions that need attention to the top of [file path].

mitchellwrosen: Done

aryairani (approved):

Issue #4556, opened 2024-01-02
parse error on "ambiguous" reference to alias


mitchellwrosen:
When investigating #4515 we came across an unrelated issue: aliases defined in scratch files can cause problems. For example, this won't compile:

```
foo.hello = 5
bar.hello = 5

donk = hello
```

The error message's precise wording is confusing, but we essentially say we think the `hello` suffix is ambiguous because you might mean `foo.hello` or `bar.hello`. But they're aliases, so we shouldn't have to disambiguate.

PR #4581, opened 2024-01-05
Work around alias parsing bug

tstat:
## Overview

Works around #4556 by printing enough segments to disambiguate aliases.

## Test coverage

A new transcript has been added

## Loose ends

This PR makes PPEs worse by printing more segments than before to work around a bug that is only present for aliases defined in a scratch file (rather than aliases in the codebase). So, we plan to follow this up with a PR that only prints these additional segments when printing names for terms/types defined in update/upgrade.

### DIFF
--- a/unison-src/transcripts/command-replace.output.md
+++ b/unison-src/transcripts/command-replace.output.md
@@ -47,7 +47,7 @@ Test that replace works with terms
 
 .scratch> view x
 
-  x : Nat
+  x : builtin.Nat
   x = 2
 
 ```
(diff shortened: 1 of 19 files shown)


unison-src/transcripts/command-replace.output.md
```diff
@@ -47,7 +47,7 @@ Test that replace works with terms
 
 .scratch> view x
 
-  x : Nat
+  x : builtin.Nat
   x = 2
```
aryairani, line 51: This seems unexpected, is it expected? @tstat

aryairani:
@tstat I'd like to estimate the level of effort for the next iteration PR before deciding whether to merge this one by itself — let me know?

tstat:
The next PR is basically ready, so we can close this one.

PR #4589, opened 2024-01-08
bugtweak: make suffixification a little worse when rendering unison files via update/upgrade

mitchellwrosen:
## Overview

This PR works around an issue we have related to aliases and suffixification of names, for which we haven't yet thought of an ideal solution.

The issue is as follows. If you write a Unison file with two terms that would hash to the same value, and they share a suffix, you are not able to refer to either definition by the shared suffix.

```unison
one.foo = 18
two.foo = 18

-- parse error; which "foo" did you mean, "one.foo" or "two.foo"?
hey = foo + foo
```

A user is probably not likely to hit this when writing Unison code from scratch. However, it's certainly possible to hit in `update` / `upgrade` code paths in which we render out a Unison file for you.

The current name suffixification mechanism _would_ render a definition like `hey` above as `foo + foo`. This is because we:

1. Take the fully-qualified name for a thing (say it's `one.foo`)
2. Consider all suffixes of the name, starting with the shortest (`foo`, then `one.foo`)
3. Stop whenever a suffix refers to the same set of references as the fully-qualified name

The workaround this PR implements is to adjust the name suffixification algorithm used when rendering a Unison file during `update` / `upgrade` to one that has a different stopping condition:

3. Stop whenever a suffix is unique among all names

## Loose ends

This adjustment to suffixification only needs to happen for names that are referring to something in the Unison file itself. It's perfectly fine to use the existing suffixification mechanism for names that are already in the namespace.

However, implementing that seems a lot trickier than just moving `update` / `upgrade` code paths over to use the worse suffixification mechanism for all names. So, we decided to stop work here (to get the fix in that prevents certain users from using `update` / `upgrade` at all), and can revisit if necessary.


aryairani:
I'd still like to estimate the level of effort for the de-suffixified-as-needed version before deciding whether to merge this one by itself — let me know?

mitchellwrosen:
@aryairani I'm not sure, a few days?

aryairani (approved):

PR #4605, opened 2024-01-10
bugfix: fix upgrade suffixifier

mitchellwrosen:
## Overview

#4589 accidentally only fully addressed the issue in `update`, not `upgrade`. This PR fixes the remaining sources of suffixify-by-name in  `upgrade`.

## Test coverage

There's a new transcript that demonstrates the behavior before and after the fix, in different commits.


aryairani (approved):

aryairani:
> before and after the fix, in different commits.

Thanks this is super helpful.

PR #4612, opened 2024-01-12
change pre-release tag name again to `trunk-build`

aryairani:
Apparently I need something that appears alphabetically later than "release", to work around a Github issue. Probably nobody cares, since we should be linking to the latest actual release on the website anyway, e.g. https://github.com/unisonweb/unison/releases/latest/download/ucm-linux.tar.gz

Loose ends: update homebrew formula too, but it's in a different repo.

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -162,7 +162,7 @@ jobs:
       - uses: "marvinpinto/action-automatic-releases@latest"
         with:
           repo_token: "${{ secrets.GITHUB_TOKEN }}"
-          automatic_release_tag: "pre-release"
+          automatic_release_tag: "trunk-build"
           prerelease: true
           title: "Development Build"
           files: |


pchiusano (approved):
I don't know if I understand this, but this is a funny GitHub quirk.

We don't want to have to remember to update the website every time there's a new release, so being able to use the dynamic url is better. Does this affect that somehow?

aryairani:
@pchiusano The link in my top comment (discovered by @ceedubs) is the dynamic url we want on the website long term. it's redirects to the latest actual release.

Currently the link on the website is to the latest _pre-release_ version, instead of the latest _release_ version, and should be switched to the link above. cc @rlmark Sorry we didn't know the new link in time for the last website update.

This PR doesn't affect any of that though, or really anyone but me — giving me the ability to tell at a glance which ucm release a given PR was first incorporated into.

PR #4613, opened 2024-01-12
feat: better suffixification in update and upgrade

mitchellwrosen:
## Overview

This PR continues the work of #4589 and #4605. Now (in `update` and `upgrade`), when rendering a collection of definitions, names that resolve to something in the namespace (not the file) will suffixify "by hash" (the familiar suffixify algorithm). It's only names that resolve to something in the file itself that are suffixified "by name" (which produces slightly worse names, but is necessary because aliases can share suffixes).

## Test coverage

See transcript updates for the gist of the user-facing changes.


aryairani (approved):
Amazing 🚀

PR #4636, opened 2024-01-24
Support tab-completion of branch relative paths

tstat:
## Test coverage

Manual testing. There are also some doc-tests on the parser.


aryairani:
@tstat I'm thinking let's merge this as-is for now and continue with the `update` fixes / `merge` work, but later and in the future, there's a [`tab-completion` transcript](https://github.com/unisonweb/unison/blob/trunk/unison-src/transcripts/tab-completion.output.md) that we should use for testing these.

aryairani (approved):

Issue #4456, opened 2023-12-01
strange type error


stew:
this code:

```
go x =
  let _ = do x
  ()
```
gives the error:

```
  IllegalContextExtension:
    context:
      Γ
    |start1|
    |𝕩2|
    'Inference Other-3
    'User "x"-5
    'Inference Ability-6
    'Inference Output-7
    x4 : x  element:
      a17 : '{𝕖} xtype ReferenceDerived (Id "00nv2kob8fp11qdkr750rlppf81cda95m3q0niohj1pvljnjl4r3hqrhvp1un2p40ptgkhhsne7hocod90r3qdlus9guivh7j3qcq0g" 0) -> ({[Var 'Inference Ability-13]} Var 'User "x"-5) is not well-formed wrt the context
```

stew:
and yeah, I got lost and forgot which language I was in when I write something like "let x = y in z" but, its a weird error

dolio:
My latest PR gives a different error message for this. Now it's:

```
  The last element of a block must be an expression, but this is a
  definition:
  
      2 |   let _ = do x
  
  Try adding an expression at the end of the block.
```

So I guess this is now some kind of parsing bug?

dolio:
I realized this is actually the correct error (I think), not a parsing bug. A more explicit parse is:

```
go x =
   (let _ = do x)
  ()
```

So, the first line is a let that ends in a definition, just like the error says.

dolio:
Should be fixed now.

Issue #4638, opened 2024-01-24
typechecker crash with `let ... in`


aryairani:
[Slack Message](https://unisoncomputing.slack.com/archives/CLGTK464E/p1706129916851039?thread_ts=1706129916.851039&cid=CLGTK464E)

dolio:
Should be fixed now.

PR #4642, opened 2024-01-25
Local definition generalization and possible TDNR improvement

### LINKED ISSUE #3752: Typechecker fails to generalize closed inner functions
Issue noticed by @SystemFw. Here's an example that doesn't typecheck. We define a polymorphic function, `id`, inside another function. If you annotate `id` it works fine, but 

```
foo = do
  id x = 
    _ = 1
    x
  id ()
  id "hi"
```

This gives a type error:

```
  The 1st argument to `id`
  
            has type:  Text
      but I expected:  Unit
  
      7 |   id "hi"
```

Implying that `id` isn't being generalized.

This example doesn't use TDNR (or I can't see how it does), which I thought could be a confounding factor,  so I'm not sure what's going on. 

The typechecker does have some logic in `synthesizeWanted` to generalize the types of inner definitions if they are closed and have no free variables, but for some reason it isn't being triggered here.

## Clues

Another clue is that if `id` is defined as `id x = x` (so without the extraneous binding `_ = 1`) then it works fine. Weird, right?

Second clue is that this example works, just removing the `do`: 

```
foo =
  id x = 
    _ = 1
    x
  id ()
  id "hi"
```

It's interesting that this second example works, since inferring the body of the `do` should go throught he same inference logic.

Possible idea: I wonder if when you add the `do` it actually ends up *checking* the body of the `do` lambda against some type signature with existentials in it, so it doesn't end up in the synthesize code path that does the generalization.

Not urgent but would be nice to get to the bottom of this.

dolio:
This PR is the result of my work investigating and fixing #3752. The fix to that was to generalize non-recursive `let` in the type _checking_ code path. However, in getting that to actually work, I had to tweak some other aspects of the type checker as well.

I factored out the code for generalizing `let` in the synthesis path. However, I believe it wasn't correct. For instance, I think it would generalize a let like:

    r = Scope.ref '(bug "whatever")

When you do this, you get `r : forall a. Ref {Scope} 'a`, which can then be used to coerce between any two types. To avoid this, I do not generalize when the right hand side requires abilities. Note though that this allows lambda terms to be generalized, so it's probably not a very significant limitation.

I think the old generalization code would also produce bad TDNR results. There was a check to see if the RHS was, 'closed,' which meant it contains free variables that weren't previously type checked. However, TDNR candidates are not actually represented as variables, they are 'blank' terms. So the old code would generalize, and probably prevent any information from flowing into the actual TDNR solution.

Instead, my new strategy is to only generalize over _some_ variables (and the 'closed' check has just been removed). The context actually keeps track of which variables come from a TDNR candidate. So, when we move out of the scope of the binding's body, I look at the discarded context, and find all variables that are used in the TDNR candidate's type. These are _not_ generalized over, while other variables are. This allows the body of the let to contribute to solving the TDNR candidate's type.

The handling of the saved TDNR data has also been improved. Whenever we discard a portion of context with a TDNR type in it, we take a snapshot of the type at that point. However, this is just as problematic as the generalization above would be. The snapshot is never refined by later instantiations. So, I have updated it so that when we do context discards, we substitute into the TDNR info, which lets us retain any information we've figured out about the types.

Fixes #3752 
Improves the error message in #4456

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -639,6 +639,16 @@ modifyContext f = do
 appendContext :: (Var v, Ord loc) => [Element v loc] -> M v loc ()
 appendContext = traverse_ extendContext
 
+markRetained :: (Var v, Ord loc) => Set v -> M v loc ()
+markRetained keep = setContext . marks =<< getContext
+  where
+    marks (Context eis) = Context (fmap mark eis)
+    mark (Existential B.Blank v, i)
+      | v `Set.member` keep = (Var (TypeVar.Existential B.Retain v), i)
+    mark (Solved B.Blank v t, i)
+      | v `Set.member` keep = (Solved B.Retain v t, i)
+    mark p = p
+
 extendContext :: (Var v) => Element v loc -> M v loc ()
 extendContext e =
   isReserved (varOf e) >>= \case
@@ -1191,7 +1201,7 @@ synthesizeWanted tm@(Term.Request' r) =
   fmap (wantRequest tm) . ungeneralize . Type.purifyArrows
     =<< getEffectConstructorType r
 synthesizeWanted (Term.Let1Top' top binding e) = do
-  (tbinding, wb) <- synthesizeBinding binding
+  (tbinding, wb) <- synthesizeBinding top binding
   v' <- ABT.freshen e freshenVar
   when (Var.isAction (ABT.variable e)) $
     -- enforce that actions in a block have type ()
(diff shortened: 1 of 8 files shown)


pchiusano (approved):
Nice! Glad this worked out.

aryairani:
Heads up that CI failed in the jit tests, not sure if it's a false positive or what: https://github.com/unisonweb/unison/actions/runs/7658151250/job/20870460454?pr=4642

dolio:
Yeah, I saw. I think something in there is more monomorphic than it previously was. So either I need to add a signature or tweak my new strategy a little more.

dolio:
I realized, the error message is accurate in 4456 with this PR. So:

Fixes #4456 
Fixes #4638 (a duplicate)

dolio:
Okay, hopefully this is complete now (once I fix some warnings).

I realized that I was missing some parts of the context that would be relevant to refining the TDNR solutions. So it was possible to confuse the checker by nesting lets and stuff within one another. For example, a case like:

```
foo i =
  bar j =
    k = i + j
    k
  bar 1
```

Would not be able to figure out `+`. So, I added a new sort of `Blank` to indicate that we should keep a variable around for informing TDNR, and re-mark some upper portions of the context with these if they're generic blanks. Let me know if something doesn't look right with that.

pchiusano:
Nice. I'd say merge once CI passes.

PR #4643, opened 2024-01-26
Don't implicitly close transclude blocks

### LINKED ISSUE #4592: Roundtrip bug in docs
version:
release/0.5.12

https://share.unison-lang.org/@alvaroc1/draw/code/main/latest/terms/@rib3qhc3jstr533nnr8vf28t71b7mv2ptaaps9k5aa4hemlnhen64p11acaeq2ec0h8g6n4dqgbg87bpbhdtqjkhgk5uu9caa5q1n1g

Renders like this, but is rejected by the parser because of the indentation:

```unison
examples.carDrawing.doc : Doc
examples.carDrawing.doc =
  {{ {{ embedSvg
    (Draw.toSvg 600 200 do
      Draw.scale 4.0 4.0
      !carDrawing) }} }}
```

the original was indented:

examples.carDrawing.doc : Doc
```unison
examples.carDrawing.doc : Doc
examples.carDrawing.doc =
  {{ {{ embedSvg
          (Draw.toSvg 600 200 do
            Draw.scale 4.0 4.0
            !carDrawing) }} }}
```

similar issue here:
https://share.unison-lang.org/@alvaroc1/draw/code/main/latest/terms/Rgba/doc

```
Rgba.doc : Doc
Rgba.doc =
  {{
  Color in RGBA where each value is between 0 to 255.
  
  * red
  * green
  * blue
  * alpha
  
  {{
  docTable
    ([ 'black
    , 'blue
    , 'brown
    , 'green
    , 'red
    , 'Rgba.tan
    , 'transparent
    , 'white
    , 'yellow
    ]
      |> (lib.base.data.List.map cases
           color ->
             c = !color
             [ Special (SpecialForm.Link (Right (Term.Term (Any color))))
             , {{
               {{ docLink (docEmbedTermLink 'c) }}
               }}
             , Rgba.toDoc !color
             ])) }}
  }}
  ```

tstat:
fixes #4592

## Overview

I'm not familiar with the lexer, but after some debugging it seems like this is a lexing bug where an additional `Close` lexeme is emitted after the deindent on the `(Draw.toSvg` line (in the code from #4592). Tracking this down, the `Close` lexeme is emitted [from `token` on this line](https://github.com/unisonweb/unison/blob/2ab2ef5142b390e99b2280a4b45b24295f58169f/unison-syntax/src/Unison/Syntax/Lexer.hs#L373) as `token` calls `token''` which may emit closing tokens as seen [here](https://github.com/unisonweb/unison/blob/2ab2ef5142b390e99b2280a4b45b24295f58169f/unison-syntax/src/Unison/Syntax/Lexer.hs#L239-L261).

So, this change instructs `token''` to not implicitly close transclude blocks. Someone more familiar with the lexer should give this list a look though. I don't know how many open block types we have since they are distinguished by strings, so I don't know if this list is complete.

## Test coverage

There is a new transcript

### DIFF
--- a/unison-src/transcripts/fix4592.md
+++ b/unison-src/transcripts/fix4592.md
@@ -1,8 +1,8 @@
 ```ucm:hide
-.> builtins.merge
+.> builtins.mergeio
 ```
 
-```unison:error
+```unison
 doc = {{ {{ bug "bug"
   52 }} }}
 ```
--- a/unison-src/transcripts/fix4592.md
+++ b/unison-src/transcripts/fix4592.md
@@ -1,0 +1,8 @@
+```ucm:hide
+.> builtins.merge
+```
+
+```unison:error
+doc = {{ {{ bug "bug"
+  52 }} }}
+```
--- a/unison-src/transcripts/fix4592.output.md
+++ b/unison-src/transcripts/fix4592.output.md
@@ -1,0 +1,15 @@
+```unison
+doc = {{ {{ bug "bug"
+  52 }} }}
+```
+
+```ucm
+
+  Loading changes detected in scratch.u.
+
+  I found a closing '}}' here without a matching 'syntax.docTransclude'.
+  
+      2 |   52 }} }}
+  
+
+```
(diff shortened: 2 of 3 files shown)


aryairani:
I asked @runarorama or @pchiusano to take a peek, and then I tagged it to merge after that.

pchiusano (approved):
This makes sense. Transclusion is closed explicitly by the `}}`, not by unindenting. I think this was just an oversight.

PR #4648, opened 2024-01-29
Delete invalid remote mappings on push

tstat:
## Overview

If a branch remote mapping points to a deleted project or branch then the user needs to update their remote mapping somehow. We currently don't have a way, but it makes sense to drop the existing mapping as it will not be useful. With this change a subsequent push without an explicit target will succeed.

## Test coverage

Manual testing

### DIFF
--- a/unison-cli/src/Unison/CommandLine/OutputMessages.hs
+++ b/unison-cli/src/Unison/CommandLine/OutputMessages.hs
@@ -1933,6 +1933,13 @@ notifyUser dir = \case
   RemoteProjectBranchDoesntExist host projectAndBranch ->
     pure . P.wrap $
       prettyProjectAndBranchName projectAndBranch <> "does not exist on" <> prettyURI host
+  RemoteProjectBranchDoesntExist'Push host projectAndBranch ->
+    pure . P.wrap $
+      "Pushing failed because the target"
+        <> prettyProjectAndBranchName projectAndBranch
+        <> "does not exist on"
+        <> P.group (prettyURI host <> ".")
+        <> "This invalid default push target has been removed."
   RemoteProjectBranchHeadMismatch host projectAndBranch ->
     pure . P.wrap $
       prettyProjectAndBranchName projectAndBranch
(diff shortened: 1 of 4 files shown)


unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1933,6 +1933,13 @@ notifyUser dir = \case
   RemoteProjectBranchDoesntExist host projectAndBranch ->
     pure . P.wrap $
       prettyProjectAndBranchName projectAndBranch <> "does not exist on" <> prettyURI host
+  RemoteProjectBranchDoesntExist'Push host projectAndBranch ->
+    pure . P.wrap $
+      "Pushing failed because the target"
+        <> prettyProjectAndBranchName projectAndBranch
+        <> "does not exist on"
```
ChrisPenner: Is there something we can say here to make it clearer in the case where I:

1. Push to `@project/branch`,
2. Delete `@project/branch`
3. `push`

Because in that case it'll say that the branch doesn't exist, but in all other cases it just creates the branch for you automatically, I think this'll be confusing because UCM hasn't stopped them from pushing branches that don't exist in other cases.

We can also probably indicate to the user that the branch doesn't exist because it was deleted, since that's the only way we can currently get into this state. E.g.

"The previous push target has been deleted from the remote server. I've deleted the invalid push target, if you'd like to push to a new branch named "@project/branch" you can run the `push` command again and I'll push to the new branch at that target or create one if it doesn't already exist"

ChrisPenner (commented):
This is a big help!

I almost wonder if we should actually just create the new branch and push in this case?

The only time we should end up in this case is if the project or branch DID exist, but then was deleted, so we should probably either tell the user that a branch by that name DID exist but was deleted and they can push again to re-create it, or just go ahead and automatically create/push-to the new branch for them.

tstat:
> I almost wonder if we should actually just create the new branch and push in this case?

I wondered the same, and initially implemented this, but thought about it a bit more and reverted. It doesn't seem to me to be clearly desirable. For example, if someone is collaborating on a project with others and the project or branch is deleted by a collaborator then it seems to me that the user ought to be alerted to this during a push, rather than recreating the project and being confused about its state.

unison-cli/src/Unison/CommandLine/OutputMessages.hs
tstat: Yeah, I agree the wording could be improved and I like explicitly saying the old branch was deleted and suggesting pushing again. How about:

push_000

aryairani (approved):
Thanks!

PR #4661, opened 2024-02-01
Fix bug in `update`

### LINKED ISSUE #4618: replace `Set.findMin` with `lookupMin` in `Update2.hs`
`findCtorNames` assumes that only `Name`s from the supplied `Names` will be passed in, but that appears to not be the case for the usage in `saveTuf`.

So I think we need to change the type of `findCtorNames` to indicate a failed lookup, but the implementer will have a better idea.

image

Context https://unisonlanguage.slack.com/archives/CLUNF0J5S/p1705350390415869

tstat:
## Overview

Fixes #4618. There was an unsafe assumption in `saveTuf` that is detailed in the code and commit message. That assumption has been removed so we shouldn't trigger this bug again. Additionally, the bug had a poor error message generated by `findMin` which has been improved.

## Test coverage

There is a new transcript.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/Update2.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/Update2.hs
@@ -207,7 +207,7 @@ saveTuf getConstructors tuf = do
 typecheckedUnisonFileToBranchUpdates ::
   (forall void. Output -> Transaction void) ->
   -- | Returns 'Nothing' if the decl isn't in namesExcludingLibdeps,
-  -- in which case we know the decl in new and do not need to generate
+  -- in which case we know the decl is new and do not need to generate
   -- delete actions for it.
   (Name -> Either Output (Maybe [Name])) ->
   TypecheckedUnisonFile Symbol Ann ->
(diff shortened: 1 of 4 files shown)


unison-cli/src/Unison/Codebase/Editor/HandleInput/Update2.hs
```diff
@@ -205,7 +206,10 @@ saveTuf getConstructors tuf = do
 --     [ ("foo.bar", insert-term("baz",<#foo>)) ]
 typecheckedUnisonFileToBranchUpdates ::
   (forall void. Output -> Transaction void) ->
-  (Name -> Either Output [Name]) ->
+  -- | Returns 'Nothing' if the decl isn't in namesExcludingLibdeps,
+  -- in which case we know the decl in new and do not need to generate
```
aryairani:
```suggestion
  -- in which case we know the decl is new and do not need to generate
```

unison-cli/src/Unison/Codebase/Editor/HandleInput/Update2.hs
```diff
@@ -362,9 +366,13 @@ forwardCtorNames names =
     ]
 
 -- | given a decl name, find names for all of its constructors, in order.
+--
+-- Precondition: 'n' is an element of 'names'
```
aryairani, line 370: 🙏

aryairani (approved):

Issue #4709, opened 2024-02-19
Ed25519 sign/verify builtins


pchiusano:
This is needed for many applications (Discord webhook auth came up), and I don't think we should try to implement in pure Unison right now, same as other crypto primitives. 
```
-- sign.impl secretKey publicKey message
crypto.Ed25519.sign.impl : Bytes -> Bytes -> Bytes -> Either Failure Bytes

-- verify.impl publicKey message signature
crypto.Ed25519.verify.impl : Bytes -> Bytes -> Bytes -> Either Failure Bool
```

The `Failure` is because not all `Bytes` correspond to a valid public key or secret key.

In base, I'd wrap these in more convenient typed interfaces:

```haskell
-- SecretKey, PublicKey, and Signature are just newtypes around Bytes

crypto.Ed25519.sign : SecretKey -> PublicKey -> Bytes ->{Exception} Signature

crypto.Ed25519.verify : PublicKey -> Bytes -> Signature ->{Exception} Boolean 
```

It might not be worth newtyping SecretKey, PublicKey, and Signature, depending on taste. There's also the question of whether we have a separate newtype for all the different public key algos, when more are added.

Following:

https://hackage.haskell.org/package/cryptonite-0.5/docs/Crypto-PubKey-Ed25519.html#t:PublicKey

@jaredly does this seem easy to implement in Racket?

jaredly:
So the racket library [crypto](https://docs.racket-lang.org/crypto/pk.html#%28idx._%28gentag._8._%28lib._crypto%2Fscribblings%2Fcrypto..scrbl%29%29%29) supports es25519, but it's [LGPL](https://github.com/rmculpepper/crypto/blob/129997ab537db2eea142b5eb25cf2fc74a19d4ec/crypto-lib/COPYRIGHT).
It [looks like](https://github.com/rmculpepper/crypto/blob/129997ab537db2eea142b5eb25cf2fc74a19d4ec/crypto-lib/private/decaf/ffi.rkt#L78) that library uses [libdecaf](https://github.com/krionbsd/libdecaf) under the hood, which shouldn't be too hard to write FFI for

jaredly:
Update on this: turns out openssl's `libcrypto` supports es25519, so I'm doing FFI to that.

jaredly:
[...]

PR #4717, opened 2024-02-23
Ed25519 builtins for Haskell

dolio:
This PR implements the Ed25519 builtins for #4709 in the Haskell interpreter.

I added a simple transcript that signs some bytes and verifies the signature. The secret/public keys are just an arbitrary pair that I generated for the test (despite the implementation only checking that the keys have the right length, you can't just use arbitrary byte strings).

I figured this was the more important part of the feature request, so the racket implementation is on hold until someone else gets to it, or I finish fixing the base tests on racket.


aryairani:
We'll also need to get the two functions and the type added to `base`.

aryairani:
Needs one more trunk merge and rerun `transcripts` due to a different merge to trunk just now. I’ll try to get to it if you don’t have it handy.

aryairani (approved):

etorreborre:
@dolio @aryairani pending the implementation of https://github.com/unisonweb/unison/issues/4709 for RSA, would it be possible to do the same for RSA signatures?

aryairani:
@etorreborre "the same" meaning... adding a test for verifying RSA signatures?

Edit: Oh, meaning adding builtins for RSA signatures

etorreborre:
Yes, exactly.

PR #4756, opened 2024-03-06
Fix a variable capture issue during let floating

### LINKED ISSUE #4746: Pattern match in ability handlers doesn't shadow constructor variables properly
Here's a quick demo of the issue:

```unison
ability Issue t where
  one : '{Issue t} () -> {Issue t} ()
  two : '{Issue t} () -> {Issue t} ()
  three : '{Issue t} () -> {Issue t} ()

x : '{Issue t} () -> {Issue t} ()
x _ = ()

works x = x + 1

run : '{Issue t} () -> '{Stream Text} ()
run s = 
  go = cases
    { one x -> resume } -> 
      emit "one"
      handle resume !x with go
    { two x -> resume } -> 
      emit "two"
      handle resume !x with go
    { three x -> resume } -> 
      emit "three"
      handle resume !x with go
    { _ } -> emit "done"
  do handle !s with go
  
> Stream.toList! <| run do
  Issue.one do
    Issue.two do
      ()
    Issue.three do
      ()
```

If you inspect the result (`["one", "done"]`), it becomes obvious that handler just skipped `two x -> resume` and `three x -> resume` cases. At first i couldn't figure out what the reason was ([neither did Runar](https://discord.com/channels/862108724948500490/1213451735977431051/1213489030956847135)), but in that example i was using `coerceAbilities`, and thought maybe it has to do something with it.

As it turns out, the issue is the definition of `x : '{Issue t} () -> {Issue t} ()
x _ = ()` outside of the handler.

I was expecting `x` to be a bound variable from constructor pattern match (`{ two x -> resume }`), and not a globally defined function. Just like it behaves in `works x = x + 1` function above.

By renaming function `x` to `xx` the result of the execution would be as expected (no cases got skipped):

```unsion
["one", "two", "three", "done"]
```

dolio:
This fixes a variable capture problem. If you had a definition like:

```unison
x = ...
h = cases
  { eff x -> resume } ->
    handle resume !x with h
```

It would get floated to:

```unison
x = ...
float resume = resume !x
h = cases
  {eff x -> resume } ->
    handle float resume with h
```

Because the binding of `x` by the match cases was not being taken into account.

Fixes #4746

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/ANF.hs
+++ b/parser-typechecker/src/Unison/Runtime/ANF.hs
@@ -234,7 +234,26 @@ enclose keep rec t@(Handle' h body)
     lamb
       | null evs = lam' a [fv] lbody
       | otherwise = lam' a evs lbody
+enclose keep rec t@(Match' s0 cs0) = Just $ match a s cs
+  where
+    a = ABT.annotation t
+    s = rec keep s0
+    cs = encloseCase a keep rec <$> cs0
 enclose _ _ _ = Nothing
+
+encloseCase ::
+  (Var v, Monoid a) =>
+  a ->
+  Set v ->
+  (Set v -> Term v a -> Term v a) ->
+  MatchCase a (Term v a) ->
+  MatchCase a (Term v a)
+encloseCase a keep rec0 (MatchCase pats guard body) =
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
Nice!

PR #4768, opened 2024-03-08
Push type checking through match statements

### LINKED ISSUE #4722: Type annotations not pushed down to branches of a `match`
I'll try to minimize this later, but given this code

CleanShot 2024-02-27 at 17 29 22@2x

I get an unexpected failure:
```

>               needs the abilities: {𝕖3480}
    but was assumed to only require: {𝕖3405, 𝕖3406, 𝕖3407, Remote, Orderator (SearchResult id3199 url3205 (Unit ->{Remote} content3196))}
  
  This is likely a result of using an un-annotated function as an argument with concrete abilities. Try adding an annotation to the function definition whose body is red.
  
    141 |       (Orderator.subtractBy SearchResult.ordering (go q1) (go q2)) --  : '{Remote, Orderator (SearchResult id url ('{Remote} content)} ())
```

Annotating the result on line 141 makes it work. 

But this seems like it shouldn't be necessary - there's a type annotation on the function, which consists of a pattern match. Therefore the type of the RHS of each pattern match should match the signature, so we could be doing a `check` here and avoiding inference.

Aside is I'm not sure why inference fails here, but I'm proposing we avoid inference entirely in this case. That is:

```
check (a -> match a with Case1 -> blah) (x -> y)
==>
check blah y
```

dolio:
This PR changes the way type checking works slightly. Previously, checking a `match` or similar would just use the default strategy of inferring and subtyping. Instead, it now will check each case against the expected type of the whole match block. This eliminates some situations where the inference would generate a type that could not be successfully subtyped against the expected type.

Fixes #4722

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -2442,6 +2442,19 @@ checkWanted want (Term.LetRecTop' isTop lr) t =
   markThenRetractWanted (Var.named "let-rec-marker") $ do
     e <- annotateLetRecBindings isTop lr
     checkWanted want e t
+checkWanted want e@(Term.Match' scrut cases) t = do
+  (scrutType, swant) <- synthesize scrut
+  want <- coalesceWanted swant want
+  cwant <- checkCases scrutType t cases
+  want <- coalesceWanted cwant want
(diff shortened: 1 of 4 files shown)


pchiusano (approved):
Nice! That was pretty straightforward. :)

PR #4770, opened 2024-03-08
list permissions after `raco distribute`

aryairani:


aryairani:
idk why that didn't work

PR #4775, opened 2024-03-11
YAML: try changing ormolu commit conditions

aryairani:
This PR (like trunk) doesn’t try to format any files that haven’t changed in the PR, so we won’t see it working yet.


ChrisPenner (approved):
Clever!

PR #4776, opened 2024-03-11
Create ormolu.yaml

aryairani:
A one-off action we can use to run ormolu on the entire codebase.
Creates a PR with the changes.

### DIFF
--- a/.github/workflows/ormolu.yaml
+++ b/.github/workflows/ormolu.yaml
@@ -25,7 +25,8 @@ jobs:
           mode: inplace
       - name: apply formatting changes
         uses: peter-evans/create-pull-request@v6
-        if: ${{ always() }}
+        # the previous step will "fail" if it makes any formatting changes
+        if: ${{ failure() }}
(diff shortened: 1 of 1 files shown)


ChrisPenner (approved):
Are you sure you want the `always()`? I think that'll create the pull-request even if ormolu fails, or am I mistaken?

aryairani:
> Are you sure you want the `always()`? I think that'll create the pull-request even if ormolu fails, or am I mistaken?

Right, good question. The ormolu step “fails” whenever it makes any formatting changes — exactly when we want the PR. 

Originally the plugin was read-only and would fail if there was anything that _needed_ to be reformatted, and that was it. I did a PR a while ago to add the option to actually _perform_ the formatting changes too, but I didn’t try to change anything about failing or not, because I figured someone would still want to distinguish “already formatted” from “formatted now”.

In hindsight, a real `output` value would be better than failing, but I didn’t know enough at the time.

PR #4778, opened 2024-03-11
fix condition on ormolu manual workflow

aryairani:
Also updates the ormolu version from 0.5.0.1 to 0.5.2.0, and tweaks the details of the PR creation to be better.

Overview: This (updated) workflow will autoformat all the Haskell files in the repo and creates a PR with the result. (In contrast to the newly re-enabled ormolu step in the CI workflow, which for performance reasons only autoformats the Haskell files that have changed in the corresponding PR, and then typically creates a new commit with the result.) This workflow can also be used on `trunk`, which is protected, because it creates a separate PR, though I could imagine updating CI.yaml to create a separate PR with updates for `trunk` where applicable.

An example of the PR it will create is https://github.com/unisonweb/unison/pull/4784, and we could merge that after this; its merge target is this branch but I didn't want to merge the diffs.

Controversial decisions:
I went back and forth between having subsequent runs of the workflow create new PRs or amend an existing PR if one exists, and ultimately chose the latter. Probably it doesn't matter, and we can always switch it if we turn out to dislike the behavior.

@ChrisPenner I'm confused about it, but I guess everything I said about the failure condition in https://github.com/unisonweb/unison/pull/4776#issuecomment-1989430770 in response to your comment on the first PR was wrong?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,7 @@ on:
   workflow_dispatch:
 
 env:
-  ormolu_version: "0.5.0.1"
+  ormolu_version: "0.5.2.0"
   racket_version: "8.7"
   ucm_local_bin: "ucm-local-bin"
   jit_version: "@unison/internal/releases/0.0.11"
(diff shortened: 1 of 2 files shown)


aryairani:
@tstat @mitchellwrosen Just a heads up that this exists, though it should rarely be needed.

ChrisPenner (approved):

PR #4779, opened 2024-03-11
Minor JIT fixes

dolio:
This fixes a couple problems that was causing nimbus to not build on the JIT.

- Universal comparison was broken for strings and bytes, causing some duplicate builtin references to be generated
- `Float.ceiling` and `unsafe.coerceAbilities` builtins were missing

With these changes, `nodeMain` _builds_. No guarantees about running.


aryairani:
Thanks @dolio cc @pchiusano

aryairani:
@dolio Do I need to re-run `@unison/internal` to use this, or can I simply substitute these changed files in?

aryairani (approved):

PR #4790, opened 2024-03-13
Fix many pattern not backtracking

dolio:
Turns out, `Pattern.many` had the same non-backtracking problem as the functions in #3540. This applies the same fix.

This change actually breaks the http library (which was accidentally relying on the erroneous behavior), so it shouldn't be merged yet. I'll let @ceedubs decide when it should be.

### DIFF
--- a/parser-typechecker/src/Unison/Util/Text/Pattern.hs
+++ b/parser-typechecker/src/Unison/Util/Text/Pattern.hs
@@ -157,7 +157,7 @@ compile (Many p) !_ !success = case p of
   Char cp -> walker (charPatternPred cp)
   p -> go
     where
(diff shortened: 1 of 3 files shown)


aryairani:
Thanks @dolio
@ceedubs Lmk how we should proceed

aryairani (approved):

ceedubs:
Should there be a test or two for this?

aryairani (changes requested):
@dolio Could you add a regression test somewhere, and then we'll get it merged.

ceedubs (approved):
Thanks!

aryairani (approved):

PR #4794, opened 2024-03-15
Update cabal.project

ceedubs:
It was missing a couple of packages and had an outdated lsp constraint.

### DIFF
--- a/contrib/cabal.project
+++ b/contrib/cabal.project
@@ -23,6 +23,8 @@ packages:
   lib/unison-util-cache
   lib/unison-util-relation
   lib/unison-util-rope
+  lib/unison-util-file-embed
+  lib/unison-util-nametree
 
   parser-typechecker
   unison-core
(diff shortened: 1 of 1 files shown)


tstat (approved):
Thanks for adding these. Dan may want to change the lsp constraint rather than dropping it for the "cabal without nix" workflow.

aryairani:
cc @dolio wdyt

dolio:
It looks fine to me. I personally use freeze files that I periodically update to match our stackage version, so I don't think removing the lsp constraint matters much for me.

ceedubs:
@dolio should your changes be committed and no longer housed within `contrib`? It seems unhelpful to leave every developer who wants to use Cabal on their own to keep these things updated.

dolio:
I don't make the file myself. stackage has freeze files for each of its resolver versions. The one that would match our current setup is:

https://www.stackage.org/lts-20.26/cabal.config

I just download that and put it in `cabal.project.freeze`. We could add instructions for that, or commit the file, I suppose.

PR #4799, opened 2024-03-17
build test and bench dependencies with "build dependencies"

aryairani:
I noticed that some dependencies were getting built after the "build dependencies" stage, and they are probably test and benchmark depdencies.


aryairani:
This should build after merging #4797

mitchellwrosen (approved):

PR #4818, opened 2024-03-22
bugfix: don't go looking for field names of a hash-only decl

mitchellwrosen:
## Overview

This PR fixes a bug wherein we at some point assert that a hash can be parsed as a `Name`, which it can't, tripping a call to `error`.

It was triggered by something like `view #somehash`, where `#somehash` is a single-constructor type declaration, for which we try to infer whether we should print with record syntax.

The issue: the "is this a record" code accepts a `HashQualified Name`, which can be `HashOnly`, to determine whether the decl with that hash is a record. But if it _is_ a `HashOnly`, we'll never infer it as a record, since we need to know its name in order to know where in the namespace to look for its fields.

So, the fix is to just bail out earlier in this record inference procedure, if we have a `HashOnly`, rather than continue on and ultimately erroneously cast the `#somehash` to a `Name`.

## Test coverage

I tested this manually.


parser-typechecker/src/Unison/DataDeclaration/Dependencies.hs
```diff
@@ -50,67 +50,76 @@ labeledDeclDependenciesIncludingSelfAndFieldAccessors selfRef decl =
...
   let vars :: [v]
-      vars = [Var.freshenId (fromIntegral n) (Var.named "_") | n <- [0 .. Type.arity typ - 1]]
-  hashFieldAccessors ppe typeName vars declRef dd
-    <&> \accs ->
-      Map.elems accs
-        & setOf (folded . _1 . to (Reference.DerivedId >>> Referent.Ref))
+      -- We add `n` to the end of the variable name as a quick fix to #4752, but we suspect there's a more
+      -- fundamental fix to be made somewhere in the term printer to automatically suffix a var name with its
+      -- freshened id if it would be ambiguous otherwise.
+      vars = [Var.freshenId (fromIntegral n) (Var.named ("_" <> tShow n)) | n <- [0 .. Type.arity typ - 1]]
```
mitchellwrosen, line 68: This just copies over a different fix, including its comment, to another place that it should have been applied, but was missed.

parser-typechecker/src/Unison/DataDeclaration/Dependencies.hs
```diff
@@ -50,67 +50,76 @@ labeledDeclDependenciesIncludingSelfAndFieldAccessors selfRef decl =
...
+  Just (setOf (folded . _1 . to Referent.fromTermReferenceId) accessors)
 
 -- | Generate Referents for all possible field accessors of a Decl.
--- Returns 'Nothing' if typechecking of any accessor fails (which shouldn't happen).
+--
+-- Returns @Nothing@ if inferring/typechecking of any accessor fails, which shouldn't normally happen, but does when
+-- record fields are higher rank, because the higher rank types can't be inferred.
+--
+-- See https://github.com/unisonweb/unison/issues/498
 hashFieldAccessors ::
```
mitchellwrosen, line 80: No changes to `hashFieldAccessors`, just refactoring.

parser-typechecker/src/Unison/Syntax/DeclPrinter.hs
```diff
@@ -166,41 +174,76 @@ fieldNames ::
...
-  Maybe [HQ.HashQualified Name]
-fieldNames env r name dd = do
-  typ <- case DD.constructors dd of
-    [(_, typ)] -> Just typ
-    _ -> Nothing
+  Maybe [Name]
+fieldNames env r hqTypename dd = do
+  -- If we only have a hash for the decl, then we can't know where in the namespace to look for the generated accessors,
+  -- so we just give up trying to infer whether this was a record (even if it was one).
+  typename <- HQ.toName hqTypename
```
mitchellwrosen, line 184: This is the fix

aryairani (approved):

PR #4823, opened 2024-03-22
A few jit nimbus related fixes/tweaks

dolio:
- Changes the naming convention for builtin types/data to avoid clashes between builtins and pregenerated data types. For instance, both had a `Value` type in scheme, because of the in-unison representation. Instead of just `unison-`, now either `builtin-` or `ref-` prefixes are used.

- Fixes several places where invalid values were being used in exceptions. Raw scheme values were being returned that do not correspond to unison values, and this was causing subsequent problems with reflection.

- Fixes some code loading operations returning in-unison `Reference` values instead of term link values. Parts of these are implemented in unison, so both types are in play, and need to be correctly mediated.

The changes involve share code, so libraries need to be regenerated.


ceedubs (approved):
Thank you!

PR #4826, opened 2024-03-22
Fix a bug in the jit's runtime code caching operation

dolio:
The jit's runtime code caching operation was written with the idea that had a return type of `Either [Link.Term] Unit`, but it just returns the list, and lets an empty list indicate success.

This should fix that issue.

Modified the jit tests to do a match on a `cache_` result that would bump into this.


ceedubs (approved):
🚀

PR #4829, opened 2024-03-24
split TH parts of unison-cli into unison-cli-main

aryairani:
## Summary

Attempt to minimize the amount that needs to be rebuilt in a "release", when we just want to make sure the `Version` is current, and possibly get the rest from a cache.

Some results (two runs each) for optimized builds, we see ~2 mins shaved off a best-case build time when everything is already cached.

|    | [trunk](https://github.com/unisonweb/unison/actions/runs/8427146937)  | [trunk (2)](https://github.com/unisonweb/unison/actions/runs/8427498957)| [split](https://github.com/unisonweb/unison/actions/runs/8427217188) | [split (2)](https://github.com/unisonweb/unison/actions/runs/8427505301) |
|  --- | ------------- | ------------- | --- | --- |
| unix       | 2m 7s  | 1m 59s | 8s  | 8s |
| mac       | 2m 57s  | 3m 33s | 1m 24s  | 40s |
| windows       | 3m 29s  | 3m 30s | 50s  | 50s |

## Background

(possibly misremembered)

We use Template Haskell to embed the git release name etc. into the UCM version.  However, build caching issues meant that the UCM version string wasn't reliably updated with each release.

So we had previously split `Main.hs` and `Version.hs` off into the `unison-cli` project, and during release builds, we `stack clean unison-cli` to make sure that `Version` current. But today, `unison-cli` includes over 100 modules and accordingly takes a while to build. An unnecessary while.

This PR splits a minimal `Main` and `Version` off into `unison-cli-main`, leaving `unison-cli` as a lib that supports both `unison-cli-main` and `unison-cli:transcripts`

It also splits off `cli-integration-tests` into its own `unison-cli-integration` package, since it only needs to be built for CI testing and not for releases, and would otherwise create a circular dependency between `unison-cli` and `unison-cli-main`.


stack.yaml
```diff
@@ -8,67 +8,68 @@ build:
   interleaved-output: false
 
 packages:
```
aryairani, line 10: hide [automatic] whitespace changes

mitchellwrosen (approved):

PR #4830, opened 2024-03-24
Split and jit

aryairani:
Combine #4828 and #4829 because it's too painful working on on #4828 without #4829, but I do want to get a review before merging.


aryairani:
I get this error: https://github.com/unisonweb/unison/actions/runs/8411703558/job/23031597844?pr=4830#step:10:26
> SubLibrary dependency is not supported, this will almost certainly fail

Although I don't see any failure, it's also not clear to me that there's any speedup yet, which is surprising, so I'm not sure what's going on.

aryairani:
> SubLibrary dependency is not supported, this will almost certainly fail

Whatever it is is supposed to be fixed in https://docs.haskellstack.org/en/latest/ChangeLog/#v2111-2023-05-18

aryairani:
merged into #4828

PR #4841, opened 2024-03-28
Flesh out jit universal comparison

dolio:
This PR adds various cases missing from the universal comparison function in the jit.

- Type and termlinks are now handled
- Partial applications and procedures are now compared based on their references
- It tries to handle different types of values equivalently to the interpreter

Some additional changes are bundled because I tweaked other stuff while debugging the equality. E.G. the `describe-value` code for debug printing will now identify unison tuples and print them in a visually nicer way. I also added more descriptive printing for `bug` errors, which were just opaque before when they bubbled up to the top level.

Tests are included for term/type link comparison, which was the original motivation for this work.

### DIFF
--- a/scheme-libs/racket/unison/boot.ss
+++ b/scheme-libs/racket/unison/boot.ss
@@ -545,11 +545,6 @@
     [(unison-termlink-con tyl i)
      (ref-referent-con (typelink->reference tyl) i)]))
 
-(define (list->unison-tuple l)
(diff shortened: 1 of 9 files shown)


aryairani (approved):

aryairani (changes requested):
Oops, I approved it, but it's missing the new `unison-src/builtin-tests/link-tests.u` file I guess?

aryairani (approved):

PR #4846, opened 2024-03-29
Split corrected many pattern into two patterns

dolio:
This rolls back the `Pattern.many` fix and instead adds a new builtin `Pattern.many.corrected` with the right behavior.

### DIFF
--- a/parser-typechecker/src/Unison/Util/Text/Pattern.hs
+++ b/parser-typechecker/src/Unison/Util/Text/Pattern.hs
@@ -12,7 +12,7 @@ data Pattern
   | Or Pattern Pattern -- left-biased choice: tries second pattern only if first fails
   | Capture Pattern -- capture all the text consumed by the inner pattern, discarding its subcaptures
   | CaptureAs Text Pattern -- capture the given text, discarding its subcaptures, and name the capture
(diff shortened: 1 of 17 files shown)


aryairani:
> boolean determines whether it works correctly

😅

parser-typechecker/src/Unison/Util/Text/Pattern.hs
```diff
@@ -12,7 +12,7 @@ data Pattern
   | Or Pattern Pattern -- left-biased choice: tries second pattern only if first fails
   | Capture Pattern -- capture all the text consumed by the inner pattern, discarding its subcaptures
   | CaptureAs Text Pattern -- capture the given text, discarding its subcaptures, and name the capture
-  | Many Pattern -- zero or more repetitions (at least 1 can be written: Join [p, Many p])
+  | Many Bool Pattern -- zero or more repetitions (at least 1 can be written: Join [p, Many p]); boolean determines whether it works correctly
```
aryairani:
```suggestion
  | Many Bool Pattern -- zero or more repetitions (at least 1 can be written: Join [p, Many p]); boolean determines whether it's the correct version (True) or the original (False).
```

aryairani:
Not time sensitive, but I think we can use a `Rename` directive to affect how `builtins.merge` names these.

aryairani (approved):

PR #4849, opened 2024-04-02
Fix some inconsistencies in the time functions in the jit

dolio:
Some racket timing functions return integers and others floating point, so some care is needed to make sure they work. This adds a designated internal unison time rep as well instead of just using the raw value returned by the racket function.

Added some jit tests for some of the builtin time functions that didn't have any.

This should fix an exception that was being encountered in nimbus, when one of the untested time functions was being called.

### DIFF
--- a/unison-src/builtin-tests/io-tests.u
+++ b/unison-src/builtin-tests/io-tests.u
@@ -20,6 +20,9 @@ io.tests = Tests.main do
     !io.test_isFileOpen
     !io.test_ready
     !io.test_now
+    !io.test_monotonic
+    !io.test_processCPUTime
+    !io.test_threadCPUTime
     !io.test_isSeekable
     !io.test_handlePosition
     !io.test_renameDirectory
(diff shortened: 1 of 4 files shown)


unison-src/builtin-tests/io-tests.u
```diff
@@ -79,6 +82,29 @@ io.test_now = do
...
+    Duration s ns
+      | (s == +0) && (ns == 0) ->
+        Tests.pass "!processCPUTime"
+      | otherwise ->
+        Tests.pass "!processCPUTime"
+
+io.test_monotonic = do
+  match !Clock.monotonic with
+    Duration s ns
+      | (s == +0) && (ns == 0) ->
```
ceedubs, line 104: Should the cases where everything returns 0 really pass the tests? Should they at least have a different message?

dolio: I didn't really put that in to test any specific case. I just put it there to possibly trigger more exceptions if the calling conventions are off. It might not be very effective, though, because it's hard to come up with things that will ensure that racket values are precisely `Nat` or `Int` due to the way the numerics work.

ceedubs:
I've been marked as a reviewer, but I don't feel like I have much relevant knowledge for this PR. As long as it's good with @aryairani and/or @stew it's good with me.

aryairani (approved):
I don't know any more than @ceedubs so we'll have to trust the tests for now.

PR #4864, opened 2024-04-05
Some jit fixes for cloud tests

dolio:
This fixes some issues that the cloud tests were bumping into on the jit.

- Some of the exception/failure related stuff was wrapping an extra `Any` around the stored value. The tests were then checking that it equaled some specific other value, but obviously it didn't because it had more wrappers.
- Some tests would fail when the suite was run consecutively too quickly. This was caused because apparently closing a TCP server socket doesn't immediately make the port available again, and you need to force using the port when creating the new socket. The Haskell runtime does this automatically, but it's configurable in racket.

This bumps the share dependency and requires re-generating the scheme libraries. The new share code no longer generates one of the builtins via wrapper code (it's been moved directly into the scheme libraries), and fixes the data info for the `Failure` data type.

I also added overrides for how a lot of internal representations of unison values are printed, so that the messages printed when exceptions are thrown should be somewhat more comprehensible in terms of unison values. E.G. links will print as their short hash or builtin name, and procedures should print as the short hash of their reference if possible.

### DIFF
--- a/.github/workflows/ci.md
+++ b/.github/workflows/ci.md
@@ -9,7 +9,7 @@ At a high level, the CI process is:
 Some version numbers that are used during CI:
 - `ormolu_version: "0.5.0.1"`
 - `racket_version: "8.7"`
-- `jit_version: "@unison/internal/releases/0.0.11"`
+- `jit_version: "@unison/internal/releases/0.0.14"`
 
 Some cached directories:
   - `ucm_local_bin` a temp path for caching a built `ucm`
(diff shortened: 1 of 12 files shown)


ceedubs (approved):
Reviewing the code is a bit beyond my abilities, but from the description this sounds good to me.

dolio:
@ceedubs I've been tagging you on these not necessarily expecting you to review them in depth. I more want you to be apprised of updates.

aryairani (approved):

PR #4872, opened 2024-04-09
try to create a `unison-runtime` mac binary for that can find `libb2`

aryairani:
hoping that this will actually cause libb2 to be copied into the redistributable
<!-- **Choose your PR title well:** Your pull request title is what's used to create release notes, so please make it descriptive of the change itself, which may be different from the initial motivation to make the change.

<!-- ## Overview

## Implementation notes

## Test coverage

## Loose ends

### DIFF
--- a/.github/workflows/bundle-ucm.yaml
+++ b/.github/workflows/bundle-ucm.yaml
@@ -166,6 +166,7 @@ jobs:
           packages: libb2-dev
           version: 1.0 # cache key version afaik
       - name: install libb2 (macos)
+        if: runner.os == 'macOS'
         run: brew install libb2
       - name: install unison racket lib
         if: steps.cache-racket-deps.outputs.cache-hit != 'true'
--- a/.github/workflows/bundle-ucm.yaml
+++ b/.github/workflows/bundle-ucm.yaml
@@ -159,17 +159,21 @@ jobs:
           distribution: "full"
           variant: "CS"
           version: ${{env.racket_version}}
-      - uses: awalsh128/cache-apt-pkgs-action@latest
+      - name: install libb2 (linux)
+        uses: awalsh128/cache-apt-pkgs-action@latest
         if: runner.os == 'Linux'
         with:
           packages: libb2-dev
           version: 1.0 # cache key version afaik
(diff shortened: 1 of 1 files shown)


.github/workflows/bundle-ucm.yaml
```diff
@@ -159,17 +159,21 @@ jobs:
...
           version: ${{env.racket_version}}
-      - uses: awalsh128/cache-apt-pkgs-action@latest
+      - name: install libb2 (linux)
+        uses: awalsh128/cache-apt-pkgs-action@latest
         if: runner.os == 'Linux'
         with:
           packages: libb2-dev
           version: 1.0 # cache key version afaik
+      - name: install libb2 (macos)
+        run: brew install libb2
```
aryairani, line 170:
```suggestion
        if: runner.os == 'macOS'
        run: brew install libb2
```

.github/workflows/bundle-ucm.yaml
```diff
@@ -159,17 +159,22 @@ jobs:
...
+      - name: install libb2 (macos)
+        if: runner.os == 'macOS'
+        run: brew install libb2
       - name: install unison racket lib
         if: steps.cache-racket-deps.outputs.cache-hit != 'true'
         run: raco pkg install --auto scheme-libs/racket/unison.zip
       - name: build unison-runtime
         run: |
           raco exe --embed-dlls --orig-exe scheme-libs/racket/unison-runtime.rkt
+          if [[ ${{runner.os}} = 'macOS' ]]; then install_name_tool -add_rpath $(brew --prefix)/lib scheme-libs/racket/unison-runtime
```
aryairani:
```suggestion
          if [[ ${{runner.os}} = 'macOS' ]]; then install_name_tool -add_rpath $(brew --prefix)/lib scheme-libs/racket/unison-runtime; fi
```

aryairani:
I think we need to do this `define-runtime-path` thing https://github.com/racket/racket/blob/f6fe40dde18004d71ece98a2d22de84e42e90bcf/racket/collects/openssl/libcrypto.rkt#L77-L86

PR #4887, opened 2024-04-16
Use a socket to communicate with the jit runtime process

dolio:
This reworks the `run.native` command and the racket backing to use a TCP socket for communication of code and results. This makes it possible for `run.native` execution to make use of standard in (previously was occupied by communication of the code to the runtime) and for actual execution results to be reported by ucm.

A little error message tweaking is included. I noticed that a lot of blank lines were being generated in some situations, so I switched to using `linesNonEmpty`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -471,12 +471,20 @@ jobs:
           echo "jit_dist_exe=$jit_dist_exe" >> $GITHUB_ENV
           echo "ucm=$ucm" >> $GITHUB_ENV
 
-      - name: cache jit binaries
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            scripts/get-share-hash.sh
+            scheme-libs
+            unison-src/builtin-tests/jit-tests.tpl.md
+            unison-src/transcripts-using-base/serialized-cases/case-00.v4.ser
+
+      - name: cache/restore jit binaries
         id: cache-jit-binaries
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
         with:
           path: ${{ env.jit_dist }}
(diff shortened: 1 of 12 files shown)


aryairani:
Looks like transcript mismatch failure (you can use `./scripts/check.sh` to test everything locally).

Also, it's hard to tell from the cli diff, but maybe one too many blank lines deleted from the output for how you want the output to ultimately look?

aryairani:
image

Could you add one blank line back between the indented error and the outdented stack trace

aryairani (approved):

aryairani:
Something (my fault) is causing the jit tests not to actually run in CI — I don't see them actually running on any of these builds, it's just getting a false pass. I will take a look into it tonight or tomorrow. Sorry :-\

dolio:
No worries.

aryairani:
@dolio Ok I think I fixed the test harness 🤞 

It's currently failing on a "contract violation" in a "cli args" test:
https://github.com/unisonweb/unison/actions/runs/8712142414/job/23897816805#step:18:118

image
image

dolio:
Okay, took me a bit to figure that one out. The problem was that the evaluated program was getting the `-p <port>` flags for the runtime communication. So I needed to remove the `unison-runtime` flags from the list after parsing them.

aryairani:
> Okay, took me a bit to figure that one out. The problem was that the evaluated program was getting the `-p <port>` flags for the runtime communication. So I needed to remove the `unison-runtime` flags from the list after parsing them.

Ah — do you like the convention `stack` and some programs use, where any args before `--` go to the runtime and anything after goes to the evaluated program?

aryairani:
Oh I see what you did there. 👍

dolio:
I think the racket `command-line` automatically handles `--`.

PR #4888, opened 2024-04-16
Decompile partially applied builtin functions

### LINKED ISSUE #4780: "Reference to unknown function" when infix function is used prefix
Scratch file:

```
> (Nat.+) 0
```

Output:

```


  ⚠️

  I had trouble decompiling some results.

  The following errors were encountered:
      A reference to an unknown function was encountered:
        ##Nat.+


    1 | > (Nat.+) 0
          ⧩
          (Nat.+)
```

Also the result is wrong.

dolio:
This adds a missing case in the decompiler for when builtin functions get partially applied. Previously the decompiler could only act on defined functions.

Fixes #4780


aryairani (approved):
Sweet.

Issue #4880, opened 2024-04-13
add proper `libb2` dependency to `unison-runtime`


aryairani:
@aryairani
>So there’s something about the structure of `unison-runtime.rkt` or a dependency that causes `libcrypto` and `libssl` to be bundled when we run `raco distribute unison-runtime`, but `libb2` is not bundled in the same way, which is causing headaches. @dolio Do you have any insights into this?

@dolio 
>Not really. I think the first two are used in 'standard' racket libraries, and maybe the last one is us using the racket FFI? So maybe it doesn't do the same bundling for the latter. Unsure if you can tell it otherwise.
>We do do the same sort of ffi import for `libcrypto` (in the same file), but I think we might also depend on racket libraries that use `libcrypto`, which might be the important factor.

@aryairani
>I skimmed the docs and didn’t see any mention; I wondered if you could take a look at the dependencies you think `libcrypto` or `libssl` might be coming from to see if there’s a magic pragma we would need to include to trigger the bundling for `libb2`

>I found this: https://github.com/racket/racket/blob/f6fe40dde18004d71ece98a2d22de84e42e90bcf/racket/collects/openssl/libcrypto.rkt#L77-L86
>Discord also says:
>>You probably need to use `define-runtime-path` to specify the location of `libb2`
>
>So we probably just need to add that.

dolio:
I created a branch `jit-bundle` that adds a `define-runtime-path` for the `libb2` library in the `crypto` file.

It builds successfully for me. However, it doesn't seem like I can test if it'll work, because `raco exe/distribute` doesn't bundle _any_ libraries in the generated directory for me. So you'll have to see if it does something on Mac, @aryairani.

aryairani:
@dolio Ok thanks, I will try it out.

It might be because I called
```
raco exe --embed-dlls --orig-exe scheme-libs/racket/unison-runtime.rkt
```
I'm not totally sure what those two flags do, or what OSs they have an effect on, but I'm guessing they're responsible for you not getting any libraries.

Or wait, I just realized you probably meant that on your local machine, when you run it without those flags, you also don't get any libraries bundled....

dolio:
Yeah, I don't get any libraries in the directory with just `raco exe .../unison-runtime.rkt` `raco distribute ...`.

I assume that on Linux all the libraries are 'system libraries' or whatever and assumed to be installed separately.

aryairani:
I just found this callout on the `raco distribute` page

[...]

PR #4892, opened 2024-04-17
Try define-runtime-path for libb2

aryairani:
See #4880 for discussion.

<!-- 

## Overview

## Implementation notes

## Test coverage

## Loose ends

### DIFF
--- a/.github/workflows/bundle-ucm.yaml
+++ b/.github/workflows/bundle-ucm.yaml
@@ -159,14 +159,15 @@ jobs:
           distribution: "full"
           variant: "CS"
           version: ${{env.racket_version}}
-      - uses: awalsh128/cache-apt-pkgs-action@latest
-        if: runner.os == 'Linux'
-        with:
-          packages: libb2-dev
-          version: 1.0 # cache key version afaik
       - name: install unison racket lib
         if: steps.cache-racket-deps.outputs.cache-hit != 'true'
         run: raco pkg install --auto scheme-libs/racket/unison.zip
+      - name: install libb2 (macos)
+        if: runner.os == 'macOS'
+        run: |
+          brew install libb2
+          ln -s "$(brew --prefix)"/lib/libb2.*.dylib \
+                "$(dirname "$(readlink -f "$(which raco)")")"/../lib/
       - name: build unison-runtime
         run: |
           raco exe --embed-dlls --orig-exe scheme-libs/racket/unison-runtime.rkt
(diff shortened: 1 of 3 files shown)


.github/workflows/bundle-ucm.yaml
```diff
@@ -159,14 +159,15 @@ jobs:
           distribution: "full"
           variant: "CS"
           version: ${{env.racket_version}}
-      - uses: awalsh128/cache-apt-pkgs-action@latest
-        if: runner.os == 'Linux'
-        with:
-          packages: libb2-dev
-          version: 1.0 # cache key version afaik
```
aryairani, line 166: We don't need libb2 for building, only for redistributing it. On Linux, we've [read](https://docs.racket-lang.org/raco/exe-dist.html) and seen that `.so` files don't get redistributed anyway like they do on macOS and Windows.

.github/workflows/ci.yaml
```diff
@@ -354,38 +354,25 @@ jobs:
...
+            scheme-libs
+      - name: look up hash for jit source generator
+        run: echo "jit_causalhash=$(scripts/get-share-hash.sh ${{env.jit_version}})" >> $GITHUB_ENV
       - uses: actions/cache@v4
-        name: cache jit source
+        name: cache auto-generated jit source
+        id: cache-generated-source
         with:
-          path: ${{ env.jit_src_scheme }}
-          key: jit_src_scheme-racket_${{env.racket_version}}.jit_${{env.jit_version}}-${{hashFiles('**/scheme-libs/**')}}
```
aryairani, line 363: We were previously using scheme source files to compute a hash for the cache key which indicated whether or not we should rebuild the jit source bundle. The scheme source files had not been checked out yet though, so the hash was empty.

.github/workflows/ci.yaml
```diff
@@ -528,15 +519,12 @@ jobs:
...
-        uses: actions/download-artifact@v4
-        with:
-          name: jit-source
-          path: ${{ env.jit_src_scheme }}
+            && (steps.cache-jit-binaries.outputs.cache-hit != 'true'
+                || steps.cache-jit-test-results.outputs.cache-hit != 'true')
+        run: |
+          brew install libb2
+          ln -s "$(brew --prefix)"/lib/libb2.*.dylib \
+                "$(dirname "$(readlink -f "$(which raco)")")"/../lib/
```
aryairani, line 527: [Apparently](https://docs.racket-lang.org/raco/exe-dist.html), `raco distribute` will only redistribute mac libraries if they are found in `<racket_dist>/lib`. So we link homebrew's libb2 there.

Separate issue, unfortunately homebrew's libb2 is not signed, which presents problems later when doing a manual mac install after downloading with the web browser.

PR #4896, opened 2024-04-19
try the buildjet-cache branch of unison/actions/stack

aryairani:


aryairani:
I merged https://github.com/unisonweb/actions/pull/2, meaning `buildjet/cache` becomes the default for Stack global and local caches without merging this PR.

PR #4897, opened 2024-04-19
Fix duplicate definitions in jit when using scratch files

dolio:
This fixes an error that would cause some definitions to be generated twice when jit evaluating with a scratch file.

Essentially, to encounter the problem you'd need to have a definition already in your codebase, _and_ the same definition in your scratch file. If the scratch file were loaded, then the jit would include both the codebase code _and_ the scratch file code in the list of definitions to be compiled. And this would result in racket throwing an error about duplicate definitions.

So, I'm de-duplicating the lists after concatenating them, preferring the scratch file, although I don't think that should actually matter.

This *might* fix some errors that @stew was seeing. I'm not sure if it can be automatically tested, though, because it requires a definition being duplicated between a codebase _and_ a scratch file.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Interface.hs
+++ b/parser-typechecker/src/Unison/Runtime/Interface.hs
@@ -31,6 +31,7 @@ import Data.Bytes.Get (MonadGet, getWord8, runGetS)
 import Data.Bytes.Put (MonadPut, putWord32be, runPutL, runPutS)
 import Data.Bytes.Serial
 import Data.Foldable
+import Data.Function (on)
 import Data.IORef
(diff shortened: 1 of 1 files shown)


aryairani:
> I'm not sure if it can be automatically tested, though, because it requires a definition being duplicated between a codebase _and_ a scratch file.

We could do a test like this with a transcript — I guess we would only run it alongside the jit tests though.

It would probably be good eventually to switch runtimes using an environment variable or a command-line flag as opposed to a command in UCM, and then we could just run all the tests with one setting for interpreter and another setting for jit.

aryairani (approved):

PR #4901, opened 2024-04-26
try libb2-1 instead of libb2-dev

aryairani:
1. Switched `libb2-dev` for `libb2-1` after #4900, to see if it would still build as well. Initially we couldn't tell because of a bug in the caching, which prevented a full build/test from being attempted.

2. However, by switching the cache keys in 473c7b4e65ecb68b1be6d6a2ca0df483e4a97350 / https://github.com/unisonweb/unison/actions/runs/8841328469, we do see a `jit-binary-macOS-12` which includes `libb2.1.dylib`.

3. Next, however, I refactored the CI workflow to try to fix the original caching bug, and the resulting builds no longer redistribute `libb2`.

4. I wondered whether this was due to #4900 breaking something for macOS, but given (2), it doesn't seem possible. I wondered whether it was due to (1), but given (2), it doesn't seem to be that either. So I must have broken something in (3), but I don't know what yet.

5. Okay, it seems that the answer is that `libb2` needs to be installed before `raco exe`; it's not enough for it to be installed before `raco distribute`.

6. And with (5), `libb2-1` with #4900 (1) is fine.

<!-- 

## Overview

## Implementation notes

## Test coverage

## Loose ends

### DIFF
--- a/.github/workflows/ci-build-jit-binary.yaml
+++ b/.github/workflows/ci-build-jit-binary.yaml
@@ -88,7 +88,7 @@ jobs:
         if: steps.cache-jit-binaries.outputs.cache-hit != 'true'
         shell: bash
         run: |
-          raco pkg install --auto --skip-installed --scope installation x509-lib
+          sudo raco pkg install --auto --skip-installed --scope installation x509-lib
           raco pkg install --auto --skip-installed "$jit_src"/unison
           raco exe --embed-dlls "$jit_src"/unison-runtime.rkt
           raco distribute -v "$jit_dist" "$jit_exe"
--- a/.github/workflows/ci-build-jit-binary.yaml
+++ b/.github/workflows/ci-build-jit-binary.yaml
@@ -99,7 +99,7 @@ jobs:
         shell: bash
         run: |
           echo "before:"
-          find "jit_dist"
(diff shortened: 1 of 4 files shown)


aryairani:
The tests got skipped because changing the yaml doesn't invalidate the jit test results. Maybe Docker is a better place to test it.

aryairani (commented):
tweak cache key

aryairani:
Some notes:
```
time brew install minimal-racket
real    0m15.298s
user    0m10.118s
sys     0m6.533s

bash-3.2$ time brew install libb2
real    0m3.627s
user    0m2.551s
sys     0m0.982s
bash-3.2$ 
```

aryairani:
```
$ time /usr/local/Cellar/minimal-racket/8.11.1/bin/raco pkg install --auto compiler-lib
real    0m11.968s
user    0m17.625s
sys     0m3.330s
```

aryairani:
```
$ time /usr/local/Cellar/minimal-racket/8.11.1/bin/raco pkg install --auto --skip-installed unison-jit-src/unison/
real    4m32.342s
user    8m10.447s
sys     0m43.179s
```

PR #4916, opened 2024-05-03
Fix representation mismatches with jit `bug` messages

### LINKED ISSUE #4914: "Contract violation" when JIT builtin test fails
If a JIT builtin fails, the output appears like this:

Screen Shot 2024-05-02 at 1 27 30 PM

Looks like its construction is

```scheme
(define-unison
  (builtin-bug x0)
  (let* ([x1 (string->chunked-string "builtin.bug")])
    (unison-POp-EROR x1 x0)))
```

but the other code involved

```scheme
(define (encode-error ex port)
  (match ex
    [(exn:bug msg val)
     (write-bytes #"\1" port)
     (write-string-bytes msg port)
     (write-value-bytes val port)]
    [else
     (write-bytes #"\2" port)
     (write-string-bytes (exception->string ex) port)])
  (void))
```

is expecting a string instead? I don't really speak Scheme. Doesn't appear to be a blocker for anything, but it did confuse me for a while trying to figure out why me passing my own chunked strings around in other parts of the code might have been triggering this.

dolio:
This fixes some mismatches in the jit with how error messages in `bug` calls were expected to be represented. Some places thought chunked strings, others thought normal strings. This was causing `bug` calls to error out the jit communication in a more severe way than intended.

Fixes #4914

### DIFF
--- a/.github/workflows/ci-build-jit-binary.yaml
+++ b/.github/workflows/ci-build-jit-binary.yaml
@@ -54,13 +54,13 @@ jobs:
         with:
           name: jit-source
           path: ${{ env.jit_src }}
-
+      
       - name: cache/restore jit binaries
         id: cache-jit-binaries
         uses: actions/cache/restore@v4
         with:
           path: ${{ env.jit_dist }}
-          key: jit-dist_${{matrix.os}}.racket_${{env.racket_version}}.jit-src_${{hashFiles(format('{0}/**.rkt',env.jit_src),format('{0}/**.ss',env.jit_src))}}.yaml_${{hashFiles('**/ci-build-jit-binary.yaml')}}
+          key: jit-dist_${{matrix.os}}.racket_${{env.racket_version}}.jit-src_${{hashFiles(format('{0}/**/*.rkt',env.jit_src),format('{0}/**/*.ss',env.jit_src))}}.yaml_${{hashFiles('**/ci-build-jit-binary.yaml')}}
 
       - name: cache racket dependencies
         if: steps.cache-jit-binaries.outputs.cache-hit != 'true'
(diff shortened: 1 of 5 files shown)


dolio:
I don't really understand what's going on with the jit tests. Is it still failing even though I put `ucm:error`?

No, I guess not. It's just not printing the good error message, and I don't understand why.

aryairani:
@dolio Do they pass for you locally? Or also fail but don't know why?

dolio:
They pass locally.

I think it's using a cached copy of the jit executable.

aryairani:
Okay, maybe so. I'll try to figure out why that would be.

aryairani:
@dolio Could you check if this jit source is up to date, or is it out of date?
https://github.com/unisonweb/unison/actions/runs/8944176269/artifacts/1471803398

dolio:
It looks up to date.

In the `build jit binary` step of the failed run, it looks to me like it's skipping the step that builds the jit executable (among others). At least if that's what the "no" symbol means.

aryairani:
> It looks up to date.
> 
> In the `build jit binary` step of the failed run, it looks to me like it's skipping the step that builds the jit executable (among others). At least if that's what the "no" symbol means.

Yes, that's what the symbol means, but it's only supposed to skip building if it already has a jit executable built from the same source files, same `.rkt` and `.ss` files. So I'm wondering if there something wrong with that test, or maybe if the source was out of date.

So you think the source is up to date, but is this binary also up to date? https://github.com/unisonweb/unison/actions/runs/8944176269/artifacts/1471804110 or does it behave differently from yours?

dolio:
Yes, it fails the test for this PR.

aryairani:
Huh, okay thanks. I'll dig a little deeper later today in that case.

aryairani:
Next steps I guess if someone wanted to try it before me, would be to use the `tmate` action to ssh in and see if the contents of the directories being hashed to decide whether to reuse a cached binary are wrong

Or we can also just delete the cached binary and re-run the tests

aryairani:
I'm trying with clearing the cache, and that seems to be working.

Seems like it will happen again until we can track it down though.

dolio:
Well, that worked for the Linux build, but the Mac build failed on `libb2` already existing or something?

aryairani:
Yeah, I'm guessing the symlink got sucked up into a cache too? Taking a look now.

aryairani:
Okay, I think I fixed both the caching issue and the libb2 issue.

dolio:
Sweet.

aryairani (approved):

PR #4918, opened 2024-05-06
Use full length hashes for jit definition names

dolio:
This should avoid any possibility of picking conflicting names for distinct functions. I'm not aware of any actual examples of this, but it could in principle happen.

I decided to use the full length hashes because picking just-long-enough names starts to become very cumbersome when doing dynamic loading. You need to keep track of all your past choices, not just the choices for the module you're loading, because the one you're loading could refer to things in previously loaded modules. It would be possible to use shorter names in a compile-time context, since you have all the names up front. But you'd still have to remember _those_ for runtime loading, and it seemed not worth it.

This doesn't include any custom error formatting, because I'd like to see error messages I get randomly before doing anything, but if they're too verbose, it's possible to override the Racket error reporting to produce shorter names there, rather than generating them up front.

### DIFF
--- a/.github/workflows/ci.md
+++ b/.github/workflows/ci.md
@@ -9,7 +9,7 @@ At a high level, the CI process is:
 Some version numbers that are used during CI:
 - `ormolu_version: "0.5.0.1"`
 - `racket_version: "8.7"`
-- `jit_version: "@unison/internal/releases/0.0.15"`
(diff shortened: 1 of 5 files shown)


aryairani:
@dolio If you merge trunk into here, it should fix CI

aryairani (approved):

PR #4922, opened 2024-05-08
Add LSP documentation for Emacs

sellout:
## Overview

Documents how to configure Unison LSP on Emacs.

This only documents Eglot usage, not lsp-mode. Eglot is much simpler and is included with Emacs as of Emacs 29.

### DIFF
--- a/docs/language-server.markdown
+++ b/docs/language-server.markdown
@@ -4,9 +4,11 @@
 
 * [Overview](#overview)
 * [Installation and setup](#installation-and-setup)
-  * [Settings](#settings)
   * [NeoVim](#neovim)
   * [VSCode](#vscode)
+  • [Helix Editor](#helix-editor)
+  * [other editors](#other-editors)
+* [Configuration](#configuration)
 
 ## Overview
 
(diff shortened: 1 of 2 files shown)


docs/language-server.markdown
```diff
@@ -207,6 +210,9 @@ Supported settings and their defaults. See information for your language server
 
 ```json
 {
+  // A suggestion for the formatter about how wide (in columns) to print definitions.
+  "formattingWidth": 80,
+
   // The number of completions the server should collect and send based on a single query.
   // Increasing this limit will provide more completion results, but at the cost of being slower to respond.
   // If explicitly set to `null` the server will return ALL completions available.
```
aryairani: I guess comments aren't legal in json? I know you didn't introduce them, and it should technically go into a separate PR, but since our eyes are here...
```suggestion
* `formattingWidth`: A suggestion for the formatter about how wide (in columns) to print definitions.
* `maxCompletions`: The number of completions the server should collect and send based on a single query.  Increasing this limit will provide more completion results, but at the cost of being slower to respond.
If explicitly set to `null` the server will return ALL completions available.

```json
{
  "formattingWidth": 80,
```

aryairani (approved):

docs/language-server.markdown
sellout: It looks Unison uses a non-squashing workflow, so I don’t have a problem with this being in the same PR (so long as the commits stay separate).

sellout: Ok, well, after applying your suggestion, I amended the commit to make the second paragraph of `maxCompletions` doc part of the list item. I _thought_ it would keep you as the author, but it didn’t (although either git or Magit at least added a `Co-authored-by` field).

I also tried to find a way to edit a suggestion before committing, but doesn’t seem like there is one.

Anyway, just some Git(Hub) bikeshedding. Hopefully not a big deal.

sellout:
I don’t know if it shows up when other people view the PR, but my fork implicitly runs [garnix CI](https://garnix.io/), and it looks like [some of the Nix packages aren’t currently building](https://github.com/sellout/unison/runs/24734320577), so CI looks like it’s failing.

aryairani:
@sellout No, it doesn't show garnix runs here.

Would you add yourself to `CONTRIBUTORS.markdown` too, acknowledging that your contributions are MIT-licensed. Sorry I forgot to mention earlier.

PR #4923, opened 2024-05-08
Repair the Nix flake

### LINKED ISSUE #4940: Nix flake is invalid
When I ran into the issue #4939 fixes, I started exploring the flake, and noticed some commands failing, notably `nix flake check`.

The flake contains nested outputs, which aren’t allowed.

sellout:
## Overview

This does the minimum to get `nix flake check` working.

The primary issue is that the standard flake outputs require flat package sets, and this flake produced nested ones.

Fixes #4940.

## Implementation notes

This merges package sets into a single one rather than making them separate entries in the outputs.

The only other change was to correct the attribute name used for the UCM docker image’s command.

This currently flattens the package sets and adds prefixes to them, for some conceptual groupings. The prefixes could be removed, though.

Also, if we wanted to maintain some of the nesting, we could use the `legacyPackages` output, but that is (as it says) legacy, and either way it’s breaking the names, so we might as well just adopt the current flat style.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,7 @@ dist-newstyle
 
 # Mac developers
 **/.DS_Store
-
 /libb2.dylib
+
+# Nix
+result
--- a/development.markdown
+++ b/development.markdown
@@ -126,9 +126,9 @@ This is specified with the normal
 
 Some examples:
 ```
-nix build '.#haskell-nix.unison-cli:lib:unison-cli'
-nix build '.#haskell-nix.unison-syntax:test:syntax-tests'
-nix build '.#haskell-nix.unison-cli:exe:transcripts'
+nix build '.#component-unison-cli:lib:unison-cli'
+nix build '.#component-unison-syntax:test:syntax-tests'
+nix build '.#component-unison-cli:exe:transcripts'
 ```
 
 ### Development environments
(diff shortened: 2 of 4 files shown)


sellout:
I can also remove the garnix config before merging. I mostly have it to make sure I’m getting decent coverage on the Nix bits. But it looks like there’s already some of that in GitHub workflows.

aryairani:
> I can also remove the garnix config before merging. I mostly have it to make sure I’m getting decent coverage on the Nix bits. But it looks like there’s already some of that in GitHub workflows.

Just spotted this. Yeah I feel a bit blocked on the garnix piece because I don't understand it and haven't set up a time to go over it.

I'd say let's remove that if it's not integral, and then we can figure out how to improve CI separately, which could include garnix, or sticking with what we have, or replacing some of what we have with something else.

sellout:
> Just spotted this. Yeah I feel a bit blocked on the garnix piece because I don't understand it and haven't set up a time to go over it.
> 
> I'd say let's remove that if it's not integral, and then we can figure out how to improve CI separately, which could include garnix, or sticking with what we have, or replacing some of what we have with something else.

Done.

aryairani (approved):

PR #4937, opened 2024-05-12
escape slashes in branch name for share api call

aryairani:
details at https://discord.com/channels/862108724948500490/1239231122924109896


aryairani:
tested in CI runs for #4935 and #4936

PR #4939, opened 2024-05-14
Add a `default` Nix package

sellout:
## Overview

development.markdown says to use `nix build` to build the `unison` executable, but it doesn’t work, because there is not a `default` package. This makes the behavior match the docs.


aryairani (approved):
¯\\\_(ツ)_/¯

PR #4944, opened 2024-05-14
feat: add optional argument to `test` command

### LINKED ISSUE #2190: Allow `test` command to take an optional namespace argument
So `test .foo.bar` runs all the tests in that namespace. Rather than needing to `cd .foo.bar` then `test` with no args.

mitchellwrosen:
## Overview

Fixes #2190 

This PR adds an optional relative path argument to `test`.

For example,

```
myproject/mybranch> test mynamespace
```

is like

```
myproject/mybranch> test
```

except only tests defined in the `mynamespace` sub-namespace of `myproject/mybranch` are run.

### DIFF
--- a/unison-cli/src/Unison/CommandLine/InputPatterns.hs
+++ b/unison-cli/src/Unison/CommandLine/InputPatterns.hs
@@ -2395,7 +2395,7 @@ test =
       help =
         P.wrapColumn2
           [ ("`test`", "runs unit tests for the current branch"),
-            ("`test foo`", "runts unit tests for the current branch defined in namespace `foo`")
+            ("`test foo`", "runs unit tests for the current branch defined in namespace `foo`")
           ],
       parse = \args ->
         maybe (Left (I.help test)) Right do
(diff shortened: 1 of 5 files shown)


aryairani (commented):
Looks good @sellout @mitchellwrosen 

Could you update the `test-command.md` transcript to demo this too?

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -2388,20 +2388,31 @@ debugNameDiff =
...
-              }
-    )
+    { patternName = "test",
+      aliases = [],
+      visibility = I.Visible,
+      args = [("namespace", Optional, namespaceArg)],
+      help =
+        P.wrapColumn2
+          [ ("`test`", "runs unit tests for the current branch"),
+            ("`test foo`", "runts unit tests for the current branch defined in namespace `foo`")
```
etorreborre: s/runts/runs

aryairani (approved):
Great!

PR #4948, opened 2024-05-14
A couple more miscellaneous JIT issues

dolio:
This fixes a couple problems I found by making ucm always use the native runtime when evaluating code normally (I.E. not replacing the 'sandboxed' runtime). I think this includes watch expressions and tests. I tried running all the transcripts this way.

Some of the problems cannot be easily fixed, because they are just things that aren't implemented yet, and aren't a priority. Most of the transcripts seem to pass, though.

The fixes are:
- Boolean deserialization wasn't quite working, because testing whether a value was a boolean wasn't quite correct on one end.
- Bytes serialization wasn't actually using the right format, so it was failing to roundtrip correctly, and could also fail to communicate back to ucm. These fixes are in `@unison/internal`.

### DIFF
--- a/.github/workflows/ci.md
+++ b/.github/workflows/ci.md
@@ -9,7 +9,7 @@ At a high level, the CI process is:
 Some version numbers that are used during CI:
 - `ormolu_version: "0.5.0.1"`
 - `racket_version: "8.7"`
-- `jit_version: "@unison/internal/releases/0.0.16"`
+- `jit_version: "@unison/internal/releases/0.0.17"`
 
 Some cached directories:
   - `ucm_local_bin` a temp path for caching a built `ucm`
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ on:
 env:
   ormolu_version: 0.5.2.0
   ucm_local_bin: ucm-local-bin
-  jit_version: "@unison/internal/releases/0.0.16"
+  jit_version: "@unison/internal/releases/0.0.17"
   jit_src_scheme: unison-jit-src/scheme-libs/racket
   jit_dist: unison-jit-dist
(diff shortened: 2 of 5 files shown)


aryairani:
> Most of the transcripts seem to pass, though.

Raises the question -- which transcripts don't seem to pass?

aryairani (approved):

pchiusano:
I'm going to guess that these were causing problems for cloud integration tests and we might get a lot further now! @ceedubs @stew

dolio:
@aryairani The transcripts that don't pass use `MVar`, `STM`, sized block buffering for file I/O (which Racket doesn't seem to have), continuation serialization, or old serialization versions.

It's possible those failures are masking other problems. But I don't know an easy way to tell.

pchiusano:
Continuation serialization doesn't work with the JIT yet? We are counting on that working to be able to do some interesting stuff, though I don't know if we rely on it anywhere currently.

aryairani:
> @aryairani The transcripts that don't pass use `MVar`, `STM`, sized block buffering for file I/O (which Racket doesn't seem to have), continuation serialization, or old serialization versions.

Apart from continuation serialization which we know we _do_ need, are there any of these listed that we don't need to support / don't plan to support in the near term? e.g. old serialization versions

Or do we need them all, but these ones are deprioritized according to what nimbus requires currently?

ceedubs:
I think that we already gave up on the serialization version that predated the change in `Nat`/`Int` serialization. That happened a little bit before Unison Cloud went public, so we thought that chances were slim that someone was depending on it.

We moved Unison Cloud away from STM in the past, with the expectation that it wouldn't be supported in the JIT. I think we might have moved off of MVar too; if not it should be easy to replace now that we have `Promise`.

Anyway that's just the Unison Cloud perspective. There are probably other views.

PR #4953, opened 2024-05-15
feat: add `lib.install` command

mitchellwrosen:
## Overview

install

Edit: the help now looks like this

2024-05-15-212736_1281x293_scrot

This PR adds a `lib.install` (alias: `install.lib`) command to replace `pull` for installing a dependency into `lib`.

It's different than `pull` in a couple ways:

- You don't provide a local name for the dependency; `pull @unison/base/releases/4.0.0 lib.base_4_0_0` just becomes `lib.install @unison/base/releases/4.0.0`, and the name is picked automatically (see screenshot).
- Because the name is picked manually, it always installs into an empty namespace. So, there's never a merge.

## Test coverage

I tested this manually

### DIFF
--- a/parser-typechecker/src/Unison/Codebase.hs
+++ b/parser-typechecker/src/Unison/Codebase.hs
@@ -91,12 +91,10 @@ module Unison.Codebase
     importRemoteBranch,
     Preprocessing (..),
     pushGitBranch,
-    PushGitBranchOpts (..),
 
     -- * Codebase path
     getCodebaseDir,
     CodebasePath,
-    SyncToDir,
 
     -- * Direct codebase access
     runTransaction,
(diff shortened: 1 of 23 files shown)


parser-typechecker/src/Unison/Codebase.hs
```diff
@@ -130,21 +124,13 @@ import Unison.Codebase.Branch (Branch)
...
 import Unison.Codebase.SqliteCodebase.Conversions qualified as Cv
 import Unison.Codebase.SqliteCodebase.Operations qualified as SqliteCodebase.Operations
-import Unison.Codebase.SyncMode (SyncMode)
-import Unison.Codebase.Type
-  ( Codebase (..),
-    GitError (GitCodebaseError),
-    PushGitBranchOpts (..),
-    SyncToDir,
-  )
+import Unison.Codebase.Type (Codebase (..), GitError)
```
aryairani, line 133: Could you add a high level description to the PR about what is happening with these

aryairani (commented):
Looks awesome!

parser-typechecker/src/Unison/Codebase.hs
aryairani: Ok I guess I found it in the commit message, or is there more to it?

aryairani (approved):

parser-typechecker/src/Unison/Codebase.hs
mitchellwrosen: The `SyncMode` argument was totally ignored, so `debug.pull-exhaustive` didn't do anything special.

hojberg:
@mitchellwrosen @aryairani 

I'm getting a PR going for share to update the install modals to this new command 🎉.
Let's coordinate when we cut a new UCM release.

aryairani:
@hojberg 
> I'm getting a PR going for share to update the install modals to this new command 🎉.
> 
> Let's coordinate when we cut a new UCM release.

Thank you!

hojberg:
Here's the PR: https://github.com/unisoncomputing/share-ui/pull/6

PR #4961, opened 2024-05-16
Fix jit `zlib-deflate`

dolio:
The procedure uses concurrency and streaming to handle the checksum, but it wasn't actually reading concurrently from the pipe with a limited buffer capacity, causing deadlocks for large enough byte strings.

### DIFF
--- a/scheme-libs/racket/unison/zlib.rkt
+++ b/scheme-libs/racket/unison/zlib.rkt
@@ -31,13 +31,14 @@
 ;; ADLER32 implementation
 ;; https://www.ietf.org/rfc/rfc1950.txt
 ;; Modified from racket/collects/net/git-checkout.rkt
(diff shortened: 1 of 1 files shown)


dolio:
FYI, I tried inflating the big value, and it worked. But maybe I should try inflating an _input_ that is bigger than 4096 bytes?

pchiusano:
Maybe add a check for `gzip` while you're at it, not sure if it uses the same approach.

dolio:
Inflating a big compressed input seems to work fine.

dolio:
The racket gzip implementation seems to have a completely different, non-streaming API, so it just works.

Actually, I guess it is stream based, but it doesn't need us to add our own checksumming to the stream like zlib does.

aryairani (approved):

PR #4962, opened 2024-05-16
Avoid printing/parsing numbered args

sellout:
## Overview

The Haskell data structures for various types are preserved through the “numbered args” processing, meaning that numbered args don’t need to be re-parsed after being expanded into the command. This avoids work, makes handling more consistent between commands, and allows for better error messages (say, when you refer to a branch when you need a definition).

## Test coverage

This is a refactoring that should be largely covered by existing transcript tests that exercise the use of numbered args. All transcripts pass without modification.


unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
```diff
@@ -1550,8 +1540,8 @@ handleDependencies hq = do
   let types = nubOrdOn snd . Name.sortByText (HQ.toText . fst) $ (join $ fst <$> results)
   let terms = nubOrdOn snd . Name.sortByText (HQ.toText . fst) $ (join $ snd <$> results)
   Cli.setNumberedArgs $
-    map (Text.unpack . Reference.toText . snd) types
-      <> map (Text.unpack . Reference.toText . Referent.toReference . snd) terms
+    map (SA.Type . snd) types
+      <> map (SA.Term . Referent.toReference . snd) terms
```
sellout: These cases (and other `SA.Ref` cases) had been displaying the `Reference` (hash), but we also have a `HashQualified Name` which seems like a much nicer thing to show to the user? Is there a reason that the HQ Name isn’t preferable here?

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -97,7 +226,372 @@ makeExampleEOS p args =
     backtick (intercalateMap " " id (P.nonEmpty $ fromString (I.patternName p) : args)) <> "."
 
 helpFor :: InputPattern -> P.Pretty CT.ColorText
-helpFor p = I.help p
+helpFor = I.help
+
+handleProjectArg :: I.Argument -> Either Text ProjectName
```
sellout: These handlers aren’t great, IMO. There is so much code here. I feel like we can probably shrink the size of the `StructuredArgument` sum type more, which would help a bit here and there are some other possibilities, too, but we probably need to discuss.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -97,7 +226,372 @@ makeExampleEOS p args =
...
+    ( \name ->
+        first
+          (const $ "“" <> Text.pack name <> "” is an invalid project name")
+          . tryInto @ProjectName
+          $ Text.pack name
+    )
+    ( \case
+        SA.Project project -> pure project
+        -- __FIXME__: Do we want to treat a project branch as a project?
+        SA.ProjectBranch (ProjectAndBranch (Just project) _) -> pure project
```
sellout: I doubt this was supported before, but when referencing numbered args, it seems like this kind of extraction is somewhat useful? E.g., you get back a list of all the `ProjectAndBranch`es in your codebase, and then you run a command to rename a project, giving it some random `ProjectAndBranch` for that project – it makes sense that it would just ignore the branch name and rename the project … right?

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -35,37 +35,119 @@ import Unison.Codebase.Editor.Input qualified as Input
...
+  SA.ProjectBranch (ProjectAndBranch mproj branch) ->
+    maybe
+      (Text.cons '/' . into @Text)
+      (\project -> into @Text . ProjectAndBranch project)
+      mproj
+      branch
+  SA.Ref reference -> Reference.toText reference
+  SA.Term termReference ->
+    -- also: ShortHash.toText . Reference.toShortHash
+    Reference.toText termReference
```
sellout: All the formatting has been localized to this one function now, but some types had previously been formatted multiple ways, in which case I added a comment like this one.

All of the existing transcripts are working unmodified (although they definitely broke along the way), so at least as much as is covered is producing the same strings, but I’m not sure I always chose the best option.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -35,37 +35,119 @@ import Unison.Codebase.Editor.Input qualified as Input
...
+import Unison.Syntax.NameSegment qualified as NameSegment
 import Unison.Util.ColorText qualified as CT
 import Unison.Util.Monoid (intercalateMap)
 import Unison.Util.Pretty qualified as P
 
+-- |
+--
+--  __FIXME__: Don’t hardcode this
+schLength :: Int
+schLength = 10
```
sellout: This seemed to be the value used for short hashes in various places, but not exposed in a good way, AFAICT. A few of the handlers need this (for commands that expect short hashes where we have a full hash). But this feels like the wrong way.

Also, this might make sense for  the formatter, but perhaps the commands that take ShortCausalHash or whatever should have it use the full length? (Even though the type has to be `Short` to allow users to type less.)

sellout (commented):
There is a bunch of cleanup to do in here (e.g., removing formatting done by wrong Ormolu version (but I think I have that sorted going forward)), but it’s feeling pretty solid with a few questions in the comments here.

I think there is some follow-up work in a separate PR to standardize some of the types expected by various commands, which would also reduce the handlers.

sellout:
This still isn’t running any checks for me …

aryairani:
> This still isn’t running any checks for me …

At [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) I see
> Workflows will not run on `pull_request` activity if the pull request has a merge conflict. The merge conflict must be resolved first.

I was unsure as to why this would matter, but reading a bit more I suppose it's because we run CI on `pull_request` (which it looks like uses the merge node, which can't exist with a conflict, to run the workflow), rather than on `push`, which just uses exactly whatever you committed.  I didn't realize this! Or I forgot.

I could see a case for running both TBH, but we don't currently, probably due to running time.

It's too bad it doesn't even mention that it would have run if not for the merge conflict.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
aryairani: We were initially planning to derive this number from the individual codebase, like git does; but we never did.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -117,7 +167,411 @@ makeExampleEOS p args =
     backtick (intercalateMap " " id (P.nonEmpty $ fromString (I.patternName p) : args)) <> "."
 
 helpFor :: InputPattern -> P.Pretty CT.ColorText
-helpFor p = I.help p
+helpFor = I.help
+
+handleProjectArg :: I.Argument -> Either Text ProjectName
```
mitchellwrosen: Suggestion: use `Either Pretty`

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -207,14 +217,82 @@ import Unison.Project
...
 import Unison.Util.Pretty.MegaParsec (prettyPrintParseError)
 
+-- |
+--
+--  __FIXME__: Don’t hardcode this
+schLength :: Int
+schLength = 10
+
+formatStructuredArgument :: StructuredArgument -> Text
+formatStructuredArgument = \case
```
aryairani: I think this function goes away altogether in the final version, right? Or maybe not?
I'm assuming it's just use for readline history and/or the reflog, and in either case don't we want the representation of the type chosen by the consumer and not the producer?

unison-cli/src/Unison/CommandLine/InputPatterns.hs
aryairani: Maybe? Sounds kind of weird to me, like if we said:
```
rename.project foo/main bar/branch
```
should that rename `foo` to `bar`?

I'd be more inclined to tell the user something like "hey I'm not sure we're on the same page here"

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -610,11 +1151,16 @@ displayTo =
         makeExample displayTo ["<filename>", "foo"]
           <> "prints a rendered version of the term `foo` to the given file."
     )
-    \case
-      file : (x : xs) ->
-        (x NE.:| xs)
-          & traverse parseHashQualifiedName
-          <&> Input.DisplayI (Input.FileLocation file)
+    $ \case
```
aryairani, line 971: We have the `LambdaCase` extension enabled, so we leave off these `$` where possible

unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -755,10 +1298,17 @@ findIn' cmd mkfscope =
...
+      p : args ->
+        Input.FindI False . mkfscope
+          <$> first P.text (handlePathArg p)
+          -- __FIXME__: This changes things a bit. Previously, `find` and
+          --            friends would just expand the numbered args and search
+          --            for them like any other string, but now it recognizes
+          --            that you’re trying to look up something you already
+          --            have, and refuses to. Is that the right thing to do? We
+          --            _could_ still serialize in this case.
+          <*> traverse (unsupportedStructuredArgument "text") args
```
aryairani: Could we see an example in a transcript for context?

aryairani (approved):
lgtm

unison-cli/src/Unison/CommandLine/InputPatterns.hs
aryairani: If there's already a transcript that demonstrates `find` it could go there

unison-cli/src/Unison/CommandLine/InputPatterns.hs
sellout: Yeah, @mitchellwrosen said the same. I think that’s right.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
sellout: Not quite. This is used for printing the list of numbered args in the first place, before there is even a consumer, so that will stick around.

Currently, it is also used for history, and that will still be true when this PR lands. The handling of number expansion and history is currently in one place. It would be a separate change to allows the `InputPattern`s to make decisions about those things.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
sellout: Whoops – there was an interim commit (8a95c5fe52f5f4395b21bacd542080ab5ff15616) that required adding those, and looks like I missed removing some in the follow-up.

unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
aryairani: Aside: I looked in the diff for these lines (e.g. old line 1553), but it doesn't match what's shown here, and I don't understand why it wouldn't? I see it on old line 1302. I see Github's marked the snippet as "Outdated", but I don't think that would matter.

I think the answer to your question is that until https://github.com/unisonweb/unison/pull/4907, the numbered args weren't shown to the user at all unless they ran `debug.numberedArgs`, making the question moot(?)

unison-cli/src/Unison/CommandLine/InputPatterns.hs
aryairani: > This is used for printing the list of numbered args in the first place, before there is even a consumer, so that will stick around.

You mean in `debug.numberedArgs`? 👍 

> Currently, it is also used for history, and that will still be true when this PR lands. The handling of number expansion and history is currently in one place. It would be a separate change to allows the `InputPattern`s to make decisions about those things.

Ah right, ok.

aryairani:
I want to understand the question about `find`, but apart from that it seems ready to merge — or did I miss addressing any other question?

unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
sellout: > Aside: I looked in the diff for these lines (e.g. old line 1553), but it doesn't match what's shown here, and I don't understand why it wouldn't? I see it on old line 1302. I see Github's marked the snippet as "Outdated", but I don't think that would matter.

Ah, I think it’s because I squashed another commit into this one when it was in draft, but after I wrote this comment.

I had originally had `Ref`, `Term`, and `Type` as three separate `StructuredArgument` cases, but that just led to three copies of identical logic, so I removed the `Term` and `Type` cases until we can reasonably preserve that information.
 
> I think the answer to your question is that until #4907, the numbered args weren't shown to the user at all unless they ran `debug.numberedArgs`, making the question moot(?)

Ah, right. I conflated a few things here and got confused.

But the question wasn’t really moot – if `Reference`s are put into numbered args state, a command expecting a `Name` couldn’t then use a numbered arg. Whereas, if `HashQualified Name`s were put into the state, the same command could then use it.

I don’t think it’s a change for _this_ PR, but it seems like using the HQ instead of the Reference would help fix some of the existing numbered args problems.

Also, I think the pretty-printed numbered results lists (e.g., the output of `find`) could eventually lean more on the numbered args handling, rather that being constructed completely independently, which is another reason using HQ instead of a Reference would be a good change.

But I don’t know what the counter-arguments might be.

aryairani: Yes, makes sense.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
sellout: I had already removed this change in other find cases, but overlooked this one. So I’ve changed this (and a couple other missed cases) to align with the others, which use `unifyArgument` to provide the string form of the structured argument.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
sellout: I have a commit at the end now that handles this a bit differently. It avoids hardcoding, and preserves the full hash as an `SCH` in a number of cases. When the length is provided to the formatter[^1] it takes advantage of it.

[^1]: `DumpNumberedArgs` gets the hash length “from the DB”, and hopefully the command history could take advantage of it soon.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
sellout: Done.

unison-cli/src/Unison/CommandLine/InputPatterns.hs
sellout: Done.

aryairani:
@sellout Go ahead and merge this when you're ready. (I was about to pull the trigger, but I had created a new merge conflict.)

PR #4966, opened 2024-05-17
feat: make `pull` call new `merge` instead of old `merge`

mitchellwrosen:
## Overview

This PR makes `pull` perform a new `merge` instead of an old `merge.old`.

## Implementation notes

`pull` now has a more restricted type: you can only target a project branch. So, this syntax no longer works

```
> pull @unison/base/releases/latest lib.base
```

that people will be familiar with from our docs. So, we'll probably want to have a better error message than the current one, which just prints the `help pull` output.

## Test coverage

No new tests, but I tested the command manually


unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -2260,26 +2285,26 @@ notifyUser dir = \case
...
         ]
-  MergeFailure path base target ->
+  MergeFailure path aliceAndBob ->
     pure . P.wrap $
       "I couldn't automatically merge"
-        <> prettyProjectBranchName (view #branch target)
+        <> prettyMergeSource aliceAndBob.bob
         <> "into"
-        <> P.group (prettyProjectBranchName (view #branch base) <> ".")
+        <> P.group (prettyProjectAndBranchName aliceAndBob.alice <> ".")
```
aryairani, line 2293: I think this does the trick

unison-src/transcripts/merge.output.md
```diff
@@ -164,7 +164,7 @@ Merge result:
 ```ucm
 project/alice> merge /bob
 
-  I merged bob into alice.
+  I merged project/bob into project/alice.
```
aryairani, line 167: 👍

aryairani (approved):
Nice!

Could you fix up (or link a new ticket up to fix up) the error message, maybe just in the `InputPattern`, so that if someone tries to `pull @<x..> lib.<z>`, it will suggest `lib.install @<x..>` instead.

Could you comment on the new help message that talks about "remote namespace" but "destination branch"; remind us why namespace in one case and branch in the other; is that purely a concession to Share loose code?

The PR description tells us the variation that no longer works, but could you remind us of the variations that do work, as of this PR?  If it ends up only needing to be short then that's good enough; if it ends up being needing to be long, then maybe a manual transcript to demo the `pull` command?

mitchellwrosen:
> Could you fix up (or link a new ticket up to fix up) the error message, maybe just in the InputPattern, so that if someone tries to pull @<x..> lib.<z>, it will suggest lib.install @<x..> instead.

Sure.

> Could you comment on the new help message that talks about "remote namespace" but "destination branch"; remind us why namespace in one case and branch in the other; is that purely a concession to Share loose code?

Just that the source can be a project branch ("branch") or arbitrary path into loose code ("namespace"), but the target can only be a project branch ("branch").

> The PR description tells us the variation that no longer works, but could you remind us of the variations that do work, as of this PR?

This: "`pull` now has a more restricted type: you can only target a project branch"
means: you can pass a `project/branch`, `/branch` or `branch` (equivalent to `/branch`)

mitchellwrosen:
I updated the `pull` error message to recommend `lib.install` if you try to pull into `lib`, and show more info about the parse error in other cases (in addition to showing help as before)

PR #4974, opened 2024-05-20
tweak: make upgrade-failed branch name more informative

mitchellwrosen:
## Overview

When an `upgrade <old> <new>` fails, we put you on a branch called `upgrade-<old>-to-<new>` to resolve the errors. But that template may form an invalid branch name, as the syntax for name segment and branch are not exactly the same!

So, there was some defensive code in place that mangled `<old>` and `<new>` by filtering down to only alphabetic characters.

That seems like too much, though - an `upgrade foo_1_0_0 foo_2_0_0` might put you on a branch called `upgrade-foo-to-foo`, but it could have been `upgrade-foo_1_0_0-to-foo_2_0_0`.

So, this PR changes the name mangling logic to first try the `upgrade-<old>-to-<new>` template without any adjustment to `<old>` or `<new>`, and if that doesn't make a valid branch name, fall back on filtering `<old>` and `<new>` just down to alphanumeric.


aryairani (approved):
Ok sounds good

PR #4977, opened 2024-05-20
feat: add `upgrade.commit` command

mitchellwrosen:
## Overview

This PR adds an `upgrade.commit` (alias: `commit.upgrade`) command that is essentially just syntactic sugar for the following:

- `update`, recording the changes that resolve the upgrade conflicts
- `switch <parent>`, switching to the branch that the upgrade was run from
- `merge <child>`, merging in the temporary upgrade branch
- `branch.delete <child>`, deleting the temporary upgrade branch

## Test coverage

I've added an example to the `upgrade-sad-path` transcript.

### Loose end 1

There is currently no explicit record of whether a branch is an "upgrade branch" on which it makes sense to run `upgrade.commit`. We just have a heuristic that will probably work fine in practice:

- Does the branch name begin with `upgrade-`?
- Does the branch have a parent?

If "yes" to both, we can proceed with the `upgrade.commit`, otherwise we'll bail with a "no upgrade in progress" message.

An improvement may be to explicitly mark branches as "is upgrade branch" in the database.

### Loose end 2

The messaging before running an `update.commit` could be improved. Currently, after resolving conflicts in the scratch file, you'll get the usual message like

```
  Loading changes detected in ~/unison/24-05-20-upgrade-commit/scratch.u.


  I found and typechecked these definitions in ~/unison/24-05-20-upgrade-commit/scratch.u. If you
  do an `add` or `update`, here's how your codebase would change:

    ⍟ These names already exist. You can `update` them to your new definition:

      honk : Nat
```

You're meant to _ignore_ this message, and run `upgrade.commit` instead, which performs that `update` and more. So, that's obviously not ideal.

### Loose end 3

An `upgrade.commit` will record three entries in the reflog: one for the initial `update` on the temp branch that gets deleted, one for the `merge` into the original branch, and one for the `branch.delete`.

Maybe it would be better to combine all of those, I'm not sure. In any case, it didn't seem important enough to make the cut for the first version of this command. The reflog is megabusted anyway.


aryairani:
Sweet!

ChrisPenner:
I'm not prepped on this, so forgive the ignorance, but I'm having trouble understanding what this does and when you'd use it.

I checked the help, but I don't know what it means to "commit" an upgrade in unison, how is it different from a regular upgrade? 

Looking at the desugaring, where does the `upgrade` happen or does it happen at all, you mention it in one of the steps, but I don't see where `upgrade` was actually run; 
Why are there multiple branches involved in an upgrade? Is this a new thing I'm missing?

Probably easiest to just chat about it in tech-chat, but also noting that these are questions most users are probably going to have too and that we need to be mindful of UX when adding new commands :)

mitchellwrosen:
@ChrisPenner If your `upgrade foo bar` fails, you're put on a new branch called `upgrade-foo-to-bar` with the conflicts in a scratch file to resolve. From there (pre-this-PR) you resolve conflicts, `update`, switch back to the original branch, merge in that upgrade branch, and delete it.

aryairani:
Loose end 1 is okay

Loose end 2 we should fix

> You're meant to _ignore_ this message, and run `upgrade.commit` instead, which performs that `update` and more. So, that's obviously not ideal.

Maybe we should leave out the `update` part of the syntactic sugar, so you run that first on your own.
Or alter the typechecking output when you're on a merge/upgrade branch, to tell you the right command.
Or both — it tells you to update on your own and `upgrade.commit` when you're finished.

Loose end 3

> Maybe it would be better to combine all of those, I'm not sure. In any case, it didn't seem important enough to make the cut for the first version of this command. The reflog is megabusted anyway.

I think having them separate is okay, and more so if the `update` is separate anyway.

Chris is working on per-project reflogs, so we should assume it's not busted for the purposes of deciding stuff like this.

mitchellwrosen:
> Or alter the typechecking output when you're on a merge/upgrade branch, to tell you the right command.

That sounds good to me.

mitchellwrosen:
Here's an example of the current typechecking output:

```

  Loading changes detected in ~/unison/24-05-20-upgrade-commit/scratch.u.


  I found and typechecked these definitions in ~/unison/24-05-20-upgrade-commit/scratch.u. If you
  do an `add` or `update`, here's how your codebase would change:

      bonk : Nat

    ⍟ These names already exist. You can `update` them to your new definition:

      honk : Nat

```

@aryairani What were you thinking it should say instead?

aryairani:
> @aryairani What were you thinking it should say instead?

If we're taking `update` out of it, then I guess the message would be after the add/update:

```
.> update

  ⍟ I've updated these definitions:
  
    foo : Nat

  If you're finished, use `upgrade.commit` to merge this work back into <branch name>.

```

If we're leaving it in,

```

  Loading changes detected in ~/unison/24-05-20-upgrade-commit/scratch.u.


  I found and typechecked these definitions in ~/unison/24-05-20-upgrade-commit/scratch.u. If you
  do an `add` or `update`, here's how your codebase would change:

      bonk : Nat

    ⍟ These names already exist. You can `update` them to your new definition:

      honk : Nat

  Once you're finished, use `upgrade.commit` to merge this back into <branch name>.
```

But I think the first option is better. What about you?

mitchellwrosen:
Hmm... well, I was thinking making `upgrade.commit` perform the `update` would be better, as it seems more similar to a `git commit` after fixing up merge conflicts - the "record the changes" and "end the merge" are the same step.

So if the software said "run `update` to update" and "once you're finished, run `upgrade.commit`", I'd think, "well I am finished... that's why I typed `update`...".

But I'm not sure, would there be a reason to sit on an `upgrade` branch and perform multiple `update` commands? Since the upgrade failures are just sitting in your ephemeral scratch.u it doesn't really seem like the kind of thing that can be done incrementally.

mitchellwrosen:
Update: `upgrade.commit` will not perform the `update` because there are other things (like `test`) you might want to run before "committing".

So, that makes me think perhaps the right place to put a reminder about `upgrade.commit` would just be when the upgrade branch is created, not at type checking time.

mitchellwrosen:
How about something like this?
2024-05-23-090711_983x549_scrot

stew:
@mitchellwrosen when you `branch.delete` with no arguments does it delete the current branch? if so, where are you left after running this command? I would hope it is in whatever branch this upgrade branch was created from and not something arbitrary like `/main`.

I would also like this behavior to not just be for upgrade branches but for any other branch. If I delete the current branch, send me back to the branch it was created from.

aryairani:
I'm guessing it sends you to fzf, but we could change that to delete the current branch.

mitchellwrosen:
When you delete the current branch, it takes to you to that branch's parent, if it had one. If you run `delete.branch` without arguments, you'll get an FZF prompt. That's why the screenshot above instructs you to provide an argument to `delete.branch`, the current branch.

Can I get some concrete feedback on the status of this PR @aryairani, are we trying to nail down a wording? I had proposed that instead of changing the typechecking feedback (on file save), we just put something in the upgrade-failure message that reminds the user about `upgrade.commit`. Is that a good idea, and if so, is that screenshot above good enough or should we continue to workshop it?

aryairani:
@mitchellwrosen I've been trying to get it merged, but I can't test my resolved merge conflicts due to a nix hiccup where it's apparently going to take hours to put my toolchain back together. You can resolve and merge it if you want, or I can push my changes to another branch to run CI on.

aryairani:
We can potentially workshop the message after merging

mitchellwrosen:
@aryairani Oh, ok sounds good – I can fix up the merge conflicts. I didn't realize it was quasi-approved ;)

aryairani:
@mitchellwrosen I was just waiting on the loose end 2.; but it looks like that's been updated, right, by splitting out the `update` part of the sequence.  I didn't realize it had been.

aryairani (approved):

PR #4986, opened 2024-05-20
feat: add `merge.commit` command

mitchellwrosen:
## Overview

This PR adds a `merge.commit` command, which is essentially the exact same thing as `upgrade.commit`, except that you're only allowed to run it on a "merge branch" (a branch that begins with the name `merge-`)

## Test coverage

I've added a test for this to the `merge.md` transcript


unison-src/transcripts/merge.output.md
```diff
@@ -579,9 +624,21 @@ baz = "bobs baz"
 ```ucm
 project/alice> merge /bob
 
-  I couldn't automatically merge project/bob into project/alice.
+  I couldn't automatically merge project/bob into alice.
```
aryairani: Was it intentional that the target project name went away? I noticed it wasn't in one of the messages PRs, but didn't notice it had been in the earlier version of the message, and I just assumed it was hard to produce for some reason.

aryairani:
@mitchellwrosen Could you show an example of `merge.commit` on a non-merge branch; looks good otherwise.

unison-src/transcripts/merge.output.md
mitchellwrosen: It was - I thought that because you are always merging into the current project branch, showing the project is a little wordy.

But it especially looks weird alongside a project+branch for Bob, so I can add it back. Or, I can make it so the project is omitted from Bob if it matches Alice's, so that the very common case is less verbose overall.

mitchellwrosen: I reverted this for now

mitchellwrosen:
@aryairani Done: https://github.com/unisonweb/unison/pull/4986/commits/e903a1c1556c6385ee5be12b0d0d22f858b6e501

aryairani (approved):

PR #5003, opened 2024-05-23
Type the `main` arg to `execute`

sellout:
## Overview

This avoids throwing away the type information from `NumberedArgs` and just
generally gets text handling out of the domain logic.

No change to the user, but some round-tripping of print/parse is removed, and remaining parsing is pushed out with other `OptParse` handling.


sellout:
Only the last commit (https://github.com/unisonweb/unison/pull/5003/commits/25d9b44b50d62705bd890416cee842c46cb21cfa) is part of this PR, the rest is because it is based on #4962. That’s also why it’s a draft.

aryairani (approved):

PR #5005, opened 2024-05-24
Remove unused `Unison.Util.Convert` module

sellout:
## Overview

Cleans up some dead code. No user impact.

## Implementation notes

The same classes and instance exist in `Unison.Name`, and that is the version already used everywhere.


mitchellwrosen (approved):

mitchellwrosen:
If memory serves, here is some background on this one:

1. We had a too-general `Convert` type class buried in `Name` for some reason.
2. I didn't like that and wanted to at the very least move it into a dedicated module, and ideally eventually just delete the type class entirely. (There's also `witch` that serves the same purpose that we could use).
3. I didn't get around to actually moving any of the call sites over to the new `Convert`, so it sat unused.
4. Now it's gone and we are back to (1)

I do think (1) is a better place to be than (3), but if anyone feels so inclined I think we'd just be better off just:

- Using explicit conversion methods
- If that's really too boilerplatey, use `witch`

sellout:
>     * Using explicit conversion methods

I think this is the way to go – ideally defined via some kind of optic structure, so we get some properties out of it.

PR #5019, opened 2024-05-29
Remove `Reference` from `StructuredArgument`

### LINKED ISSUE #4898: delete.term with numbered arg fails and with a cryptic message
```
@unison/cloud/fix-harder> dependents internal.Connection
 
  Dependents of: type internal.Connection
  
    Terms:
  
    1. connectTcp
    2. internal.Connection.stream
    3. internal.connectTls
  
  Tip: Try `view 3` to see the source of any numbered item in the above list.

@unison/cloud/fix-harder> delete.term 1
 
  1:1:
    |
  1 | #0nvef7e18virteotmgt6154hu7ve15pldl6v4uqqod71384811csd7tcko3qla9q3fgrpdo5jiepaeiatvnvqb4rl4h2ljvocjj7uj0
    | ^
  unexpected '#'
  expecting '.', '`', or operator (valid characters: !$%&*+-/:<=>\^|~)
```

sellout:
## Overview

Almost everywhere we produce a `Reference` for numbered args, we also have a
`HashQualified Name` handy, which is much more consumable by commands.

The only case we don’t have an `HQ` is in the `todo` command output, so that now explicitly builds
a `HQ.HashOnly`.

This also fixes an issue with `StructuredArgument` handling where `alias.term`
and `alias.type` wouldn’t make an alias to a `HQ.HashOnly` `StructuredArgument`.

Fixes #4898.

## Test coverage

There is a new transcript to replicate the reported issue.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
@@ -1285,12 +1285,10 @@ handleDependencies hq = do
     let types = [(PPE.typeName suffixifiedPPE r, r) | LabeledDependency.TypeReference r <- toList dependencies]
     let terms = [(PPE.termName suffixifiedPPE r, r) | LabeledDependency.TermReferent r <- toList dependencies]
     pure (types, terms)
-  let types = nubOrdOn snd . Name.sortByText (HQ.toText . fst) $ (join $ fst <$> results)
-  let terms = nubOrdOn snd . Name.sortByText (HQ.toText . fst) $ (join $ snd <$> results)
-  Cli.setNumberedArgs $
-    map (SA.Ref . snd) types
-      <> map (SA.Ref . Referent.toReference . snd) terms
-  Cli.respond $ ListDependencies suffixifiedPPE lds (fst <$> types) (fst <$> terms)
+  let types = fmap fst . nubOrdOn snd . Name.sortByText (HQ.toText . fst) . join $ fst <$> results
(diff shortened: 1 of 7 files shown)


aryairani:
The following comments aren't specifically about this PR, but:

> Almost everywhere we produce a `Reference` for numbered args, we also have a `HashQualified Name` handy, which is much more consumable by commands.

Makes me think we should have something isomorphic to `(Name, Reference)` rather than `These Name ShortHash`.

> The only case we don’t have an `HQ` is in the `todo` command output, so that now explicitly builds a `HQ.HashOnly`.

`todo` does have a name available that it prints though, it just doesn't load it into numbered-args? or no?

aryairani (approved):

sellout:
> > The only case we don’t have an `HQ` is in the `todo` command output, so that now explicitly builds a `HQ.HashOnly`.
> 
> `todo` does have a name available that it prints though, it just doesn't load it into numbered-args? or no?

Yes, so I’m actually a bit confused about when exactly the `Reference` would get used. The change in question is https://github.com/unisonweb/unison/pull/5019/files#diff-db89d574f836d421e840d72b91dad3f90f89f3a897d448c812ee1d441a3e7630L1442-R1442, and you can see just two lines below `setNumberedArgs` is `respondNumbered $ TodoOutput`. And when that TodoOutput is processed, it calls `setNumberedArgs` using `HashQualified Name` … so is the linked `setNumberedArgs` just dead code, or is that there in case something else fails along the way and the later `HashQualified` args never get set?

aryairani:
Oh weird. It looks like dead code to me.

PR #5029, opened 2024-05-31
Update .mergify.yml

aryairani:
Mergify is configured to look for specifically-named checks to complete before auto-merging, but the names of those tests change as I refactored the CI code. This PR tries to bring .mergify.yml in line with the current CI check names.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -11,9 +11,12 @@ pull_request_rules:
       - check-success=run interpreter tests (macOS-12)
       # - check-success=run interpreter tests (windows-2019)
       - check-success=generate jit source
-      - check-success=build jit binary (ubuntu-20.04)
-      - check-success=build jit binary (macOS-12)
-      - check-success=build jit binary (windows-2019)
+      - check-success=build jit binary / build jit binary (ubuntu-20.04)
+      - check-success=build jit binary / build jit binary (macOS-12)
+      - check-success=build jit binary / build jit binary (windows-2019)
+      - check-success=test jit / test jit (ubuntu-20.04)
+      - check-success=test jit / test jit (macOS-12)
+      # - check-success=test jit / test jit (windows-2019)
       - label=ready-to-merge
       - "#approved-reviews-by>=1"
     actions:


aryairani:
Unfortunately, it looks like the rules applied to this PR are the rules as they are configured on `trunk`, not on this branch. This makes me think we just have to merge it and hope for the best?

sellout:
> Unfortunately, it looks like the rules applied to this PR are the rules as they are configured on `trunk`, not on this branch. This makes me think we just have to merge it and hope for the best?

Yeah, I think so 🤞

It looks like Mergify can at least validate the config changes on a PR (https://docs.mergify.com/configuration/file-format/#validation), but probably Mergify has to be enabled for that to happen, and we don’t want to enable it until the new config is in place.

I’ve fought with this kind of poor workflow too many times – where something only occurs on the default branch, and so you end up merging a bunch of not-quite-fixes to trunk until you get it right. One workaround I’ve done (that probably doesn’t work on a repo as active as this one) is to temporarily make a PR branch the default branch while sorting out the changes and then switching the default back to `trunk` before merging.

sellout (approved):
I checked the new list (including unchanged entries) against the list of checks on this PR, and they all seem to match up.

I just want to make sure: in a PR, the builds are suffixed with “(pull request)” and they’re prefixed by the workflow name, “CI /”, but I don’t know if those strings aren’t part of what’s checked by mergify, or if mergify does a substring check. I’m guessing since this worked at some point in the past, that it’ll be successful without those bits or some kind of wildcard, but wanted to make that assumption explicit in the review.

aryairani:
> I checked the new list (including unchanged entries) against the list of checks on this PR, and they all seem to match up.
> 
> I just want to make sure: in a PR, the builds are suffixed with “(pull request)” and they’re prefixed by the workflow name, “CI /”, but I don’t know if those strings aren’t part of what’s checked by mergify, or if mergify does a substring check. I’m guessing since this worked at some point in the past, that it’ll be successful without those bits or some kind of wildcard, but wanted to make that assumption explicit in the review.

Oh, good observation. I do not see that documented anywhere on in the Github or Mergify docs.

I was comparing to the existing check results https://github.com/unisonweb/unison/pull/5029/checks?check_run_id=25658950197 and updating based on that.  So I think it will be okay but appreciate the call-out.

PR #5035, opened 2024-06-01
Add a Nix formatter and run it

sellout:
## Overview

This sets the flake’s `formatter` output to [Alejandra](https://kamadorueda.com/alejandra/).

It’s just easier to make changes if there’s an autoformatter.

There are at least three popular options for Nix formatters. I have reasons for preferring this one, but I’m happy to use whatever one people want.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -133,5 +133,7 @@
             // {
               default = self.devShells."${system}".only-tools-nixpkgs;
             };
+
+          formatter = pkgs.alejandra;
         });
 }
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,7 @@
 {
   description = "Unison";
   nixConfig = {
-    extra-substituters = [ "https://unison.cachix.org" ];
+    extra-substituters = ["https://unison.cachix.org"];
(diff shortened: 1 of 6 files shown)


aryairani:
What does `formatter = pkgs.alejandra;` do? When does formatting occur?

aryairani (approved):

sellout:
> What does `formatter = pkgs.alejandra;` do? When does formatting occur?

The `nix fmt` command runs whatever derivation `formatter` is set to, The formatter has very few restrictions (e.g., it should take a filename or directory as an optional argument). I use [treefmt-nix](https://github.com/numtide/treefmt-nix#readme) on my own projects, which is a whole-repo formatter – it runs Ormolu, hlint, prettier, alejandra – whatever formatters and linters you want. But for our purposes, the configuration is a bit much and I’m just trying to simplify the Nix for now.

Right now formatting only occurs when someone runs `nix fmt`, but we could also add a flake check that runs it on `nix flake check` (which runs a basic flake validity check (that now passes after #4923) and any other checks you define (like the test suites that are added by #5036)).

PR #5036, opened 2024-06-01
Run Cabal test-suites as Nix checks

sellout:
## Overview

`nix flake check` will now ensure both that the flake is valid, and that all tests pass.

### DIFF
--- a/nix/haskell-nix-flake.nix
+++ b/nix/haskell-nix-flake.nix
@@ -73,10 +73,25 @@
       };
     }
     // localPackageDevShells;
+
+  checks =
+    haskell-nix-flake.checks
+    // {
+      ## This check has a test that tries to write to $HOME, so we give it a fake one.
+      "unison-cli:test:cli-tests" = haskell-nix-flake.checks."unison-cli:test:cli-tests".overrideAttrs (old: {
+        ## The builder here doesn’t `runHook preBuild`, so we just prepend onto `buildPhase`.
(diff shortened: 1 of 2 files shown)


nix/haskell-nix-flake.nix
```diff
@@ -73,10 +73,25 @@
...
+      ## This check has a test that tries to write to $HOME, so we give it a fake one.
+      "unison-cli:test:cli-tests" = haskell-nix-flake.checks."unison-cli:test:cli-tests".overrideAttrs (old: {
+        ## The builder here doesn’t `runHook preBuild`, so we just prepend onto `buildPhase`.
+        buildPhase =
+          ''
+            export HOME="$TMP/fake-home"
+            mkdir -p "$HOME"
+          ''
+          + old.buildPhase or "";
+      });
```
sellout, line 89: The failing test is “clearWatchCache.CodebaseFormat2”. I feel like this might be a bug in the test suite – tests probably shouldn’t affect any codebase or other files in `HOME`.

aryairani: Do you wanna include it with this patch, or else we can try to get to the bottom of the test writing to `$HOME`?

sellout: I think with https://github.com/unisonweb/unison/issues/5040 I’m fine to merge as-is.

aryairani (approved):

PR #5042, opened 2024-06-03
Add a transcript illustrating that #1532 is fixed

### LINKED ISSUE #1532: `delete.namespace` does not assign to numberedArgs
When using `delete.namespace` ucm will not assign `numberedArgs` which is unexpected.

It also assigns multiple things to the same number (possibly related).

## Transcript showing the problem

First, lets create two namespaces. `foo` and `bar`, and add some definitions.

```unison
x = 42
y = 100
```

```ucm

  change:
  
      x : Nat
      y : Nat

```
```ucm
  ☝️  The namespace .foo is empty.

.foo> add

    x : Nat
    y : Nat

```
```unison
z = x + y
```

```ucm

  change:
  
      z : Nat

```
```ucm
  ☝️  The namespace .bar is empty.

.bar> add

    z : Nat

```
Let's see what we have created...
```ucm
.> ls

  1. bar/     (1 definition)
  2. builtin/ (168 definitions)
  3. foo/     (2 definitions)

```
Now, if we try deleting the namespace `foo`, we get an error, as expected.

```ucm
.> delete.namespace foo

  ⚠️
  
  I couldn't delete
  
    1. foo.x : builtin.Nat
    2. foo.y : builtin.Nat
    
  
  because it's still being used by these definitions:
  
    1. bar.z : builtin.Nat
    

```
However, the numbered arguments are not assigned by that command.
```ucm
.> debug.numberedArgs

  1. .bar
  2. .builtin
  3. .foo

```
As you can see, the earlier output from `ls is still there.

## Transcript source here:
[delete-namespace-does-not-assign-numberedArgs.md.zip](https://github.com/unisonweb/unison/files/4588342/delete-namespace-does-not-assign-numberedArgs.md.zip)

sellout:
## Overview

The issue has already been fixed, so this just adds a transcript that shows that and prevents regression.

Closes #1532.

### DIFF
--- a/unison-src/transcripts/fix1532.md
+++ b/unison-src/transcripts/fix1532.md
@@ -1,99 +1,40 @@
-When using `delete.namespace` ucm will not assign `numberedArgs` which is unexpected.
-
-It also assigns multiple things to the same number (possibly related).
-
-## Transcript showing the problem
+```ucm
+.> builtins.merge
+```
 
 First, lets create two namespaces. `foo` and `bar`, and add some definitions.
 
 ```unison
-x = 42
-y = 100
+foo.x = 42
+foo.y = 100
+bar.z = x + y
 ```
 
 ```ucm
-
-  I found and typechecked these definitions in scratch.u. If you
-  do an `add` or `update`, here's how your codebase would
-  change:
-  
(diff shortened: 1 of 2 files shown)


aryairani (approved):
Great!

PR #5061, opened 2024-06-07
Improve `ls` with numbered arg behavior

sellout:
## Overview

This fixes a couple issues with `ls` and numbered args. They are illustrated in the two transcripts included in this PR.

The two changes are to avoid `Absolute` paths in `ShallowListEntry` and to convert numbered args representing `Name`s to `Path'` for commands that expect `Path'`s.

This replaces a number of `Path.prefix*` operations, introducing ones that are more precise in their types (given the existing set of pathy types we have to work with).

## Test coverage

There are two transcripts here that exercise these changes. One is a modification of an existing transcript.

## Loose ends

There should be much more coverage of numbered arg handling in transcripts.

### DIFF
--- a/parser-typechecker/src/Unison/Codebase/Path.hs
+++ b/parser-typechecker/src/Unison/Codebase/Path.hs
@@ -22,9 +22,12 @@ module Unison.Codebase.Path
     relativeEmpty',
     currentPath,
     prefix,
+    prefixAbs,
+    prefixRel,
+    maybePrefix,
     unprefix,
-    prefixName,
-    prefixName2,
+    maybePrefixName,
+    prefixNameIfRel,
     unprefixName,
     HQSplit,
     Split,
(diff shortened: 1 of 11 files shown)


sellout:
I am happy to add a more comprehensive transcript for numbered args to this PR, but not until at least tomorrow. I figured I would get this up quickly, though, since it fixes current issues, and you reviewers can decide whether it’s worth getting merged more immediately.

sellout:
Oh, and https://github.com/unisonweb/unison/pull/5061/commits/ce3305738564864ac54fc6c4e651262cb114b00e is probably worth reviewing in isolation. It makes the most significant changes and has a fairly comprehensive commit message.

aryairani (approved):

ceedubs:
I've been trying this out for a few minutes and I haven't encountered any of the issues that I was hitting before!

aryairani:
That's good enough for me

PR #5067, opened 2024-06-10
bugfix: updating and deleting aliases erroneously tripped "conflicted alias" merge precondition

mitchellwrosen:
## Overview

Previously, if `foo` and `bar` were aliases in the LCA of a merge, and `foo` was updated while `bar` was deleted, we'd erroneously throw a "conflicted alias" precondition violation.

But this error should only trigger if `foo` and `bar` are both _updated_ to different things. An update+delete is fine.

@runarorama found the bug!

## Test coverage

I've added a test that demonstrates the bug (commit 1) and fixes it (commit 2)

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/Merge2.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/Merge2.hs
@@ -974,6 +974,8 @@ findConflictedAlias defns diff =
             g hashed1 alias =
               case Map.lookup alias diff of
                 Just (DiffOp'Update hashed2) | hashed1 == hashed2.new -> Nothing
+                -- If "foo" was updated but its alias "bar" was deleted, that's ok
+                Just (DiffOp'Delete _) -> Nothing
                 _ -> Just (name, alias)
 
 -- Given a name like "base", try "base__1", then "base__2", etc, until we find a name that doesn't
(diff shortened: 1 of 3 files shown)


unison-cli/src/Unison/Codebase/Editor/HandleInput/Merge2.hs
```diff
@@ -974,6 +974,8 @@ findConflictedAlias defns diff =
             g hashed1 alias =
               case Map.lookup alias diff of
                 Just (DiffOp'Update hashed2) | hashed1 == hashed2.new -> Nothing
+                -- If "foo" was updated but its alias "bar" was deleted, that's ok
+                Just (DiffOp'Delete _) -> Nothing
```
aryairani, line 978: Is there a symmetric case that needs to be considered; i.e. when "foo" was deleted but it's alias "bar" was updated? No, because we're iterating over updated names?

aryairani (approved):

unison-cli/src/Unison/Codebase/Editor/HandleInput/Merge2.hs
mitchellwrosen: That symmetric case hits this code path to return `Nothing` (meaning "no problem") https://github.com/unisonweb/unison/pull/5067/files/528c2a1cd6930f36e957fc04543b1f3232f3febb..e3440a91c6832fb01a617eb48d7297df05af309f#diff-56bc494d4c38b12f3b10d9e6ac7aba28556377ede79ab99ccf723f1c91c17881R965

PR #5070, opened 2024-06-10
Update merge precondition message: Conflicted aliases

### LINKED ISSUE #5069: merge precondition message: Conflicted aliases

```
Sorry, I wasn't able to perform the merge:

On the merge ancestor, bar and foo were aliases for the same definition; but on project/alice the names have different definitions currently. I'd need just a single new definition to use in their dependents when I merge.

Please fix up project/alice to resolve this. For example,

    * `update` the definitions to be the same again, so that there's nothing for me to decide.
    * `rename` or `delete` one of the definitions; I'll use the remaining name when propagating updates. 

and then try merging again.
```

```
Sorry, I wasn't able to perform the merge:

On the merge ancestor, bar and foo were aliases for the same definition; but on project/alice the names have different definitions currently. I'd need just a single new definition to use in their dependents when I merge.

Please fix up project/alice to resolve this. For example,

    * `update` the definitions to be the same again, so that there's nothing for me to decide.
    * `rename` or `delete` one of the definitions; I'll use the remaining name when propagating updates, 
       and you can change the name back after the merge.

and then try merging again.
```

https://www.diffchecker.com/

aryairani:
fixes #5069 
proposal from https://discord.com/channels/862108724948500490/1249809359882555462/1249823678359801920

### DIFF
--- a/unison-cli/src/Unison/CommandLine/OutputMessages.hs
+++ b/unison-cli/src/Unison/CommandLine/OutputMessages.hs
@@ -1372,6 +1372,7 @@ notifyUser dir = \case
                       <> "or"
                       <> IP.makeExample' IP.delete
                       <> "all but one of the definitions; I'll use the remaining name when propagating updates."
+                      <> "(You can `rename` it back after the merge.)"
                   )
               ]
           )
--- a/unison-src/transcripts/merge.output.md
+++ b/unison-src/transcripts/merge.output.md
@@ -976,7 +976,8 @@ project/alice> merge /bob
     * `update` the definitions to be the same again, so that
       there's nothing for me to decide.
     * `move` or `delete` all but one of the definitions; I'll
-      use the remaining name when propagating updates.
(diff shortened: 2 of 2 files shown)


unison-cli/src/Unison/CommandLine/OutputMessages.hs
```diff
@@ -1372,6 +1372,7 @@ notifyUser dir = \case
                       <> "or"
                       <> IP.makeExample' IP.delete
                       <> "all but one of the definitions; I'll use the remaining name when propagating updates."
+                      <> "(You can `rename` it back after the merge.)"
```
mitchellwrosen, line 1375: I think `` `rename` `` should be `IP.makeExample' IP.moveAll` here, as above

aryairani: Shoot, I agree.

PR #5102, opened 2024-06-18
Implement custom Word64 delay

### LINKED ISSUE #3957: sleep returns instantly for large values
The following program terminates almost instantly:

```
main = do
  sleep maxNat
```

It instead sleeps for a long time if you change it to:

```
main = do
  sleep (truncate0 maxInt)
```

I presume this is because the `Nat` argument is being interpreted as a signed integer by the runtime.

dolio:
The builtin is specified to take a `Nat`, but Haskell's `threadDelay` takes an `Int`. So the previous implementation was erroneously treating large numbers as negatives and not sleeping at all. Instead we split the wait into multiple delays for large numbers (although practically the second delay won't be reached).

Fixes #3957


aryairani (approved):

ceedubs:
I've confirmed that this fixes the issue for me. Thanks! :tada:

PR #5108, opened 2024-06-20
Improve some lexer error messages

### LINKED ISSUE #5060: Confusing error message about reserved keyword
This error message is really confusing.

```
     The identifier used here isn't allowed to be a reserved keyword: 
      
          1 | this.is.a.long.namespace.path.foo = "foo1"
```

What it's _trying_ to say is that I can't use `namespace` as a namespace segment. But it doesn't tell me which word it's unhappy about, and it's telling me that it isn't allowed to be a keyword as opposed to telling me that I can't use it because it _is_ a reserved keyword.

sellout:
## Overview

The message mentioned in #5060 was incorrect. This also uses the passed-in `id` for that message and related messages (removing the reliance on ANSI colors for conveying where the error is). And it adds a suggestion on how to avoid the error.

It also does some minor adjustment of highlighting – styling individual identifiers rather than a list of them.

Fixes #5060.

### DIFF
--- a/unison-src/transcripts/error-messages.output.md
+++ b/unison-src/transcripts/error-messages.output.md
@@ -19,8 +19,8 @@ x = 1. -- missing some digits after the decimal
   
       1 | x = 1. -- missing some digits after the decimal
   
-  I was expecting some digits after the '.', for example: 1.0 or
-  1.1e37.
+  I was expecting some digits after the `.` , for example: `1.0`
+  or `1.1e37`.
 
 ```
 ```unison
@@ -36,7 +36,7 @@ x = 1e -- missing an exponent
       1 | x = 1e -- missing an exponent
   
   I was expecting some digits for the exponent, for example:
-  1e37.
+  `1e37`.
 
 ```
 ```unison
@@ -52,7 +52,7 @@ x = 1e- -- missing an exponent
       1 | x = 1e- -- missing an exponent
   
   I was expecting some digits for the exponent, for example:
-  1e-37.
+  `1e-37`.
 
 ```
 ```unison
@@ -68,7 +68,7 @@ x = 1E+ -- missing an exponent
       1 | x = 1E+ -- missing an exponent
   
   I was expecting some digits for the exponent, for example:
-  1e+37.
+  `1e+37`.
 
 ```
 ### Hex, octal, and bytes literals
@@ -343,7 +343,7 @@ use.keyword.in.namespace = 1
 
   Loading changes detected in scratch.u.
 
-  The identifier namespace used here is a reserved keyword: 
+  The identifier, `namespace`, used here is a reserved keyword: 
   
       1 | use.keyword.in.namespace = 1
   
--- a/unison-src/transcripts/error-messages.output.md
+++ b/unison-src/transcripts/error-messages.output.md
@@ -343,10 +343,12 @@ use.keyword.in.namespace = 1
 
   Loading changes detected in scratch.u.
 
-  The identifier used here isn't allowed to be a reserved keyword: 
+  The identifier namespace used here is a reserved keyword: 
   
       1 | use.keyword.in.namespace = 1
   
+  You can avoid this problem either by renaming the identifier
+  or wrapping it in backticks (like `namespace` ).
 
 ```
 ```unison
--- a/unison-src/transcripts/error-messages.output.md
+++ b/unison-src/transcripts/error-messages.output.md
@@ -343,7 +343,7 @@ use.keyword.in.namespace = 1
 
   Loading changes detected in scratch.u.
 
-  The identifier, `namespace`, used here is a reserved keyword: 
+  The identifier `namespace` used here is a reserved keyword: 
   
       1 | use.keyword.in.namespace = 1
   
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -1366,31 +1366,45 @@ renderParseErrors s = \case
             <> style ErrorSite (fromString open)
             <> ".\n\n"
             <> excerpt
-        L.InvalidWordyId _id ->
+        L.InvalidWordyId id ->
           Pr.lines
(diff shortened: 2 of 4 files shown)


ceedubs:
Thanks!

Should the example in #5060 (or something like it) be added to a transcript to avoid a regression?

sellout:
There were two transcripts that already tested this – I just forgot to commit the changed output.

And I accidentally rebased on trunk. C’est la vie.

unison-src/transcripts/error-messages.output.md
```diff
@@ -343,10 +343,12 @@ use.keyword.in.namespace = 1
 
   Loading changes detected in scratch.u.
 
-  The identifier used here isn't allowed to be a reserved keyword: 
+  The identifier namespace used here is a reserved keyword:
```
ceedubs: Hmm I like the idea of including the text here so we don't just rely on colors. But I could see this getting confusing if it doesn't include quotes or backticks or something around the identifier.

sellout: Ah right. I had originally hoped to modify `style` to wrap `Code` in something, but it requires not-insignificant changes to support anything other than ANSI colors there.

It’s probably good English to change

> The identifier namespace used here is a reserved keyword:

to

> The identifier, namespace, used here is a reserved keyword:

And I maybe defining a function like
```haskell
styleCode code = "‘" <> style Code code <> "’"
```
to use in the interim would be good enough? (It avoids hardcoding the quotes everywhere, but also doesn’t require generalizing all the `ColorText` stuff.)

> The identifier, ‘namespace’, used here is a reserved keyword:

ceedubs: I don't feel qualified to answer that. @aryairani or @mitchellwrosen WDYT?

mitchellwrosen: I'm not familiar enough to know about generalizing `ColorText` / `SyntaxText` / whatever to support a different solution, but that `styleCode` combinator does look fine and good.

aryairani: I guess I would use backticks if anything, like in `backticked`. 

Also we have other places where we rely on color that gets lost in transcript/non-ansi output, and I'm not sure what to do about it. :-\  Once upon a time, we scrapped having separate renderers depending on the terminal mode, but long term we could look at that again.

sellout: > I guess I would use backticks if anything, like in `backticked`.

Done. The reason I prefer Unicode quotes is because you often need to distinguish between literal backticks in the code vs backticks used to surround code. E.g., in this particular error message, it says “You can avoid this problem […] by […] wrapping it in backticks (like \`namespace\` ).” where we want those backticks to be considered part of the code (and they are colorized as such in contexts that allow it). But this is exactly why we have these style abstractions. If you decide to switch the way it’s quoted down the road, you just change `quoteCode` and not dozens of string literals.

> Also we have other places where we rely on color that gets lost in transcript/non-ansi output, and I'm not sure what to do about it. :-\ Once upon a time, we scrapped having separate renderers depending on the terminal mode, but long term we could look at that again.

Yeah, it’s tough. Apple’s pasteboard lets you store multiple versions of the content depending on where you paste to, but that doesn’t work (at least not easily) from the terminal, and it also relies on the receiving app having pretty sophisticated pasteboard handling. I think you mostly need to accept that anything other than text is going to be tough. Another approach could be to use ANSI escapes to hide the quotes in the terminal, so that copying the text still results in them being in the clipboard. But, yeah, a conversation for another time.

aryairani: Oh, interesting.

unison-src/transcripts/error-messages.output.md
```diff
@@ -343,10 +343,12 @@ use.keyword.in.namespace = 1
 
   Loading changes detected in scratch.u.
 
-  The identifier used here isn't allowed to be a reserved keyword: 
+  The identifier, `namespace`, used here is a reserved keyword:
```
aryairani: I think breaking up "The identifier used here" with commas is weird / possibly a mistake.

Maybe:

> The identifier `namespace` used here is a reserved keyword

or

> The identifier used here, `namespace`, is a reserved keyword

While we're in here, could you add cases in the transcript for the ones that didn't have coverage, so we can scrutinize their commas too? 🙏 
`ReservedSymbolyId`, `InvalidSymbolyId`, `InvalidWordyId`.

sellout: > I think breaking up "The identifier used here" with commas is weird / possibly a mistake.

You’re right – I thought appositives always needed to be offset with commas, but not if there is already some other punctuation (the backticks in this case).

> While we're in here, could you add cases in the transcript for the ones that didn't have coverage, so we can scrutinize their commas too? 🙏 `ReservedSymbolyId`, `InvalidSymbolyId`, `InvalidWordyId`.

`InvalidWordyId` isn’t even used, so I removed that case. The other two I’m not sure are reachable. E.g., [the transcript case that is ostensibly a `ReservedSymbolyId` error actually fails differently](https://github.com/unisonweb/unison/pull/5108/files#diff-00211f24524c732d516cdbe237e28deb0c74f20efe6f87ef55e2ceeb32e973abR355-R374).

aryairani: I see.  Another variant gave another error:

```
(!) a b = 1
```
```
  I got confused here:
  
      1 | (!) a b = 1
  
  
  I was surprised to find a ( here.
  
  * ability
  * newline or semicolon
  * type
  * use
```
Not sure what's going on there.

aryairani (approved):

unison-src/transcripts/error-messages.output.md
aryairani: ~(whoops this was with the wrong verson)~

aryairani: Disregard about [the message having been generated] with the wrong verson, it looks the same in this branch.

PR #5111, opened 2024-06-20
Support branch-relative paths for `docs.to-html`

### LINKED ISSUE #4402: docs.to-html should support projects arguments
Currently the docs.to-html command expects a path to namespaces like ".myPath.to.namepsace" but in the projects ecosystem there is no way to refer to the root of the current project. It would be great if the docs.to-html command could support turning items like top-level README's into html. 

Also currently, if you follow the suggested help text and give a path argument like `~/my/path` it will create a directory called `~` in your codebase and that is a bit dangerous.

sellout:
Fixes #4402.


sellout:
The new transcript needs to be able to check the output directory before this is ready to merge.

aryairani (approved):
👍

PR #5118, opened 2024-06-21
feat: add `alias.term.force`

mitchellwrosen:
## Overview

This PR adds `alias.term.force`, which is exactly like `alias.term`, except:

- It's hidden from `help` output
- It won't complain if you ask it to make a conflicted name

## Test coverage

I've added a transcript.

## Loose ends

`alias.type.force` isn't implemented yet


aryairani:
per https://unisoncomputing.slack.com/archives/C03HW54CYP4/p1718291333921389?thread_ts=1718289802.482839&cid=C03HW54CYP4 let's prefix it with `debug.` so it doesn't show up in tab-complete for `alias.term`.

mitchellwrosen:
@aryairani Sounds good, renamed

aryairani (approved):
We may decide that `debug.alias.term` is enough to imply "force" too.

PR #5120, opened 2024-06-22
Fix roundtrip bug when sub-docs contain headings

### LINKED ISSUE #4476: DocTable no longer supports Header elements in cells
I've been trying to get some website updates done and fighting what appear to be a few Docs regressions. 

This minified Doc shows the problem where header elements `#` used to be able to be table headings, but are now rejected. 

```
whatProblemsDoesUnisonSolve : Doc
whatProblemsDoesUnisonSolve = {{

  # What problems does Unison solve?

    {{ docTable [ [{{ # The problem }}, {{ # Our remedy }}]
      , [ {{
          a
          }}
        , {{
          b
          }}
        ]
      ] }}
  }}
```

The error is 

```
  offset=128:
  unexpected "# "
  expecting "}}" or white space
```

### LINKED ISSUE #4729: Roundtrip bug in syntax.docCallout.doc
This parses fine:

```
foo =
  docCallout
    (Some
      {{
      # Title
      
        
      }}) {{ This is a callout with a title }}
```

But this doesn't:

```
syntax.docCallout.doc : Doc
syntax.docCallout.doc =
  {{
  A callout is a block of text that is visually distinct from the surrounding
  text. It is used to draw attention to important information.
  
  # Examples
  
    ```
    docCallout None {{ This is a callout }}
    ```
    
    ```
    docCallout (Some {{ 👋 }}) {{ This is a callout with an emoji }}
    ```
    
    ```
    docCallout
      (Some
        {{
          # Title
        
          
        }}) {{ This is a callout with a title }}
    ```
  }}
```

Seems to fail to parse doc syntax inside the example block in a doc

neduard:
Currently the parser errors out when we have headings inside sub-docs eg:

```
    x = {{ # Heading A
    
    ```
    {{ # Heading B }}
    ```
    }}
```

Fails with 

```
I got confused here:

      4 | {{ # Heading B }}


  I was surprised to find a #  here.

  * white space
  * }}
```

I think this is because in `Lexer.hs` we have a single `parentSection :: Int` to keep track of the heading levels and to know how many `#` characters to expect.  However, when we enter a sub-doc, that number needs to be saved somehow and reset back to `0`. 

## Overview

Use a List (read: stack) to push a new `0` each time we enter a doc section.

Closes #4476
Closes #4729 

## Test coverage

I've added three tests in `reparses-with-same-hash.u` - not sure if they are really necessary but I figured it's better to err on the side of caution.

## Loose ends

N/A?

### DIFF
--- a/unison-syntax/src/Unison/Syntax/Lexer.hs
+++ b/unison-syntax/src/Unison/Syntax/Lexer.hs
@@ -78,7 +78,7 @@ data ParsingEnv = ParsingEnv
     -- are we inside a construct that uses layout?
     inLayout :: Bool,
     -- Use a stack to remember the parent section and
-    -- allow docSections within docSessions.
+    -- allow docSections within docSections.
     -- 1 means we are inside a # Heading 1
     parentSections :: [Int],
     -- 4 means we are inside a list starting at the fourth column
--- a/unison-syntax/src/Unison/Syntax/Lexer.hs
+++ b/unison-syntax/src/Unison/Syntax/Lexer.hs
@@ -71,11 +71,18 @@ type BlockName = String
 type Layout = [(BlockName, Column)]
 
 data ParsingEnv = ParsingEnv
-  { layout :: !Layout, -- layout stack
-    opening :: Maybe BlockName, -- `Just b` if a block of type `b` is being opened
-    inLayout :: Bool, -- are we inside a construct that uses layout?
-    parentSection :: Int, -- 1 means we are inside a # Heading 1
-    parentListColumn :: Int -- 4 means we are inside a list starting at the fourth column
+  { -- layout stack
+    layout :: !Layout,
+    -- `Just b` if a block of type `b` is being opened
+    opening :: Maybe BlockName,
+    -- are we inside a construct that uses layout?
+    inLayout :: Bool,
+    -- Use a stack to remember the parent section and
+    -- allow docSections within docSessions.
+    -- 1 means we are inside a # Heading 1
+    parentSections :: [Int],
+    -- 4 means we are inside a list starting at the fourth column
+    parentListColumn :: Int
   }
   deriving (Show)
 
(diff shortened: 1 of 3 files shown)


unison-syntax/src/Unison/Syntax/Lexer.hs
```diff
@@ -71,11 +71,18 @@ type BlockName = String
...
-    parentSection :: Int, -- 1 means we are inside a # Heading 1
-    parentListColumn :: Int -- 4 means we are inside a list starting at the fourth column
+  { -- layout stack
+    layout :: !Layout,
+    -- `Just b` if a block of type `b` is being opened
+    opening :: Maybe BlockName,
+    -- are we inside a construct that uses layout?
+    inLayout :: Bool,
+    -- Use a stack to remember the parent section and
+    -- allow docSections within docSessions.
```
aryairani:
```suggestion
    -- allow docSections within docSections.
```
(was this a typo or intentional?)

aryairani:
Hey thanks for this PR @neduard.

I'm not super familiar with the doc parser, and I was wondering what the  `parentSection` field is even for? cc @pchiusano 
Shouldn't any heading level be supported after any other?

unison-syntax/src/Unison/Syntax/Lexer.hs
neduard: Oh well spotted!  Total typo I'm afraid :sweat_smile:

neduard:
Good question @aryairani , the way I understand it is this is indirectly caused by us parsing markdown into a syntax tree.  For example:

```
f = {{

#### HA

## HB

### HC
  
}}
```

parses to:

```
Open "/home/ed/code/unison/scratch.u"
  WordyId (NameOnly (Name Relative (NameSegment {toUnescapedText = "f"} :| [])))
  Open "="
    Open "syntax.docUntitledSection"
      Open "syntax.docSection"
        Open "syntax.docParagraph"
          Open "syntax.docWord"
            Textual "HA"
          Close
        Close
      Close
      Open "syntax.docSection"
        Open "syntax.docParagraph"
          Open "syntax.docWord"
            Textual "HB"
          Close
        Close
        Open "syntax.docSection"
          Open "syntax.docParagraph"
            Open "syntax.docWord"
              Textual "HC"
            Close
          Close
        Close
      Close
    Close
  Close
Close
```

or in other words:
```
> display f

  # HA
  
  # HB
  
    # HC

```

Notice we don't actually store the heading level in the tokens.  As such, we need to remember the heading level of the parent such that we know whether to create a sub-section or "close" the current one if that makes sense.

I agree though this might be overkill, so happy to hear alternatives!

aryairani:
@neduard Ok I see, thanks.

aryairani (approved):

PR #5121, opened 2024-06-22
Fix roundtrip bug in Bytes.splitAt.doc

### LINKED ISSUE #4727: Roundtrip error in Bytes.splitAt.doc
```
foo = {{ `` 0xs12 `` }} 
```

Roundtrips as:

```
foo = {{ `` 0xs 12 `` }}
```

Which doesn't parse

neduard:
Closes https://github.com/unisonweb/unison/issues/4727

## Overview

Wraps the printing of `BytesLiteral` in a `F.Group` this removes the space between elements.

**NOTE** I am not 100% sure this is the right fix.  I haven't traced exactly how the `Pretty` datatype gets converted to a String.  I just noticed we use `group` in other cases where we don't want separation eg:

```
prettyRemoteBranchInfo :: (URI, ProjectName, ProjectBranchName) -> Pretty
prettyRemoteBranchInfo (host, remoteProject, remoteBranch) =
  -- Special-case Unison Share since we know its project branch URLs
  if URI.uriToString id host "" == "https://api.unison-lang.org"
    then
      P.group $
        "https://share.unison-lang.org/"
          <> prettyProjectName remoteProject
          <> "/code/"
          <> prettyProjectBranchName remoteBranch
  ...
```

Without the fix we would get this failure:

```
    
    
      This looks like a function call, but with a Bytes where the function should be.  Are you missing an operator?
      
        302 | fix_4727 = {{ `` 0xs 900dc0ffee `` }}

```

## Loose ends

I added the test in `reparses-with-same-hash.u` mainly because there isn't a dedicated test for `TermPrinter.hs` (there is only one for `TypePrinter.hs`) I'm wondering if it would be better add it though?

### DIFF
--- a/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
+++ b/parser-typechecker/src/Unison/Syntax/TermPrinter.hs
@@ -459,7 +459,7 @@ pretty0
                     go tm = goNormal 10 tm
                 PP.hang kw <$> fmap PP.lines (traverse go rs)
               (Bytes' bs, _) ->
-                pure $ fmt S.BytesLiteral "0xs" <> PP.shown (Bytes.fromWord8s (map fromIntegral bs))
+                pure $ PP.group $ fmt S.BytesLiteral "0xs" <> PP.shown (Bytes.fromWord8s (map fromIntegral bs))
               BinaryAppsPred' apps lastArg -> do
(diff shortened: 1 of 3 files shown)


aryairani:
I think this is probably the right fix. Just need to merge in `trunk` because there was a conflict with the other PR. (I'm trying as well, but don't mind if you beat me to it.)

aryairani (approved):

Issue #5125, opened 2024-06-24
`merge` claims something is a dependent, but it isn't


mitchellwrosen:
To reproduce:

```
> clone @mitchellwrosen/bug-5125/main
> clone @mitchellwrosen/bug-5125/topic
> switch @mitchellwrosen/bug-5125/main
> merge /topic
```

PR #5130, opened 2024-06-25
add transcript that demonstrates bug in merge

mitchellwrosen:
## Overview

This PR adds a transcript that demonstrates #5125

### DIFF
--- a/unison-src/transcripts/merge.output.md
+++ b/unison-src/transcripts/merge.output.md
@@ -1793,3 +1793,166 @@ project/main> view Foo
   type Foo = Bar
 
 ```
+### Dependent that doesn't need to be in the file
+
+This test demonstrates a bug.
+
+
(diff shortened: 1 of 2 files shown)


unison-src/transcripts/merge.output.md
```diff
@@ -1793,3 +1793,166 @@ project/main> view Foo
...
+
+bar : Nat
+bar =
+  use Nat +
+  foo + foo
+
+
+```
+
+But `bar` was put into the scratch file instead.
```
aryairani, line 1957: Instead of what; instead of being left out I guess?

aryairani (approved):
I'd rephrase/clarify that one line, otherwise good demo.

unison-src/transcripts/merge.output.md
mitchellwrosen: Instead of being put into the underlying namespace (as mentioned above).

PR #5131, opened 2024-06-25
Add IOG cache to substituters

sellout:

## Overview

## Implementation notes

## Test coverage

## Loose ends

### DIFF
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -25,13 +25,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: cachix/install-nix-action@v27
-      if: runner.os == 'Linux'
-      with:
-        extra_nix_config: |
-          extra-trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
-          extra-substituters = https://cache.iog.io
-    - uses: cachix/install-nix-action@v27
-      if: runner.os != 'Linux'
     - uses: cachix/cachix-action@v15
       with:
         name: unison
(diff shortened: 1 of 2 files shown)


.github/workflows/nix-dev-cache.yaml
```diff
@@ -25,13 +25,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: cachix/install-nix-action@v27
-      if: runner.os == 'Linux'
```
aryairani, line 28: This is the substituter that's poisoned on Mac (unless we have some way to confirm that it no longer is, but that seems hard/impossible), so don't think we want it for non-Linux.

sellout: Ah, wait – from input-output-hk/haskell.nix#2018[^1] I thought the poisoned cache would result in a build (or possibly runtime) failure, not a malicious injection or something.

[^1]: which I should link to in the description above, once I actually write it, as well as #4251.

The reason this is a draft is that I wanted to test it after #5142 lands, and see if we can restore that cache to the flake, which could lower the pressure on our own cache[^2].

[^2]: Using the IOG cache on Linux here also seems to be a problem – it means our cache will avoid storing anything that is found in IOG’s cache, but then if they’re needed by contributors, they’ll need to be rebuilt locally, because that cache isn’t not made available to users via the flake (and, as it seems you know, we can’t provide substituters in the flake based on the system).

I also think that _maybe_ the reason you were running into it with IOG is that 9.2.8 isn’t one of their “officially supported“ versions (neither is 9.6.5), so it is probably selected by fewer users, and thus no one notices/cares.

aryairani: Yeah, a runtime error in the build tools, wasn't worried about malicious injection; and we don't actually test them here, right?

My worry was that the bad packages would then be mirrored to the cachix cache, and then sadness when we use that one as well.

aryairani:
likely incorrect due to misunderstanding cachix, or still dangerous due to https://github.com/input-output-hk/haskell.nix/issues/2018

Issue #3513, opened 2022-10-17
Composition of effectful functions typechecks when it shouldn't


runarorama:
This typechecks, but shouldn't:

```
  IO.tryEval : '{g, IO, Exception} a ->{g, IO, Exception} a
  IO.tryEval = reraise << catchAll.impl

  IO.catchAll.impl : '{IO, Exception} a ->{IO} Either Failure a
  IO.catchAll.impl thunk =
    handle tryEval.impl '(catch thunk)
    with
      cases
        { x }                    -> x
        {Exception.raise f -> _} -> Left f
```

The argument to `IO.tryEval` requires an ability `g`, but the type of `catchAll.impl` should prohibit that.

Note that this does not typecheck:

```
  IO.tryEval : '{g, IO, Exception} a ->{g, IO, Exception} a
  IO.tryEval x = reraise (catchAll.impl x)
```

The type of `<<` is: 
```
  (Function.<<) : (b ->{𝕖} c) -> (a ->{𝕖} b) -> a ->{𝕖} c
  (Function.<<) f g x = f (g x)
```

PR #5133, opened 2024-06-25
Don't relax type when doing an instantiateL during subtyping

### LINKED ISSUE #5129: type checker not catching disallowed ability

Below I have a `forkThread` function that takes a thunk that is only allowed to use the `IO` ability. But when I call into it via `List.foreach`, I am able to pass in a thunk that requires another ability (in this case `Exception`, but I can add arbitrary abilities to the same effect). The type checker should catch that this is not allowed.

Input:
````
```ucm
.> project.create-empty proj
proj/main> builtins.mergeio
```

```unison:error
forkThread :
  '{IO} ()
  ->{IO} ()
forkThread thunk = todo "not important"

List.foreach : (a -> {g} ()) -> [a] -> {g} ()
List.foreach = todo "not important"

main : '{IO, Exception} ()
main = do
  thunk : '{IO, Exception} ()
  thunk = todo "you shouldn't be able to pass this into forkThread because it requires the Exception ability"
  List.foreach forkThread [thunk]
```
````

Output:
````
```ucm
.> project.create-empty proj

  🎉 I've created the project proj.

  🎨 Type `ui` to explore this project's code in your browser.
  🔭 Discover libraries at https://share.unison-lang.org
  📖 Use `help-topic projects` to learn more about projects.
  
  Write your first Unison code with UCM:
  
    1. Open scratch.u.
    2. Write some Unison code and save the file.
    3. In UCM, type `add` to save it to your new project.
  
  🎉 🥳 Happy coding!

proj/main> builtins.mergeio

  Done.

```
```unison
forkThread :
  '{IO} ()
  ->{IO} ()
forkThread thunk = todo "not important"

List.foreach : (a -> {g} ()) -> [a] -> {g} ()
List.foreach = todo "not important"

main : '{IO, Exception} ()
main = do
  thunk : '{IO, Exception} ()
  thunk = todo "you shouldn't be able to pass this into forkThread because it requires the Exception ability"
  List.foreach forkThread [thunk]
```

```ucm

  change:
  
      List.foreach : (a ->{g} ()) -> [a] ->{g} ()
      forkThread   : '{IO} () ->{IO} ()
      main         : '{IO, Exception} ()

```

```



🛑

The transcript was expecting an error in the stanza above, but did not encounter one.
````

 - `ucm --version` 7172bb8e4
 - OS/Architecture: x86 NixOS

**Additional context**

I ran into both this and #5128 in the same definition for a confusing double whammy.

dolio:
This is the case of `a < T` for some structured `T`. By relaxing, we are actually allowing `a` to be a _supertype_ of `T` as far as abilities go, which is not correct. Seems like it was just erroneously mirrored from the opposite case, where `T < a` means that `a` _could_ be more general than `T`.

Fixes #5129 and #3513.

### DIFF
--- a/parser-typechecker/src/Unison/Typechecker/Context.hs
+++ b/parser-typechecker/src/Unison/Typechecker/Context.hs
@@ -2561,8 +2561,7 @@ subtype tx ty = scope (InSubtype tx ty) $ do
     go ctx (Type.Var' (TypeVar.Existential b v)) t -- `InstantiateL`
       | Set.member v (existentials ctx)
           && notMember v (Type.freeVars t) = do
-          e <- extendExistential Var.inferAbility
-          instantiateL b v (relax' False e t)
+          instantiateL b v t
     go ctx t (Type.Var' (TypeVar.Existential b v)) -- `InstantiateR`
       | Set.member v (existentials ctx)
           && notMember v (Type.freeVars t) = do
--- a/unison-src/transcripts-using-base/fix5129.md
+++ b/unison-src/transcripts-using-base/fix5129.md
@@ -1,0 +1,45 @@
+```ucm:hide
+.> builtins.mergeio
+```
+
+Checks for some bad type checking behavior. Some ability subtyping was
+too lenient when higher-order functions were involved.
+
+```unison:error
+foreach : (a ->{g} ()) -> [a] ->{g} ()
+foreach f = cases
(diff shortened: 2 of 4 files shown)


ceedubs (approved):
Nice thanks!

aryairani (approved):

PR #5140, opened 2024-06-26
Add numbered results to `test`, `io.test`, and `io.test.all`

### LINKED ISSUE #5080: numbered results for test and io.test.all output
**Is your feature request related to a problem? Please describe.**

When a test fails, I often want to look at the source code for the test.

**Describe the solution you'd like**

Right now the output of the `test` command looks like:

```
   ◉ time.patterns.asctimeFormat.test        Passed
   ◉ time.patterns.rfc7231DateTime.test850   Passed
   ◉ time.patterns.rfc7231DateTime.testIMF   Passed
   
   ✅ 3 test(s) passing
```

It would be nice if the output looked like:

```
   1. ◉ time.patterns.asctimeFormat.test        Passed
   2. ◉ time.patterns.rfc7231DateTime.test850   Passed
   3. ◉ time.patterns.rfc7231DateTime.testIMF   Passed
   
   ✅ 3 test(s) passing
```

And I could do `view 2` to view `time.patterns.rfc7231DateTime.test850`.

**Describe alternatives you've considered**

N/A

**Additional context**

I think that this applies to both `test` and `io.test.all`.

The test output can be a bit more complicated than what I've pasted above. There can be one section for cached results and another section for new results. Also there can be failing tests.

sellout:
## Overview

The test results are now numbered, and make it possible to view, edit, delete, etc. the definitions.

Fixes #5080.

## Implementation notes

`TestResult` is moved to `NumberedOutput`, and given corresponding numbered args.

## Test coverage

There are a number of tests that have updated output, showing the numbers for all affected commands, and there’s a new transcript specifically to show that `numberedArgs` is set.


aryairani (commented):
Great!

aryairani (dismissed):

aryairani:
It looks like a transcript had whitespace changes that are upsetting CI?

We often see whitespace churn I can't explain, not sure if this is an instance.

sellout:
> It looks like a transcript had whitespace changes that are upsetting CI?

Fixed. The issue was that I opened the output in my editor before committing, and my editor likes to trim trailing whitespace.

ceedubs:
Ooh nice thanks!

The result type for tests is `[Result]` (a list). If I recall correctly, each term is displayed in the output once for _each_ test result. Is that what the `testABunchOfNats` output is showing? If so, should each `testABunchOfNats` output show a different numbered arg, or should each term only get a single number? Probably not a big deal either way: just making sure it's a conscious decision.

aryairani:
> > It looks like a transcript had whitespace changes that are upsetting CI?
> 
> Fixed. The issue was that I opened the output in my editor before committing, and my editor likes to trim trailing whitespace.

Ah... that may be what I have been seeing happen on various PRs in the past, too.

aryairani:
> Ooh nice thanks!
> 
> The result type for tests is `[Result]` (a list). If I recall correctly, each term is displayed in the output once for _each_ test result. Is that what the `testABunchOfNats` output is showing? If so, should each `testABunchOfNats` output show a different numbered arg, or should each term only get a single number? Probably not a big deal either way: just making sure it's a conscious decision.

Good question...
@sellout Would it be easy to dedup these? (probably RLE style is enough)

ceedubs:
If it's much effort I wouldn't worry about it too much. Also TBH repeating the term name a bunch of times isn't particularly helpful. If it's easier to add a section for each term that might be better. But I suspect it is not easier, and it is a larger change.

sellout:
There are other outputs that do some similar deduplication. E.g.:
```ucm
zestful-tiger/main> delete.term tmp.moo tmp.meh
  
  I didn't delete the following definitions because they are still in use:
  
  Dependency   Referenced In
  meh          1. tmp.baz
               
  moo          2. tmp.baz
               3. tmp.bar
```
However this is still a bit different – the numbered results aren’t deduped, but the terms we tried to delete are, and correspondingly, the numbered args are in the second column.

I would like to do the deduplication, but ideally in a separate PR, as I feel like there might be some back and forth on exactly what it should be.

aryairani (approved):

PR #5144, opened 2024-06-27
Group test results by definition

sellout:
## Overview

Avoids duplication in the output.

Follow-up to #5140.

With this new display, I’m wondering if it makes sense to _not_ split the failures from the successes, so all the terns for a definition are together rather than having the definition duplicated in each list.

## Test coverage

This is reflected in a number of transcripts.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput/Tests.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput/Tests.hs
@@ -14,7 +14,6 @@ import Data.Map qualified as Map
 import Data.Set qualified as Set
 import Data.Set.NonEmpty (NESet)
 import Data.Set.NonEmpty qualified as NESet
-import Data.Tuple qualified as Tuple
 import Unison.ABT qualified as ABT
 import Unison.Builtin.Decls qualified as DD
 import Unison.Cli.Monad (Cli)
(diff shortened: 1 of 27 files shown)


sellout:
So, what do you think about moving to
```ucm
  Cached test results (`help testcache` to learn more)
  
    1. fix5080.tests  ◉ got the right answer
                      ✗ [1, 2, 3]
   
   🚫 1 test(s) failing, ✅ 1 test(s) passing
```
instead of
```ucm
  Cached test results (`help testcache` to learn more)
  
    1. fix5080.tests  ◉ got the right answer

    2. fix5080.tests  ✗ [1, 2, 3]
   
   🚫 1 test(s) failing, ✅ 1 test(s) passing
```

ceedubs:
Nice!

@sellout I agree that for the small example, not printing the successes and failures separately looks better. But I do think that there is some benefit in putting all of the failures at the top or bottom so it's easy to find them if you have a bunch of tests run. No super strong opinions, though.

ceedubs (approved):

sellout:
> I agree that for the small example, not printing the successes and failures separately looks better. But I do think that there is some benefit in putting all of the failures at the top or bottom so it's easy to find them if you have a bunch of tests run.

Yeah, it can be nice to have the failures all together at the end. The output is also colorized, though, so it does make scanning a long list for failures pretty easy.

Regardless, it can stay the way it is for now, and it’s easy enough to change later if we want.

ceedubs:
Just gave this a try and it's working great!

PR #5177, opened 2024-07-03
feat: report merge precondition violations in `todo`

### LINKED ISSUE #5088: update `todo`

mitchellwrosen:
## Overview

Fixes #5088

This PR amends `todo` to report merge precondition violations (constructor aliases, missing constructor names, nested decl aliases, and stray constructors).

## Test coverage

I've added transcripts that demonstrate this feature


unison-src/transcripts/todo.md
```diff
@@ -3,19 +3,15 @@
 When there's nothing to do, `todo` says this:
 
 ```ucm
-project/main> todo
```
aryairani, line 6: fwiw I think `project` is fine in a transcript too, possibly clearer than `scratch`; I know we are de facto standardizing on `scratch`.

unison-src/transcripts/todo.output.md
```diff
@@ -39,16 +35,16 @@ bar = foo + foo
...
   ⍟ I've added these definitions:
   
     bar : Nat
     foo : Nat
 
-project/main> todo
+scratch/main> todo
 
-  These terms call `todo`:
+  These terms call `todo`.
```
aryairani: I think the `:` was more correct here?

unison-src/transcripts/todo.output.md
```diff
@@ -97,10 +93,251 @@ project/main> delete.namespace.force foo
   Dependency   Referenced In
   bar          1. baz
 
-project/main> todo
+scratch/main> todo
 
-  These terms do not have any names in the current namespace:
+  These terms do not have any names in the current namespace.
```
aryairani: Same (`:` vs `.`)

unison-src/transcripts/todo.output.md
```diff
@@ -97,10 +93,251 @@ project/main> delete.namespace.force foo
...
+scratch/main> debug.alias.term.force foo bar
+
+  Done.
+
+scratch/main> todo
+
+  ❓
+  
+  The term bar has conflicting definitions: 1. foo 2.
+  bar#cq22mm4sca
```
aryairani: * Is this just a PPE issue that causes it to show `foo` instead of a kind of `bar`?
* Would it be hard to get it to show them both as `bar`-something?
* This should be an indented, newline separated list, right?

unison-src/transcripts/todo.output.md
```diff
@@ -97,10 +93,251 @@ project/main> delete.namespace.force foo
...
+
+  ⍟ I've added these definitions:
+  
+    lib.foo : Nat
+
+scratch/main> todo
+
+  There's a type or term at the top level of the `lib`
+  namespace, where I only expect to find subnamespaces
+  representing library dependencies. Please move or remove it.
```
aryairani, line 183: It would be nice if this said the name of the bad thing, but hopefully terms in `lib` are going away altogether before we care too much. So nevermind.

We should add a ticket to modify `move`/`alias` to not allow destinations in `lib` and hope that it doesn't mess up any hacky-but-necessary workflows.

unison-src/transcripts/todo.output.md
```diff
@@ -97,10 +93,251 @@ project/main> delete.namespace.force foo
...
+scratch/main> delete.term Foo.Bar
+
+  Done.
+
+scratch/main> todo
+
+  These types have some constructors with missing names.
+  
+    1. Foo
+
```
aryairani: Could use a hint here, to `view` and then... can they use `alias` to create these names by hash if they don't already exist? Or what's the fix in this case?

unison-src/transcripts/todo.output.md
```diff
@@ -97,10 +93,251 @@ project/main> delete.namespace.force foo
...
+  Done.
+
+scratch/main> todo
+
+  These constructors are not nested beneath their corresponding
+  type names. Please either move or delete them.
+  
+    1. Baz
+
```
aryairani: Would it be hard to show what type they belong to?

aryairani (changes requested):
Nice!
I left a couple questions/comments

unison-src/transcripts/todo.output.md
aryairani: https://github.com/unisonweb/unison/issues/5184

unison-src/transcripts/todo.output.md
mitchellwrosen: I thought it wouldn't be that bad to pipe things through, but it's looking like more piping than I anticipated. I'm going to make a ticket to track.

mitchellwrosen:
@aryairani I think I addressed everything except this one https://github.com/unisonweb/unison/issues/5197, for which I made that ticket

aryairani (approved):

PR #5180, opened 2024-07-05
move pretty-print helpers out of CommandLine.hs

aryairani:
to avoid a potential cycle between CommandLine.hs and InputPatterns.hs that otherwise came up on another branch.

InputPatterns.hs also has formatting helpers that could be used in CommandLine.hs. Maybe they should be moved too, but I haven't thought about it.


aryairani:
This will also allow using `makeExample` in #5054; in general we should be using that, instead of hard-coding command names into strings.

PR #5181, opened 2024-07-05
initial transcripts to track help and parse error messages

aryairani:
Part of improving our transcript coverage of user-facing output.

The second transcript is obviously a mess, arguably not worth committing, but I want to get something started that we can build off of.


aryairani:
These are included in #5054

PR #5187, opened 2024-07-05
`doc2` is a parser in lexer’s clothing

### LINKED ISSUE #5076: parser bug on docs with double backticks
Input:
````
```unison:hide
x = {{
    ``let "me"`` live
    ``do "me"`` in      
  }}
```
````

Output:
````
  I got confused here:
  
      2 |     ``let "me"`` live
  
  
  I was surprised to find a syntax.docWord here.
  
  * :
  * and
  * bang
  * do
  * false
  * handle
  * if
  * infixApp
  * let
  * or
  * quote
  * termLink
  * true
  * tuple
  * typeLink
````

This is causing a round trip error with our website currently for this term:

https://share.unison-lang.org/@unison/website/code/main/latest/terms/learn/languageReference/delayedComputations/syntacticPrecedence

sellout:
## Overview

This exposes the `Doc` parser and then converts directly from the Doc AST to Unison `Term`s.

This makes the permitted structure of Docs much clearer, and should have no impact on users. It should make working on the Doc parser much easier in future.

Part of the separation is that the Doc parser has no notion that it is wrapped in `{{`/`}}`. That is handled by the Unison lexer, and Doc just processes the part of the stream that it is allowed to.

The Doc parser is also completely independent of the Unison lexer/parser, so this could be used to add support for Unison Doc to any PL that has a Megaparsec parser..

Fixes #5076.

## Implementation notes

This was already the case, but previously the parser directly emitted a list of `Lexeme`s as it went. This now preserves the tree structure created by the parser, and adds a new `Lexeme` case called `Doc`, which passes that structure to the Unison parser to be converted directly to Unison `Term`s.

Doc can contain Unison code in a number of places, and those are stored as `[Token Lexeme]` in the Doc structure. The Unison parser then runs “sub-parses” on these chunks as it encounters them.

It also makes sense to review this commit-by-commit. E.g., the fourth commit separated the parsing (in Lexer.hs) from creating the `Lexeme` stream, which allows you to see that the parsing logic is almost entirely untouched, and that we only switch from building a stream to building a tree. Then the fifth commit removes the stream production in favor of handing the tree to the Unison parser and allowing it to produce the `Term`s directly from the tree.

Probably a number of things. One is just the data representation of `Doc`, which is more complicated than would be ideal, but is informed by what is possible to parse. Simplifying the data model requires changing the parser to match it.

I had also considered making it a GADT, which would simplify some things, but then that would make it harder to parallel in Unison’s data model of `Doc`, and would hide some of the complexities that should be eliminated.

## Test coverage

All transcripts which include `{{ }}` test this change.

## Loose ends

There are a lot of improvements to the data model to be made, but this change was intended strictly as a refactor, to clarify some things.


pchiusano:
Oh, cool. I don't know why I didn't think of this.

The whole time I was doing the doc support, I was finding it annoying that the lexer and parser were separated (they were previously combined, but we moved away from that due to it being overly complicated and hard to debug to deal with layout and whitespace issues in the parser), but this PR seems like it gets the best of both worlds.

aryairani (approved):

Issue #5190, opened 2024-07-06
`fzf` runs, even in transcript mode, on under-saturated commands


aryairani:
**Describe and demonstrate the bug**
Please attach a [ucm transcript](https://www.unison-lang.org/docs/tooling/transcripts/) if possible, calling out the unexpected behavior in the text. e.g.

Input:
````
```ucm:error
scratch/main> display
```

```unison
a = 1
```

```ucm
scratch/main> add
scratch/main> display
```
````

Output:
````
```ucm
scratch/main> display

⚠️

Sorry, I was expecting an argument for the definition to display, and I couldn't find any to suggest to you. 😅

```
```unison
a = 1
```

```ucm

  Loading changes detected in scratch.u.

  I found and typechecked these definitions in scratch.u. If you
  do an `add` or `update`, here's how your codebase would
  change:
  
    ⍟ These new definitions are ok to `add`:
    
      a : ##Nat

```
```ucm
scratch/main> add

  ⍟ I've added these definitions:
  
    a : ##Nat

scratch/main> display

```
````

In the first `ucm` block, the `display` is about failing to launch fzf due to lack of suggestions, suppressing an error about not having provided any parameters. In the second `ucm` block, `fzf` launches and I hit Esc to exit, and it seems like the command is aborted with no output, instead of returning to the command's argument parser:

```haskell
display :: InputPattern
display =
  InputPattern
    "display"
    []
    I.Visible
    [("definition to display", OnePlus, definitionQueryArg)]
    ( P.lines
        [ "`display foo` prints a rendered version of the term `foo`.",
          "`display` without arguments invokes a search to select a definition to display, which requires that `fzf` can be found within your PATH."
        ]
    )
    $ maybe
      (wrongArgsLength "at least one argument" [])
      (fmap (Input.DisplayI Input.ConsoleLocation) . traverse handleHashQualifiedNameArg)
      . NE.nonEmpty
```

**Screenshots**

<img width="358" alt="image" src="https://github.com/unisonweb/unison/assets/538571/093a1d74-fd01-46fd-a1e0-3b91fb4650d2">

**Environment (please complete the following information):**
ucm 0.5.23, macOS

**Additional context**

PR #5192, opened 2024-07-06
avoid numeric vs textual clash in args length error

aryairani:
## Overview

image

## Implementation notes

Just calls `Numerals.us_cardinal` instead of `show` on the number of arguments to be mentioned in the error message.

* `numerals` doesn't seem to be maintained anymore, but maybe that doesn't matter

* "zero" (the default) vs special-casing "none"

## Test coverage

I attempted to add additional transcript tests to cover the under-saturated case, but couldn't due to #5190. I would create them as soon as it's possible.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -64,8 +64,13 @@ extra-deps:
   - recover-rtti-0.4.2@sha256:c179a303921126d8d782264e14f386c96e54a270df74be002e4c4ec3c8c7aebd,4529
   - lsp-2.2.0.0@sha256:82fbf4b69d94d8d22543be71f89986b3e90050032d671fb3de3f8253ea1e5b6f,3550
   - lsp-types-2.0.2.0@sha256:a9a51c3cea0726d91fe63fa0670935ee720f7b31bc3f3b33b2483fc538152677,29421
+  - numerals-0.4.1@sha256:f138b4a0efbde3b3c6cbccb788eff683cb8a5d046f449729712fd174c5ee8a78,11430
   - row-types-1.0.1.2@sha256:4d4c7cb95d06a32b28ba977852d52a26b4c1f695ef083a6fd874ab6d79933b64,3071
   - network-udp-0.0.0
+
+allow-newer: true
+allow-newer-deps:
+  - numerals
 
 ghc-options:
   # All packages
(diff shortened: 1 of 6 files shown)


unison-cli/src/Unison/CommandLine/InputPatterns.hs
```diff
@@ -342,7 +344,11 @@ wrongStructuredArgument expected actual =
 
 wrongArgsLength :: Text -> [a] -> Either (P.Pretty CT.ColorText) b
 wrongArgsLength expected args =
-  Left . P.text $ "I expected " <> expected <> ", but received " <> Text.pack (show $ length args) <> "."
+  let foundCount =
+        if null args
+          then "none"
+          else fromMaybe (tShow $ length args) $ Numeral.us_cardinal defaultInflection (length args)
```
sellout:
```suggestion
        case length args of
          0 -> "none"
          n -> fromMaybe (tShow n) $ Numeral.us_cardinal defaultInflection n
```

stack.yaml
```diff
@@ -64,9 +64,14 @@ extra-deps:
   - recover-rtti-0.4.2@sha256:c179a303921126d8d782264e14f386c96e54a270df74be002e4c4ec3c8c7aebd,4529
   - lsp-2.2.0.0@sha256:82fbf4b69d94d8d22543be71f89986b3e90050032d671fb3de3f8253ea1e5b6f,3550
   - lsp-types-2.0.2.0@sha256:a9a51c3cea0726d91fe63fa0670935ee720f7b31bc3f3b33b2483fc538152677,29421
+  - numerals-0.4.1@sha256:f138b4a0efbde3b3c6cbccb788eff683cb8a5d046f449729712fd174c5ee8a78,11430
```
sellout, line 67: I think this probably isn’t important to anyone other than me, but I’m never sure exactly which parts of stack.yaml are picked up by haskell.nix. And the nix dev GitHub workflow doesn’t yet actually build Unison, so I’m running `nix build github:unisonweb/unison/topic/numerals` right now just to make sure it’s all good.

But again, I don’t think `nix build` should hold up the process for anyone else.

sellout (approved):
Ah, wonderful – I had made a note to look into what packages might exist for this. I was hoping for one that was also parameterized on “style” (e.g., `10 -> "ten"`, but `1023 -> "1,023"`), but I don’t think that matters for this use case.

Just one minor suggestion.

stack.yaml
sellout: Oh, I forgot to update this – it built fine.

aryairani:
> 1023 -> "1,023"

Lol, I did not consider "I was expecting one thousand twenty three." But yeah I think we are safe in this case.

stack.yaml
aryairani: No problem. Does a passing "Nix development cache" run not indicate success here?

sellout: No – it currently only builds the dev tools. With the LTS upgrade PR it will build everything.

PR #5194, opened 2024-07-09
bugfix: synhash var debruijn indices, not names

mitchellwrosen:
## Overview

This PR changes the syntactic hashing of variables to be the hash of their nameless debruijn index, not the var name.

Also included in this PR: a `debug.synhash.term` command that was useful for debugging.

## Test coverage

There are no tests at the moment... I haven't thought of a reliable way of showing off this change as a syntactic hashing "feature" in a transcript.

It would be trivial to simply _capture_ the syntactic hash of any term that binds a variable, just to show off that it changes before and after this change to the syntactic hashing code. But that would probably require some tweaks to `debug.synhash.term`, as right now it prints the syntactic hash to the console and also writes all of the tokens to a file.

I have manually tested these changes against @ceedubs's big merge branch, though, and the results look good: no more superfluous-looking conflicts.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
+++ b/unison-cli/src/Unison/Codebase/Editor/HandleInput.hs
@@ -59,6 +59,7 @@ import Unison.Codebase.Editor.HandleInput.CommitMerge (handleCommitMerge)
 import Unison.Codebase.Editor.HandleInput.CommitUpgrade (handleCommitUpgrade)
 import Unison.Codebase.Editor.HandleInput.DebugDefinition qualified as DebugDefinition
 import Unison.Codebase.Editor.HandleInput.DebugFoldRanges qualified as DebugFoldRanges
+import Unison.Codebase.Editor.HandleInput.DebugSynhashTerm (handleDebugSynhashTerm)
 import Unison.Codebase.Editor.HandleInput.DeleteBranch (handleDeleteBranch)
 import Unison.Codebase.Editor.HandleInput.DeleteProject (handleDeleteProject)
 import Unison.Codebase.Editor.HandleInput.EditNamespace (handleEditNamespace)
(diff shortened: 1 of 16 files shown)


aryairani:
> There are no tests at the moment... I haven't thought of a reliable way of showing off this change as a syntactic hashing "feature" in a transcript.

I don't think we need to capture the change or the specific syntactic hashes per se, but in general more like the overall behavior that the PR. Is the issue that we don't really know/have a minimal test case that would have failed before and doesn't fail now? I'm imagining something in `merge.md`.

aryairani:
Got this error on `pull_request` CI; I'm re-running it to see if it happens twice.
```
argument-parsing.can compile, then run compiled artifact.unison transcript unison-cli-integration/integration-tests/IntegrationTests/transcript.md FAILURE 
ExitSuccess
** did not equal expected value **
ExitFailure 1
  crash, called at src/EasyTest.hs:75:10 in easytest-0.1-HXNbGChvVO01IYAwkLBSx9:EasyTest
  expectEqual, called at integration-tests/IntegrationTests/ArgumentParsing.hs:78:3 in main:IntegrationTests.ArgumentParsing
💥  argument-parsing.can compile, then run compiled artifact.unison transcript unison-cli-integration/integration-tests/IntegrationTests/transcript.md
```

mitchellwrosen:
> I don't think we need to capture the change or the specific syntactic hashes per se, but in general more like the overall behavior that the PR. Is the issue that we don't really know/have a minimal test case that would have failed before and doesn't fail now? I'm imagining something in merge.md.

Let me see if it's easy to capture the failure mode here just by propagating an update to a term with a `_` variable binding. If that reliably (i.e. always, or probably always) changes the variable's name (e.g. from `_1000` to `_2000`) then the regression test would be useful to have.

mitchellwrosen:
I added [this test](https://github.com/unisonweb/unison/pull/5194/commits/94c6bfca628049b2918d47d6f804aaddcd945c8e) and confirmed it fails on trunk.

aryairani (approved):

PR #5196, opened 2024-07-09
Pin Haskell tool versions for VS Code

sellout:
This should help VS Code users get the same versions of various tools as Nix users do when working on Unison.

This also has the flake get its version pins from the VS Code settings. And we pin Cabal now, too.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "haskell.toolchain": {
+    "cabal": "3.10.3.0",
+    "hls": "2.8.0.0",
+    "stack": "2.15.7"
+  }
+}
--- a/flake.nix
+++ b/flake.nix
@@ -26,11 +26,15 @@
       "aarch64-darwin"
     ]
     (system: let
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "haskell.toolchain": {
+    "cabal": "3.10.3.0",
+    "hls": "2.8.0.0",
+    "stack": "2.15.7"
```
sellout, line 5: I wonder if VS Code or the extension will complain if we put the other version pins here, so they’re all in one place.

Or is that a bad idea because VS Code users might think that it’s _actually_ pinning Ormolu, etc.

sellout:
This is on top of the LTS upgrade (#5142). I don’t intend to merge it there, but just to block until that is merged.

sellout:
Well, that’s interesting … I guess I shouldn’t point a merge at a branch that might get deleted.

sellout:
Trying to re-point this at trunk now …

PR #5204, opened 2024-07-10
Replace transcript parser with `cmark`

### LINKED ISSUE #1421: indented `unison` block in transcript angers parser
```ucm:hide
      .builtin> alias.type ##Nat Nat
      .builtin> alias.term ##Nat.+ Nat.+
      ```
      ```unison:hide
      unique type A = A Nat
      unique type B = B Nat Nat
      ```
^-- 2 spaces of indent; convenient for aligning heredoc-style transcript strings in Haskell files.

It complains with:

    ⚙️   Processing stanza 3 of 3.  ```unison
    unique type A = A Nat
      unique type B = B Nat Nat
      ```

Indented `ucm` stanzas seem to be okay, indented `unison` stanzas break down.

An indented .u file is okay though, just not an indented `unison` stanza in a transcript.

### LINKED ISSUE #1809: Transcript parser occasionally leaves out a trailing backtick
As in, the input file has `` `foo` `` but the output looks like `` `foo ``. Example is here: https://github.com/unisonweb/unison/pull/1808/files#diff-49f174e20f1f6d21f42d2210d20568b37501cee5b6d789d4e622385b9fe5686bR33

Note to self, the transcript parser could make use of `manyTill` combinator, which I didn't know about at the time I wrote the transcript parser originally. So the current parser has some weird logic where it temporarily pauses whenever it hits a backtick, in case that one backtick is followed by two more backticks. 

Doing ````manyTill anyChar (string "```")```` would probably fix this and be simpler to boot.

### LINKED ISSUE #2826: Transcripts don't support fenced codeblocks in Docs
Because the triple backtick syntax is used to open transcript stanzas AND fenced codeblocks in the docs, any fenced codeblock within a Doc snippet is interpreted as the closure of a stanza. 

A workaround for the curious: 
You can write your unison code in a separate text file, add a `load` statement in the ```ucm block, and then add it to the codebase.

### LINKED ISSUE #5141: Transcript parser ignores comment delimiters
The following transcript errors (“I don't know how to invalid.command”) as if there were no comment delimiters.

````
<!--
```ucm
.> invalid.command
```
-->
````

 - ucm: 0.5.23
 - OS/Architecture: macOS 14.5, M1

**Additional context**

I came across this when trying to disable individual blocks in a transcript I was debugging.

### LINKED ISSUE #5168: transcript parser blows up on more backticks

In a transcript, this valid Markdown blows up:
`````
The `edit` seems to suppress a following ```` ```unison ```` block:
```unison
b = 2
```
`````

```
    An error occurred while parsing the following file: /tmp/transcript-error.md
  
    /tmp/transcript-error.md:5:1:
      |
    5 | <empty line>
      | ^
    unexpected end of input
    expecting "api", "ucm", "unison", or spaces
```
ucm v0.5.23

**Additional context**
This is because the transcript parser doesn't actually understand Markdown. We should use an off-the-shelf Markdown processor (e.g. cmark, mmark) instead of building our own.

### PR DESCRIPTION
**Pull request by sellout**
## Overview

We don’t need a very rich parser for transcripts, but we _do_ need to reliably identify fenced code blocks, and that implies a number of subtle cases. Using a polished CommonMark parser/printer handles those subtleties for us.

This
- fixes #1421;
- fixes #1809;
- fixes #2826;
- fixes #5141;
- fixes #5168;
- fixes an unreported issue where fenced code blocks without an info string would use the beginning of the content as the info string;
- fixes an unreported issue where output blocks for generated `unison` stanzas (which could contain nested fenced blocks) might not have long-enough fences;
- fixes an unreported issue where Unison errors were reported on the wrong line number (and thus the printed error lines were also incorrect);
- fixes an unreported issue where API blocks were missing trailing newlines, producing broken output blocks; and
- corrects a few transcripts that had incorrect Markdown in them.

## Implementation notes

We had also considered unifying the transcript and `Doc` parsers, but that was undesirable for a few reasons, including
- we want transcripts to match common Markdown usage and
- we want to be flexible in what our own `Doc` language supports (regardless of how Markdown-like it may be at various points).

I chose `cmark` for a few reasons:
- it’s a wrapper around `libcmark`, which is the reference implementation of CommonMark, so it should be correct;
- it provides both a parser and a printer (unlike MMark); and
- it is extremely fast (about 20x faster than MMark), so the fact that our home-rolled parser got to skip over everything that’s not a block isn’t an issue.

This only _partially_ uses the `cmark` printer. I think it should use it entirely, but for the cases where we do streaming output (processing UCM commands, etc.) it’s a more involved change. So I think it should be handled separately.

## Test coverage

This affects every transcript. I tried to keep the changes to the outputs as minimal as possible.

## Loose ends

Unison still pretty-prints its own code blocks in some cases (basically, whenever we use the streaming output, like with executing `ucm` commands), but those should also be changed to use `cmark`.

### CODE REVIEW

**Comment by sellout** [comment]
It might not be a bad idea to review this one commit at a time, but also the changes are generally isolated enough to do it at once.

File: unison-cli/src/Unison/Codebase/TranscriptParser.hs [hunk 1]
```diff
@@ -657,44 +655,17 @@ fenced = do
           hide <- lineToken hidden
           err <- lineToken expectingError
           fileName <- optional untilSpace1
-          blob <- spaces *> untilFence
-          pure $ Unison hide err fileName blob
+          P.setInput body
+          blob <- spaces *> (Text.init <$> P.getInput)
```
sellout, line 659: This `Text.init` is the fly in this ointment.

I’m not sure why there’s an extra newline parsed when I write it this way. But removing it makes the output line up with what already exists (and also match the input blocks).

I had done the `fenced` parsing a bit differently at first, and it didn’t have this issue. But that code was more complicated and slower.

pchiusano:
Fixing the bugs seems good, but I have a concern that many people are not going to have libcmark as a shared library on their system. Does this matter? Maybe it only matters if you're running transcripts, or could it cause a crash on startup even if you aren't using transcripts?

Parsing of transcripts is a negligible contribution to most transcript runtime, so I wouldn't worry about that angle. I also wouldn't worry about exact adherence to CommonMark specification, since transcripts are mostly a tool for our testing purposes. Basically any reasonable markdown parser will do.

aryairani:
@pchiusano no, 
> This package [...] includes sources for [libcmark](http://github.com/jgm/cmark) and does not require prior installation of the C library.

unison-cli/src/Unison/Codebase/TranscriptParser.hs
aryairani: What's this comment in reference to?

sellout:
> I have a concern that many people are not going to have libcmark as a shared library on their system. Does this matter? Maybe it only matters if you're running transcripts, or could it cause a crash on startup even if you aren't using transcripts?

Well, @aryairani beat me to this one. But I do think that it matters, even if users may never run a transcript.

> Parsing of transcripts is a negligible contribution to most transcript runtime, so I wouldn't worry about that angle. I also wouldn't worry about exact adherence to CommonMark specification, since transcripts are mostly a tool for our testing purposes. Basically any reasonable markdown parser will do.

I agree. I think it’s important that it also has a Markdown _printer_, too, which is less common. Many tools just parse and then provide some output formats like HTML. Rolling our own Markdown printer is just as fraught (and some of the fixed issues reflect that).

I also think there are some situations that might not be handled well by all Markdown tools – e.g., handling of complicated fence nesting, due to our frequent use of them in `Doc`.

That said, swapping out one implementation for another should be pretty simple. I actually did MMark first, realized I couldn’t print with it, then switched to `cmark`, and that was easy.

pchiusano:
Cool!

unison-cli/src/Unison/Codebase/TranscriptParser.hs
sellout: So, what’s happening here is that we have a code block with Unison code. We do nothing in particular here, just grab the text and hold onto it to parse later.

But for some reason, there is an extra `\n` in the text. So I call `Text.init` to drop the last character. I just don’t know _why_, but it works consistently. I mean, it ”works” without the `init`, but every `unison` block in the output ends up with an extra newline at the end.

aryairani: I see, so it's just a weird thing to drop a character we don't know where it's coming from; but that all the transcripts look okay as long as we keep it. (?)

aryairani (approved):

PR #5210, opened 2024-07-11
Use pretty text in MsgTrace case of debug toText

dolio:
This changes the behavior of the `DBTX` instruction to use the provided 'pretty' text in cases where decompilation isn't complete. I think when this was first written it wouldn't provide much output, but since we have made it so that decompilation always produces something relatively useful, even if there are parts of the value that cannot be decompiled into any term.

This should improve e.g. cases that @ceedubs was seeing where looking at debug text for a value with a promise in it would produce massively more verbose `show` output rather than somewhat more comprehensible 'pretty' output.

### DIFF
--- a/parser-typechecker/src/Unison/Runtime/Machine.hs
+++ b/parser-typechecker/src/Unison/Runtime/Machine.hs
@@ -384,7 +384,7 @@ exec !env !denv !_activeThreads !ustk !bstk !k _ (BPrim1 DBTX i)
       ustk <- bump ustk
       bstk <- case tracer env False clo of
         NoTrace -> bstk <$ poke ustk 0
-        MsgTrace _ tx _ -> do
+        MsgTrace _ _ tx -> do
           poke ustk 1
           bstk <- bump bstk
           bstk <$ pokeBi bstk (Util.Text.pack tx)


aryairani (approved):

dolio:
Example output difference:

```
examples/main> run testo
DataB2 (ReferenceDerived (Id "2lg4ah6ir6t129m33d7gssnigacral39qdamo20mn6r2vefl
iubpeqnjhejai9ekjckv0qnu9mlu3k9nbpfhl2schec4dohn7rjhjt8" 0)) 3735552 (DataU1 (
ReferenceBuiltin "Nat") 1441792 5) (DataB2 (ReferenceDerived (Id "2lg4ah6ir6t1
29m33d7gssnigacral39qdamo20mn6r2vefliubpeqnjhejai9ekjckv0qnu9mlu3k9nbpfhl2sche
c4dohn7rjhjt8" 0)) 3735552 (DataU1 (ReferenceBuiltin "Nat") 1441792 5) (DataB2
 (ReferenceDerived (Id "2lg4ah6ir6t129m33d7gssnigacral39qdamo20mn6r2vefliubpeq
njhejai9ekjckv0qnu9mlu3k9nbpfhl2schec4dohn7rjhjt8" 0)) 3735552 (Foreign (Wrap 
ReferenceBuiltin "Promise" _)) (Enum (ReferenceDerived (Id "00nv2kob8fp11qdkr7
50rlppf81cda95m3q0niohj1pvljnjl4r3hqrhvp1un2p40ptgkhhsne7hocod90r3qdlus9guivh7
j3qcq0g" 0)) 3801088)))
```

```
examples/main> run testo
(5, 5, bug "<Foreign>")
```

aryairani:
@dolio What are we even seeing in that example?

dolio:
The exact same value rendered two different ways. :)

The first is `show` output of the value in the second thing. The last value of the triple is an empty promise.

aryairani:
So they both represent a triple? But there was a bug encountered in the last element that was causing the original message to print? But now the error that gets printed is better?

dolio:
There's no way to decompile a promise to a unison term, so the tracer gives back both the output of Haskell's `show` for the value, and a pretty printed value containing dummy `bug` calls for the things it couldn't decompile. The two snippets are the corresponding text you see for each of those. I changed the `DBTX` case to use the second text instead of the first.

If you're actually debugging the runtime you might want to see the first thing, but most people don't.

ceedubs (approved):
Oh wow that's a small diff :)

Like @aryairani the `bug` confused me. It might be nice to make the output a little clearer. This doesn't seem like a slam dunk in its current form. The old output was super noisy, but at least I could tell that there was a `Promise` in there. Is `Foreign` necessarily this opaque?

Anyway this shouldn't affect most people, so I think that we can give it a try.

PR #5216, opened 2024-07-12
Use builtin reference names in foreign decompiling

dolio:
This change makes e.g. a printed promise look like `bug "<Promise>"` instead of `bug "<Foreign>"`. I don't think there are any foreign values that have hash based references, but if there are, they'll fall back to just printing `<Foreign>`.


aryairani:
Nice!

aryairani:
But would you want to show more, if you have more, in the non-`Builtin` case, too?

ceedubs (approved):
Nice, thanks!

PR #5227, opened 2024-07-15
Avoid duplicating YAML info in Markdown

sellout:
## Overview

Reviewing the last merge into #5142, I noticed some duplicated and out-of-date information. This brings things up-to-date and slightly reduces the duplication.

## Implementation notes

- bumped Ormolu used by ci.yaml from 0.5.2.0 to 0.7.2.0, to match flake.nix
- removed Markdown that claimed ci.yaml was using Ormolu 0.5.0.1
- moved description from `base-codebase` in Markdown to comment on `runtime_tests_codebase` in ci.yaml (and updated it to refer to builtin-tests/interpreter-tests.md instead of builtin-tests/base.md)
- removed `unison_src_test_results` as it’s no longer managed as a single variable
- moved other comments from Markdown to ci.yaml
- added Markdown recommending to look in ci.yaml for specifics
- rearranges the order of vars in ci.yaml to match the order they were presented in Markdown

## Test coverage

This bumps the Ormolu version in ci.yaml, but otherwise just rearranges some items in that file. Running CI should catch if any of that is incorrect.

## Loose ends

One thing I wasn’t sure how to map over: Markdown claims Racket 8.7 is used in CI, but ci.yaml doesn’t mention any Racket version.


.github/workflows/ci.yaml
```diff
@@ -12,19 +12,26 @@ on:
   workflow_dispatch:
 
 env:
-  ormolu_version: 0.5.2.0
-  ucm_local_bin: ucm-local-bin
+  ## Some version numbers that are used during CI
```
aryairani, line 15: I feel like some editor told me having comments in here was illegal, but I guess it's not true

aryairani (approved):

PR #5237, opened 2024-07-17
Support native compilation on non-GNU Unix systems

neduard:
## Overview

Currently only tested on OpenBSD, but the hope is to more easily support non-GNU systems.  Two main changes:

1. Use `libb2` for all blake2 related functions.  This is because blake2 is not supported [at the moment](https://github.com/libressl/portable/issues/199) by LibreSSL
2. Use the more portable `sh` in `jit-tests.sh`.  This ensures it can run on systems where bash isn't installed by default.

(cc @dolio as I think you've been quite involved in the native compilation?  Curious to hear your thoughts on this!)

One could argue that the vast majority of users will be on Windows/Linux/Mac.  However, I think it's important in general to strive for portable software.  The more systems software can run on, the less chances of system/implementation-specific issues cropping up.

## Test coverage

Relying mostly on CI .  Manually ran `jit-tests.sh` locally on Mac.

## Loose ends

I've also updated the README for the scheme-libs here #5238

Tests are currently failing because of outdated scripts.  That is fixed in #5236

### DIFF
--- a/scheme-libs/racket/unison/crypto.rkt
+++ b/scheme-libs/racket/unison/crypto.rkt
@@ -96,6 +96,7 @@
                     (error 'blake2 "~a failed with return value ~a" fn r))))))
 
 (define blake2b-raw (libb2-raw "blake2b"))
+(define blake2s-raw (libb2-raw "blake2s"))
 
 (define HashAlgorithm.Md5 (lc-algo "EVP_md5" 128))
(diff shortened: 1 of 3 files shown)


sellout:
I haven’t tried this out, but the changes look good to me in principle, and I agree with the motivations.

dolio (approved):
Cool.

Seems like simple enough changes to make things more portable. Those two hashes appear to be in the tests, so it seems like it should still be working on the usual platforms.

PR #5242, opened 2024-07-22
Don't limit max line length on .u scratch files

ceedubs:
The recently added `.editorconfig` sets a max line length of 120 characters for all file types. I found this to be
pretty frustrating when writing Unison scratch files. While writing Unison code my editor would seemingly randomly
create a new line, generating invalid Unison code. This came up more often when writing `Doc` values, since I tend to
write longer lines in prose than in code.

The formatting in scratch files is ephemeral, so I don't think that there is any good reason to force a max line length
on it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,6 @@ indent_style = space
 insert_final_newline = true
 max_line_length = 120
 trim_trailing_whitespace = true
+
+[*.u]
+max_line_length = off


sellout (approved):
This looks good.

The original settings were created primarily for Haskell, but I figured it’s good for many programming languages, so I left it with `*`. But we could also constrain `max_line_length` (and possibly other settings) to `*.hs` and some other explicit extensions (e.g., I like to write Markdown without line breaks … but I don’t know if that’s the style at Unison or not).

In any case, I think this change on its own is an improvement.

PR #5251, opened 2024-07-24
Ensure transcript parser consumes entire stanzas

sellout:
## Overview

With the switch to `cmark`, the “second phase” parsing of individual stanzas omitted an EOF check to ensure that the entire stanza had been parsed. This resulted in parses where we end up with truncated sets of UCM commands or API requests, which could either result in premature success or failures occurring later in the transcript, where they’d complain about the wrong thing.

## Implementation notes

This additionally simplifies the `ucmLine` parser (using an existing project/branch parser directly, rather than indirecting via `tryFrom @Text`)

## Test coverage

There are two new transcripts in transcripts/errors to test the two changed cases.

### DIFF
--- a/unison-src/transcripts/errors/no-abspath-in-ucm.md
+++ b/unison-src/transcripts/errors/no-abspath-in-ucm.md
@@ -1,0 +1,5 @@
+``` ucm:error
+scratch/main> builtins.merge
+-- As of 0.5.25, we no longer allow loose code paths for UCM commands.
+.> ls
+```
--- a/unison-cli/src/Unison/Codebase/Transcript/Parser.hs
+++ b/unison-cli/src/Unison/Codebase/Transcript/Parser.hs
@@ -24,11 +24,10 @@ where
 import CMark qualified
 import Data.Char qualified as Char
 import Data.Text qualified as Text
-import Data.These (These (..))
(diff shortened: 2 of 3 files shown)


unison-src/transcripts/errors/no-abspath-in-ucm.md
aryairani, line 1: I forget, do these "errors/" transcripts get checked in CI?

aryairani (approved):

unison-src/transcripts/errors/no-abspath-in-ucm.md
sellout: Yeah, they’re run as part of the `transcripts` command.

https://github.com/unisonweb/unison/actions/runs/10082431729/job/27877734245?pr=5251#step:8:299

aryairani: I guess the "errors/" ones wouldn't have output; they just get the error code confirmed?

ceedubs (approved):
Thank you!

unison-src/transcripts/errors/no-abspath-in-ucm.md
sellout: Yeah, that’s right. It would be nice to be able to compare against stderr, though.

PR #5056, opened 2024-06-05
Convert `ShallowListEntry` in `InputPattern` handlers

sellout:
## Overview

`ShallowListEntry` as returned by `ls` weren’t being handled by the new `InputPattern` handlers, and so the results of `ls` couldn’t be used as numbered args. This adds support for `ShallowListEntry` to the handlers.

Fixes #5055.

## Interesting/controversial decisions

This additionally makes `hqSplitFromName'` total, which triggers a few other minor adjustments.

Most uses of `ShallowListEntry` in the handler join the entry with the absolute prefix indicating where the `ls` was performed, however there is one handler that only allows relative paths (and is used by `alias.many`). That handler ignores the prefix and provides just the relative `ls` result.

## Test coverage

There is a new transcript that exercises this as exhibited in #5055.

Issue #3977, opened 2023-05-11
pretty-printer: roundtrip error with bang and long line


ceedubs:
I couldn't think of a concise way to describe this in the issue title...

This transcript shows some code that fails to parse after being pretty-printed:

# Raw transcript

````
```ucm:hide
.> builtins.merge
```

```unison:hide
failure msg context = Failure (typeLink Unit) msg (Any context) 

foo = Left (failure ("a loooooooooooooooooooooooooooooooooong" ++ "message with concatenation") ())
```

```ucm
.> add
.> edit foo
.> load scratch.u
```

````

# Output

```unison
failure msg context = Failure (typeLink Unit) msg (Any context) 

foo = Left (failure ("a loooooooooooooooooooooooooooooooooong" ++ "message with concatenation") ())
```

```ucm
.> add

  ⍟ I've added these definitions:
  
    failure : Text -> context -> Failure
    foo     : Either Failure b

.> edit foo

  ☝️
  
  I added these definitions to the top of
  /home/cody/code/unison/scratch.u
  
    foo : Either Failure b
    foo =
      use Text ++
      Left
        !(failure
          ("a loooooooooooooooooooooooooooooooooong"
            ++ "message with concatenation"))
  
  You can edit them there, then do `update` to replace the
  definitions currently in this namespace.

.> load scratch.u

  offset=18:
  unexpected (
  expecting , or newline or semicolon
      6 |       ("a loooooooooooooooooooooooooooooooooong"
  

```

🛑

The transcript failed due to an error in the stanza above. The error is:

  offset=18:
  unexpected (
  expecting , or newline or semicolon
      6 |       ("a loooooooooooooooooooooooooooooooooong"

sellout:
This round-trips now.
````
```unison
failure msg context = Failure (typeLink Unit) msg (Any context)

foo = Left (failure ("a loooooooooooooooooooooooooooooooooong" ++ "message with concatenation") ())
```

```ucm
.> add

  ⍟ I've added these definitions:
  
    failure : Text -> context -> Failure
    foo     : Either Failure b

.> edit foo

  ☝️
  
  I added 1 definitions to the top of scratch.u
  
  You can edit them there, then run `update` to replace the
  definitions currently in this namespace.

.> load scratch.u

  Loading changes detected in scratch.u.

  I found and typechecked the definitions in scratch.u. This
  file has been previously added to the codebase.

```
```unison:added-by-ucm scratch.u
foo : Either Failure b
foo =
[...]

PR #5252, opened 2024-07-24
Add transcripts for some already-fixed issues

### LINKED ISSUE #1327: `alias.many` suffers from lingering parsing issues, and also incompatible with numberedArgs of some commands
I think this should work, but doesn't:

```
.ns1> find
.ns1> alias.many 1-28 .ns1_nohistory
```

Reason this doesn't work - `find` produces hash qualified numbered args and `alias.many` doesn't accept hash qualified arguments.

@aryairani I had thought this was a major use case for the command, copying a namespace without its history. I remember we'd considered a `merge.nohistory` command and went with this instead after thinking they had equivalent power.

Seems like `alias.many` could just be made to work with hash qualified names and that would address the problem?

sellout:
This closes #1327 (which was actually fixed by #5056 or #5061).

 #3977 was already closed, but without the transcript to avoid regression.

### DIFF
--- a/unison-src/transcripts/fix1327.md
+++ b/unison-src/transcripts/fix1327.md
@@ -1,0 +1,11 @@
+```unison
+foo = 4
+
+bar = 5
+```
+
+```ucm
+scratch/main> add
+scratch/main> ls
+scratch/main> alias.many 1-2 .ns1_nohistory
+```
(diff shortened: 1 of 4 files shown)


aryairani:
I know the existing transcripts are not great in this regard, but would you add just some minimal text to the transcripts as well, to indicate what's being demonstrated (or "why", if "what" is obvious)?

aryairani (approved):

aryairani (approved):

PR #5253, opened 2024-07-24
Convert Doc v1 transcripts to v2

sellout:
This does a very superficial conversion – it doesn’t correct all of the verbiage that describes how to do things in v1, just changes the actual syntax.

It also merges in trunk, for the `ProjectPath` changes.


unison-src/transcripts/doc-formatting.md
```diff
@@ -20,10 +20,10 @@ scratch/main> add
 scratch/main> view foo
 ```
 
-Note that `@` and `:]` must be escaped within docs.
+Note that `{{`, `@`, and `}}` must be escaped within docs.
 
 ```unison
-escaping = [: Docs look [: like \@this \:] :]
+escaping = {{ Docs look --doc open-- like --amphora--this --doc close-- }}
```
sellout: I wasn’t sure how to convert this bit – AFAICT, there’s no support for escaping characters in Doc v2?

sellout: I just changed this to use verbatim (``{{ Docs look `{{ like @this }}` }}``). It’s still not exactly an escape, but much closer than what I had in there before.

sellout:
Everything other than the last two commits is from merging in (a now very old version of) trunk.

unison-src/transcripts/doc-formatting.md
aryairani: Yeah, ok thanks.

sellout:
Replaced by #5794.

.gitignore
```diff
@@ -1,9 +1,14 @@
...
 .unison*
 test-output
-transcript-*
-scratch.u
 unisonLocal.zip
+*.uc
+# Ignore all scratch files...
+*.u
+# Except those in unison-src
+!unison-src/**/*.u
```
aryairani, line 9: good idea

sellout: Heh, it’s not even part of this PR … it’s just a consequence of merging trunk into this to get CI working again (since these branches are too old to function any more). The mess of that is why I ended up closing this and re-applying these changes on top of trunk in #5794.

PR #5257, opened 2024-07-30
feat/refactor: 3-way diff API

mitchellwrosen:
## Overview

- [x] Depends on #5250 

This PR refactors the 3-way diff API a bit, making it suitable to call from Unison Share.

Previously, it was hard-coded to take a record of callbacks that hit SQLite. Now, the database access is removed from the API; it's just pure functions.

## Test coverage

No new tests, but everything should work as before.

## Loose ends

TODO soon: rework local merge to use the new API. I left around the old way for now.


aryairani (approved):

aryairani:
cc @ChrisPenner

PR #5259, opened 2024-08-01
feat/refactor: add merge api

mitchellwrosen:
## Overview

This PR refactors the merge API to make it more  pure and suitable to call from Share.


unison-src/transcripts/merge.output.md
```diff
@@ -818,7 +820,7 @@ project/alice> merge /bob
 type Foo = Baz Nat Nat | Qux Text
 
 -- project/bob
-type Foo = Baz Nat | BobQux Text
+type Foo = BobQux Text | Baz Nat
```
aryairani, line 823: Any idea what caused these to flip?

mitchellwrosen: Not exactly, but I did investigate and they seem to have been wrong before. I manually walked through the transcript and observed that on `project/bob` (in this example) the constructors do render in the new order. 🤷

aryairani (approved):

unison-src/transcripts/merge.output.md
mitchellwrosen: Updated explanation: in this PR we no longer reuse unique type UUID in merge, so any merge performed in a transcript can affect subsequent unique type constructor orders. We were going to circle back on that in a follow-up PR.

PR #5260, opened 2024-08-01
refactor: clean up merge

mitchellwrosen:
## Overview

- [ ] Depends on #5257 

This PR moves the merge implementation over to the pure API, and deletes the old API that uses a record of database-accessing functions


mitchellwrosen:
working this into another pr

PR #5264, opened 2024-08-02
move /nix to /mnt/nix on Linux workflow

### LINKED ISSUE #5266: nix-dev-cache keeps failing
see https://github.com/unisonweb/unison/actions/runs/10219694266/job/28278461583?pr=5250

["No space left on device"](https://github.com/user-attachments/assets/28b1c398-7aea-4de5-8b49-c22b0a8bc13b)

`ubuntu-20.04` runner:
`/` has 18GB available, this is where `/nix` is installed by default
`/mnt` has 66GB available

`macos-12` runner:
`/` has 63GB available

`macos-14` runner:
`/` has 55GB available

So can we put the nix store on `/mnt` in `ubuntu-20.04`?

aryairani:
## Overview

nix is installed in /nix, which on the ubuntu-20.04 runner doesn't have enough space to build everything from scratch, resulting in CI failures on trunk and some PRs, eg #5265 

This PR causes the linux runner install nix onto a disk that's bigger than `/`

fixes https://github.com/unisonweb/unison/issues/5266

## Implementation notes

adds a step to `mount -o bind /mnt/nix /nix` before installing nix, and this seems to work.

## Test coverage

the nix steps still pass

## Loose ends

### DIFF
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -29,7 +29,7 @@ jobs:
       if: runner.os == 'Linux'
       run: |
         sudo mkdir /nix /mnt/nix
-        sudo mount -B /mnt/nix /nix
+        sudo mount --bind /mnt/nix /nix
     - uses: cachix/install-nix-action@v27
       if: runner.os == 'Linux'
       with:
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -27,8 +27,8 @@ jobs:
     - name: set up mount points on Linux
       if: runner.os == 'Linux'
       run: |
-        mkdir /nix /mnt/nix
-        mount -B /mnt/nix /nix
+        sudo mkdir /nix /mnt/nix
+        sudo mount -B /mnt/nix /nix
     - uses: cachix/install-nix-action@v27
       if: runner.os == 'Linux'
       with:
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -25,6 +25,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: mount Nix store on larger partition
+      # on the Linux runner `/` doesn't have enough space, but there's a `/mnt` which does.
       if: runner.os == 'Linux'
       run: |
         sudo mkdir /nix /mnt/nix
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -24,7 +24,7 @@ jobs:
           - macOS-14
     steps:
     - uses: actions/checkout@v4
-    - name: set up nix mount points on Linux
+    - name: mount Nix store on larger partition
       if: runner.os == 'Linux'
       run: |
         sudo mkdir /nix /mnt/nix
--- a/.github/workflows/nix-dev-cache.yaml
+++ b/.github/workflows/nix-dev-cache.yaml
@@ -24,6 +24,11 @@ jobs:
           - macOS-14
     steps:
     - uses: actions/checkout@v4
+    - name: set up mount points on Linux
+      if: runner.os == 'Linux'
+      run: |
+        mkdir /nix /mnt/nix
+        mount -B /mnt/nix /nix
     - uses: cachix/install-nix-action@v27
       if: runner.os == 'Linux'
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/nix-dev-cache.yaml
```diff
@@ -24,6 +24,11 @@ jobs:
           - macOS-14
     steps:
     - uses: actions/checkout@v4
+    - name: set up nix mount points on Linux
+      if: runner.os == 'Linux'
```
sellout, line 29: Why limit this to Linux?
- Because macOS `mount` doesn’t support `--bind`?
- Because we don’t need it on macOS? (I try to do things as similarly across platforms as possible, though)
- Because there’s no `/mnt` partition on macOS runners?

Could there be a comment about whatever it is?

.github/workflows/nix-dev-cache.yaml
```diff
@@ -24,6 +24,11 @@ jobs:
...
+      run: |
+        sudo mkdir /nix /mnt/nix
+        sudo mount -B /mnt/nix /nix
```
sellout:
```suggestion
        sudo mount --bind /mnt/nix /nix
```
I find it helpful to use expanded flags in scripts, more obvious what they’re doing.

.github/workflows/nix-dev-cache.yaml
```diff
@@ -38,3 +43,5 @@ jobs:
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
     - name: build all packages and development shells
       run: nix -L build --accept-flake-config --no-link --keep-going '.#all'
+    - name: print disk free status
+      run: df -h
```
sellout, line 48: I’m guessing this is intentional and not just leftover debugging.

.github/workflows/nix-dev-cache.yaml
```diff
@@ -24,6 +24,11 @@ jobs:
           - macOS-14
     steps:
     - uses: actions/checkout@v4
+    - name: set up nix mount points on Linux
```
sellout:
```suggestion
    - name: mount Nix store on larger partition
```
To give more of a reason why.

sellout (approved):
This seems very helpful. I made a few non-blocking suggestions.

.github/workflows/nix-dev-cache.yaml
aryairani: Do you know if this definitely works on Github? I had tried `mount -o bind` initially and either it didn't work or I messed something else up. `mount -B` is new to me.

.github/workflows/nix-dev-cache.yaml
aryairani: Yeah I added this intentionally for future debugging

.github/workflows/nix-dev-cache.yaml
aryairani: Yeah it's both 2) and 3); not sure about 1) but skimming `man mount` on mac, it also seems to be true.

.github/workflows/nix-dev-cache.yaml
aryairani:
```suggestion
    - name: switch nix mount point on Linux
      # On the Linux runner `/` doesn't have enough space, but `/mnt` does.
```

.github/workflows/nix-dev-cache.yaml
```diff
@@ -24,6 +24,11 @@ jobs:
           - macOS-14
     steps:
     - uses: actions/checkout@v4
+    - name: mount Nix store on larger partition
```
aryairani, line 27:
```suggestion
    - name: mount Nix store on larger partition
      # on the Linux runner `/` doesn't have enough space, but there's a `/mnt` which does.
```

.github/workflows/nix-dev-cache.yaml
aryairani: I realized I can re-run an old commit to check.

ChrisPenner (approved):
Thanks for handling the annoying CI stuff!

.github/workflows/nix-dev-cache.yaml
stew: This seems super weird to me. /mnt is only supposed to be used transiently by a system administrator, not something that should be scripted against. There is no reason (if you care about the FHS) that there should be anything mounted to /mnt to make it a separate filesystem than /. Is /mnt just some other big disk on GHA?

If this works, that's great, I'm just curious, as this seems kinda wrong to me...

PR #5277, opened 2024-08-13
merge API tweaks

mitchellwrosen:
## Overview

This PR contains a small number of tweaks to the merge API from the last round of review:

1. Reuse unique type GUIDs rather than make up new ones.
2. Remove requirement to provide a `UniqueName` to the parser.
3. Distinguish between term and type dependencies (for the step of merge that needs to make a `TypeLookup`.

Other misc cleanup: deleted a couple unused record fields.

### DIFF
--- a/unison-src/transcripts/merge.output.md
+++ b/unison-src/transcripts/merge.output.md
@@ -1024,7 +1024,7 @@ Bob, meanwhile, first deletes the term, then sort of deletes the type and re-add
 ``` ucm
 project/bob> view Foo.Bar
 
-  type Foo.Bar = Hello Nat Nat | Baz Nat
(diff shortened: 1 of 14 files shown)


unison-src/transcripts/merge.output.md
```diff
@@ -1024,7 +1024,7 @@ Bob, meanwhile, first deletes the term, then sort of deletes the type and re-add
 ``` ucm
 project/bob> view Foo.Bar
 
-  type Foo.Bar = Hello Nat Nat | Baz Nat
```
mitchellwrosen, line 1027: These are expected because we're reusing an existing unique type GUID, which affects constructor ordering.

aryairani (approved):

PR #5285, opened 2024-08-16
feat: namespace directive

mitchellwrosen:
## Overview

This PR implements the "namespace directive" feature: a Unison file can contain an optional `namespace foo` line that affects parsing as follows:

1. All bound names (type names, constructor names, generated accessor names, term names, and watch expression names) are prefixed with the namespace. 
2. Variables that refer to locally-bound names exactly are prefixed with the namespace.

That is, 

```
namespace a.b.c

type Foo = Bar
type Qux = { quaf : Foo }

baz = ...

honk = ... baz ...
```

is equivalent to

```
type a.b.c.Foo = Bar
type a.b.c.Qux = { quaf : a.b.c.Foo }

a.b.c.baz = ...

a.b.c.honk = ... a.b.c.baz ...
```

Note that the existence of a namespace directive therefore _prevents_ one from referring to a name outside of the file whose suffix matches something in the file.

For example, without a namespace directive, I could write

```
foo.factorial = ... foo.factorial ... factorial ...
```

and refer to both my locally-bound `foo.factorial` as well as a term literally named `factorial` (no other prefix) in the namespace. (N.B that's not true in trunk today, but it will be fixed soon: https://github.com/unisonweb/unison/issues/5268).

However, _with_ a namespace directive...

```
namespace foo

factorial = ???
```

there's no way to write the same function (because `factorial` resolves to the locally-bound `factorial`, which will be expanded to `foo.factorial` during parsing).

For this reason, it would be unsafe to implement `edit.namespace foo` naively by putting a `namespace foo` directive at the top of the block.

## Test coverage

I've added a transcript to demonstrate the feature.

### DIFF
--- a/unison-src/transcripts/generic-parse-errors.output.md
+++ b/unison-src/transcripts/generic-parse-errors.output.md
@@ -30,12 +30,32 @@ namespace.blah = 1
 
   Loading changes detected in scratch.u.
 
-  The identifier `namespace` used here is a reserved keyword: 
+  I got confused here:
   
       1 | namespace.blah = 1
   
-  You can avoid this problem either by renaming the identifier
-  or wrapping it in backticks (like `namespace` ).
+  
+  I was surprised to find a = here.
+  I was expecting one of these instead:
+  
+  * ability
+  * bang
+  * binding
+  * do
+  * false
+  * force
+  * handle
+  * if
+  * lambda
+  * let
+  * newline or semicolon
+  * quote
+  * termLink
+  * true
+  * tuple
+  * type
+  * typeLink
+  * use
 
 ```
 ``` unison
--- a/parser-typechecker/src/Unison/Syntax/FileParser.hs
+++ b/parser-typechecker/src/Unison/Syntax/FileParser.hs
@@ -62,10 +62,6 @@ file = do
   (namesStart, imports) <- TermParser.imports <* optional semi
   (dataDecls, effectDecls, parsedAccessors) <- declarations
 
-  let unNamespacedTypeNames :: Set v
-      unNamespacedTypeNames =
-        Set.union (Map.keysSet dataDecls) (Map.keysSet effectDecls)
-
   env <-
     let applyNamespaceToDecls :: forall decl. Iso' decl (DataDeclaration v Ann) -> Map v decl -> Map v decl
         applyNamespaceToDecls dataDeclL =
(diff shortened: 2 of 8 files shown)


unison-src/transcripts/generic-parse-errors.output.md
```diff
@@ -30,12 +30,32 @@ namespace.blah = 1
 
   Loading changes detected in scratch.u.
 
-  The identifier `namespace` used here is a reserved keyword: 
+  I got confused here:
```
mitchellwrosen, line 33: This regression is weird... I don't know why the message is so much worse now. The only change I made to the lexer was adding "namespace" to the list of keywordy things here:

https://github.com/unisonweb/unison/pull/5285/files#diff-06c08c644fa09b2d2218485a64c9965261fb6bf70d88cf0faf62c1cbe16443f2R576

"namespace" was already partially considered a keyword because it's in this set: https://github.com/unisonweb/unison/blob/f2adc77e39d72d5125f7b2f0a43279aa26119588/unison-syntax/src/Unison/Syntax/ReservedWords.hs#L27

Have I done the wrong thing here?

pchiusano (approved):
Awesome!! That was fast.

I'm not sure why the transcript error message regression is happening. Though it might not matter too much if the only behavior is a worse error message in the event you call a definition `namespace`. Still, would be nice to keep the good message.

hojberg:
Very cool! Can you have multiple namespace directives in a file?

mitchellwrosen:
@hojberg You cannot – zero or one

mitchellwrosen:
@pchiusano I looked into this briefly, it really isn't more complicated than we're now parsing `namespace` as a `Reserved` lexeme rather than an `Identifier` lexeme, and go down different error paths from there. This could/should be cleaned up for sure, but I'm inclined to leave it alone for now as it also affects other reserved keywords in the same way.

pchiusano:
@mitchellwrosen cool. Yeah I wouldn’t sweat it for this PR then.

unison-src/transcripts/generic-parse-errors.output.md
aryairani: What happens when you use a different reserved word as a name segment?

aryairani: Maybe @sellout would have an idea too, since they've been looking at these?

aryairani: Disregard, I guess this is superseded by https://github.com/unisonweb/unison/pull/5285#issuecomment-2294455794.

PR #5290, opened 2024-08-20
Improve transcript runner output

sellout:
## Overview

Previously, it would output messages like

    ⚙️   Processing stanza Just 6 of 7.

and

    ⚙️   Processing stanza Nothing of 7.

which was especially confusing when there was text or some other non-Unison block at the end of the transcript.

Now the messages look like

    ⏩   Skipping non-executable Markdown block.
    ⚙️   Processing stanza 6 of 7.
    ✔️   Completed transcript.

The one shortcoming is that I don’t know how to clear the line after the carriage return, so I added whitespace padding to make sure the previous messages get overwritten.

## Test coverage

I’ve only tested this by hand.

### DIFF
--- a/unison-cli/src/Unison/Codebase/Transcript/Runner.hs
+++ b/unison-cli/src/Unison/Codebase/Transcript/Runner.hs
@@ -300,16 +300,21 @@ run isTest verbosity dir stanzas codebase runtime sbRuntime nRuntime config ucmV
             _ <- liftIO (writeIORef mStanza maybeStanza)
             case maybeStanza of
               Nothing -> do
-                liftIO (putStrLn "")
+                liftIO (putStrLn "\r✔️   Completed transcript.                   ")
                 pure $ Right QuitI
-              Just (s, idx) -> do
+              Just (s, midx) -> do
                 unless (Verbosity.isSilent verbosity) . liftIO $ do
                   putStr $
-                    "\r⚙️   Processing stanza "
-                      ++ show idx
-                      ++ " of "
-                      ++ show (length stanzas)
-                      ++ "."
+                    maybe
+                      "\r⏩   Skipping non-executable Markdown block."
+                      ( \idx ->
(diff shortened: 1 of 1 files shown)


unison-cli/src/Unison/Codebase/Transcript/Runner.hs
```diff
@@ -300,16 +300,21 @@ run isTest verbosity dir stanzas codebase runtime sbRuntime nRuntime config ucmV
...
-                      ++ show (length stanzas)
-                      ++ "."
+                    maybe
+                      "\r⏩   Skipping non-executable Markdown block."
+                      ( \idx ->
+                          "\r⚙️   Processing stanza "
+                            ++ show idx
+                            ++ " of "
+                            ++ show (length stanzas)
+                            ++ ".          "
```
aryairani: 😑
Could we pull out the text constants and use `strlen` and `max` and arithmetic instead of hard-coding the spaces? 😄
Otherwise looks good.

sellout: Ah, it looks like we have [`Unison.PrettyTerminal.clearCurrentLine`](https://github.com/unisonweb/unison/blob/trunk/lib/unison-pretty-printer/src/Unison/PrettyTerminal.hs#L34), which should do it.

There’s also `^[[2K`, but I don’t know how we feel about VT100 escapes.

aryairani: [`clearCurrentLine`](https://github.com/unisonweb/unison/blob/trunk/lib/unison-pretty-printer/src/Unison/PrettyTerminal.hs#L34) sounds good for now; I wouldn't wanna print escapes directly without conditioning on some sort of terminal detection. (Though I wouldn't be shocked to hear that we currently do this with ANSI codes).

aryairani (approved):

PR #5294, opened 2024-08-22
Numbers binary notation

SimaDovakin:
## Overview

This PR adds support for the binary notation for Nat and Int types (e.g., `0b101`, `-0b11`). At first, I added a binary literal in the transcript that had invalid binary characters in the middle of the literal (e.g., `0b1210`), and I got this error:
```
This looks like a function call, but with a Nat where the function should be. Are you missing an operator?
```
But the same error is thrown with hexadecimal and octal notations. So I assumed that this is an expected behavior for now. Also, I added syntax highlighting for Vim. I found that Atom's config doesn't distinguish number notations. Did I miss something?

In addition, I have to add binary notation in documentation for Nat and Int in Unison Share, right?

### DIFF
--- a/CONTRIBUTORS.markdown
+++ b/CONTRIBUTORS.markdown
@@ -87,3 +87,4 @@ The format for this list: name, GitHub handle
 * Dan Doel (@dolio)
 * Eric Torreborre (@etorreborre)
 * Eduard Nicodei (@neduard)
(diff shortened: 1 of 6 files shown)


aryairani (approved):
Looks great, thanks!

Issue #5179, opened 2024-07-04
unexpected "impossible" exception in lexer


ceedubs:
**Describe and demonstrate the bug**

I see this error in my terminal, though seems seem to continue to run fine. I think that it may be coming from an LSP thread similar to #4737.

```
 Uh oh, an unexpected exception brought the process down! That should never happen. Please file a bug report.

Here's a stringy rendering of the exception:

  impossible
  CallStack (from HasCallStack):
    error, called at src/Unison/Syntax/Lexer.hs:241:22 in unison-syntax-0.0.0-3wBkjZ3Vjv91L4a1yNKHwZ:Unison.Syntax.Lexer
```

I'm not _exactly_ sure what caused this. But I think that it was something like this:

- `add` a type to the project and a function that matches on one of its constructors
- Bring the type and the function into your scratch file
- Add an argument to the constructor, and modify the function to use it. Save the file but don't `update`.
- Delete or comment out the updated type, but leave the function depending on the modified constructor in the scratch file.

I'm using LSP; not sure how relevant that is

**Environment (please complete the following information):**
 - `ucm --version` a74d4e8cb
 - OS/Architecture: x86 NixOS
 - Editor: Neovim with lspconfig

sellout:
Here’s the code at that commit: https://github.com/unisonweb/unison/blob/a74d4e8cb9cd860ebd79e348f8012f57625e2239/unison-syntax/src/Unison/Syntax/Lexer.hs#L229-L241

A first step might just be returning a parser failure rather than `error` (and maybe changing the description).

aryairani:
Is this closed as of #5297?

sellout:
I don’t think so – this has just hopefully made the failure more informative when it occurs again (since it‘s clearly not impossible).

Since this is an failure in the lexer, I don’t think Cody’s sequence of steps is helpful to replicate it, especially since it seems to be a result of the background LSP thread.

PR #5297, opened 2024-08-22
Turn a possible “impossible” into a parse failure

sellout:
## Overview

From #5179, there’s a case where we hit an `error "impossible"`, which doesn’t provide much context. This turns it into a parse failure, so we have #thte state of the lexer when this happens again. It also adds a comment that describes when this “impossible” case gets hit.

## Test coverage

No coverage – this is to improve the diagnostics when this case gets hit again.

## Loose ends

Ormolu is bad at formatting comments on conditional branches.


ceedubs (approved):

pchiusano:
Thanks @sellout !

PR #5298, opened 2024-08-23
Support pattern matching on empty types

### LINKED ISSUE #4731: Match with no cases in `Void.absurdly`
This used to work:

```
Void.absurdly : '{e} Void ->{e} a
Void.absurdly v = match !v with 
```

But no longer compiles:

```
I expected some patterns after a match / with or cases but I didn't find any.
```

There aren't any cases though, so this seems like an oversight.

The following used to also work, but no longer does:

```
  Void.absurd : Void -> a
  Void.absurd x =
```

sellout:
## Overview

Previously, `match` and `cases` expressions needed to have at least one pattern to match on. This allows them to work with zero patterns, which is useful for matching on empty types.

Since `EmptyMatch` is no longer a failure case, errors that previously said “I expected some patterns after a match / with or cases but I didn't find any,” now say “Pattern match doesn't cover all possible cases”.

Fixes #4731.

## Test coverage

There is a new transcript that covers the new cases. The existing error-messages transcript captures the change in error message with empty `cases` that are applied to non-empty types.

### DIFF
--- a/parser-typechecker/src/Unison/PatternMatchCoverage.hs
+++ b/parser-typechecker/src/Unison/PatternMatchCoverage.hs
@@ -35,7 +35,6 @@ module Unison.PatternMatchCoverage
   )
 where
 
-import Data.List.NonEmpty (nonEmpty)
 import Data.Set qualified as Set
 import Debug.Trace
 import Unison.Debug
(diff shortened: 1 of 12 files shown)


pchiusano (approved):
Looks great! I'll go ahead and merge while Arya's out.

PR #5303, opened 2024-08-28
bugfix: not-found or ambiguous constructor treated as var

### LINKED ISSUE #5301: out of scope data constructor in pattern is treated as a var
Input:

````
```ucm
scratch/main> builtins.merge
```

```unison
type Foo = Bar Nat

foo : Foo -> Nat
foo = cases
    Bar X -> 5
```
````

Output:
````
``` ucm
scratch/main> builtins.merge

  Done.

```
``` unison
type Foo = Bar Nat

foo : Foo -> Nat
foo = cases
    Bar X -> 5
```

``` ucm

  change:
  
      type Foo
      foo : Foo -> Nat

```
````

In the above example, the `cases Bar X -> 5` line should have failed with a "data constructor `X` not found" error, but instead it succeeded (parsing `X` as an unused variable).

I believe this isn't correct, even though we do allow variables to begin with a capital letter. The code appears to be _trying_ to commit to the current branch, causing a parse failure: https://github.com/unisonweb/unison/blob/6299fc3eb40ee2ad5c785060b76984f73955dedc/parser-typechecker/src/Unison/Syntax/TermParser.hs#L304

But our parser is incredibly fragile; megaparsec doesn't have a good "commit" or "cut" primitive. We could (and perhaps should) use something like [headed-megaparsec](https://hackage.haskell.org/package/headed-megaparsec) for that.

But in the meantime, some surrounding `try` (probably [this one](https://github.com/unisonweb/unison/blob/6299fc3eb40ee2ad5c785060b76984f73955dedc/parser-typechecker/src/Unison/Syntax/TermParser.hs#L332)) is causing our branch that called `failCommitted` to act as though it didn't consume any tokens. Then, we proceed to try other alternatives (ultimately succeeding to parse `X` as a variable).

### PR DESCRIPTION
**Pull request by mitchellwrosen**
## Overview

Fixes #5301 

This PR adjusts the parser to properly commit to a branch that fails to resolve a constructor pattern (either due to 0 or 2+ matches in the namespace).

Previously, code like

```
cases Foo X -> Y
```

would erroneously treat `X` as a variable binding if `X` didn't resolve to a single constructor in the namespace.

## Test coverage

There's a transcript that captures the fix

### CODE REVIEW

File: unison-src/transcripts/pattern-match-coverage.md [hunk 1]
```diff
@@ -465,7 +465,7 @@ result : '{e, Give T} r -> {e} r
 result f = handle !f with cases
        { x } -> x
        { give _ -> resume } -> result resume
-       { give A -> resume } -> result resume
+       { give T.A -> resume } -> result resume
```
mitchellwrosen, line 468: This test wasn't quite testing the right thing: since the `A` constructor is ambiguous in the namespace, it ended up as a variable, essentially testing:

```
{ give _ -> resume } -> result resume
{ give _ -> resume } -> result resume
```

rather than the intended

```
{ give _ -> resume } -> result resume
{ give T.A -> resume } -> result resume
```

but both produce the same output (complaining that the last pattern is redundant)

aryairani (approved):

PR #5324, opened 2024-09-03
bugfix: fix suffixification logic in upgrade

mitchellwrosen:
## Overview

Same as #5311, but for `upgrade`.

## Test coverage

Transcript included


aryairani (approved):
Nice, thanks.

Could you add a few words to the transcript about what is being demonstrated? Was it `b.y` being rendered wrong in `c`?

mitchellwrosen:
@aryairani Yep, see the first commit for the busted transcript.

aryairani:
> @aryairani Yep, see the first commit for the busted transcript.

I did see that—thanks for including it—but still generally want the final checked-in transcripts to be self-documenting. I know we have other tech debt in that area so I'm only singling this one out because we're in the process of adding it.

mitchellwrosen:
@aryairani Yep, I'm on it

PR #5333, opened 2024-09-06
Various fixes that enable running cloud on the JIT

dolio:
This PR includes significant changes that facilitate running cloud on the JIT. With them the cloud integration tests should pass. There's also some debugging infrastructure that I added while trying to figure out why the integration tests weren't passing.

- `define-unison` will accept a hint that turns on racket tracing for the specified definition. There's no mechanism for generating such definitions from ucm or the like at the moment, but it lets you easily trace if you're debugging via tweaking generated code.
- The compiled code generator has support for wrapping the main entry in racket profiling. It uses periodic printing so you can observe the behavior while the program is running. Again, there's no mechanism for enabling this except manual intervention, but it's useful for debugging the JIT right now.

- The server socket queue limit has been increased to 2048 to match the Haskell package we use. The racket default is 4, which causes performance problems in scenarios with a lot of connections.
- `define-unison` has been reworked to use a fixed set of 'curry' functions for enabling partial/overapplication of unison functions. This fixes an optimization problem that was introduced by generating closure structures and using the structure application for the slow path of application, but also avoids generating a lot of code for each function.
- Several issues with dynamic code loading and lookup have been fixed.  It had previously been neglected for standalone executables, but we didn't have anything that really tests that case. I've also added proper synchronization to the code loading to avoid race conditions when multiple threads load things at the same time.
- Abilities have been reworked to use symbols as the keys in the continuation. Symbols are 'interned' which makes them work properly with regard to equality, which is necessary for the correct behavior. The previous approach used termlink structures, and arranging for a single, unique termlink to be used so that they are all `eq?` instead of merely `equal?` is a bit much to ask.

Some additional minor fixes are included, but I believe the above are the major changes.

I'm anticipating that I'll have to merge with trunk and fix some stuff before this can be merged, since it's pretty far behind at the moment.


dolio:
Apparently I was wrong, and this just merges cleanly. Amazing.

aryairani:
Awesome. Is there any specific review you want before merging?

aryairani (approved):

ceedubs (approved):
I am not really qualified to review the code changes, but the fact that this makes the Unison Cloud integration tests go from all failing to all passing is a strong indicator that it is a strict improvement. Good work @dolio !

dolio:
I don't know exactly what code review is appropriate. I know most of you don't really do much scheme, and aren't experts on how the JIT works. Maybe just if someone would look over the code and ask if they see anything that seems strange to them.

pchiusano (approved):

PR #5338, opened 2024-09-10
bugfix: pretty printer should prefer direct dependencies

### LINKED ISSUE #5267: Pretty-printer should prefer direct dependencies
On Share, direct dependencies are preferred: 

CleanShot 2024-08-05 at 14 06 02@2x

https://share.unison-lang.org/@unison/routes/code/releases/6.2.2/latest/terms/Route/test/harness/doc

That is, even though there is a transitive dependency for a different hash called `parseOrBug`, that doesn't matter, because Share prefers the direct dependency.

However, if you `edit test.harness.doc`, you'll see:

CleanShot 2024-08-05 at 14 07 46@2x

Notice that `test.verify` and `parseOrBug` are fully qualified even though there is only one thing those could refer to:

```
@unison/routes/main> names test.verify

  Term
  Hash:   #99bqpp5tki
  Names:  lib.unison_base_3_13_0.test.verify
  
  Tip: Use `names.global` to see more results.

@unison/routes/main> names parseOrBug

  Term
  Hash:   #2u9hjme3hu
  Names:  lib.unison_base_3_13_0.IO.net.URI.parseOrBug
  
  Tip: Use `names.global` to see more results.
```

The fix is to just have the pretty-printer use the same logic that parser uses and only add qualifying name segments to distinguish among direct dependencies.

I'm pretty sure this used to be the case and this might be a regression, so a regression would be good to catch this in the future.

mitchellwrosen:
## Overview

Fixes #5267

This PR gets the pretty printer in sync with the parser with respect to the preference for terms and types that are not indirect dependencies.

Now, a suffix `foo` that matches one local definition (outside `lib`) or one direct dependency (inside `lib` but outside `lib.*.lib`) will resolve to that definition, regardless of how many indirect dependencies (inside `lib.*.lib.*`) it also matches.

## Test coverage

I've added a transcript.

### DIFF
--- a/unison-core/src/Unison/Name.hs
+++ b/unison-core/src/Unison/Name.hs
@@ -354,22 +354,36 @@ preferShallowLibDepth = \case
   [x] -> Set.singleton (snd x)
   rs ->
     let byPriority = List.multimap (map (first minLibs) rs)
-        minLibs [] = NamePriorityOne
+        minLibs [] = NamePriorityOne ()
         minLibs ns = minimum (map classifyNamePriority ns)
-     in case Map.lookup NamePriorityOne byPriority <|> Map.lookup NamePriorityTwo byPriority of
+     in case Map.lookup (NamePriorityOne ()) byPriority <|> Map.lookup (NamePriorityTwo ()) byPriority of
           Nothing -> Set.fromList (map snd rs)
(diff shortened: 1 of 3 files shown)


aryairani (approved):
Great

PR #5343, opened 2024-09-11
bugfix: fix name resolution logic bug related to indirect dependencies

### LINKED ISSUE #5340: Name binding logic doesn't prefer direct deps to indirect deps
If your namespace contains

```
my.foo = ...
lib.dep.lib.dep.foo = ...
```

then a reference to suffix `foo` should resolve to `my.foo` (since it's prioritized over indirect dependencies).

And it does:

```
something = foo + foo -- resolves to my.foo
```

Unless `my.foo` is also bound in the file:

```
my.foo = ...
something = foo + foo -- ambiguous
```

This is because the suffix `foo` matches one thing in the file (`my.foo`) and one thing in the namespace-shadowed-by-file (`lib.dep.lib.dep.foo`). The name binding logic should note that here too one name should be prioritized over the other.

mitchellwrosen:
## Overview

Fixes #5340

This PR cleans up some name binding logic and fixes a bug related to indirect dependencies causing ambiguity errors when they shouldn't.

## Test coverage

I've added a transcript

### DIFF
--- a/unison-src/transcripts/fix-5340.md
+++ b/unison-src/transcripts/fix-5340.md
@@ -1,0 +1,28 @@
+```ucm:hide
+scratch/main> builtins.merge lib.builtin
+```
+
+```unison
+type my.Foo = MkFoo
+type lib.dep.lib.dep.Foo = MkFoo
+
+my.foo = 17
+lib.dep.lib.dep.foo = 18
+```
+
+```ucm
+scratch/main> add
+```
+
+These references to type `Foo` and term `foo` should be unambiguous (resolving to the `my.Foo` and `my.foo` in the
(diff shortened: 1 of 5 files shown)


unison-src/transcripts/fix-5340.md
```diff
@@ -0,0 +1,28 @@
...
+
+my.foo = 17
+lib.dep.lib.dep.foo = 18
+```
+
+```ucm
+scratch/main> add
+```
+
+These references to type `Foo` and term `foo` are be unambiguous (resolving to the `my.Foo` and `my.foo` in the
```
aryairani:
```suggestion
These references to type `Foo` and term `foo` are unambiguous (resolving to the `my.Foo` and `my.foo` in the
```

aryairani (approved):
Great

PR #5355, opened 2024-09-18
Some JIT tweaks for improved general performance

dolio:
This PR implements some changes to the JIT to improve performance for more cases.

The biggest change is that it implements a way of generated unison code to directly call into the fast path for known-saturated cases, so long as the function that is being called is in the same compilation unit as the caller. This is important for getting good performance on loops, and it was difficult to get the racket compiler to do it for us with the wrappers we have on functions in general.

I've also turned on the inlining hint for all unison builtins, so that it isn't necessary to sprinkle it manually on every builtin definition, which might be about what we want. Obviously this can be revisted later.

The arity stuff could also be extended in the future to allow faster calls to functions outside a single compilation unit, but that would require storing the arity information, and exporting the fast paths from modules, so I'm leaving that a future time when we decide it actually matters.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ on:
 env:
   ## Some version numbers that are used during CI
   ormolu_version: 0.7.2.0
-  jit_version: "@unison/internal/releases/0.0.19"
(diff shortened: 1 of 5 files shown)


pchiusano (approved):
Nice!!

PR #5361, opened 2024-09-20
add transcript with `edit.namespace` output for `base` and `http`

### LINKED ISSUE #5597: certain multiline strings don't round-trip
It's hard to demonstrate in a transcript, but these two definitions have different hashes, yet they pretty-print the same way (as the latter). So if you round-trip `s`, its hash will change.

```
s = "HTTP/1.1 200 OK\r\n"
t =
  """
  HTTP/1.1 200 OK
  
  """
```

aryairani:
## Overview
A place to notice pretty-printing changes.

Closes #5383
Also fixes #5597 

* It takes a while to run, we may want to split them out into a different test that utilizes some codebase caching.
* It produces big output which may clog Github Actions log viewing.
* It doesn't try to round-trip them, as they would both fail.
* It's pinned to specific releases to avoid spurious output changes

## Loose ends


aryairani:
It adds to the transcript running time (especially with the `clone`) but I figure it's better to get it in first, so we can spot how changes affect the output, and speed it up later.

mitchellwrosen (approved):

aryairani:
Windows showing a change to the golden transcripts in http output, not sure why yet.

PR #5364, opened 2024-09-23
tweak: don't truncate long names in some output messages

mitchellwrosen:
## Overview

This PR just deletes the logic that truncates a long name if it can't fit in the preferred terminal width.

Before:

```
  I couldn't figure out what contentLength.get refers to here:

      2 | Body.decodeNonChunkedBody headers = match contentLength.get headers with

  The name contentLength.get is ambiguous. Its type should be: Headers -> Optional Nat

  I found some terms in scope that have matching names and types. Maybe you meant one
  of these:

  ...ers.standard.contentLength.get : Headers -> Optional Nat
  standard.contentLength.get : Headers -> Optional Nat
```

After:

```
  I couldn't figure out what contentLength.get refers to here:

      2 | Body.decodeNonChunkedBody headers = match contentLength.get headers with

  The name contentLength.get is ambiguous. Its type should be: Headers -> Optional Nat

  I found some terms in scope that have matching names and types. Maybe you meant one
  of these:

  Headers.standard.contentLength.get : Headers -> Optional Nat
  standard.contentLength.get : Headers -> Optional Nat
```

## Test coverage

I tested this change manually

### DIFF
--- a/parser-typechecker/src/Unison/PrintError.hs
+++ b/parser-typechecker/src/Unison/PrintError.hs
@@ -1127,10 +1127,7 @@ renderContext env ctx@(C.Context es) =
 
 renderTerm :: (IsString s, Var v) => Env -> Term.Term' (TypeVar.TypeVar loc0 v) v loc1 -> s
 renderTerm env e =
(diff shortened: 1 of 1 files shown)


aryairani (approved):
Awesome

PR #5370, opened 2024-09-25
bugfix: fix tdnr + name resolution bug

### LINKED ISSUE #5369: there's something buggy about TDNR and name resolution
This transcript shows odd behavior:

````
```ucm
scratch/main> builtins.merge
```

```unison
one.foo : Nat
one.foo = 17

two.foo : Text
two.foo = "blah"
```

```ucm
scratch/main> add
```

```unison
one.foo : Nat
one.foo = 18

bar : Nat
bar = foo + foo
```
````

Here, we expect `foo` to resolve to the `one.foo` in the file, because although it could also refer to `two.foo` in the namespace, that wouldn't type check.

Yet the transcript fails, and with a confusing/wrong error message, too:

```

  I couldn't figure out what foo refers to here:
  
      5 | bar = foo + foo
  
  The name foo is ambiguous. Its type should be: Nat
  
  I found some terms in scope that have matching names and
  types. Maybe you meant one of these:
  
  one.foo : Nat
  one.foo : Nat
```

### PR DESCRIPTION
**Pull request by mitchellwrosen**
## Overview

Fixes #5369 

This PR fixes the bug demonstrated in #5369. As of this PR, name resolution (logically) works as follows:

- When resolving a name `foo`,
  - Start with all names in scope (the project local definitions, direct dependencies, and indirect dependencies)
  - Throw away names that don't end in `foo`
  - Throw away indirect dependencies, unless they're all indirect dependencies
    - This implements the "prefer direct deps to indirect deps" logic
  - (for terms only) Throw away names that don't typecheck (i.e. do TDNR) 

Previously, the last step did not properly shadow namespace names with file names, and furthermore was a little out of sync with the "prefer direct deps" rule.

## Test coverage

I've added a regression test for #5369, and another big transcript that shows how TDNR works in a number of different situations

### DIFF
--- a/unison-src/transcripts/fix-5369.output.md
+++ b/unison-src/transcripts/fix-5369.output.md
@@ -1,0 +1,62 @@
+``` ucm
+scratch/main> builtins.merge
+
+  Done.
+
+```
+``` unison
+one.foo : Nat
+one.foo = 17
+
+two.foo : Text
+two.foo = "blah"
+```
+
+``` ucm
+
+  Loading changes detected in scratch.u.
+
+  I found and typechecked these definitions in scratch.u. If you
+  do an `add` or `update`, here's how your codebase would
+  change:
+  
+    ⍟ These new definitions are ok to `add`:
+    
+      one.foo : Nat
+      two.foo : Text
+
+```
+``` ucm
+scratch/main> add
+
+  ⍟ I've added these definitions:
+  
+    one.foo : Nat
+    two.foo : Text
+
+```
+``` unison
+one.foo : Nat
+one.foo = 18
+
+bar : Nat
+bar = foo + foo
+```
+
+``` ucm
+
+  Loading changes detected in scratch.u.
+
+  I couldn't figure out what foo refers to here:
+  
+      5 | bar = foo + foo
+  
+  The name foo is ambiguous. Its type should be: Nat
+  
+  I found some terms in scope that have matching names and
+  types. Maybe you meant one of these:
+  
+  one.foo : Nat
+  one.foo : Nat
+
+```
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: unison-src/transcripts/fix-5369.output.md
Comment (aryairani) [inline unison-src/transcripts/fix-5369.output.md:1]: I'd love just a bit of text describing what's being demonstrated here without having to look up #5369. But I also don't want to hold up the PR for it.

File: unison-src/transcripts/tdnr.output.md [hunk 1]
```diff
@@ -0,0 +1,1008 @@
...
+    
+      thing : Nat
+    
+    ⍟ These names already exist. You can `update` them to your
+      new definition:
+    
+      bad.foo : Text
+
+```
+TDNR selects local term (shadowing namespace) that typechecks over local term (in file) that doesn't.
```
aryairani, line 268: On the examples that select the shadowing namespace, do we need to show the result to know that it's selecting in the shadowing version?

unison-src/transcripts/tdnr.output.md
```diff
@@ -0,0 +1,1008 @@
...
+    
+      thing : Nat
+    
+    ⍟ These names already exist. You can `update` them to your
+      new definition:
+    
+      good.foo : Nat
+
+```
+TDNR not used to select local term (in file) that typechecks over indirect dependency that also typechecks.
```
aryairani, line 576: Is this example _showing_ TDNR not being used? Or what is this example showing?

aryairani: Same with the next two "TDNR not used" examples

aryairani: Is it just informative, saying the local term (in file) is selected over indirect dependency, before TDNR happens? If so, maybe we should print the result again to demonstrate it.

aryairani (approved):
Looks great, I left a few suggestions/questions about the transcripts.

aryairani:
I'm going to merge anyway to help with dogfooding, but we can still tweak the transcripts.

PR #5375, opened 2024-09-26
bugfix: fix suffixification of indirect dependencies

### LINKED ISSUE #5374: `suffixifyByHash` doesn't properly suffixify an indirect dependency name
Input:

````
```ucm
scratch/main> builtins.merge lib.builtin
```

```unison
lib.direct.foo = 17
lib.direct.lib.indirect.foo = 18

thing = indirect.foo + indirect.foo
```

```ucm
scratch/main> add
scratch/main> view thing
```
````

Output:

````
thing : Nat
thing =
  use Nat +
  foo + foo
````

mitchellwrosen:
## Overview

Fixes #5374

This PR fixes a bug in suffixification of indirect dependencies. The bug was caused by the _unnecessary-but-harmless_ optimization to the stopping condition of suffixification (`Set.size matchingRefs == 1`), which has been around forever, and which suddenly became _necessary-not-to-do-and-decidedly-harmful-if-done_ after suffixification was amended to prefer direct deps to indirect deps in https://github.com/unisonweb/unison/pull/5338. The fix was simply to delete that optimization.

## Test coverage

I've added a transcript (failing in the first commit) that demonstrates the fix.

### DIFF
--- a/unison-core/src/Unison/Name.hs
+++ b/unison-core/src/Unison/Name.hs
@@ -575,7 +575,7 @@ suffixifyByHash fqn rel =
 
     isOk :: Name -> Bool
     isOk suffix =
-      Set.size matchingRefs == 1 || matchingRefs == allRefs
+      matchingRefs == allRefs
       where
         matchingRefs :: Set r
         matchingRefs =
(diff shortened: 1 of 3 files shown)


ceedubs:
I can confirm that this fixes the original issue that I reported for this!

aryairani (approved):

PR #5384, opened 2024-09-30
Two fixes for the infix application term printer

runarorama:
This fixes two issues with the term printer when printing infix operator application.

1. *Actual bug*: Printer would re-associate expression to the left when the expression on the right has the same precedence as the expression on the left. For example `1 - (2 - 3)` would get printed as `1 - 2 - 3` which is wrong.

2. *Formatting regression*: The old term printer would lay out infix operator applications in a table. This was somewhat nicer in some cases, but bad for readability in some others. This PR reverts the printer to the old version *only in the case when several applications of the same operator are chained together in a left-associated expression*. This renders e.g. 

```
[1,2,3] 
   |> foo
   |> bar
   |> baz
```

...when space is limited, instead of e.g. 

```
[1,2,3] |> foo |> bar
  |> baz
```

If the operators are associated to the right, or if multiple operators are mixed together, the printer reverts to the normal way of printing.


unison-src/transcripts-round-trip/reparses-with-same-hash.u
```diff
@@ -605,6 +605,8 @@ fixity = do
     (%) = Nat.mod
     ($) = (Nat.+)
     c = 1 * (2 + 3) * 4
+    minus = 1 Nat.+ 2 Nat.+ 3
+    minus2 = 1 Nat.+ (2 Nat.+ 3)
```
ceedubs: Is there a reason that this is called `minus` but does addition? It is probably harmless but it seems peculiar.

runarorama: It used to be `Nat.-` :)

The main thing is that the whole expression roundtrips to the same hash.

aryairani:
@runarorama Looks like there is some interaction with rewrites in `rewrites.md` that is failing CI; could you take a look? Feel free to pull in whoever.

aryairani (approved):
lgtm, thanks!

PR #5388, opened 2024-10-03
bump racket to 8.14 in CI

aryairani:
to match homebrew-installed version

### DIFF
--- a/.github/workflows/bundle-ucm.yaml
+++ b/.github/workflows/bundle-ucm.yaml
@@ -13,7 +13,7 @@ on:
         required: true
 
 env:
-  racket_version: "8.7"
+  racket_version: "8.14"
 
 defaults:
   run:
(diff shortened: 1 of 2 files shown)


ChrisPenner (approved):

dolio (commented):
Seems good to me. My only concern previously was that we shouldn't do CI against a newer version than we can tell users to install, in case we rely on something fixed in the newer version.

dolio (approved):

kylegoetz:
Re @dolio comment about relying on something in 8.14: iirc the UDP implementation I wrote is broken on 8.13 but works on 8.14. I think that was the best explanation we had at the time.

aryairani:
I think we can tell users to install 8.14; it's the latest version at https://download.racket-lang.org/
Homebrew even has it now too, someone added a patch to use openssl3.
I think we have to keep bundling our own though, because we need to be able to link `libb2.{so,dylib}` into racket's `lib/` in order to include libb2 in the redistributable bundle.

Issue #5214, opened 2024-07-11
change transcript info string format


sellout:
Right now transcripts use an info string that looks like `language:flag:other-flag filename`, roughly – e.g. ` ```unison:added-by-ucm scratch.u`.

> The first word of the info string is typically used to specify the language of the code sample, and rendered in the class attribute of the code tag. However, this spec does not mandate any particular treatment of the info string.
—[CommonMark Spec §4.5](https://spec.commonmark.org/0.31.2/#info-string)

The GFM spec says the same thing. Neither seems to define “word”, unfortunately, so it’s hard to say what it means. But, in practice it seems to be basically space-delimited. What that means is that with the current syntax we use in transcripts, we end up with
````markdown
```unison:added-by-ucm
foo = 2 + 2
``` 
````
becoming
```html
<pre>
  <code class="language-unison:added-by-ucm">
    foo = 2 + 2
  </code>
</pre>
```
when it _should_ contain `<code class="language-unison">`.

Despite GitHub not yet recognizing Unison, other places (e.g., my editor) can, but don’t highlight `unison:added-by-ucm` blocks, but my editor _does_ highlight ` ```unison added-by-ucm` blocks.

I propose that we change the syntax. The minimal change would be to replace the first `:` with a space, so ` ```unison added-by-ucm scratch.u` … but there are already cases where we have the file name, but no flags provided, which would be ambiguous. So I wonder if we can come up with a more consistent set of flags that are amenable to Unison adding more in the future.

sellout:
There are related changes suggested in #5199, in terms of what flags should exist.

sellout:
It seems like there’s [a](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444) [light](https://pandoc.org/MANUAL.html#extension-fenced_code_attributes) [consensus](https://michelf.ca/projects/php-markdown/extra/#fenced-code-blocks) toward something like
````markdown
```ucm {hide=all}
scratch/main> builtins.merge
```

```unison {error=intentional file="skritch.u"}
foo = bar
```
````

A starting point for discussing the set of options:
- `result` can be applied to any code block (currently `api` doesn’t support it) and has the values
  - `success` (the default)
  - `error`
  - `failure` (for #5350)
  - `ignored` (called `skip` in https://github.com/unisonweb/unison/issues/5199#issuecomment-2223658861)
- `hide` can be applied to any code block (currently `api` doesn’t support it) and has the values
  - `none` (the default)
  - `output`
[...]

PR #5394, opened 2024-10-07
Add support for “known failures” to transcripts

### LINKED ISSUE #5350: support “known failures” in transcripts
**Is your feature request related to a problem? Please describe.**

I mentioned this in https://github.com/unisonweb/unison/pull/5263#issuecomment-2269837881, so there is some ensuing discussion there.

@aryairani and I have both triaged issues, writing transcripts to see if they’re still issues or not, but then there is nothing to do with the transcript if they _are_ still issues. The next person to triage needs to write those transcripts again and test them out.

This would prevent that work and also allow the org to be more pro-active about resolving issues as soon as they are addressed.

**Describe the solution you'd like**

Right now you can do `` ```unison:error `` to indicate that a block is _supposed to_ result in an error. What I would also like is a way to indicate that a block would ideally succeed, but there is a not-yet-fixed bug that causes it to fail.

This is helpful for identifying when open issues are fixed indirectly. When a ticket is opened, we can add a known failure case to the repo immediately. When the issue is addressed – either directly or indirectly – that transcript will fail, forcing removal of the known-failure flag, and implying that the associated issue should be reviewed and either be closed or have a still-failing test case added.

I think that `:error` might be best used for this new purpose. This is so as issues are opened _containing_ transcripts, they can be passing transcripts, and `:error` is probably the most obvious flag for contributors to use) while a less obvious flag like `:intentional-error` (open to suggestions[^1]) can be used for the relatively small number of cases where we have blocks where an error _is_ the correct behavior.

[^1]: Maybe like `:hide:all`, `:error:intentional`?

**Describe alternatives you've considered**

As suggested in https://github.com/unisonweb/unison/pull/5263#issuecomment-2271642843, we could have another transcripts directory that holds known failures. I think this has two issues
1. it’s too coarse – having multiple blocks that represent known failures in a single transcript allows for commentary attached to them, and allows issues to have only one such transcript.
2. it’s also too coarse in that it doesn’t constrain where in the transcript the failure occurs, or what the failure is, which the current `:error` does nicely.
3. contributors can’t specify it in issue transcripts (ideally we can just copy issue descriptions directly to transcript files unedited).

**Additional context**

Pay attention to #5214 and perhaps also #5199 when addressing this, as it would be good to minimize future flag name changes.

sellout:
## Overview

This allows issues that haven’t yet been resolved to be reflected in the transcripts. This makes it easier to close tickets as soon as issues have been fixed, and avoids having to manually re-try transcripts from a ticket to see if the situation is the same.

Fixes #5350.

## Implementation notes

This adds a new transcript tag `:bug` to represent when ucm thinks a block is an error when it shouldn't, or isn't when it should.

- (no tag) stanza is expected to succeed; the transcript runner will complain if it fails
- `:error` stanza is expected to fail; the transcript runner will complain if it succeeds
- `:bug` stanza expected to succeed, but currently fails (a bug); the transcript runner will notify you if this changes and starts succeeding (bug is fixed).
- `:error :bug`  stanza is expected to fail, but currently succeeds (a bug); the transcript runner will notify you if this changes and starts failing (bug is fixed).

This PR also renames `:hide:all` to `:hide-all` to clarify that it's a single tag and not a combination of tags.

## Test coverage

This adds three new transcripts for open tickets exercising the two new tags.

## Loose ends

There are lots of open tickets that should also get transcripts added.


aryairani:
I wanna think a little more about what the label names should be, because I think `error` `failure` and `incorrect` are hard to distinguish; but otherwise this looks great.

sellout:
I agree. It might be worth doing #5214 before this.

aryairani:
How about:
* \<no-tag> : stanza is expected to succeed; runner will complain if it doesn't succeed
* `:error` : stanza is expected to fail, runner will complain if it doesn't error
* `:bug` : stanza is expected to succeed, but currently doesn't, due to a known ucm bug. runner should notify us if the behavior changes
* `:error :bug` : stanza is expected to error, but currently doesn't, due to a known ucm bug. runner should notify us if the behavior changes

And then also if it's idempotent, we want to preserve any unsuccessful input or unprocessed.

sellout:
Ok, I’ve switched to `:error`/`:bug`, which is a big improvement. Ready for final review.

unison-cli/src/Unison/Codebase/Transcript/Runner.hs
```diff
@@ -204,12 +205,23 @@ run isTest verbosity dir codebase runtime sbRuntime nRuntime ucmVersion baseURL
...
-          then writeIORef hasErrors True
-          else dieWithMsg msg
+        liftIO $ writeIORef hasErrors True
+        liftIO (liftA2 (,) (readIORef allowErrors) (readIORef expectFailure)) >>= \case
+          (False, False) -> liftIO . dieWithMsg $ Pretty.toPlain terminalWidth msg
+          (True, True) ->
+            transcriptFailure
+              out
+              ( Text.pack . Pretty.toPlain terminalWidth $
+                  "The stanza above previously had an incorrect successful result, but now fails with"
```
aryairani: my brain melts with "incorrect successful result".  

```suggestion
                  "The stanza above previously passed, but was marked as a `:bug` that should be `:error`. Now it fails with"
```

unison-cli/src/Unison/Codebase/Transcript/Runner.hs
```diff
@@ -204,12 +205,23 @@ run isTest verbosity dir codebase runtime sbRuntime nRuntime ucmVersion baseURL
...
+                    <> "\n"
+                    <> Pretty.border 2 msg
+                    <> "\n"
+                    <> "if this is the expected result, remove `:bug`, otherwise remove `:error`."
```
aryairani:
```suggestion
                    <> "If it is supposed to fail, you can remove `:bug` and hopefully close a corresponding bug ticket."
                    <> "Otherwise, if it is supposed to succeed, remove `:error`, leave `:bug`, and open a new ticket, since it is currently failing."
```
These are hard to write; let me know if I have it wrong.

unison-src/transcripts/errors/unison-hide-all-error.output.md
```diff
@@ -4,7 +4,7 @@ When an expected error is not encountered in a `unison :hide:all:error` block
 then the transcript parser should print the stanza
 and surface a helpful message.
 
-``` unison :hide:all:error
+``` unison :hide:all :error
```
aryairani: Just curious, why does `:error` get a space but `:all` doesn't?

aryairani (approved):

unison-src/transcripts/errors/unison-hide-all-error.output.md
sellout: Because `:hide:all` is really just one tag. It’s never been parsable as `:hide :all`. I just changed the printer to put spaces between, but didn’t modify the parsers.

Might be worth renaming to `:hide-all` (perhaps allowing `:hide:all` still for backward compat).

unison-cli/src/Unison/Codebase/Transcript/Runner.hs
sellout: I think we should come up with some consistent terms, e.g.:
- “succeeded” – had a non-error result, regardless of whether that was intended
- “errored” – had an error result, regardless of whether that was intended
- “passed” – had the intended result, regardless of whether it was success or error
- “failed” – had an unintended result, regardless of whether it was success or error (or just a different error than expected)

Doesn’t have to be those terms in particular, but I think the succeeded/errored pair should be complementary terms, and the passed/failed pair should also be complementary.

“Succeeded” is the only term there that I’m not particularly happy with.

In this case, the stanza already has `:error :bug` on it, and we want to suggest the removal of `:bug`. My original version doesn’t satisfy my terminology above because it says “fail” when it should say “error”. Maybe something like, “The stanza above is now erroring with […] as expected. Please remove `:bug`.”?

unison-cli/src/Unison/Codebase/Transcript/Runner.hs
sellout: Oh, this is the tail of the same message – I missed that.

So, I don’t think my original or the new one are correct here. Removing `:error` is probably not what we want to do. I think what I was trying to capture here is something like “this is now erroring, which means we should remove `:bug` … but is the error the error that we’re actually expecting, or is it still the wrong error message?”

unison-src/transcripts/errors/unison-hide-all-error.output.md
sellout: I just added a commit with this change, I can back it out if you prefer.

unison-cli/src/Unison/Codebase/Transcript/Runner.hs
aryairani: I think having four terms is too many, I wonder if we can do it with no more than two.

Yes to "is the error the error we're actually expecting?" that's where I was heading with the "open a new ticket if needed".

I'd be satisfied with this, wdyt:

```suggestion
                  "The stanza above marked with `:error :bug` is now failing with"
                    <> "\n"
                    <> Pretty.border 2 msg
                    <> "\n"
                    <> "so you can remove `:bug`, close any appropriate Github issues by using the commit message or PR description, and/or open a new issue if the error message is still not correct."

unison-src/transcripts/errors/unison-hide-all-error.output.md
aryairani: I like `:hide-all`

unison-cli/src/Unison/Codebase/Transcript/Runner.hs
aryairani: I opened a new thread below that doesn't split the message up into two conversations.

sellout: Oh, yeah – mentioning closing the related issue is important! I’ll do the same on the the other `:bug` case. And yeah, this message looks good to me.

I think there’s still four terms, but different axes, and ones that match our tags better: passing/failing (`:error`) and expected/unexpected (`:bug`). So yeah, I like that direction.

sellout: Ok, I changed the phrasing a bit again, so see what you think in the latest commits.

I also realized that the only tests added were for actual outstanding bugs, so I added an additional transcript that will continue testing the “positive” cases after those outstanding bugs are fixed, as well as two error tests for fixed bugs (which show the messages we’ve been fiddling with).

PR #5405, opened 2024-10-09
Some tweaks related to caching top level values.

dolio:
This includes a couple changes related to the recently merged top-level value caching PR.

1. It changes the value serialization format back to an extension of the `v4` format. That way, all old values are parsed/serialized in exactly the same way as before. Only code that is newly recognized as cacheable is serialized in a different way. Old implementations would be unable to recognize such serialized code, but they will know they are failing to recognize something and give an error, rather than continuing to parse things in an erroneous way.
    - This was motivated by cloud, and the above error case shouldn't be a problem, because the server will not typically be sending _code_ values back to the client. So, not incrementing the version allows it to communicate the values it does send back in a format that the client should be able to understand. It doesn't fix a similar problem of servers communicating with one another while partially upgraded, but that seems less urgent.
3. Also included is some code support for having the values cached in the JIT. All the code generation should be there (as far as I know), but it's turned off at the moment because I cannot easily sort the definitions in a way that racket will accept. Once that problem is fixed it can be enabled to get similar performance enhancements on the JIT. I already had this written, so it was easier to just include it now than leave it rotting somewhere.

Here's an additional question. I could make it so that the for-hash serializer pays no attention to the cacheability status. This would make it so that code values that are newly recognized as being hashable retain the same hash as they did before. Should I do that or not bother?


pchiusano (approved):
> Here's an additional question. I could make it so that the for-hash serializer pays no attention to the cacheability status. This would make it so that code values that are newly recognized as being cacheable retain the same hash as they did before. Should I do that or not bother?

Yes, I'd do this, so everything hashes the same as before.

ceedubs (approved):
I can't claim to have done much of a review, but I agree with the premise. I'll run it through the Unison Cloud integration tests once it's in a dev build.

.github/workflows/ci-test-jit.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_call:
 
 env:
-  runtime_tests_version: "@unison/runtime-tests/releases/0.0.2"
+  runtime_tests_version: "@unison/runtime-tests/releases/0.0.1"
```
aryairani, line 7: Is this the switch back to v4?

unison-src/builtin-tests/interpreter-tests.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 set -ex
 
-ucm=$(stack exec -- which unison)
+ucm=$(cabal exec -- which unison)
```
aryairani, line 4: was this meant to be merged? asking late sorry

dolio: No, that was an accident. I'll fix it in a PR for the hashing change.

Maybe I'll fix it to work like the jit script, so you can pass that in on the command line.

.github/workflows/ci-test-jit.yaml
dolio: Yeah, the change to 0.0.2 was just to point it at different files.

PR #5409, opened 2024-10-10
Make hash of cacheable code the same as if it were uncached

dolio:
This implements the change that throws away the cacheability information when serializing for hashing, so that it doesn't influence the hash.

Also tweaks the `interpreter-tests.sh` file to accept a ucm on the command line, and falls back to using stack to find the file, which was accidentally changed in the last PR. This will avoid me accidentally changing it in the future.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ on:
 env:
   ## Some version numbers that are used during CI
   ormolu_version: 0.7.2.0
-  jit_version: "@unison/internal/releases/0.0.22"
+  jit_version: "@unison/internal/releases/0.0.23"
   runtime_tests_version: "@unison/runtime-tests/releases/0.0.1"
 
   ## Some cached directories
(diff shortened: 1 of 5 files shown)


aryairani (approved):

pchiusano:
@dolio @SystemFw before merging this, I'm uncertain how your test in cloud client passed before (that the hashes of things didn't change). It seems like until this PR, the hash of a bunch of things would have changed, and that would have caused your test to fail.

dolio:
What is being hashed? Most stuff didn't change. Only the hash of `Code` values.

pchiusano:
Oh okay, what about the hash of a Unison function? As in:

```
blah x = x + 938

> crypto.hash Sha3_256 blah
```

You're saying that never was changed, even before this PR?

dolio:
Right, that is a partial application whose hash is based on the reference for `blah` and its arguments (but there are 0 in this case). The only thing that would change hash is if you asked for the `Code` of `blah` and then hashed it.

PR #5416, opened 2024-10-18
Decouple JIT builtins from interpreter wrappers

dolio:
This PR restructures the JIT builtins to avoid reliance on the interpreter wrappers. This means we can freely change the wrappers without having to worry about how it affects the JIT (and fixing changed calling conventions and whatnot). It's also just a better implementation for many of the changed builtins, because it avoids extra indirections, possibly through unnecessary data structures that won't be optimized away.

As part of this, I've moved most of the builtin implementations to a `unison/primops` subdirectory, and now the `unison/primops` module mostly re-exports these modules. It also contains a couple special cases of instruction-like things that weren't eliminated (but should still not cause problems when working on the interpreter).

I also added native versions of the test commands so that you can try running tests on the JIT without hacking ucm. They're hidden at the moment. Beware: running `test.native` incurs the JIT startup overhead for every individual test, so running tests in e.g. `@unison/base` takes quite a while (~800 tests = 800 startups).

I've got a couple base tests to finish fixing up, but otherwise I think this is almost ready to go. Most of them pass except for some RSA tests that haven't been implemented yet, and I'm not going to bother right now.


pchiusano (approved):
😎

dolio:
FYI, the base tests are more or less fixed. Apparently the discrepancy was actually due to a problem in a previous commit. So I think this is ready to go from my perspective.

aryairani (approved):
rubber stamp

PR #5442, opened 2024-11-05
fix #5441 - malformed Unison examples in source code

### LINKED ISSUE #5441: TermParser source code has example Unison code that isn't valid
TermParser.hs indicates a restructuring bind with LHS being a special list form (like `[a] ++ [b]`) being valid, but docs disagree.

Also contains an example where LHS doesn't have surrounding parens.

```haskell
-- example:
--   (x, y)   = foo
--   hd +: tl | hd < 10 = [1,2,3]
--   stuff
--
-- desugars to:
--
--   match foo with
--     (x,y) -> match [1,2,3] with
--       hd +: tl | hd < 10 -> stuff
```

and

```haskell
-- We have to look ahead as far as the `=` to know if this is a bind or
  -- just an action, for instance:
  --   Some 42
  --   vs
  --   Some 42 = List.head elems
```

Fix would be remove or correct these examples.

kylegoetz:
## Overview

This changes no behavior. It only removes/updates a couple Unison examples in Haskell comments so they're no longer malformed Unison code.

The docs claim all destructuring bind LHSs must be parenthesized. However, the Haskell code has two examples where this is not true. One of them even uses a special list pattern on the LHS of a restructuring bind. The docs explicitly say this is not supported.

Reading through TermParser, I couldn't find a single reason this *would* fail, and yet I couldn't get it to succeed in UCM.

Suggesting the docs are right and the comments are not.

## Test coverage

No tests because it's just comment changes.

## Loose ends

One of the comments I edited makes a claim about how destructuring binds are desugared to a `match` structure. I have not independently verified this is true; I only edited the malformed restructuring binds.

### DIFF
--- a/parser-typechecker/src/Unison/Syntax/TermParser.hs
+++ b/parser-typechecker/src/Unison/Syntax/TermParser.hs
@@ -1243,22 +1243,20 @@ verifyRelativeName' name = do
 
 -- example:
 --   (x, y)   = foo
---   hd +: tl | hd < 10 = [1,2,3]
 --   stuff
 --
 -- desugars to:
 --
 --   match foo with
---     (x,y) -> match [1,2,3] with
---       hd +: tl | hd < 10 -> stuff
+--     (x,y) -> stuff
 --
 destructuringBind :: forall m v. (Monad m, Var v) => P v m (Ann, Term v Ann -> Term v Ann)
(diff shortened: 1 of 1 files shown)


aryairani (approved):

aryairani:
Good catch, thanks!

PR #5443, opened 2024-11-05
add some merge progress output messages

mitchellwrosen:
## Overview

`merge` can take a long time (which should be addressed of course), but in the meantime we want more feedback so users don't think it's broken.

This PR just sticks in output messages in a few places. A merge with this PR looks like this:

Screen Shot 2024-11-05 at 2 15 44 PM

Once we finalize exactly what these messages should say, how many there should be, etc, I can run all the transcripts and get CI passing.


mitchellwrosen:
I can't take a screenshot, but this PR now puts all of those progress messages in a console region that updates in-place and disappears at the end. @aryairani

aryairani:
> I can't take a screenshot, but this PR now puts all of those progress messages in a console region that updates in-place and disappears at the end. @aryairani

Cool!

P.S. `asciinema` is a nice option for capturing that sort of thing.

aryairani:
Looks great though

aryairani (approved):

